PR #2, opened 2018-06-22
Fixes for Issue #1

jstraus59:
This branch has fixes for items 1 thru 3 in Issue #1.

Item 4 was not addressed because it  is unclear to me what the intended difference is between the ISA and RISCV_DEVICE variables.


jeremybennett:
Thanks for these Jim.  All obvious fixes, so I've merged them.

jeremybennett:
Merging in commits as being non-controversial

PR #14, opened 2018-11-06
fix for ma_fetch test in rv32mi

neelgala:
The ma_fetch test has been written in a way that implementations with and without "C" support should generate the same signature. The aim is to avoid 
```
j fail
```
at multiple places. 

However, currently the signatures generated by implementation with and without C support are different. This is because the trap handler sets the return address in mepc to point to 

```
SWSIG(?, TESTNUM)
```
instead of 

```
li TESTNUM, ?
```

This merge request fixes that and also updates the signature. Has been verified with e-class, riscvOVPsim and spike.


neelgala:
@aswaterman 
Would this work?

eroom1966:
Disregard the previous request, I have merged from the master
Thx
Lee

PR #19, opened 2018-11-15
New signature format

neelgala:
As discussed in the CTG meeting, I have changed the signature file format for all references.
Old format: 
```
00000000000000010000000000000000
00000000000000100000000f0000001f
```
New format:
```
00000000
00000000
00000001
00000000
0000001f
0000000f
00000010
00000000
```
This change makes sense as a minimum riscv device would be rv32 which will always support lw/sw instructions. Thus a 4-byte data per word instead of 16-bytes.

Also put in a patch for spike-device Makefiles where the old-signature format is post-processed to generate a signature in the new format at the end of each test. 
Spike now passes all the tests.

Also made a small modification in the root-dir Makefile. 
Default was to run all tests on the riscv-test-suite on a particular device if RISCV_ISA is not set. However, the default should be to run all tests that are supported by the device. This would be the list of suites mentioned in riscv-target/RISCV_TARGET/device.

Other Targets:
riscvOVPsim and Codasip-simulator will have to modify their targets to meet this new change. 
A quick hack would be to simply use the same patch as used in spike - **but this is highly discouraged.**

ChangeLog:

2018-11-21 Neel Gala <neelgala@incoresemi.com>
   	* riscv-test-suite/rv_/references/_.reference_output: changed signature 
	  format for all tests to include only 4-bytes per line starting with the
  	  most significant byte on the left.
    	* riscv-target/spike/device/rv_/Makefile.include: Added a patch for 
	  spike-device Makefiles where the old-signature format is post-processed 
      	  to generate a signature in the new format at the end of each test.
	* riscv-target/riscvOVPsim/device/rv_/Makefile.include: same patch as above.
    	* Makefile: default target for Makefile is now to run all tests supported by 
	  the target mentioned defined by RISCV_TARGET variable.

doc/ChangeLog:

2018-11-21  Neel Gala  <neelgala@incoresemi.com>
	* README.adoc: Added new signature format spec.


davidvconnelly:
This might work but if you do do an 'objdump -s' on the elf file you will see a different format.  It displays four 32-bit words per line - should we match that to be consistent everywhere?

neelgala:
'objdump -D' however prints according to the proposed signature format. Any reason why '-s' should be chosen over '-D'
Also the current format is how gdb would also print. This would be important because to capture signature of a test from silicon/fpga prototypes you will typically use a debugger+openocd framework which would make it easy to dump in the current format that is proposed.

davidvconnelly:
The objdump output using the -s option was a suggestion to bridge the gap from the original format (16 byte line with no spaces) to an updated format that is more readable.  The suggested 4-byte word per line looks good and is readable and as mentioned advantageous to gdb users so it should work going forward....as long as there is confidence this will stick for rv64 and beyond as well

neelgala:
owners merging this request should probably make suitable entries in the changelog since its not part of this request.

neelgala:
when will this get merged?

jeremybennett:
Neel,

Thanks for this. I don't believe I was at the TG meeting where this was agreed, but given the review comment and response I am happy to merge, but we need three things

1. Can you extend the documentation (doc/README.adoc) to cover your changes as appropriate.

2. Can you verify that all tests still pass with OVPsim and Spike.

3. Can you add ChangeLog and doc/ChangeLog entries appropriately. The ChangeLog entries should be included in your commit message thus:

```
<summary line>

        <description of your patch>

ChangeLog:

<ChangeLog entry>

doc/ChangeLog:

<doc/ChangeLog entry>
```

Once we have this, I'll merge the commit.

neelgala:
The pull-request includes changes to make spike pass all tests.
I can make relevant changes in the OVPsim device Makefiles so that it can pass all tests as well.
@simon5656 @eroom1966 can I go ahead with making those changes for OVPsim?

neelgala:
@jeremybennett 
Made updates to ChangeLog, doc/ChangeLog and README.adoc. Are they okay to merge?

jeremybennett:
Hi Neel,

Thanks for this. The main ChangeLog is not strictly in the defined format, but it captures what we need, so lets go with it.

Can you confirm that all the Spike and OVPsim tests still pass, and then I'll commit.

For future reference, the spec of ChangeLogs is [here](https://www.gnu.org/prep/standards/html_node/Change-Logs.html). I'm a great fan of them as imposing discipline to a change control process, and have blogged about it [here](https://www.embecosm.com/2014/06/12/change-logs-why-they-still-matter/). Others argue that git logs can do the same, but this only works if you apply the same engineering discipline&mdash;I've yet to meet a project which does!

jeremybennett:
Nice. Now just need to know that the Spike & OVPsim tests still pass and I can commit.

neelgala:
@jeremybennett All done. riscvOVP and Spike both pass all the tests.

Thanks for ChangeLog pointer :)

jeremybennett:
Thanks for all the work on this.

PR #34, opened 2019-01-29
Rocket Chip as target

debs-sifive:
**Goal:** Add rocket chip as a target for the compliance tests. Using `rocket` as a target will by default use the emulator generated for `DefaultConfig` or `DefaultRV32Config` depending on the test suite. Users can build their own configs and run them against the compliance test using the `ROCKET_CONFIG` environment variable.

**Issues:**
- This is currently failing 3 tests, listed below.
- The `DefaultRV32Config` does not support double precision, and thus will not pass the `rv32ud` test suite. For these tests, a `ROCKET_CONFIG` is required, and will not default to `DefaultRV32Config`.

**TODO:**
- [ ] fix for [rv32i/I-MISALIGN_JMP-01.S](https://github.com/riscv/riscv-compliance/issues/31)
- [ ] fix for [rv32mi/breakpoint.S](https://github.com/riscv/riscv-compliance/issues/32)
- [ ] fix for [rv32si/ma_fetch.S](https://github.com/riscv/riscv-compliance/issues/33)
- [x] add documentation on how to use rocket cores as targets
- [x] update change log


jeremybennett:
Nice work. Only minor polish is that it is usual to leave a blank line between the header of a ChangeLog entry and the contents, but I don't propose worrying about that - the important thing is the content.

I'm a great believer in "commit early and commit often", so I'll now merge. You never know, someone else may then fix the failing tests!

PR #35, opened 2019-01-29
This is a simplification of the Logging and Asserting macros.

eroom1966:
The only reserved register used now is x31
This register is used as a pointer to a bank in memory used for pushing
and popping registers at the prolog/epilog of the calls for logging and
asserts

This addresses some of the issues raised regarding the problem of assert macros not able to assert on some registers. Only the value of x31 is not preserved after a call to either the macros for assert or message logging


simon5656:
Radek -
1) maybe I am missing something here, - in any text you write you can choose to use x31 - and then you cant use the ASSERT macro - as it would clash - so there is no restriction on using x31 in your tests if you want just dont use both.
2) as far as I can see the pull request is only about tests and infrastructure and has nothing to do with riscvOVPsim - yes it is always good policy to keep pull requests separate for different functional parts of the repo.
thanks
Simon


Imperas:
Hi Radek
>     1. I don't agree that we should completely remove register x31 from all tests. Assertions are optional debugging feature - it must not influence the tests in a negative way.

There is no requirement to remove register x31 from tests
If using Assertions, the value of x31 will not be preserved after the call

In the existing infrastructure there are 4 registers which are not preserved, so this pull requests reduces that to 1 register which is not preserved

Thx
Lee

hajekra:
@simon5656 
In this pull request I see 3 commits:
> This is a simplification of the Logging and Asserting macros.
change riscv-ovpsim to a submodule
add riscv-ovpsim as submodule 

@simon5656 @eroom1966 
I see that x31 was replaced by another register in existing tests. I can now implement only 31 registers in my architecture and it will pass compliance tests, because x31 is used only within assertions which are optional and can be skipped...

eroom1966:
@hajekra 
> @simon5656 @eroom1966
> I see that x31 was replaced by another register in existing tests. I can now implement only 31 registers in my architecture and it will pass compliance tests, because x31 is used only within assertions which are optional and can be skipped...

We can create an additional register test that moves values between registers and into the memory signature file. This will prove x31 is available, and operational, so for instance if we had 
```
la x2, <signature_start>
li x31, 0x1f
sw x31, 0(x2)
```
We now have the value via the x31 register in the signature file, so now you can no longer implement your architecture with only 31 registers.
I am happy to add this simple test (probably to an existing test) to provide a solution
Would that be sufficient ?

Thx
Lee

hajekra:
Well, adding new test for x31 would probably find the missing register. But there are still other issues:
1) you have to choose between using x31 and assertions. But it is not obvious and may cause some troubles in the future. Or I can implement assertions using different register...
2) coverage - in my opinion, all instructions should use x31. So, it is not only about testing whether x31 is present.
3) Assertions using x31 are not compatible with E.

I would prefer different solution - do not modify tests but change only assertions. In other words, do not reserve any register for tests. In every single test there is at least one register available, so I would pass it to assertion:
```
RVTEST_IO_ASSERT_GPR_EQ("register to compare", "expected value")
->
RVTEST_IO_ASSERT_GPR_EQ("register to compare", "expected value", "helping register")
```

eroom1966:
@hajekra 

> Well, adding new test for x31 would probably find the missing register. But there are still other issues:
>     1. you have to choose between using x31 and assertions. But it is not obvious and may cause some troubles in the future. Or I can implement assertions using different register...
>     2. coverage - in my opinion, all instructions should use x31. So, it is not only about testing whether x31 is present.

Regarding coverage, I raised an issue months ago to discuss what coverage should comprise in the compliance WG meetings
This has never been discussed and definitively decided, and so until there is agreement, there is only personal opinion, for example
Should every register be used as a target and a source register
Should tests include instructions using the same target AND source register
Should every set of source combinations target every set of destination options
Should all data values be provided in all registers, is a toogle, walking 1, walking 0 sufficient
Should all data values be provided in target register results, is a toggle, walking 1, walking 0 sufficient
The list is endless and a base susbset should be agreed, before it should be enforced
 
>     3. Assertions using x31 are not compatible with E.
agreed

> I would prefer different solution - do not modify tests but change only assertions. In other words, do not reserve any register for tests. In every single test there is at least one register available, so I would pass it to assertion:
> ```
> RVTEST_IO_ASSERT_GPR_EQ("register to compare", "expected value")
> ->
> RVTEST_IO_ASSERT_GPR_EQ("register to compare", "expected value", "helping register")
> ```
I think this is a good suggestion, and have implemented this and have it working, the one difference is in the argument list order which is
```
RVTEST_IO_ASSERT_GPR_EQ("helping register", "register to compare", "expected value" )
RVTEST_IO_WRITE_STR("helping register",  "text string");
```

I do not propose to go back and change the tests to cover x31 in all operations, without a test specification
This should be part of the discussion in the Compliance-WG to agree what a test should target in terms of register and data coverage, including cross-coverage - until that is agreed this is totally in flux, and in fact the completion of the tests was supposed to be part of the contracted tender for a 3rd party, not for the infrastructure prototype

Thx
Lee

hajekra:
Well, I wouldn't change the tests if it is not necessary. I would prefer to only change the macros and their usage.

eroom1966:
superceeded by #36

PR #36, opened 2019-01-30
Simplification of macros and addition of riscv-ovpsim as a submodule

eroom1966:
This pull request supercedes #35 
includes simplification of macros for RVTEST_IO_ASSERT_GPR_EQ and RVTEST_IO_WRITE_STR
This only requires the usage of a single (volatile) register which is user selectable

Additionally risc-ovpsim is provided as a submodule referring to
https://github.com/riscv/riscv-ovpsim

Thx
Lee


eroom1966:
@jeremybennett 
Jeremy, we have decided to heed your advice to not use submodules, in which case lets ignore this pull request, and I will shortly submit another
Thx
Lee

Issue #32, opened 2019-01-26
breakpoint.s undesired behavior when trigger does not exist?


debs-sifive:
When a trigger doesn't exist, the test goes to the `pass` label. The result is that the signature is not written to after the trigger check, and the target will fail `make verify`. This doesn't seem like the ideal action, since it's valid for a target to have no trigger.

For context, I'm trying to add rocket chip's `Default[RV32]Config` as a target, and it only has one trigger.

allenjbaum:
IF this has been fixed, can we close the issue?

PR #43, opened 2019-02-05
breakpoint.S fix

debs-sifive:
**Goal:** solve issue #32

**Changes:** Instead of going to the `pass` label if a breakpoint does not exist, branch to `sign_pass`, which writes the signature, then goes to the `pass` label. The end result is that the signature will match even if breakpoints are not implemented.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,7 @@
 
 2019-02-05 Deborah Soung <debs@sifive.com>
     * [Issue #33] fixing rv32si/ma_fetch.S test
+    * [Issue #32] fixing breakpoint test
 
 2019-02-01 Lee Moore <moore@imperas.com>
     * updated Infrastructure macros to support non-volatile registers
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -1,5 +1,6 @@
 2019-02-05  Deborah Soung  <debs@sifive.com>
 	* README.adoc: Update documentation for rocket chip as target (fixed rv32si/ma_fetch.S).
+	* README.adoc: Update documentation for rocket chip as target (fixed breakpoint.S).
 
 2019-01-29  Deborah Soung  <debs@sifive.com>
 	* README.adoc: Documentation for rocket chip as target.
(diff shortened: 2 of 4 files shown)


jeremybennett:
Does the standard allow breakpoints not to be implemented?

BTW, I notice you use quite a lot of force pushes in your branch. Is there a reason for this?

debs-sifive:
Ack sorry, I force pushed to squash/clean up my messy commit messages.

I guess using the term "breakpoint" is a bit misleading. From what I can gather, the code expects to hit a "trigger", which can then cause a breakpoint according to the debug spec, and I believe the implementation of triggers is platform-specific. Maybe @aswaterman can confirm?

allenjbaum:
I haven't looked into this deeply, but since this test is using mmode-only
resources (tdata, tselect), it should be located in the privileged test
suite, not the unprivileged suite.
Ditto for any test that uses other MMode or SMode CSRs, or that may trap as
part of the test.
Tests that modify their behavior to avoid traps, but can determine whether
they should skip the trapping behavior without resorting to accessing
an M/SMode CSR can remain in the unprivileged test suite.

aswaterman:
The debug specification defines some standard triggers, but the debug spec is itself entirely optional, so it's valid to have a RISC-V implementation that has no breakpoints.

Allen's right that tests that manipulate the trigger CSRs belong in the privileged suite (and also in the debug suite, since there are some trigger features that are only available to debug mode, but maybe that's outside the scope of this project).

allenjbaum:
On the other hand, the name of the test is breakpoint.s, which i think
implies the presence of S-mode, so is therefore is automagically in the
privileged test suite?
So it might already be there; though we have to ensure that the framework
knows that.

jeremybennett:
Deborah - this is still showing rebase conflicts. Are you able to resolve this. It might be simpler to just submit a new pull request with a single quashed commit.

jeremybennett:
Deborah - could you resubmit as a new pull request on the latest master and I'll approve and commit.

debs-sifive:
@jeremybennett sorry for the wait--was very ill. Please try pushing now.

jeremybennett:
Thanks - glad you have recovered.

Issue #33, opened 2019-01-29
rv32si/ma_fetch.S produces a different signature depending on RVC support


debs-sifive:
I think this is a similar issue to [fix for ma_fetch](https://github.com/riscv/riscv-compliance/issues/15), but in `rv32si`.

When I run the test with target `spike` with `--isa=rv32ic`, the signature I get is:

```
00000001
ffffffff
00000002
00000003
00000004
00000005
00000006
00000007
ffffffff
  .
  .
  .
```

Which does not match up with the signature produced with `--isa=rv32i`, which matches the [current reference signature](https://github.com/riscv/riscv-compliance/blob/master/riscv-test-suite/rv32si/references/ma_fetch.reference_output).

allenjbaum:
IF this is fixed, can we close it?

PR #44, opened 2019-02-06
rv32si/ma_fetch fix

debs-sifive:
**Goal:** solve issue #33 

**Changes:** `stvec_handler` should now jump to the `li` instruction before `SWSIG` so that the signature is written to regardless of `C` extension support. See pull #14.


jeremybennett:
This looks routine to me. I'll leave it for 3 days for comments, and if there are no objections I shall approve and commit.

PR #49, opened 2019-02-21
Fixed issue of asserting on register t0

eroom1966:
This is to fix the issue of T0 asserts not working correctly


eroom1966:
@jeremybennett @hajekra 
Hi Jeremy, Radek has gone ahead and reviewed - this is simply a set of bug fixes to the previous changes for the new ASSERT style macro
Could you please merge the request.
Thx
Lee

riscv-test-suite/rv32i/src/I-ADDI-01.S
```diff
@@ -294,7 +294,7 @@ RV_COMPLIANCE_CODE_BEGIN
     sw      x5, 4(x2)
 
     RVTEST_IO_ASSERT_GPR_EQ(x2, x0, 0x00000000)
-    RVTEST_IO_ASSERT_GPR_EQ(x2, x5, 0x00000000)
```
neelgala: I guess this value should be 0 and not 1. Atleast that is what the signature value is.

hajekra: I believe 1 is correct - it is result of "addi x5, x0, 1", isn't it?

neelgala: Ohh crap.. you are right. I was looking at ADD and ADDI tests simultaneously and got mixed up.

neelgala:
It might make sense to add the following in the I-FENCE.I-01.S [here](https://github.com/riscv/riscv-compliance/blob/2fdea4a96cac9846d426242079be9528c267d877/riscv-test-suite/rv32i/src/I-FENCE.I-01.S#L79)
```
RVTEST_IO_ASSERT_GPR_EQ(x17, x1,  0x00000030)
```

PR #50, opened 2019-02-21
RISC-V ISA Formal Specification as target

debs-sifive:
For reference: [RISC-V ISA Formal Specification](https://github.com/sifive/RiscvSpecFormal)

The test environment is provided inside the target directory.

Suites `rv32i`, `rv32ua`, `rv32uc`, `rv32uf`, and `rv32ui` are currently supported by the formal model. Some tests in `rv32i` will not pass due to their usage of privileged CSRs (not yet implemented in the formal model).


jeremybennett:
Thanks for this Deborah.  I'm hoping someone else will review this - I'm outside my area of confidence.  Because I've merged two other commits, this pull request now has conflicts.  You should be able to resolve these by rebasing your branch on master and dealing with the conflicts. Almost certainly in the ChangeLog files, since git seems always to make a mess of these.

simon5656:
The dir name of the target/device needs to be more than just eg 'formal', it needs be something like SiFive_ISA_formal or something... so that it is distinguishable from the other 5 formal models...

debs-sifive:
Thanks for pointing that out, @simon5656! I've renamed it now.

jeremybennett:
Gah - we have conflicts again. Could you rebase and I'll merge it. Thanks

debs-sifive:
@jeremybennett done! Thanks for your help as always :)

PR #54, opened 2019-04-02
Grift

benjaminselfridge:
Added riscv-compliance targets for GRIFT. Fails the misaligned-ldst test in rv32i, but that is because the test currently assumes that misaligned loads/stores trap. All other tests that I have provided devices for are passing.

Still need to update README.


benjaminselfridge:
@jeremybennett Can we get this merged soon?

jeremybennett:
@benjaminselfridge I'm frustrated by this. I have asked twice for a reviewer for this patch at meetings. Perhaps you would join the Zoom meeting next Wednesday to press  the case. This is way out of my comfort zone, so I'm not a suitable reviewer. Sorry about the delay.

benjaminselfridge:
Yep -- happy to join the meeting, that sounds fine.

PR #57, opened 2019-05-17
Macro proposals, grammar/wording changes

allenjbaum:
Added wording to clarify spec intent (work in progress/goal rather than final) 
Proposes macros to ease test authoring .
Added detail on proposals for connection to framework (how framework selects tests).
Expanded definition of signature format
Changed the (proposed) directory structure and naming convention to eliminate ambiguities, add consistancy and slightly better match existing structure
Added many "future work" items related to the above
Added examples
Added comments to code examples to indicate how proposed macros would be used


.gitignore
```diff
@@ -5,3 +5,5 @@
 /.project
 /work/
 /setup.env
+#ignore Mac file system artifacts
+*.DS_Store
```
pdonahue-ventana: Shouldn't Mac users do something like this:
echo "*.DS_Store" > ~/.gitignore
git config --global core.excludesfile ~/.gitignore

That solves the problem for all git workspaces rather than just riscv-compliance.  That's the best way for emacs users to ignore tilde files, etc.

spec/TestFormatSpec.adoc
```diff
@@ -112,6 +114,7 @@ Allen Baum |
 More grammar wording, syntax corrections,
 Added detail regarding the 2 approaches for test selection: central database, or embedded conditions embedded in macros
 Added detail of proposed standard macros RV_SIGBASE, RVTEST_SIGUPDATE, RVTEST_PART
+Mre expanation of spec status in initial f\uture work paragraph (i.e. goal, not yet accomplished)
```
pdonahue-ventana: Does this contain typos?

allenjbaum:
Let's just say I was more careful checking my changes than checking the check-in message.

-Allen

.gitignore
allenjbaum: That requires all Mac users to do something - which is a lot larger number of actions than having it added to each riscv repository. Users get added at a much higher rate than repositories, and this fixes it no matter how many people start added to a repository (and who use a Mac)

pdonahue-ventana: There are tens of millions of git repositories on GitHub alone.  While we could potentially add it to each riscv repository, what happens if you clone some other repo?  The only general solution is for each Mac user to control their own user-specific settings.

However, your solution will work for you if you never use git except for riscv repositories.

I don't think that this is a big problem, it's just not the cleanest solution.

pdonahue-ventana:
> Let's just say I was more careful checking my changes than checking the check-in message.

The changes themselves were shorter than the check-in message so I didn't actually read the check-in message.  The typos are in the actual file that you checked in at the line I indicated.

.gitignore
allenjbaum: Personally, I'm only worried about riscv repositories (and more particularly, the one I'm most responsible for, compliance.) Personally, I'd rather put the onus on the creators/users of a repository rather than individual users, but either works. In any case, it is now fixed for this repository. And, I'll take your advice so I don't have to do it for any other repositories I'm using.

PR #61, opened 2019-08-09
Added Ibex as a target (under Verilator)

eroom1966:
Also modified the prolog code for targets not supporting supervisor mode
certain CSR's are only written if privilege modes are supported


ChangeLog
```diff
@@ -1,4 +1,8 @@
-2016-06-21 Ben Selfridge <benselfridge@galois.com>
+2019-08-08 Lee Moore <moore@imperas.com>
+    * Added support for lowRISC/ibex RTL as a target using Verilator.
+      In conjunction with Philipp Wagner of lowRISC imphil@philipp-wagner.com
```
asb: Probably better to use Philipp's phw@lowrisc.org work email

eroom1966: thanks updated in new pull request

eroom1966:
> I suspect this needs breaking into two pull requests, one for Ibex and
> one for some general tidying up. Could you let me know your thoughts
> on the above and I'll then review again.

No problem, let me take another look

Thx
Lee

eroom1966:
need to resubmit

PR #62, opened 2019-08-13
Update of riscv-ovpsim, and inclusion of new config-yaml

eroom1966:
This updates only the riscv-ovpsim portion of the riscv-compliance tree


eroom1966:
Update Changelog

neelgala:
Would suggest using the 1.0.2 version of riscv-config. That has logging capabilities which is missing in 1.0.0

jeremybennett:
@neelgala 

Is the riscv-config suggestion orthogonal to this, in which case it should be a separate pull request.

@eroom1966 

As discussed offline, this is making two changes, one to OVPsim, the other to add Ibex. It has highlighted a problem of a global linker script, instead of making this target platform specific.

I'll give Neela and you a short while to respond, but then I'm minded to approve and merge so we get Ibex support working.

neelgala:
riscv-ovpsim/config-yaml/requirements.txt: this file is being added as part of this pull-request. I suggest to edit it to use 1.0.2 for riscv-config.

eroom1966:
@neelgala 
The commit referencing riscv-config, is from @simon5656 code, I suggest we go ahead with this as it stands, and Simon can update later if he wishes

neelgala:
Okay. I am fine with that.

jstraus59:
Neel. We specifically fixed config-yaml to riscv-config version 1.0.0 as that is the one we tested. At some stage with new yaml features for CSRs we will move it forward.
Simon

PR #68, opened 2019-10-02
Update I-MISALIGN_JMP-01.S to reflect ratified CSR

AaronKel:
As per 3.1.17 Machine Trap Value (mtval) Register of the privileged spec:
"The mtval register replaces the mbadaddr register in the previous specification. In addition
to providing bad addresses, the register can now provide the bad instruction that triggered an
illegal instruction trap"


eroom1966:
@jeremybennett Agreed I am fine with this update as per the latest name.

PR #69, opened 2019-10-03
Ibex: Write instruction trace into log file

imphil:
To help debugging test runs, Ibex can now write an instruction trace
into a log file. Create one test-specific log file for each run.

Also create a .objdump-noalias file for each test which calls objdump
with arguments to produce a file closer to what you'd see in the log
files (no ABI names for registers, and no instruction aliases). Using
this file together with the Ibex-generated trace is helpful to debug
test cases.

Depends on https://github.com/lowRISC/ibex/pull/368

Closes lowRISC/ibex#217

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2019-10-07 Philipp Wagner <phw@lowrisc.org>
+    * When executing the test suite, Ibex always writes an instruction
+      log. Update the Makefile to write it to a test-specific location
+      (next to all other log files).
+    * On Ibex, provide an additional .objdump-noalias disassembly file
+      with no aliases and numeric register names (instead of ABI names).
+      This file matches the Ibex trace and can be used to debug the test
+      runs.
+
 2019-08-29 Robert Balas <balasr@iis.ee.ethz.ch>
     * Added support for using RI5CY as a target.
     * Added subdirectory riscv-target/ri5cy
--- a/riscv-target/ibex/device/rv32imc/Makefile.include
+++ b/riscv-target/ibex/device/rv32imc/Makefile.include
@@ -9,6 +9,7 @@ RUN_TARGET=\
     $(TARGET_SIM) \
         --term-after-cycles=100000 \
         --raminit=$(work_dir_isa)/$<.vmem \
+        +ibex_tracer_file_base=$(work_dir_isa)/$(*)_ibex_trace \
         > $(work_dir_isa)/$(*).stdout \
         2> $(work_dir_isa)/$@; \
         grep "^SIGNATURE: " $(work_dir_isa)/$(*).stdout | sed 's/SIGNATURE: 0x//' \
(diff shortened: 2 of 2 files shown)


jeremybennett:
Hi Philipp,

Thanks for the contribution. Some comments:
-  I don't think compliance tests should be generating trace output. I don't think any other compliance target does, and this seems to be beyond the role of a compliance test suite. If it does it should certainly be optional and off by default.
- we need a ChangeLog entry, which should also be in the commit message.
- should there be some documentation update to go with this change?

Let me know your thoughts.

imphil:
Having traces available was a request from @eroom1966 (Lee) in https://github.com/lowRISC/ibex/issues/217. It doesn't slow down the execution noticeably and writes files only to the work directory, so I thought it wouldn't hurt to have it unconditionally enabled. I can try adding a switch (it's a bit tricky given the current makefile structure).

Changelog and documentation are good points, I'll add them once we have agreement on how/if to implement this feature.

jeremybennett:
Ho hum. I'd like to get the wider group's view on this. The original idea was that compliance should only produce a yes/no result as output. I'll send an emal

eroom1966:
@imphil @jeremybennett The request I made was unrelated to the riscv-compliance setup
This was to enable the riscv-dv ongoing work to use verilator.
As Jeremy says for compliance - it is simply the signature output.
However, I have also made trace output an *optional* feature from riscvOVPsim, which is off by default. So I would think that a similar mode (as suggested by Jeremy), to have off by default, but can be enabled, would be fine.

imphil:
Asking the other way around: What do we loose by always having the trace around, like other log files, objdump intermediate representations, etc.? It doesn't hurt and makes debugging and development of tests easier.

I fully agree that this trace should not in any way considered for the result of the test run.

eroom1966:
@imphil 
Cannot disagree with your comment. I am not overly concerned to be honest, I presume verilator does not see an IO impact with the trace generation, due the the fact it is an RTL simulation and will be thousands of instructions rather than billions of instructions per second.

jeremybennett:
Let's see if there are any other feedback from the wider community.

Depending what you are doing, tracing can clobber Verilator performance, although in this case I can't see that matters since the compliance tests are tiny.

imphil:
Note that these traces are instruction traces (effectively a printf per instruction, rather cheap in simulation time), not signal traces (which slow down Verilator considerably). We run all verification which instruction traces enabled and haven't noticed any real slowdown.

I'll try to get a version running with optional tracing however.

imphil:
I updated the PR to include a changelog entry. 

In my previous comments, I forgot that the Ibex simulation actually always emits the trace. What this PR does it only move it to a sensible directory next to the other log files.

jeremybennett:
Thanks - I wasn't at the last meeting and was waiting for confirmation

PR #71, opened 2019-10-14
Added the ability to select a single test from a sute of test for

eroom1966:
Added the ability to pass a set of FLAGS to the target, via a command
line variable to make RISCV_TARGET_FLAGS

Added the ability to select a single test from a suite of tests for compilation and execution

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,15 @@
+2019-10-14 Lee Moore <moore@imperas.com>
+    * Added Ability to run a single test by using the Make Variable RISCV_TEST
+    for example, to only run the test I-ADD-01 from the rv32i suite
+        make RISCV_ISA=rv32i RISCV_TEST=I-ADD-01
+    * Added Top Level Variable to Makefile RISCV_TARGET_FLAGS, 
+    in the case of the RISCV_TARGET this can be passed and appended to the invocation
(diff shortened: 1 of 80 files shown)


jeremybennett:
Lee - can you get someone to review this. Perhaps Allen

jeremybennett:
BTW - will need to be rebased on the upstream.

allenjbaum:
I'd like to ensure that this is (or can be made) compatible with riscof

jstraus59:
Allen. 
What?

We cannot stop usability enhancements of the current framework while we start discussing future evolution. 

Riscof is not something we need to be compatible with. 

Riscof is something that _might_ fulfil our future requirements when we agree them. 

So, please lets not stop evolving the current framework. (You have heard of the Osborne effect?)

Thnx
Simon

allenjbaum:
It's a moot issue, since it should work. I certainly have no objection to
be having a method to select a specific test instead of the whole suite - I
think its a good idea.

jeremybennett:
Since this is reviewed, I am happy to merge once the conflicts are addressed.

neelgala:
Just to add.. RISCOF allows selecting of specific tests as well.

eroom1966:
Abandoning this pull request will re-submit

PR #73, opened 2019-11-05
Restructured RV32I to move Zicsr and Zifencei into their own suites

eroom1966:
Also added a note on the RISCV compliance WG
This is a first step to repartition the Zicsr and Zifencei instructions into their own suites
Further work needs doing in order to move the tests which cause exceptions


jeremybennett:
Merging, since this a approved.

PR #79, opened 2020-02-01
Ignore case when comparing signature outputs

Xiretza:
VHDL's `hwrite()` function produces values using uppercase hex literals, while the reference signatures are lowercase. Verifying the signatures would thus either require a separate lowercasing step, or the comparisong can simply be done in a case-insensitive manner.

### DIFF
--- a/riscv-test-env/verify.sh
+++ b/riscv-test-env/verify.sh
@@ -21,7 +21,7 @@ do
         echo    "Check $(printf %16s ${stub}) ... IGNORE"
         continue
     fi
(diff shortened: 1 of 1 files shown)


jeremybennett:
This looks a perfectly sensible extension to me. @Xiretza can you confirm that all existing results are unaffected?

If there are no contrary comments in the next 3 days, I'll approve and commit.

Xiretza:
What would be the best way to check that? I have tested with `make RISCV_TARGET=spike`, and all tests still pass.

Issue #70, opened 2019-10-08
Target error exit condition


towoe:
I noticed that the tests do not fail immediately if there is an error in `$(TARGET_SIM)`.
So for example in [Ibex](https://github.com/riscv/riscv-compliance/blob/master/riscv-target/ibex/device/rv32imc/Makefile.include#L13) the termination is done by `;` and not `&&`.
I was just wondering if this a feature so all the other tests are also executed.
If that is the case maybe this can be improved by using `&&` and ignore the error with `-` in the calling Makefile so it is more obvious in the `make` output.
Or maybe this is not the intended way. I am happy to create a PR but wanted to check before what the intentions here are.

allenjbaum:
I believe the intention is (or should be) to run (or attempt to run) all tests, reporting failures for each. I have seen riscof have this kind of behavior, and it makes it easier to see a class of problems that might be fixed all at once instead of one by one.
Please close this if this answers your issue,

towoe:
Following the discussion in #80 (v2 of framework is riscof, change for all cores to be avoided) closing this issue.

PR #80, opened 2020-02-04
Report error in Ibex run target

towoe:
Thanks for the feedback in #70 @allenjbaum 
I created this PR which reflects the description / improvement detailed in the issue.

The current output for an error in the execution of the Ibex target is
not reported. All tests are executed and at the verification stage a
difference in the output signatures will result in a failure of the test
case.
As no error of `RUN_TARGET` is produced, it is difficult to investigate
the actual problem.

This change will allow `make` to report an error but continue the
execution of the following tests.
As the output of a test case now shows that an error occurred, it is
easier to find the cause for a failed test if an error happened in the
execution of the target.

Closes riscv/riscv-compliance#70

### DIFF
--- a/riscv-target/ibex/device/rv32imc/Makefile.include
+++ b/riscv-target/ibex/device/rv32imc/Makefile.include
@@ -7,11 +7,12 @@ TARGET_SIM   ?= Vibex_riscv_compliance
 TARGET_FLAGS ?= $(RISCV_TARGET_FLAGS)
 
 RUN_TARGET=\
-    $(TARGET_SIM) $(TARGET_FLAGS) \
+    echo '' > $(work_dir_isa)/$(*).signature.output; \
+	$(TARGET_SIM) $(TARGET_FLAGS) \
         --term-after-cycles=100000 \
         --raminit=$(work_dir_isa)/$<.vmem \
         > $(work_dir_isa)/$(*).stdout \
-        2> $(work_dir_isa)/$@; \
+        2> $(work_dir_isa)/$@ && \
         grep "^SIGNATURE: " $(work_dir_isa)/$(*).stdout | sed 's/SIGNATURE: 0x//' \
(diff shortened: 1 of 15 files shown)


allenjbaum:
Ah, it was unclear what it was you wanted. This is the case where a test fails to run at all, and therefore no result is produced, no comparison against the reference signature is made, and therefore it appears to pass. Can you explain what causes a test to fail? A concern expressed at the meeting is that to be consistent between models, we would have to do this for every implementation, and we would prefer to avoid that. In addition, v.2 of the framework (riscof) does properly test for this case.

towoe:
The problem was that the simulation itself had an error. It was not even possible to start it, and as you said, no results are produced in this case and it was really not straightforward where the tests failed.
I guess this problem could also be caused for other cores if for example the path is not correctly set.
I agree that different implementations should have the same behavior, just wanted to show code so it can be discussed before changing too much.
I did not know that riscof is the successor to riscv-compliance, will have a look at it. And as this is not the most important fix I will close this PR.

allenjbaum:
I'm not sure (yet) this should have been closed. If a failure to run at all is not reported, then this is a bug that should be fixed. It shouldn't be fixed such that it fails immediately (as opposed to running all tests and reporting which failed for incorrect results and which failed to run at all. As I have been reminded, riscof is an example framework that (so far) is the only implementation of the latest test spec format, and the only example (so far) that actually compares two models instead of relying on in-line assertions. So, while these kinds of issues will not occur once (some v.2) framework is in place, if it is happening now, then some level of fix is warranted - but probably not one that halts the process after the first failure.

towoe:
Sorry for closing the PR, but from

> A concern expressed at the meeting is that to be consistent between models, we would have to do this for every implementation, and we would prefer to avoid that.

I thought the solution provided here is not the correct way forward.
I will create a new issue so a solution can be properly discussed.
If necessary I can then reopen this PR.

PR #82, opened 2020-02-06
Fix bad expected values in I-SRL-01 asserts

### LINKED ISSUE #81: Undefining `RVTEST_IO_QUIET` for OVPsim uncovers SRL test bugs
Commenting out [this line](https://github.com/riscv/riscv-compliance/blob/master/riscv-target/riscvOVPsim/compliance_io.h#L33) to enable more verbose test output, the `I-SRL-01` test in the `rv32i` suite fails an assertion:

```
Test Begin
# Test number 1
Assertion violation: file src/I-SRL-01.S, line 49: x3(0000007f) != 0xf0000000
```

Looking at the code for this test case:

```
TEST_RR_OP(srl, x3, x28, x13, 0xf0000000, 0x7ff, 0x4, x5, 12, x6)   # Testcase 3
```

This expands to:

```
li x28, 0x7ff
li x13, 0x4
srl x3, x28, x13
sw x3, 12(x5)
RVTEST_IO_ASSERT_GPR_EQ(x6, x3, 0xf0000000)
```

This should shift the value 0x7ff right by 0x4 bits, which should yield the value 0x7f, which is what we see `x3` contains in the assertion.

I discovered this while testing my own implementation, only to find out that OVPsim uncovers the same error under the same conditions. I think the expected value should be `0x7f`.

When commenting out lines in the test that fail, we see similar errors later:

```
Assertion violation: file src/I-SRL-01.S, line 60: x5(00000000) != 0x1000
Assertion violation: file src/I-SRL-01.S, line 61: x6(00000765) != 0x43210000
Assertion violation: file src/I-SRL-01.S, line 62: x7(00000000) != 0xfffffffe
Assertion violation: file src/I-SRL-01.S, line 74: x10(00000000) != 0x2468
Assertion violation: file src/I-SRL-01.S, line 75: x11(08000000) != 0x0
Assertion violation: file src/I-SRL-01.S, line 76: x12(00ffffed) != 0xcc000000
Assertion violation: file src/I-SRL-01.S, line 77: x13(00000001) != 0xfffffffe
Assertion violation: file src/I-SRL-01.S, line 78: x14(0000ffff) != 0xf8010000
Assertion violation: file src/I-SRL-01.S, line 92: x19(0000007f) != 0xf0000000
Assertion violation: file src/I-SRL-01.S, line 103: x21(00000000) != 0x1000
Assertion violation: file src/I-SRL-01.S, line 104: x22(00000765) != 0x43210000
Assertion violation: file src/I-SRL-01.S, line 105: x23(00000000) != 0xfffffffe
Assertion violation: file src/I-SRL-01.S, line 117: x26(00000000) != 0x2468
Assertion violation: file src/I-SRL-01.S, line 118: x27(08000000) != 0x0
Assertion violation: file src/I-SRL-01.S, line 119: x28(00ffffed) != 0xcc000000
Assertion violation: file src/I-SRL-01.S, line 120: x29(00000001) != 0xfffffffe
Assertion violation: file src/I-SRL-01.S, line 130: x30(0000ffff) != 0xf8010000
```

It seems the simulator is producing correct output in these cases, but the expected values in the test are wrong. This only triggers failures _without_ the [quiet macro defined](https://github.com/riscv/riscv-compliance/blob/master/riscv-target/riscvOVPsim/compliance_io.h#L33), as otherwise, the register values are written to the result area, but not asserted to contain the expected values shown here.

yupferris:
These were hidden due to the fact that #define RVTEST_IO_QUIET in compliance_io.h masks out all assertions. The test does appear to produce correct values and a correct signature, however the expected values in the assertions are wrong.

Fixes #81.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+2020-02-06 Jake Taylor <yupferris@gmail.com>
+    * [Issue #81] fixing rv32i/I-SRL-01.S test
+
 2019-11-05 Lee Moore <moore@imperas.com>
     * Restructured RV32I to move Zicsr and Zifencei into their own suites
 
--- a/riscv-test-suite/rv32i/src/I-SRL-01.S
+++ b/riscv-test-suite/rv32i/src/I-SRL-01.S
@@ -46,7 +46,7 @@ RV_COMPLIANCE_CODE_BEGIN
 	TEST_RR_OP(srl, x0, x31, x16, 0x0, -0x1, 0x0, x5, 0, x6)   # Testcase 0
 	TEST_RR_OP(srl, x1, x30, x15, 0x1, 0x1, 0x0, x5, 4, x6)   # Testcase 1
 	TEST_RR_OP(srl, x2, x29, x14, 0x0, 0x0, -0x1, x5, 8, x6)   # Testcase 2
-	TEST_RR_OP(srl, x3, x28, x13, 0xf0000000, 0x7ff, 0x4, x5, 12, x6)   # Testcase 3
+	TEST_RR_OP(srl, x3, x28, x13, 0x7f, 0x7ff, 0x4, x5, 12, x6)   # Testcase 3
 	TEST_RR_OP(srl, x4, x27, x12, 0x0, 0x0, 0x8, x5, 16, x6)   # Testcase 4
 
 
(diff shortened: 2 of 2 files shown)


allenjbaum:
Frankly, many of the assertions are wrong, when run on RV64.

-Allen

yupferris:
> Frankly, many of the assertions are wrong, when run on RV64.

To clarify, this is the only test with bad assertions for rv32i (the issue is explicit about this but I forgot to mention that in this PR, I can fix that if desired).

eroom1966:
@yupferris 
Thanks for bringing this to our attention
We have a fix for this, which is more comprehensive, because it also needs to handle souble and single word load stores in the PUSH POP macros.
I will try to get this committed today
Lee

yupferris:
Fantastic! Feel free to close this when that's up. I look forward to the fixes.

yupferris:
Fixed in #85 instead

PR #91, opened 2020-02-17
Add Support for Parallel make

Imperas:
As the test suites increase in size, the option to perform parallel make becomes more important
This PR includes the ability to run the tests in parallel, if resources are available


jeremybennett:
Agree it's needed. A couple of comments

1. This seems more than just adding parallel make. Could there be an explanation of what the extra changes are - a more detailed commit message would suffice. If there are multiple changes being made, the should be split into multiple pull requests.

2. This seems to be hardcoding either `-j 8` or `-j 4` for the parallelism. It would be better to choose a value based on the machine capacity (our nightly testing is typically on 40-48 core Xeons). `-j <ncpus>` is commonly used. Providing a user variable to overrride this would then be useful.

HTH

Imperas:
Sure some additional comments:

The JOBS variable can be supplied via the command line, the default when running riscvOVPsim is -j8 --max-load=4, but equally the JOBS variable can be command line set to Make, so for example

$ make PARALLEL=1 JOBS="-j64"

Also this currently will only parallelize riscvOVPsim, this is because the other targets currently share a temporary file during execution, which means that each process stamps on the same file - this will need to be changed for the other targets, if parallel is to be enabled.

As @simon5656 mentions, this is critical to supporting the vector tests, the suite is huge in comparison to anything that has been produced up until this point

Thx
Lee

jeremybennett:
@eroom1966 @Imperas 

This makes things clear. I notice that one thing missing is any documentation. Could you update `doc/README.adoc` with instructions on how to run parallel make, including use of the `JOBS` paremeter.  Then I should be happy to merge.

Thanks,


Jeremy

Imperas:
> 
> 
> @eroom1966 @Imperas
> 
> This makes things clear. I notice that one thing missing is any documentation. Could you update `doc/README.adoc` with instructions on how to run parallel make, including use of the `JOBS` paremeter. Then I should be happy to merge.
> 
> Thanks,
> 
> Jeremy

Additional description added
Thx
Lee

jeremybennett:
Looks good to me. Thanks

byllgrim:
Hi.
Is the intended use of makefiles now to use `$(@)` instead of `$(work_dir_isa)/$@`?
And also to use `$(1)` instead of `$(2)`?

PR #92, opened 2020-02-18
Fix Ibex target following parallel changes

tomroberts-lowrisc:
- Update Ibex Makefile to match changes made for parallel make
- Ibex CI is currently failing without this fix


eroom1966:
Approved

PR #104, opened 2020-04-08
fix issue with test writing signature data outside of data_being/end area

allenjbaum:
This adds a padding word before the first signature label, and prefixes the reference signature with the expected value


byllgrim:
I think(?) this works for me at least.
Previously the simulation would loop endlessly and time out.
But with your fix all tests run fine.

_Except_ my core fails the new tests :upside_down_face: so 1) I have some work to do on my own, and 2) I can only be a fragment in the vow to this pull request's integrity.

neelgala:
@allenjbaum 
Couple of changes required for the test to pass:

1. You have padded 5 regions in the test but updated only a single line in the reference-signature. There need to be 4 more lines at the top with `ffffffff`
2. Please also update the changlog

allenjbaum:
Which new tests are failing?

allenjbaum:
No, I padded only test1 with a single word - that's the minimum amount that
is needed.
All the rest are unpadded - that just means testN writes into the signature
of testN-1 at offset 16 -- but that is OK, since test N-1 never touches
anything beyond offset 7.
Changelog now updated, added, committed.

byllgrim:
> You have padded 5 regions in the test but updated only a single line in the reference-signature. There need to be 4 more lines at the top with ffffffff

I think @neelgala might be right about this.
At least that's what my output looks like.

neelgala:
missed checking the replies on this post.

@allenjbaum 
the following line you have added in the two tests: I-SB-01.S and I-SH-01.S 
```
	.fill 5, 4, -1   //padded for neg offset
```
means the signature now includes 5 more lines than the previous version of the test.
However, in your PR you have updated th signatures with just one line:
```
ff01ffff
```
it should rather be:
```
ffffffff
ffffffff
ffffffff
ffffffff
ff01ffff
```

Once you fix this I could re-run spike and riscvovpsim as targets

allenjbaum:
That's odd; my local versions have

	.fill 1, 4, -1   //padded for neg offset

and I thought I made all the changes at the same time. I think this is
fixed now

allenjbaum:
I'm not sure how it got there, but it should be fixed now.

allenjbaum:
Sigh- I thought I had made that change.

-Allen

allenjbaum:
change has been made

eroom1966:
@allenjbaum @neelgala Looks like it has conflicts
**This branch cannot be rebased due to conflicts**

neelgala:
I usually squash and merge. But I am still waiting for Allen to update the changelog.

PR #127, opened 2020-07-17
fixed coverage usage

duncangraham-Imperas:


simon5656:
two commits - one for riscv-ovpsim coverage and second to clarify license info - so ok.

PR #128, opened 2020-07-29
grift: update for new directory structure

nmeum:
The directory structure was changed in 5a978cfd444d5e640150d46703deda99057b2bbb. However, as stated in
this commit it was only tested with riscvOVPsim and spike. Unfortunately, the commit broke the grift target as paths weren't adjusted accordingly for it. This commit updates the paths accordingly.

### DIFF
--- a/riscv-target/grift/device/rv32i/Makefile.include
+++ b/riscv-target/grift/device/rv32i/Makefile.include
@@ -8,7 +8,7 @@ RUN_TARGET=\
     $(TARGET_SIM) $(TARGET_FLAGS) --arch=RV32I \
         --mem-dump-begin=begin_signature --mem-dump-end=end_signature \
 	--halt-pc=grift_stop_addr \
-        $(work_dir_isa)/$< > $(work_dir_isa)/$(*).signature.output 2> $(work_dir_isa)/$@;
+        $< > $(*).signature.output 2> $@;
 
 RISCV_PREFIX   ?= riscv32-unknown-elf-
 RISCV_GCC      ?= $(RISCV_PREFIX)gcc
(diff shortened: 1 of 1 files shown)


MarcKarasek (approved):

riscv-target/grift/device/rv32i/Makefile.include
```diff
@@ -8,18 +8,18 @@ RUN_TARGET=\
...
 
 COMPILE_TARGET=\
-	$$(RISCV_GCC) $(2) $$(RISCV_GCC_OPTS) \
+	$$(RISCV_GCC) $(1) $$(RISCV_GCC_OPTS) \
 		-I$(ROOTDIR)/riscv-test-env/ \
 		-I$(ROOTDIR)/riscv-test-env/p/ \
 		-I$(TARGETDIR)/$(RISCV_TARGET)/ \
-		-T$(ROOTDIR)/riscv-test-env/p/link.ld $$< \
-		-o $(work_dir_isa)/$$@; \
-	$$(RISCV_OBJDUMP) -D $(work_dir_isa)/$$@ > $(work_dir_isa)/$$@.objdump
```
MarcKarasek, line 25: i think you missed one of the $$@ that should be here.  The first one in the line will only be $@ instead of $$@

MarcKarasek (changes requested):
See inline comments

riscv-target/grift/device/rv32i/Makefile.include
nmeum: Could you elaborate a bit? Which exact line you are referring to? I am a bit confused as there are a lot of `$$@` in this diff :sweat_smile:

allenjbaum: MArc is pointing at he first $$@, which is at the end of line 24, and it should be just $@ if I'm interpreting his comments correctly.
IF this is correct Marc, could you comment? 
And please update this and close this if Marc agrees.

**Note** that the entire riscv-target/grift directory should be moved back into the grift home repo as we will be removing all target directories  at some point in the (near) future, replacing it with a pointer to your repo (preferrable as a top level arch_test_target directory to avoid this kind of issue entirely. 
See github.com/riscv/riscv-arch-test/riscv-target/spike/ for what will remain
and github.com/riscv/riscv-isa-sim/arch_test_target/spike/README.md as example documentation

allenjbaum:
This needs to be resolved, or I'm going to close it without the PR being merged

benjaminselfridge:
I think it's safe to just remove support for grift -- grift was dormant for a bit but is now going through some significant updates. When those are done, I will circle back and see what needs to be done to add support for the compliance framework. Until then, I suggest you don't worry about it. :)

allenjbaum:
This is not being actively worked on, so removing by request.

PR #129, opened 2020-07-31
sail-riscv-ocaml: Disable RVC extension on all devices not using it

nmeum:
Especially when using the `rv32i` device it might be unexpected that compressed instructions are supported (i.e. 2-byte aligned PCs).

### DIFF
--- a/riscv-target/sail-riscv-ocaml/device/rv32Zifencei/Makefile.include
+++ b/riscv-target/sail-riscv-ocaml/device/rv32Zifencei/Makefile.include
@@ -6,7 +6,7 @@ ifeq ($(shell command -v $(TARGET_SIM) 2> /dev/null),)
 endif
 
 RUN_TARGET=\
-    $(TARGET_SIM) $(TARGET_FLAGS) -isa=rv32 \
+    $(TARGET_SIM) $(TARGET_FLAGS) -isa=rv32 -disable-rvc \
         -test-signature=$(*).signature.output \
         $(<) 2> $@
 
--- a/riscv-target/sail-riscv-ocaml/device/rv32i/Makefile.include
+++ b/riscv-target/sail-riscv-ocaml/device/rv32i/Makefile.include
@@ -6,7 +6,7 @@ ifeq ($(shell command -v $(TARGET_SIM) 2> /dev/null),)
 endif
 
 RUN_TARGET=\
-    $(TARGET_SIM) $(TARGET_FLAGS) -isa=rv32 \
+    $(TARGET_SIM) $(TARGET_FLAGS) -isa=rv32 -disable-rvc \
         -test-signature=$(*).signature.output \
         $(<) 2> $@
 
--- a/riscv-target/sail-riscv-ocaml/device/rv32im/Makefile.include
+++ b/riscv-target/sail-riscv-ocaml/device/rv32im/Makefile.include
@@ -6,7 +6,7 @@ ifeq ($(shell command -v $(TARGET_SIM) 2> /dev/null),)
 endif
 
 RUN_TARGET=\
-    $(TARGET_SIM) $(TARGET_FLAGS) -isa=rv32 \
+    $(TARGET_SIM) $(TARGET_FLAGS) -isa=rv32 -disable-rvc \
         -test-signature=$(*).signature.output \
         $(<) 2> $@
 


nmeum:
Maybe also pass `--disable-compressed` for `sail-riscv-c`?

allenjbaum:
Can't you simply use the Makefile.include that builds RV32IM instead of RV32IMC?
Are you saying that the Sail-OCAML model cannot run C-extension code?

nmeum:
>  Are you saying that the Sail-OCAML model cannot run C-extension code?

No.

> Can't you simply use the Makefile.include that builds RV32IM instead of RV32IMC?

I am not sure if I understand your question correctly. The problem here is the following: The `RV32IM` and `RV32I` device do not include the `C` extension. As such, I would expect a simulator for these devices to not allow compressed instructions. This is relevant for compliance tests which test the behavior of unaligned jumps, with the C extension a PC can be 2-byte aligned without it it must be 4-byte aligned (IIRC). For Sail the C extension is enabled by default and needs to be explicitly deactivated for the `RV32I` and `RV32IM` device, otherwise sails behavior will differ (compared to other simulators for this device) when considering unaligned PCs.

allenjbaum:
OK, I was fooled by the reference to the Sail-Ocaml target. 
I think this fails on any model that actually supports the C-extension and doesn't (I think) implement HW Ld/St misaligned handling. 

The test framework doesn't do anything special if the model supports C, but you elect not to test the C-extension, except it won't select C-extension tests.  The misalign jmp test isn't specific to C-extension, so it is run - but C-extension is present in the model (even if not tested) and tests are written assuming misalign support, so those jumps won't trap.

The existing framework knows only a single configuration of harts (specifically misalign support, which can only be discovered by trying it a seeing if it traps). Framework 2.0- will fix this problem by configuring the SAIL model to the configuration of the device under test, using a configuration file in YAML format. The SAIL model model will be updated to configure itself based on that configuration file. That work is just beginning.

allenjbaum:
I believe I understand what is going on; it is a known problem that is not
fixable in the current framework, explained in github comment.

allenjbaum:
Re-explaining this: there are explicit tests for misaligned execution that should give different results depending on either the C-extension is implemented (should execute) or not (should trap). The existing framework only has a single reference signature, so can only tests a single kind of DUT. This will be fixed  in a short while when we cut over to the riscof framework .
In order to make this work properly, you will need to creat a riscv-config YAML description of your DUT.
See the README regarding riscof here:  https://riscof.readthedocs.io/en/stable/
I will close this after I have confirmation that this has been tests on an RVI and RVIC and both work.

allenjbaum:
Since target directories are no longer used, this will be closed on Aug 01 unless it can be shown there are still failures when run under the new framework with the proper YAML configuration file (that is, with C extension enabled or not).

This raises an interesting question, however.
Normally it is not necessary to disable any extension that is not supported, since it is legal to implement them even if the support not declared (e.g. if just a few of the ops are supported, but others aren't). The compatibility tests just avoid testing C-extension ops in that case.  But if a custom C-extension is implemented (e.g. only a partial implementation, or some different ops), then branching to a halfword aligned address won't trap.  I'm going to see if I can get clarity about whether we need to explicitly declare (in the YAML) that halfword unaligned addresses won't trap when C-extension isn't implemented.

allenjbaum:
I've confirmed that misaligned accesses to a halfword address is not dependent on the C-extension, but instead on IALIGN=16 (which is required if C-extension is implemented and enabled)

I would expect Sail to set IALIGN to 16 only when C-extension is implemented and enabled (misac.C==1)
If not, that is a Sail bug (not that IALIGN==16 at reset if C-extension is enabled).

But, that precludes an implementation that has their own custom compressed extension.
ACTs won't check custom ops, but we will check that 2B aligned inst. fetches execute properly.
That means that the YAML must have an IALIGN configuration parameter to force IALIGN=16 for implementations that have custom compressed extensions 
If that YAML force IALIGN==16, we can't disable it though, because that's a custom instruction.
Theoretically, you could implement custom compressed ops AND the C-extension, 
in which case disabling the C-extension would still allow 2B aligned instruction fetches even when C-extension is disabled.

pawks:
This behaviour can also pop up in-case of custom 48 bit instructions without the C extension. SAIL will need to take IALIGN as a CLI argument to address this correctly. The IALIGN in the yaml should also be able to capture whether it is possible to change IALIGN at runtime. So it should probably be a list.

allenjbaum:
I don't think we need to worry about changing IALIGN at runtime.
The op to do that would be custom, and we won't test custom op[s in the
ACTs.
When we get ratified 48bit ops, then Sail will know how to deal with it, so
no list should be necessary.
Otherwise, it is custom brhavior we shouldn't test.

allenjbaum:
We are now using riscof, which doesn't use Makefile.inclue files, so closing this PR.

PR #134, opened 2020-08-19
Update of Coverage and OpenHW/CV32E40P as a target

eroom1966:


simon5656 (approved):
this adds OpenHW Groups RTL core as target and fixes some coverage reporting options

PR #140, opened 2020-10-28
xori on 64 bit compiler needs neg number

MarcKarasek:
Compiling with 64 bit compiler does not
see a 32 bit number as negative.  Need to
implicitly set this as a negative number so
both 32 and 64 bit compilers behave properly.


neelgala (approved):
works with spike.

allenjbaum:
Thanks!

PR #152, opened 2020-12-03
RI5CY: Fix assembly for clang

### LINKED ISSUE #108: RI5CY's `compliance_io.h` fails to compile with clang
Errors with
```
riscv32-unknown-elf-clang --target=riscv32-unknown-elf -march=rv32i -c
-o I-ENDIANESS-01.o -I ../../../riscv-test-env -I
../../../riscv-test-env/p -I ../../../riscv-target/ri5cy I-ENDIANESS-01.S

../../../riscv-target/ri5cy/compliance_io.h:187:12: error: operand must
be a bare symbol name
    la t3, 0x10000000; sw a0, (0)(t3);;
```
because strictly speaking `la` doesn't accept a constant.

See also https://github.com/riscv/riscv-compliance/issues/105

bluewww:
Older clang versions can't handle it
clang-10.1 works with it and gcc doesn't complain either.

Fixes #108

### DIFF
--- a/riscv-target/ri5cy/compliance_io.h
+++ b/riscv-target/ri5cy/compliance_io.h
@@ -89,7 +89,7 @@
     jal         FN_WriteA0;						\
 
 #define LOCAL_IO_PUTC(_R)                                               \
-    la          t3, RVTEST_CUSTOM1;					\
+    li          t3, RVTEST_CUSTOM1;					\
     sw          _R, (0)(t3);						\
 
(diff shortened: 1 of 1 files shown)


MarcKarasek (commented):
Instead of changing this we should instead say that we support clang version x or greater.

bluewww:
Ok in that case I withdraw this pr for #155

PR #158, opened 2020-12-17
Bump to v2.1

neelgala:
ChangeLog:

* remove env folder symlinks from all riscv-test-suite src folders 
* fixed assertion macros for ovpsim
* renamed RVTEST_ASSERT to RVMODEL_ASSERT in the Makefile and ovpsim macros
* tests updated with right set of "correctvals"


MarcKarasek (approved):
Looks good

PR #161, opened 2020-12-21
Update for support of riscvOVPsim

duncangraham-Imperas:
Tidied Makefile.include so it is consistent
Re-integrated the 64-bit device targets for riscvOVPsim
Add a postverify target to allow target specific post processing of data

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,8 +89,12 @@ simulate:
 verify:
 	riscv-test-env/verify.sh
 
-cover:
-	riscv-test-env/cover.sh
+postverify:
+ifeq ($(wildcard riscv-target/$(RISCV_TARGET)/postverify.sh),)
+	$(info No post verify script found riscv-target/$(RISCV_TARGET)/postverify.sh)
+else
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -89,8 +89,12 @@ simulate:
 verify:
 	riscv-test-env/verify.sh
 
-cover:
```
MarcKarasek, line 92: why did u remove the cover: target?

MarcKarasek (commented):
See inline comment

Makefile
MarcKarasek: If we remove this and add the postverify the documentation should also be changed/added to

neelgala: The cover target was ovpsim specific to start with. I think moving it to the target makes sense.

Imperas: Mark - I think Neel's comment covers it - so can this pull request please be merged. Thanks. Simon

MarcKarasek (approved):

PR #166, opened 2021-01-15
Deduplicate makefiles

MarcKarasek:
All makefiles in riscv-test-suite were 99% the same file copy-pasted
again and again. Separate out the common parts into a Makefile.include
and include that from all the makefiles instead.


MarcKarasek:
Neel can you review this.   I pulled this one commit from Xiretza 3 commits.

This is a good idea

neelgala:
This won't work.. you have removed all the makefiles but I don't see a riscv-test-file/Makefile.include in this commit. I think #163 is good enough to merge..

MarcKarasek:
I do not want the other two commits in #163

MarcKarasek:
This is weird..  I see the Makefile.include in my forked repo. 
It just is not in the PR????

MarcKarasek:
The Makefile.include is now there... 

Have no idea why it got left out.  It is in the patch file I applied to my fork??  Oh well

neelgala:
you will need to update the Changelog and Also the change from #163 in the root level Makefile is worth taking it. it forces verify to occur after simulate is done. Not having that patch is currently breaking things on the master when run in parallel. 

Which essentially will makes this PR the same #163. You can squash the commits on the other one if required. I don't see how this PR is any different from #163 at the end

Xiretza (changes requested):
It seems the authorship information on my commits has been lost. I'm honestly not sure how this happened, but using `git cherry-pick` to pick my commits into a fresh branch, and optinally `git commit --amend` to change them, should fix it.

MarcKarasek:
This will be done in another PR

PR #167, opened 2021-01-16
Fix NARG macro.

pawks:
This pull request fixes the NARG macro definition. The previous definition was always evaluated to `length+1` instead of `length` due to a missing argument in the instantiation of the `_ARG5` macro.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,6 @@
+2021-01-16 S Pawan Kumar <spawan1999@gmail.com>
+  	* Fixed NARGS macro defintion to work correctly. 
+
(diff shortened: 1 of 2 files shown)


neelgala:
update Changelog

neelgala (approved):

neelgala:
@MarcKarasek if you agree.. please merge this.

pawks:
@MarcKarasek There was a bug in the implementation which caused the `NARG` macro to be evaluated to 1 even when the argument list was empty. I have fixed it in the latest commit. I have verified that both `sail` and `spike` work.

neelgala:
shouldn't you be creating a new tag for v2.1.1 with this merged ?

PR #183, opened 2021-04-10
commented out addition of K-ext target test

allenjbaum:


neelgala:
I doubt this is necessary. The Makefile chooses suites based on the target's device directory structure. Keeping this uncommented shuold not affect devices who don't have K support.

ben-marshall:
Yeah please don't get rid of this! It makes it unnecessarily hard to demo working K tests to people on things like Spike/Sail/OVPSim and cores which support the extension.

allenjbaum:
OK, then - reject it - no problem.

PR #184, opened 2021-04-15
updating http reference for condstr

dansmathers:
is this the correct link? https://riscof.readthedocs.io/en/latest/testformat.html#rvtest-case-condition-formating
 
https://riscof.readthedocs.io/en/latest/cond_spec.html#cond-spec is not found


allenjbaum (approved):
This is the correct link, 
I expect that this will all be merged into the arch-test repo instead of the riscof readthedocs repo when we switch over to riscof. 
When we do that, let's please fix "formating"  to "formatting" to triggering my OCD.

allenjbaum:
Yes, this broken in the original doc as well, and you have the correct link

https://riscof.readthedocs.io/en/latest/testformat.html#rvtest-case-condition-formating

neelgala:
@dansmathers thanks for fixing this. Can you create a CHANGELOG.md entry for 2.4.2 . After which I can merge this.

@allenjbaum Yes I'll make the change as part of the transition to RISCOF.

PR #185, opened 2021-04-16
changelog update for pull #184 fixing http path

dansmathers:
changelog update for pull #184 fixing http path.


neelgala:
I think you mistakenly created a new PR instead of pushing to the old one #184 . Closing this one.

PR #187, opened 2021-04-21
issue107 fix

allenjbaum:
This should fix the renamed SPTBR CSR to be SATP and fix problems with anyone that has updated and is using these files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## [2.4.2] - 2021-04-20
+  - changed all occurances of SPTBR to the new name SATP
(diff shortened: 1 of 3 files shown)


neelgala:
Please create an entry in the CHANGELOG.md for 2.4.2 and i'll merge it.

allenjbaum:
I think I did - but not sure how to merge the update to the old pull
request - only know how to create a new one.
Although, it looks like it is done; I expected the pull request to have a
different comment, but I see the second commit in there.

Something is very wonky with my git setup, because things that used to work
don't seem to any more.

neelgala (approved):

PR #217, opened 2021-10-14
fix the lower case `i` in the `RVTEST_CASE` macros used in the shift …

neelgala:
fix the lower case `i` in the `RVTEST_CASE` macros used in the shift operation tests.


MarcKarasek (commented):
Looks good

PR #225, opened 2021-12-08
K ext: up-to-date the K extension for the V.1.0.0 ratified spec

phthinh:
This PR makes some small changes for K extended instruction set so that it is up-to-date with the ratified v.1.0.0 Scalar cryptography extension spec. Particularly, this PR changes the names of xperm.n -> xperm4 and xperm.b -> xperm8 instructions and remove an unsupported instruction packu. It also added the K extension support for the ibex target.


allenjbaum:
Not that support for individual targets will be removed in V3 of the framework riscof) and you should be prepared for that eventuality.

phthinh:
Thanks a lot for the note @allenjbaum! is it means the riscv-arch-test repo will be superseded by the riscof repo?
Actually, I am testing an implementation of the K extension for the ibex core. I found the riscv-arch-test (previously know riscv-compliance test) is useful and supports the K_unratified spec. It needs some small changes to up-to-date with the ratified spec sothat I can use to test my implementation. I though it would be good to update it here too. 
I looks forward to using the riscof framework.

neelgala:
I will need an entry in the CHANGELOG.md file for a major-version update - 2.7.0.

phthinh:
I have done the update for rv63i_m too. And I added an entry in the CHANGELOG.md file for the update. It's ready for review.

allenjbaum:
The tests in the current arch-test repo are the same tests that riscof will
use; riscof is the framework, and that framework doesn't use a target
directory.
It does use a python plugin, and there are definitions of those for various
models (including Sail and Spike) in the repo.

allenjbaum:
This looks good to go. For both rv32 and rv64, it
  - renames the K_unratified directory files to be simply K
  -updates the makefile to point at the new directory name
 - removes both .S and reference output for packu and packuw
 - renames xperm.b and .n mnemonics to be xperm.8 and .4 in the .S files
 - It also adds the K directory to IBEX target directory

neelgala (approved):

PR #228, opened 2022-01-24
Bitmanip test suite, stats

anku-anand:
Added test suites, coverage report, dpr for instructions in bitmanip sepc.


riscv-test-suite/rv32i_m/Bitmanip/Makefile
```diff
@@ -0,0 +1,3 @@
+include ../../Makefile.include
+
+$(eval $(call compile_template,-march=rv32ib -mabi=ilp32 -DXLEN=$(XLEN)))
```
pawks: The march `rv32ib` isn't supported on the gnu-toolchain yet. The march should be based on the subextensions.

riscv-test-suite/rv64i_m/Bitmanip/Makefile
```diff
@@ -0,0 +1,3 @@
+include ../../Makefile.include
+
+$(eval $(call compile_template,-march=rv64im -mabi=lp64 -DXLEN=$(XLEN)))
```
pawks: The march string is wrong here.

pawks (changes requested):

riscv-test-suite/rv64i_m/Bitmanip/Makefile
anku-anand: Hi
Updated the Makefile  with  -march=rv32i_zba_zbb_zbc_zbs for RV32 and -march=rv64im_zba_zbb_zbc_zbs  for RV64

Thanks.

anku-anand:
Updated cgf based on review comments, updated src, signature and coverage.html

neelgala:
@allenjbaum let me know if the coverpoints and tests are satisfactory so I can merge this.

allenjbaum:
you overlooked RV64 ANDN uniform_random parameter is different than (XNOR,ORN) 
 I'm not sure its worth going through all the trouble to fix

allenjbaum (approved):
I don't think it's worth quibbling about ANDN having few (I think) random values than XNOR or ORN, so this is good to go from my point of view.

fatimasaleem:
Hi @anku-anand, when this PR is going to be merged? cc @allenjbaum @pawks

ZenithalHourlyRate:
Some note on the "B" folder name: https://github.com/riscv/riscv-opcodes/pull/101#issuecomment-1038067612
> there's still no such thing as "the B extension".

I wonder why it changed from "Bitmanip" to "B". I think maybe "Zb" is more appropriate?

allenjbaum:
There is no B extension, as you say. There is Aba, Zbb, Zbk*, etc.
Tests for all of these are in B folder... except when they're not.
(e.g.some zbk* extensions will be in K folder, because the ops in those
extensions were never ratified as part of K)
And, of course, There is no K extension, There is Zkn, Zks, Zone, Zknd,
Zkt, etc
Those tets are in the K folder.
The riscof framework doesn't care much which folder a tests is in; tests
are labelled and filtered according to the extension name.

Moral: start using riscof. We officially deprecate the current framework
and make riscof the main branch at the end of the month.

ZenithalHourlyRate:
Thanks for your kind explanation!

> The riscof framework doesn't care much which folder a tests is in; tests are labelled and filtered according to the extension name.

Yes I understand that. My point is that a more clear folder name will cause less confusion and have more consistency with the standard.

> Moral: start using riscof. We officially deprecate the current framework and make riscof the main branch at the end of the month.

Does it mean this PR is stucked by this transition? (i.e. instead of PR on master, the author should PR on riscof-dev)

allenjbaum:
I'll let Anand comment on whether this PR is being held up by directory naming conventions. 
I expect that the riscof-dev branch is keeping up with this one way or another.
The intent is that the framework changes, but the tests don't so if it works with the current framework, regardless of which directory a test is in, it should work with riscof as well.

M-Ijaz-10x:
Hi @anku-anand,
I was going throught the clmul.S test file and I found that some of the instructions have wrong values of operand_a and operand_b in comments, like instruction number 2 and 3, it needs to be fixed before it gets merged into riscv-arch-test master
Thanks.

neelgala:
The RVTEST_ISA and the RVTEST_CASE strings are still wrong here. For example, based on the ratified spec, there is no extension `K` and `B`. A simple grep and check in the [riscv-opcodes](https://github.com/riscv/riscv-opcodes) repo will help you figure out which extensions apply to which instruction. 

For example consider the `andn` instruction. Run the command `grep "andn" rv*` in riscv-opcodes and you will get the following result: 
```
rv_v:vmandn.mm      31..26=0x18 vm vs2 vs1 14..12=0x2 vd 6..0=0x57
rv_v_aliases:$pseudo_op rv_v::vmandn.mm vmandnot.mm  31..26=0x18 vm vs2 vs1 14..12=0x2 vd 6..0=0x57
rv_zbb:andn       rd rs1 rs2 31..25=32 14..12=7 6..2=0x0C 1..0=3
rv_zbkb:$import rv_zbb::andn
rv_zk:$import rv_zbb::andn
rv_zkn:$import rv_zbb::andn
rv_zks:$import rv_zbb::andn
```

Which shows that `andn` is part of Zbb, Zbkb, Zk, Zkn and Zks which is not captured in the RVTEST_ISA and RVTEST_CASE strings of the andn.S file in your PR.

More importantly these changes must be incorporated in RISCV-CTG (I am guessing [this PR](https://github.com/riscv-software-src/riscv-ctg/pull/36))

Once these changes are done, please ping here for another review.

anku-anand:
Updated ISA strings in yaml files and generated .S files.
CGF and template.yaml were also updated in riscv-ctg (https://github.com/riscv-software-src/riscv-ctg/pull/36)

neelgala:
@anku-anand will need an entry in CHANGELOG.md for 3.3.0 version bump. Will merge once that is pushed.

pawks (approved):

anku-anand:
@neelgala resolved all conflicts, please review and merge the PR

neelgala:
as per the [versioning policy](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/CONTRIBUTION.md#versioning) the changelog version will need to be 3.4.0

Issue #227, opened 2022-01-17
Makefile for rv32e_unratified


davidharrishmc:
It appears that the riscv-arch-test Makefile is only designed to run for rv32i_m and rv64i_m.  For example, SUITEDIR has i_m hardwired into the name. Will it be extended in the near future to also build rv32e_unratified in a standardized fashion?

allenjbaum:
RV32E has not been ratified yet; it certainly will at that point. But, I
don't think we should standardize anything that hasn't been ratified.
We may want to talk about what we can do to make unratified tests more
accessible.

davidharrishmc:
I built the E tests.  There are a number of issues to report.

1) The signature.output and reference_output files differ by 1-3 final 00000000 entries for most of the tests.  I had to manually change the reference to match the signature.

2) To get the E tests to compile, I had to make the following changes:

Add #define RVTEST_E to the start of each file so it doesn’t try to initialize registers x16-x31
Modify riscv-isa-sim/arch_test_target/spike/model_test.h “sw x1, tohost , t5” to “sw x1, tohost , t1” to use a RV32E register for the HTIF write.

3) The whole process was a major pain because of the way that the makefiles tie the riscv-arch-test directories to the riscv-isa-sim directories.  I had to make some substantial hacks. 

David Harris

neelgala:
For issue1: signatures must start and end at a 16-byte boundary [as per spec](https://github.com/riscv-non-isa/riscv-arch-test/blob/master/spec/TestFormatSpec.adoc#36-the-test-signature). Those unwanted unwanted regions need to be initialized with 0s. To fix this make sure to add the `.align 4` before the begin/end signature labels as is done here [in spike](https://github.com/riscv-software-src/riscv-isa-sim/blob/master/arch_test_target/spike/model_test.h#L32-L36)

For issue2: I think this has to be fixed and probably the best solution is to add -DRVTEST_E at the end of [this line](https://github.com/riscv-non-isa/riscv-arch-test/blob/307c77b26e070ae85ffea665ad9b642b40e33c86/riscv-test-suite/rv32e_unratified/E/Makefile#L3) and all other Makefile withins the rv32e_unratified director. If you can try this fix and let me know it works, I'll put up a quick PR.

[...]

PR #234, opened 2022-02-24
Misc fixes

### LINKED ISSUE #214: Test Format Spec doesn't specify the order of line in the signature file
That should be clarified to be: 
  in the order it is stored in the signature region, from lowest address to highest (from  RVMODEL_DATA_BEGIN to RVMODEL_DATA_END)

### LINKED ISSUE #230: comments in reference_output
As we are writing custom tests by hand, it's challenging to keep track of the meaning of the values in the reference_output.  We decided to add comments delimited by # in the reference_output.  We modified one line of the verify.sh script to ignore these comments when running a diff against signature.output, and think this feature might be helpful to other people.

verify.sh
    diff --ignore-case --ignore-trailing-space --strip-trailing-cr <(grep -o '^[^#]*' ${ref}) ${sig} &> /dev/null

neelgala:
- modified verify.sh to ignore comments in reference signature during diff operation [close #230]
 - udpated test-format spec to include the order of lines in the signature file [close #214]
 - RVTEST_E macro to be enabled for all rv32E tests. [#227]
 - Added a disclaimer in README about migration to RISCOF


pawks (approved):

neelgala:
@allenjbaum added disclaimer in this PR. Please let me know if any changes required.

allenjbaum (commented):
Didn't we just add an option for outputting 64bits (16 character) per line?
I think that needs to be documented here, along with the flag that enables it.

neelgala:
@allenjbaum no we don't have any such option/support where the signature can be 64bits per line.

allenjbaum:
Um, it was just added to Sail by Bilal. See PR#136

neelgala:
but how does that affect the current framework ? you want to host multiple references for each test ? For RISCOF I understand that will change the SAIL plugins and config options- but don't really see what should happen in the current framework/spec ?

allenjbaum:
This was requested by Marc, who needed it for his environment. It's an
option someone can use, not a requirement. It shouldn't affect the current
framework, just an environment like Marc's.

neelgala:
well.. then can we have Marc submit a PR for that ? Does the current PR have to be gated by that ?

allenjbaum:
OK, I was confused. 
What he fixed was in sail-riscv, not in arch-test. Sorry for the confusion.
For some reason I thought that this was affected by his changes - but it isn't.

bilalsakhawat-10xe:
Hi,
Does this mean I should change 
https://github.com/riscv-non-isa/riscv-arch-test/blob/307c77b26e070ae85ffea665ad9b642b40e33c86/riscv-test-suite/rv32e_unratified/E/src/add-01.S#L31
and remove `def RVTEST_E = True;` from every rv32e test (change riscv-ctg) ? Because is unnecessary now.
Otherwise change for rv32e_unratified LGTM.
Thanks!

pawks:
> Hi, Does this mean I should change
> 
> https://github.com/riscv-non-isa/riscv-arch-test/blob/307c77b26e070ae85ffea665ad9b642b40e33c86/riscv-test-suite/rv32e_unratified/E/src/add-01.S#L31
> 
> 
> and remove `def RVTEST_E = True;` from every rv32e test (change riscv-ctg) ? Because is unnecessary now.
> Otherwise change for rv32e_unratified LGTM.
> Thanks!

No, this is a temporary fix for the bash scripts based framework in this repo. To run the tests with riscof, that line in the tests is necessary.

bilalsakhawat-10xe (approved):

PR #237, opened 2022-02-28
Zfh extension support

hemanthkumar17:
PR to add Zfh extension


WeeBull:
> {separately: I don't know why this is a 1300 line diff when there only about a dozen short differences]

...because of [this change](https://github.com/riscv-non-isa/riscv-arch-test/commit/69b3fa0c36283b12ff30851dd0a1083c8b8c3d21) on the main branch.

allenjbaum:
1. this was clearly not resolved before riscof was deployed, and we are no longer using Makefile.include files
2. artifacts should not be in the repo, and we will be removing the entire riscv-test-stats directory from the repo, and putting them into a google docs folder ("for risc-v -members -> Workgroups  -> ACT Test Development Artifacts" or  https://drive.google.com/drive/folders/1C70-DJPSV2HxNPbHg6qL9wvVadlmAB2h) New tests that are merged should place stats files here instead of in the repo
3. the arch_test.h file has been extensively modified, and will undergo several more changes in the very near future including splitting off any test specific macros (e.g. TEST_FPRR_OP) into their own file.
 
After the new arch_test.h and test_macro.h files have been merged, if this is still relevant, pull from the latest, and try again.

allenjbaum:
I looked in a lot of detail about the changes in arch_test.h.
There have been updates to these test macros, in addition to splitting them off into a separate test_macros.h file.
I don't believe this PR changes those definitions, but it's hard to tell because there have been additions and corrections since then, but none that I can see that would affect Zfh.

the only substantial change is adding the definitions of FLREG = flh,FSREG =fsh, and FREGWIDTH =2. I'll make sure those go into the new arch_test.h. 
This also redefinea SREG, LREG, and REGWIDTH to match , which is not correct; the width of integer registers does NOT change  when  the floating point extension changes.

I'm going to close this for now. It can be reopened if you find something missing, but that should not include Makefiles nor anything from the riscv-test-stats directory, which should be copied into the googledocs folder: 
for riscv Members -> Workgroups -> ACT Test Development Artifacts -> PRs
(direct link https://drive.google.com/drive/folders/153nIRznXwzu7N1rWl9CesqwAXLqwHoEx)
 in a new folder named PR#xx

PR #238, opened 2022-03-11
RV64F Extension's all instructions.

ptprasanna:
Hi,
These are the current reports that we are working on. Few three operand instructions takes significant amount of time to run. Hence they will be uploaded in another few days.

Thanks
Prasanna T


allenjbaum:
There are test-by-test coverages; shouldn't there be a summary coverage
document of all tests?

PR #242, opened 2022-03-17
Patch for issue 240. Fixes #240

### LINKED ISSUE #240: Edge signatures are 0xdeadbeef now
Since [this patch](https://github.com/riscv-non-isa/riscv-arch-test/issues/240) in every test that has >= 2048 bytes of signature that edge block fails.
This works if you change last signature block address (2044) to zero (0).
But I think that this should be fixed somehow different.
Reverting "[Automatically adjust base and offset if offset gets too big](https://github.com/riscv-non-isa/riscv-arch-test/commit/49e3e0b2efcbf3cda315d5faf6d3dfb82fc0f505)" works for me.

pawks:


allenjbaum (approved):
It fixes the problem, so its good to go

PR #244, opened 2022-03-23
Added helper macro based sig offset update.

pawks:
- Modified the SIGUPD* macros to use a helper macro for offset sanitation.

### DIFF
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -109,6 +109,8 @@
 #endif
 
 
+
+
 // ----------------------------------- CODE BEGIN w/ TRAP HANDLER START ------------------------ //
 
 .macro RVTEST_CODE_BEGIN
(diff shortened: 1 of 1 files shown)


allenjbaum (changes requested):
SIGALIGN must be defined for SIGUPD_F, and it must check that the alignment begins at a SIGALIGN, and issue a warning if it doesn't:
  .if (offset&(SIGALIGN-1))!=0				;\
/* Throw warnings rather than modify offset to target */;\
     .warning "Incorrect Offset Alignment for signature." ;\
     .set offset, offset&(SIGALIGN-1)+SIGALIGN		;\
  .endif						;\

The CHK OFFSET macro for SIGUPD_F doesn't use REGWIDTH, it uses SIGALIGN. This allows it to handlesthe RV32_D case
Since the CHK_OFFSET macro is updating the offset between every store, the final line of each UPD macro should only update by 1*REGWIDTH/SIGALIGN), not 2*

pawks:
> I don't think this works. SIGALGN should _always_ be max(FLEN,XLEN), and you have simply be XLEN when FLEN is 64 (line 89)

Oops. Copy pasta error. Fixed it.

PR #245, opened 2022-03-29
Crypto scalar updates

anku-anand:
1. rev.b is changed to brev8 in spec, updated related files.
2.  removed references to K_unratified in 
       riscv-arch-test/tree/master/riscv-test-stats/coverage/rv32i_m 
       riscv-arch-test/tree/master/riscv-test-stats/coverage/rv64i_m
3. Updated ISA spec strings in following yaml files to be same as mentioned in the spec
     riscv-arch-test/blob/master/riscv-test-stats/coverage/rv32ik.yaml
     riscv-arch-test/blob/master/riscv-test-stats/coverage/rv64ik.yaml
4. Added more coverpoints and datasets to zbkb instructions.


allenjbaum:
Could you fix this CHANGELOG conflict?

pawks:
The config strings for the instructions from bitmanip extensions are wrong. They contain `zk*` values and no `zbb`. The tests for these instructions are also replicated across both `B` and `K` directories. There should only be 1 instance of these tests which are picked up for both `B` and `K`. The tests should only be present in the directory of the extension which appears first canonically.

This PR also seems to add Makefiles, Makefrags and references for the B extension. These aren't necessary anymore.

Several links in the coverage.md file are broken for B and K.

anku-anand:
Since bitmanip instructions and crypto scalar instructions overlap, all changes for bitmanip and crypto scalar are put i n

PR #228 Bit manip test suites and stats.
Closing this PR

PR #247, opened 2022-03-31
fix for LLVM compiler support

jsleroy:
Update arch_test.h code so that code compiles also with LLVM compilers.

`.if` changed to `.ifc` due to an issue in LLVM. This construct raise an "expected absolute expression" error. See similar issue reported in LLVM https://github.com/llvm/llvm-project/issues/53728

`.set` used to declare/update variables.

Modified code still run with GCC.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## [2.7.2] - 2022-03-31
+  - Update framework to support test suite compilation with LLVM.
+
 ## [2.7.1] - 2022-03-18
 - Fix bug in auto-offset update for SIGUPD macros.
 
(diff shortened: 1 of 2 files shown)


neelgala:
Can you add an entry in the CHANGELOG so that I can merge it ?

jsleroy:
Yes sure, will update the branch.

jsleroy:
Updated CHANGELOG. Not sure about the version/date so feel free to correct them.

neelgala:
I don't see the CHANGELOG update in this PR. You can create a 2.7.3 version entry

jsleroy:
Sorry, fixed. I had put already 2.7.3, so we're fine here.

PR #251, opened 2022-04-18
Cleaning fflags csr after each test

misterjdrg:
Replace `csrrs flagreg, fflags, x0` -> `csrrw flagreg, fflags, x0`, so fflags part of fcsr set to 0 after each test.


allenjbaum (approved):
This reads and clears the FCSR after each test which is necessary to enable testing for the correct settings of the summary status. Note that this will require that reference signature will need to be updated for every test in the current framework, but will not be required for the riscof framework. Also note that we still need to add  the infrastructure to measure the CSR side-effects, the coverpoints that test them, and possibly new tests to exercise the coverpoints.

misterjdrg:
I tested all extensions for rv32i_m and rv64i_m, except K (gcc throws `Error: unrecognized opcode `aes64ds x12,x10,x11'`)

pawks:
@misterjdrg Please update `CHANGELOG.md` with an entry for this.

misterjdrg:
@pawks Done

PR #252, opened 2022-04-21
fix error on BSD diff

kuopinghsu:
BSD diff does not support option --ignore-trailing-space, change to --ignore-all-space.


allenjbaum:
Can you please either resolve this conflict, or close the issue?

kuopinghsu:
This is the file in old-framework-2.x, it does not exist in master branch. close it.

PR #264, opened 2022-06-29
Added all the PMP configuration and address CSRs

Abdulwadoodd:


allenjbaum:
Hmm, encoding.h   says
/*
* This file is auto-generated by running 'make
../riscv-isa-sim/riscv/encoding.h' in
* https://github.com/riscv/riscv-opcodes (e8f0392)
*/
I'm not sure we should have our own version at all; we probably should just
be linking to a version of it.
The most complete one is almost certainly here:

   https://github.com/riscv-software-src/riscv-isa-sim/  in file
 riscv/encoding.h

Abdulwadoodd:
True, I have pushed the autogenerated encoding.h and removed the own version.
Thanks

Abdulwadoodd:
The file that I just pushed is generated by running `make` in  [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes) and the generated file is named as `encoding.out.h` which i renamed as `encoding.h` and then pushed here.
Apparently, this autogenerated file from [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes) seems to be more updated than the file present in [riscv-isa-sim/riscv](https://github.com/riscv-software-src/riscv-isa-sim/tree/master/riscv). 
Couldn't grasp the linking part though.

And I'm learning. Thanks!

Abdulwadoodd:
True, that seems the right thing to do. The latest one is present in [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes) which is generated on running `make`. So, we must be pulling from here and not from [riscv-isa-sim/riscv](https://github.com/riscv-software-src/riscv-isa-sim/tree/master/riscv).
I try to find a way to do this

Abdulwadoodd:
PR #268 is opened which handles this issue in a better way. So, time to close this one.

PR #267, opened 2022-07-01
Add changes to enable multiple Privilege Modes in Trap Handler

bilalsakhawat:
Co-authored-by: allenjbaum <allen.baum@esperantotech.com>

To test a core with hardware misaligned support:
Add --enable-misaligned to the execute command on line 148 in riscof-plugins/sail_cSim/riscof_sail_cSim.py

### DIFF
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -1,40 +1,435 @@
+// This file is divided into the following sections:
+//	RV Arch Test Constants
+//	general test and helper macros, required,  optional, or just useful
+//	   _ARGn, SIG[BASE/UPD[_F/ID]],BASE_UPD,BIT,LA[U],LI[U],RVTEST_[INIT/SAVE]_GPRS, XCSR_RENAME
+//	RV ARCH Test Interrupt Macros ****FIXME:spec which regs must not be altered
+//	primary macros used by handle: RVTEST_TRAP{_PROLOG/_HANDLER/_EPILOG/SAVEAREA}
+//	required test format spec macros:      RVTEST_{Code)/DATA/SIG}{_BEGIN/_END}
+//	macros  from Andrew Waterman's risc-v test macros
+//	deprecated macro name aliases, just for migration ease
+
+//  The resulting memory layout of the trap handler is (MACRO_NAME, label [function])
+//****************************************************************
+//  (code section)
+// RVMODEL_BOOT
+//      rvtest_entry_point: [boot code]
+// RVTEST_CODE_BEGIN
+//      rvtest_init:	   [TRAP_PROLOG]   (m, ms, or msv)
+//			   [INIT_GPRS]
+//      rvtest_code_begin:
+//*****************************
+//********(body of tests)******
+//*****************************
+// RVTEST_CODE_END
+//      rvtest_code_end:   [*optional* SAVE_GPRS routine]
+//			   [GOTO_MMODE ] **FIXME** this won't work if MMU enabled unless VA=PA
+//      cleanup_epilogs	   [TRAP_EPILOG   (m, ms, or msv)] (jump to exit_cleanup)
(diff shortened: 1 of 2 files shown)


pawks:
Having a single `arch_test.h` file is becoming difficult to maintain and make changes to. Its better to have the trap handler in a separate file and add an include in `arch_test.h` to include it. That way all changes to the trap handler are localised. Can you split it into different files?

riscv-test-suite/env/arch_test.h
```diff
@@ -159,22 +500,367 @@
...
+.endm
+
+/************************************************************************************/
+
+.macro  RVTEST_SIG_END
+    RVMODEL_DATA_END		//model specific stuff
+
+
+ .global rvtest_sig_end		//defines end of signature area
+  rvtest_sig_end:
```
pawks, line 654: The `rvtest_sig_end` label should probably be placed before the `RVMODEL_DATA_END` macro.

riscv-test-suite/env/arch_test.h
```diff
@@ -1,40 +1,435 @@
...
+.set SMODE_SIG,1
+  //#define MMODE_SIG 3
+  //#define VMODE_SIG 2
+  //#define SMODE_SIG 1
+	/* these macros need to be defined because mode is uppercase in mode specific macros */
+	/* note that vs mode uses smode return */
+#define MRET	mret
+#define VRET	sret
+#define SRET	sret
+#ifndef rvtest_sig_end
```
pawks, line 197: I'm not sure if this piece of code is relevant. The `rvtest_sig_end` is an assembler label and these statements are preprocessor directives. Hence a `rvtest_sig_end` macro will get generated, which when used with the `LA` macro will always cause the register to have the address `mtrap_sigptr+6*REGWIDTH`. This can cause problems in certain scenarios where this value is used to make decisions in the trap handler.

allenjbaum: The intent was to default a label for the end of the trap signature area if it wasn't defined.
The 6*reg offset was the artifact of pre-incrementing by adding a single maximal sized entry guardband. so you could recursively trap
 That label is used in 2 places: 
 -  to ensure that writes to the trap signature don't go past the end of the signature area and corrupt things, 
   (that's just a nice-to have)
 -  to check whether xTVAL is pointing into the signature area (really required for tests that write data directly into the signature area and have relocated the signature area to someplace other than immediately following the rvtest data area)
I'll simplify this by 
 - pre-incrementing by the exact amount (instead of the maximal amount)
 - conditionally implementing the overrun check if rvtest_sig_end isn't explicitly defined
   (because there is simply no way to know where the signature area ends in current tests - no label is defined). 
- conditionally implemented relocation relative to the signature area if rvtest_sig_end and rvtest_sig_begin 
  aren't defined  (because we can't know where the signature area is)
The new macros RVTEST_SIG_BEGIN/END macros will fix that partially; they can specify where the signature ends, but not where it starts, without modifying all existing tests that trap. We will need to do that also - I think it is only misaligned tests in the privilege directory.

pawks: I think that all the tests in the privilege directory will need to undergo a change. It is best to require all tests which use the trap handler to use `RVTEST_SIG_*` macros instead of `RVMODEL_SIG_*` macros while letting the unprivileged tests use either(to ensure backward compatibility).

allenjbaum:
Yes; the changes I've made will accommodate that. All new tests must also
define rvtest_sig_begin: explicitly in the test, immediately after
RVMODEL_DATA_BEGIN, else we can't properly relocate even unpriv tests (e.g.
store tests that store directly into the signature area to be safe, and I
think that will make it easier for new cores to interface to the tests. (we
can argue whether it should be signature_begin (which is recommended in the
spec) or rvmodel_sig_begin: or rvtest_sig_begin:, but the spec will need to
be updated.

bilalsakhawat:
closing this PR as the modified version is there in PR #277

Issue #263, opened 2022-06-29
Complete PMP configuration CSRs and address CSRs are not supported for 64 PMP entries


Abdulwadoodd:
Arch-tests do not have the support of following CSRs in `encoding.h` file.
- pmpcfg`x` where x is [4,15]
- pmpaddr`x` where x is [16,63]

Abdulwadoodd:
Little correction: The file present by default in [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes) is not complete but the most up-to-date version (even latest than what is in [riscv-isa-sim/riscv](https://github.com/riscv-software-src/riscv-isa-sim/tree/master/riscv)) of the file can be generated by running `make` in [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes).

allenjbaum:
I've filed an issue to try to ensure that the version in the repository is always the latest.
The question is how the framework pulls it in, or should the framework always regenerate it as you did
In either case it doesn't need to be in the repo at all.

Abdulwadoodd:
According to Andrew's comment on your issue ([here](https://github.com/riscv/riscv-opcodes/issues/132#issuecomment-1170642498)) for now, we need to regenerate it every time. How about writing a makefile that clone [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes)[riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes) and build encoding.h locally to place in [riscv-arch-test/riscv-test-suite/env](https://github.com/riscv-non-isa/riscv-arch-test/tree/main/riscv-test-suite/env). What do you say

allenjbaum:
We could either
 - 1. clone it as you said
 - 2. add a CI script to the riscv-opcodes repo to produce it as a
release and simply copy it from as part of make
 - 3.copy it from the spike directory.

Short term, your solution is best.
Longer term: I'd like to see the CI script solution so others can use the
same approach, if that is possible

Abdulwadoodd:
Hi @allenjbaum, I have implemented a long-term solution by writing a CI script. Kindly review PR #268

PR #268, opened 2022-07-03
Add a CI script to update the `encoding.h` file

Abdulwadoodd:
As discussed in the issue [here](https://github.com/riscv-non-isa/riscv-arch-test/issues/263), the latest version of `encoding.h` file can be generated from [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes) repository by running `make`. So, instead of maintaining a separate copy of `encoding.h` in this repository, we mush by pulling the latest version of this file from the parent [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes) repository. 
So, I have written a CI script that does the same. On every `push` or `pull_request`, CI will invoke and replaces the old encoding with the latest one.


pawks:
Looks like the CI is failing. It might require some additional changes. Note that the main branch is protected and direct pushes to main are disalowed.

Abdulwadoodd:
I've changed the script so that it would trigger on the pull request, should that work fine? Also, what is the alternative to direct pushes to the main branch if it is disallowed

pawks:
Only Merge/pull requests are allowed to master.

neelgala:
I am not completely comfortable with this automated check-in of files. The encoding.h here should be updated when the tests in this repo require those updates in encoding.h. This will make sure that the modifications in encoding.h are in sync with changes in tests. 

Also, the encoding.h generated in riscv-opcodes is specifically designed to be consumed by spike (certain pseudo-ops are ommitted) and the changes that happen there will be specific to that. Automated check-ins have the potential of pulling in something that is non-backward compatible and break things here - which shouldn't be the case.

Solution should be to manually check-in encoding.h as an when required.

allenjbaum:
I'm not sure if you're uncomfortable with the  concept of automated
check-in, or the conditions under which it is checked in.
E.g. is limiting it to when merges  are approved implementable, and if so -
why would that make you uncomfortable?

I'm more concerned with how to ensure that the right toolchain(s) get used,
which seems to be a more overriding problem,
but that is out of our hands.

neelgala:
having an untested code-check in is what is scary. Right now all merges are manually verified (by contributors and owners of this repo). Having a CI to commit a code without due testing = recipe for chaos. Having the CI test before committing has the  issue of which toolchain should be used for that test - because unratified tests will probably work with custom toolchains which are not upstream yet. If a PR warrants a change in the encoding.h then it should be done as part of that PR. Steps to generate the encoding.h from riscv-opcodes can be captured in the docs somewhere.

PR #274, opened 2022-08-08
Added tests for Floating point memory operations. 

pawks:
Added tests for aligned floating point load and store operations.


allenjbaum:
This line 918 of arch_test.h should be split into 2 lines.
+918.  LI(testreg, fcsr_val); csrw fcsr, testreg                                                ;\
 It's technically correct, but a bit jarring

neelgala:
@allenjbaum can I merge this ?

allenjbaum:
yes, go ahead,

allenjbaum:
This one is ready to go

allenjbaum (approved):
This should be good to go now. We do need to add comments to macros that explain how they are used, however, but that shouldn't hold up merging for now; that can be done when we merge in the split arch_test.h / test_macros.h file

PR #278, opened 2022-09-10
fix `arch_test.h`

leecannon:
I'm getting the below error since ea3f151c0aaca474baa45537f09a02c756ab43c1
```
ERROR | In file included from /home/lee/src/zriscv/riscof/riscv-arch-test/riscv-test-suite/rv64i_m/I/src/add-01.S:19:
/home/lee/src/zriscv/riscof/riscv-arch-test/riscv-test-suite/env/arch_test.h:918:90: warning: backslash and newline separated by space
  918 | LI(testreg, fcsr_val)                                                                   ;\
      |
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## [3.4.1] - 2022-09-11
+  - Fix trailing space in arch_test.h
+
 ## [3.4.0] - 2022-08-18
   - Added tests for Bitmanip and Crypto scalar extensions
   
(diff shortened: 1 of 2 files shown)


allenjbaum (changes requested):
oops; this should have caught in CI. Is this a toolchain issue (that it wasn't caught)

Please update the change log as well and this should go through.

allenjbaum (approved):
For the record: I'd like to know how this slipped through, but the fix is correct

pawks:
I think it was missed because its just a warning.

PR #284, opened 2022-09-19
Pmp tests

UmerShahidengr:
Dear All,
This PR contains Priv Arch compliance tests. This repo will be updated weekly with new priv arch tests. These short PRs will be easier to review on weekly basis. In the first iteration, a few basic PMP compliance tests have been pushed for review.


allenjbaum:
What is going to be important in order to merge these tests are that
 - they compare between Sail and Spike (I think you already verified this)
 - that risc-isac can verify coverage. That means you need to be able to
describe the coverage using a YAML ctg schema description.
I think you'll find that this will be as hard as writing the tests, but it
is important.
Your written descriptions are a start, but the need to be fleshed out with
the real corner cases  in that YAML format (e.g.
 -  2 matches from both first and last PMP entry,
 -  intervening entries being non-matching (including off)
 -  the first is a wider range than the last, or a smaller  or overlapping
 - 4 combinations of illegal/legal
 - for each of the permission bits.

Another might be 3 entries: 2 abutting, and a 3rd overlapping the other two
for just 8bytes
(which I think can only occur on an RV64 with a doubleword data fetch) with
- 6 different prioriity orders,
- 3 different accesses (matching entirely in the 1st, 2nd or 3rd entry)
- 8 different legal/illegal combinations,
 - both loads and stores

pawks:
> What is going to be important in order to merge these tests are that - they compare between Sail and Spike (I think you already verified this) - that risc-isac can verify coverage. That means you need to be able to describe the coverage using a YAML ctg schema description. I think you'll find that this will be as hard as writing the tests, but it is important.

There is a small hitch in this case. The csr coverpoints are currently defined over the values in the csrs. So the coverpoints for the `pmpaddr` csr can be defined only based on the value contained in them. However, it is perfectly legal for the values to be different based on the implementation defined address space, meaning that the coverpoint itself changes between implementations. There are two possible ways to handle this:
- Allow for the coverpoints to be different based on the implementation and assume a default for SAIL. Since the coverage is primarily measured on SAIL, this allows us to easily measure coverage but the coverpoints are tied to a specific memory map. 
- Define the coverpoints based on the behaviour of the pmp, making them independent of the actual addresses i.e use the effective address of the access and write expressions which mimic the pmp match and priority functions to check for coverage. Doing so will require some more discussion and adding support for the same in isac. 

> Your written descriptions are a start, but the need to be fleshed out with the real corner cases in that YAML format (e.g. - 2 matches from both first and last PMP entry, - intervening entries being non-matching (including off) - the first is a wider range than the last, or a smaller or overlapping - 4 combinations of illegal/legal - for each of the permission bits. Another might be 3 entries: 2 abutting, and a 3rd overlapping the other two for just 8bytes (which I think can only occur on an RV64 with a doubleword data fetch) with - 6 different prioriity orders, - 3 different accesses (matching entirely in the 1st, 2nd or 3rd entry) - 8 different legal/illegal combinations, - both loads and stores

The tests also should be easily configurable to account for the following:
- Granularity of the pmps
- Write-ability of the pmp* csrs
- Number of pmp csrs defined

Another corner case which needs to be considered is partial matches. This particular case is applicable for mem accesses only when the granularity of the pmps is 4 and the `XLEN` is `64` or the `D` extension exists. However, this is applicable for instruction accesses when the `C` extension exists regardless of the granularity of the csrs.

UmerShahidengr:
> - that risc-isac can verify coverage. That means you need to be able to describe the coverage using a YAML ctg schema description. I think you'll find that this will be as hard as writing the tests, but it is important.

Since ISAC has limitation on adding coverpoints. I added csr coverpoints which were defined over the values in the pmp registers, I submitted that PR [at riscv-ctg with updated cfg files](https://github.com/riscv-software-src/riscv-ctg/pull/41), I will update those cfg files according to the test cases for each upcoming test

UmerShahidengr:
> * Allow for the coverpoints to be different based on the implementation and assume a default for SAIL. Since the coverage is primarily measured on SAIL, this allows us to easily measure coverage but the coverpoints are tied to a specific memory map.
> * Define the coverpoints based on the behaviour of the pmp, making them independent of the actual addresses i.e use the effective address of the access and write expressions which mimic the pmp match and priority functions to check for coverage. Doing so will require some more discussion and adding support for the same in isac.

Dear @pawks , I am not sure if I understand it fully. I'll dig in to it, but I might need more help in this regard. I will ping you offline in case of any query. In the meantime, can you refer me to some good documentation on understanding cgf files in more detail?

UmerShahidengr:
> * Granularity of the pmps
> * Write-ability of the pmp* csrs
> * Number of pmp csrs defined
> 
> Another corner case which needs to be considered is partial matches. This particular case is applicable for mem accesses only when the granularity of the pmps is 4 and the `XLEN` is `64` or the `D` extension exists. However, this is applicable for instruction accesses when the `C` extension exists regardless of the granularity of the csrs.

Point noted. We are actually following [this testplan](https://docs.google.com/spreadsheets/d/1TwiNvnGmAvSCAXhnsTnn6E295kD4p97CLVjh49zRVF8/edit?usp=sharing), you can review our testplan and give your feedback, we will add more tests to cover the corner point which you mentioned.

pawks:
> Dear @pawks , I am not sure if I understand it fully. I'll dig in to it, but I might need more help in this regard. I will ping you offline in case of any query. In the meantime, can you refer me to some good documentation on understanding cgf files in more detail?

A description of the various types of coverpoints and their functions can be found [here](https://riscv-isac.readthedocs.io/en/stable/cgf.html).

pawks:
The tests should also mention the conditions for it to be enabled in the `TEST_CASE` macros. It would probably be better if the tests are organized in such a way that all related test cases for a particular pmpentry in a single test. All the tests in the PR for the `L` bit only test their warl behaviour. However, the same test can be leveraged to test for the matching and permission logic too.

pawks:
Another thing which should probably be accounted for is the Smepmp spec as it changes the behavior of the pmps. They will definitely influence the conditions for these tests.

allenjbaum:
I think this is a perfectly reasonable approach for development.

UmerShahidengr:
In this week's commit, the following changes are made:

- Based on Allen's comments, in [PMP-CFG-reg.S](https://github.com/Abdulwadoodd/riscv-arch-test/blob/pmp_tests/riscv-test-suite/rv32i_m/privilege/src/pmp32/PMP-CFG-reg.S) tests for locked bit and zero bits are merged.
- code size is reduced by adding macros and .rept directive.
- coverpoints for pmpcfg tests are added in [rv32i_priv.cgf](https://github.com/Abdulwadoodd/riscv-arch-test/blob/pmp_tests/coverage/rv32i_priv.cgf).
Hopefully, all other coverpoints and pmp tests will be merged till next week, contributors are requested to review the draft.

PR #291, opened 2022-10-30
Make env CLIC-compatible

### LINKED ISSUE #285: CLIC or vectored mode leads to misaligned load
This line
https://github.com/riscv-non-isa/riscv-arch-test/blob/da638bafbc6d69f75ffc09769cfe7e3a45b41a3a/riscv-test-suite/env/arch_test.h#L250
suggests that mtvec (thus t2) contains a correct address.
This is not always true, e.g. if `MODE (MTVEC[1:0]) == 3 (CLIC)`.
`andi t2, t2, -4` should be added before this line.

marcfedorow:
Fixes #285

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## [3.5.2] - 2022-10-31
+  - Make env CLIC-compatible
+
 ## [3.5.1] - 2022-10-28
   - Add test cases for P-ext
   - Correct TEST_PKRR_OP() macro in arch_test.h 
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -233,9 +233,11 @@
   	la	t1, mtrampoline    
   	la	t4, mtvec_save
   	csrrw	t2, CSR_MTVEC, t1		          // swap mtvec and trap_trampoline
+	andi    t2, t2, -64
   	SREG	t2, 0(t4)		              // save orig mtvec
   	csrr	t3, CSR_MTVEC		              // now read new_mtval back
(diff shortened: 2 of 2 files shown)


allenjbaum (changes requested):
First: this is being completely superseded by a new version of arch-test.h (PR267, RealSoonNow) 
What this is doing is taking original value of xTVEC and clearing the low 6 bits.
That means when we return from the test with all CSRs restored to their original value - this one has been destroyed. So, not acceptable.

Secondly, you're using this to force some particular behavior (clic behavior, I assume) regardless of whether clic is implemented or not. Also not acceptable.
We can enforce that trampolines be at 64b boundaries - I'll update PR267 to do that.

Third: you're clearing the upper 12 bits of the mcause register, which (among other things) destroys the bit that indicates where this was an interrupt or an exception.  Obviously not acceptable.

Please close this and wait until PR267 is merged, then try again - but any change you make must be completey compatible with non-clic systems.

marcfedorow:
> regardless of whether clic is implemented or not. Also not acceptable.

Definitely not. IDK how this follows from my patch. Clearing some bits before writing them to signature will not definitely break anything (except actually clearing the meaningful bits of addres, and that is workaronded with the proper alignment).

Using masks instead of shifts is not that hard I suppose, so this PR is easily fixable.

allenjbaum:
Ah, I see what you mean. But, the trap handler restores configured CSRs to
the state they were in before the trap handler was installed, and this
doesn't.
Take a look at PR 267 and 277, and branch off that instead of off the
existing code.
Not that the expectation is that if clic is implemented and enabled, both
Sail and the DUT will be configured such that writes to xTVEC will treat
bits 5:0 as 0
(which I assume is what CLIC does, I don't recall the details) so all this
masking is unnecessary.

PR #302, opened 2022-12-17
Canary patch

### LINKED ISSUE #190: the 16-byte signature boundary issue
Consider the [div-01.s](https://github.com/riscv/riscv-arch-test/blob/master/riscv-test-suite/rv32i_m/M/src/div-01.S) test which has 590 (starts with inst_0) test-cases. So the signature region created is 590*4 bytes = 2360 bytes in size initialized to 0xdeadbeef. **Now this is not a multiple of 16**.

Now if we go back to the test-format-spec here it says: 
- `the signature should always begin at a 16-byte` 
- `"it should also end at a 16-byte boundary`. 
Due to the above requirements,  the [spike model_test.h](https://github.com/riscv/riscv-isa-sim/blob/9d4f45c2ebf105503974fc80a42590ca1584c354/arch_test_target/spike/model_test.h#L32) constraints the alignment of the the RVMODEL_DATA_BEGIN and RVMODEL_DATA_END to 16-byte boundaries. This means the RVMODEL_DATA_END is at an offset of 2368 bytes from RVMODEL_DATA_BEGIN. 

Therefore, there are 8 bytes of data in the signature which are not initialized by the test and can be any value the model decides (which in GCC is 0x0). Hence, if you look at the last [2 entries of the reference signature](https://github.com/riscv/riscv-arch-test/blob/b436dd0939c968f2c3da86bb9b63bb2dfe03b134/riscv-test-suite/rv32i_m/M/references/div-01.reference_output#L591-L592) of div-01.reference_output they are going to 0's instead of 0xdeadbeef

Now, since the extra padded values are controlled by the DUT, they can be anything. 

Possible solutions:

1. the spec needs to be updated to mandate that the values being padded to meet alignment must be 0xdeadbeef. It shouldn't be 0x0 since the spec mentions [here ](https://github.com/riscv/riscv-arch-test/blob/master/spec/TestFormatSpec.adoc#451-common-test-format-rules)that the signature must be loaded with 0xdeadbeef This will require updating the references and all ported targets. The targets could achieve the above by simply using `.align 4,0xdeadbeef` instead. No changes to tests.

2. Change the spec to mandate that the signature should end at a 4-byte boundary. This will require all targets to remove any `.align 4` from their macros, since by default all test signatures end at a 4-byte boundary. This will require references to be updated. No change to tests.

3. Explicitly padd the tests to align at 16-byte boundaries. This will require updating the tests and the corresponding references. No changes to the model side (or atleast that is what I hope).

Would like to know which of the above is best suited? or if there is another solution.

### LINKED ISSUE #280: Missing information about the `RVTEST_CASE` argument format in test format spec.
The arguments for the `RVTEST_CASE` are in a specific format and the test format spec needs to be updated to document the format. Something akin to [here](https://github.com/riscv-software-src/riscof/blob/a003752ff95dd5f9153c323dc2eb8ff7a0426dd9/docs/source/testformat.rst#rvtest_case-condition-formatting). Previously it was included in the riscof documentation along with the test format spec. But the test format spec was removed from riscof as it is actively maintained here.

### LINKED ISSUE #299: Misaligned floating-point signature causes hang
I recently pulled the latest repo (first time since July) and riscof now hangs when I run.  I've also updated to the latest riscof but not changed any of my other files such as the linker scripts.

The hang occurs on rv32 D tests.  The root cause seems to be misaligned doubles in the signature.

For example, I traced the problem to spike hanging on fadd.d_b1-01.S.  The hang is caused by a misaligned store address on the first test case.  The store is to 8000b614, which is not a multiple of 8. The Spike trace and disassembled code are shown below.

I added .align 3 before signature_x1_0, which fixed the issue.  But there must be something more fundamental wrong.  Any pointers?

Thank you,

David Harris

core   0: 0x8000012c (0x00e0b027) fsd     fa4, 0(ra)
core   0: exception trap_store_address_misaligned, epc 0x8000012c
core   0:           tval 0x8000b614
: 
core   0: exception trap_instruction_access_fault, epc 0x00000000
core   0:           tval 0x00000000
: reg 0 ra
0x8000b614

80000110:	50808093          	addi	ra,ra,1288 # 8000b614 <signature_x1_0>

80000114 <inst_0>:
80000114:	0001b707          	fld	fa4,0(gp)
80000118:	0081b707          	fld	fa4,8(gp)
8000011c:	00000113          	li	sp,0
80000120:	00311073          	fscsr	sp
80000124:	02e77753          	fadd.d	fa4,fa4,fa4
80000128:	00302273          	frcsr	tp
8000012c:	00e0b027          	fsd	fa4,0(ra)
80000130:	0040a423          	sw	tp,8(ra)

### LINKED ISSUE #300: Divide and Remainder are missing overflow corner case test
The RISC-V specification Section 7.2 defines a corner case for divide and remainder for the most negative integer divided by -1.  

It appears that this case is not tested in div-01 and other divide and remainder tests.

pawks:
## Description
- Fix Canary definition to depend on sigalign.
- Fix SIGALIGN definition to work in all cases of FLEN and XLEN definitions.
- Fix inconsistencies in tests (missing trap regions and incomplete ISAs).
- Add Zicsr to ISA in RV32 privilege tests
- Modify signature size requirement to multiple of 4.

### Related Issues
- Fixes #190 
- Fixes #280
- Fixes #299


spec/TestFormatSpec.adoc
```diff
@@ -300,6 +315,98 @@ significant amount of the framework shall depend on the existence of these macro
     - the test-case must be delimited with an #ifdef CaseName/#endif pair +
     - the format of CondStr can be found in https://riscof.readthedocs.io/en/latest/testformat.html#rvtest-case-condition-formating
 
+===== *Writing the arguments for RVTEST_CASE macro*
+
+This section describes the format for the conditions `CondStr` to be followed while writing the `RVTEST_CASE` macro.
+
+====== *Vocabulary*
+A keylist is a string of '>' separated words(keys) which is used to navigate the supplied specs. The schema may be used to specify them. Only valid keys and their combinations are allowed(as present in the scema).
```
allenjbaum, line 323: typo:  scema -> schema

spec/TestFormatSpec.adoc
```diff
@@ -300,6 +315,98 @@ significant amount of the framework shall depend on the existence of these macro
...
+
+There are two types of valid statements allowed.
+
+1. `check` statements:
+  These statements get translated into the condtions which need to be true for the part to be enabled.
```
allenjbaum, line 328: typo condtions->conditions

allenjbaum: should "part" be "test case" or "test_case"?

spec/TestFormatSpec.adoc
```diff
@@ -300,6 +315,98 @@ significant amount of the framework shall depend on the existence of these macro
...
+  The `keylist` specifies the path to the field in the ISA YAML dictionary whose value needs to be checked. 
+  The `value` is the value against which the entry in the input yaml is checked.
+  The `value` can be a regular expression as well, in which case it should be specified as `regex("expression")`
+  
+  Example: 
+  check ISA:=regex(.*I.*Zicsr.*);  # checks if ISA node supports I and Zicsr extensions
+  check hw_data_misaligned_support:=True; # checks if the misaligned support is available.
+
+  * `keylist=key` The `keylist` specifies the path to the field whose keys needs to be checked. 
+        The `key` is the key whose presence needs to be checked in the field specified by the keylist.
```
allenjbaum, line 341: IS "key" really a key, or a value associated with a key?

spec/TestFormatSpec.adoc
```diff
@@ -300,6 +315,98 @@ significant amount of the framework shall depend on the existence of these macro
...
+
+
+2. `def` statements 
+    * These statements specify which macros to be defined for the part to run and their values(optional) in the following format.
+    
+        def macro(s)(=value/keylist/function);
+    
+    * The macro specifies the name of the macro.
+    * Multiple macros can be specified using a comma in-between them.
+    * A keylist specifying the path of the field whose value has to be passed as the value of the macro can also be given.
```
allenjbaum, line 379: This is really hard to interpret. A shorter description might be clearer:
   A keylist specifying the field path whose value is passed as the macro value can also be given.
I thought this would be something like macro_name =value; is "macro(s)" supposed to be a keyword or a macro name? IF its a name, then call it "macro_name". Also, give an example of when there are more than one. I'm not sure if the syntax above requires the "(" , ")" or not,

spec/TestFormatSpec.adoc
```diff
@@ -300,6 +315,98 @@ significant amount of the framework shall depend on the existence of these macro
...
+    * A function along with the arguments can also be specified. At runtime the function is called using the specified arguments and its return values are assigned to the macro(s) specified. The list of functions supported are as follows.
+        
+            - `getlegal(dependency_values_list, num_vals, key_list_for_field)` -> `list(int)`
+                This function is valid only for _WARL_ fields in the csrs. It 
+                returns a list of legal values for the specified field when the values 
+                of the fields listed as dependency for the field in question on is equal 
+                to the _dependency_values_list_. The length of the list returned is equal 
+                to _num_vals_. Each entry in the list is assigned to the corresponding 
+                macro listed on the left hand side of the `=` sign.
```
allenjbaum, line 388: It  returns a list of legal values for the specified field when the dependency field values are equal to the _dependency_values_list_. 
The length of the list returned is equal to _num_vals_.  <--This doesn't sound right; The length of the list should be the number of dependency value in the list, shouldn't it? Why do you need to specify that? 
Are the legal values randomly generated from the ranges of legal values?

spec/TestFormatSpec.adoc
pawks: For line 341, it is a key by itself because the underlying node being accessed is a dictionary node by itself. For example, the `base` field of the `mtvec` register can be implemented as a `ro_constant`, `ro_variable` or `warl`. For each of these types, the description is also included by making it a dictionary node. You can use the following statement to identify whether the behavior is WARL: ` misa>rv32>base>type=warl`

spec/TestFormatSpec.adoc
pawks: Added.

spec/TestFormatSpec.adoc
pawks: Fixed

spec/TestFormatSpec.adoc
pawks: > Can "value" be a list if there is more than one macro name? 

No. Multiple macro based LHS is supported only for function calls. 

> Ditto for function Can you mix values, keylists, and functions? If you can't, then the num_vals parameters aren't needed, as they can be determined from the number of macro_names

You cant mix and match. The reason for having the `num_vals` is because of the way the functions are evaluated. This particular feature is still rudimentary in riscof. It will become more robust as we start writing tests for the priv spec.

spec/TestFormatSpec.adoc
pawks: Yes they are random. Updated the description to indicate the same.

spec/TestFormatSpec.adoc
```diff
@@ -308,24 +308,35 @@ significant amount of the framework shall depend on the existence of these macro
...
 
 ===== *Writing the arguments for RVTEST_CASE macro*
 
 This section describes the format for the conditions `CondStr` to be followed while writing the `RVTEST_CASE` macro.
 
 ====== *Vocabulary*
-A keylist is a string of '>' separated words(keys) which is used to navigate the supplied specs. The schema may be used to specify them. Only valid keys and their combinations are allowed(as present in the scema).
+The `keylist` is used to navigate and access the various nodes in the input configuration files.
+A keylist is a string of '>' separated words(keys) which is used to navigate the supplied
+configuration files. A list of keys indicates the hierarchy of nodes to be accessed
```
allenjbaum, line 325: I'm picky :-). Yu say it is a string of ">" separated words, and then refer to it as a list. That makes me think there a two different things, so you should make the terminology consistent

spec/TestFormatSpec.adoc
```diff
@@ -308,24 +308,35 @@ significant amount of the framework shall depend on the existence of these macro
...
 ===== *Writing the arguments for RVTEST_CASE macro*
 
 This section describes the format for the conditions `CondStr` to be followed while writing the `RVTEST_CASE` macro.
 
 ====== *Vocabulary*
-A keylist is a string of '>' separated words(keys) which is used to navigate the supplied specs. The schema may be used to specify them. Only valid keys and their combinations are allowed(as present in the scema).
+The `keylist` is used to navigate and access the various nodes in the input configuration files.
+A keylist is a string of '>' separated words(keys) which is used to navigate the supplied
+configuration files. A list of keys indicates the hierarchy of nodes to be accessed 
+in the input yaml configuration files(
```
allenjbaum, line 326: add a space bewteen "files" and "("

spec/TestFormatSpec.adoc
```diff
@@ -368,30 +380,31 @@ There are two types of valid statements allowed.
                 is equal to the _dependency_values_list_.
 
 
-
 2. `def` statements 
     * These statements specify which macros to be defined for the part to run and their values(optional) in the following format.
```
allenjbaum, line 384: These statements define and (optionally) give values to macros required for the test case to run in the following format.

spec/TestFormatSpec.adoc
```diff
@@ -368,30 +380,31 @@ There are two types of valid statements allowed.
...
     
-        def macro(s)(=value/keylist/function);
-    
-    * The macro specifies the name of the macro.
+        def <macro-name>[=value/keylist/function];
+
+    These statements specify which macros to be defined for the part to run and their values(optional).
```
allenjbaum, line 388: this is just a copy of what is above and can be deleted

spec/TestFormatSpec.adoc
```diff
@@ -368,30 +380,31 @@ There are two types of valid statements allowed.
...
+                specified for the csr. The length of the list returned is equal 
                 to _num_vals_. Each entry in the list is assigned to the corresponding 
                 macro listed on the left hand side of the `=` sign.
 
             - `getillegal(dependency_values_list, num_vals, key_list_for_field)` -> `list(int)`
                 This function is valid only for _WARL_ fields in the csrs. 
-                It returns a list of illegal values for the specified field when the 
-                values of the fields listed as dependency for the field in 
-                question on is equal to the _dependency_values_list_. The length of 
+                It returns a list of illegal values(random) for the specified field when the values
```
allenjbaum, line 405: add space between "values" and "("

spec/TestFormatSpec.adoc
```diff
@@ -368,30 +380,31 @@ There are two types of valid statements allowed.
...
-    * A function along with the arguments can also be specified. At runtime the function is called using the specified arguments and its return values are assigned to the macro(s) specified. The list of functions supported are as follows.
+    * A keylist can be used to extract the value from the input configuration files. 
+    * A function(with arguments) can also be specified. At runtime the function is called using the specified arguments and its return values are assigned to the macro(s) specified. The list of functions supported are as follows.
         
             - `getlegal(dependency_values_list, num_vals, key_list_for_field)` -> `list(int)`
                 This function is valid only for _WARL_ fields in the csrs. It 
-                returns a list of legal values for the specified field when the values 
-                of the fields listed as dependency for the field in question on is equal 
-                to the _dependency_values_list_. The length of the list returned is equal 
+                returns a list of legal values(random) for the specified field when the values
```
allenjbaum, line 397: add space between "values" and "("

spec/TestFormatSpec.adoc
allenjbaum: You use the syntax "=" for key, but ":=" in line 338 for value. 
The example
     misa>rv32>base>type=warl  
which would evaluates to "true" if the node exists, but  you could also write
    check misa>rv32>base>type:=warl
also evaluates to true also? OR is "=" used to check if a subnode exists and ":="  used for node with deeper structure?
Note line 343 and 351 are duplicates

spec/TestFormatSpec.adoc
```diff
@@ -368,30 +380,31 @@ There are two types of valid statements allowed.
...
     * These statements specify which macros to be defined for the part to run and their values(optional) in the following format.
     
-        def macro(s)(=value/keylist/function);
-    
-    * The macro specifies the name of the macro.
+        def <macro-name>[=value/keylist/function];
+
+    These statements specify which macros to be defined for the part to run and their values(optional).
+    * The _macro-name_ specifies the name(s) of the macro(s). Multiple comma separated names can be
+used to indicate multiple macros.
```
allenjbaum, line 390: You should say that multiple macro names can only be used in the case of function calls here

spec/TestFormatSpec.adoc
```diff
@@ -368,30 +380,31 @@ There are two types of valid statements allowed.
...
     * Multiple macros can be specified using a comma in-between them.
-    * A keylist specifying the path of the field whose value has to be passed as the value of the macro can also be given.
-    * A function along with the arguments can also be specified. At runtime the function is called using the specified arguments and its return values are assigned to the macro(s) specified. The list of functions supported are as follows.
+    * A keylist can be used to extract the value from the input configuration files. 
+    * A function(with arguments) can also be specified. At runtime the function is called using the specified arguments and its return values are assigned to the macro(s) specified. The list of functions supported are as follows.
```
allenjbaum, line 393: add space between "function" and "("

spec/TestFormatSpec.adoc
pawks: `=` is used to test for the presence of a subnode and `:=` is used to check whether the value of a particular node is exactly the same as the RHS. To put it simply, `=` tests for the presence of a key in a dictionary.

allenjbaum (approved):
This makes the changes I thought were needed for clarity

PR #318, opened 2023-02-28
Updated trap handler

### LINKED ISSUE #313: test compilation warnings
I'm getting following warnings:

```
riscv-arch-test/riscv-test-suite/rv32i_m/C/src/cadd-01.S: Assembler messages:
riscv-arch-test/riscv-test-suite/rv32i_m/C/src/cadd-01.S:37: Warning: shift count out of range (-1 is not between 0 and 63)
riscv-arch-test/riscv-test-suite/rv32i_m/C/src/cadd-01.S:47: Warning: shift count out of range (-1 is not between 0 and 63)
riscv-arch-test/riscv-test-suite/rv32i_m/C/src/cadd-01.S:62: Warning: shift count out of range (-1 is not between 0 an
```
...
is this expected?

UmerShahidengr:
This PR is co-authored by Allen Baum @allenjbaum. In this update, we have made the following changes:
- Split LI() macro for RV32 and RV64 to eliminate warning messages which should resolve #313 .
- Reduce the use of LA() macro by declaring fixed size offsets in save area.
- Updated RVTEST_GOTO_MMODE macro
- Updated \__MODE__\()trap_sig_sv, replaced an LA(rvtest_trap_sig) by calculating initial_Xtrap_sigptr + (Mtrap_sigptr-initial_Mtrap-sigptr) 
- Similar changes are made in chk_\__MODE__\()trapsig_overrun macro.
- Updated RVTEST_TRAP_SAVEAREA.


allenjbaum:
There were 2 primary reasons for this update.
The first is to reduce the reliance on the LA macro, which can dramatically reduce code in various places.
The second is to enable the use of cross-priv_level, cross VM-mode tests. When transitioning to a lower or high mode,
the test would like to continue operating where it changed modes - but in a different addressing routine. This change enables that by storing copies of the operating modes CODE, DATA, and SIG pointers in the respective local save area.
Then, one mode's EPC or TVAL can be translated to or from Mmode by looking at the difference between the BEGIN pointers and adding that to the  EPC or TVAL value, e.g. mEPC = sEPC+(mBEGIN-sBEGIN). 
Note that the prolog will initialize all of these to the Mmode values, but when a test establishes a new VM mode, these values (and the value store in xSCRATCH which points to DATA_BEGIN) need to be updated to the new mappings.

UmerShahidengr:
The latest commit is co-authored by [Abdul Wadood](https://github.com/Abdulwadoodd) who is working on Virtual Memory ACTs. He pointed out some bugs in trap handler while running some VM ACTs ([available here](https://github.com/Abdulwadoodd/riscv-arch-test/tree/vm_dev)). The latest changes are successfully setting up non-identity mapped page tables for both data memory and instruction memory.

allenjbaum:
Great, more evidence this is working. Thx.

Abdulwadoodd:
Update: Found a bug under `spcl_\__MODE__\()2mmode_test` label. The code does not switch to Mmode (upon calling RVTEST_GOTO_MMODE in S-mode with virtualization enabled), instead executes the trap when returns back to S-mode on subsequent instruction.

allenjbaum:
I'm not understanding the exact issue here.
RVTEST_GOTO_MMODE executes an ECALL with x2=0, which should cause a trap.
ECALL must not delegate, else you can't ever get to Mmode(that is a test
restriction)
That means it goes to the Mmode trap handler, regardless of which mode it
started in.

In spcl_\__MODE__\()2mmode_test:, it checks to see whether MCAUSE is in the
range 8..11: that's ECALL.
 if it isn't,  it  branches  to \__MODE__\()trapsig_ptr_upd: and proceeds
normally.
if it is, it further tests if x2 == 0.
if it isn't, it falls through to \__MODE__\()trapsig_ptr_upd:, and proceeds
normally.
 but if x2==0, it branches rtn2mmode (which exists only in Mmode) which
    figures out which mode GOTO_MMOFE was called from
    calculates the correct relocation amount from the calling mode's EPC to
Mmode's EPC
    adjusts EPC by that amount, and branches to the adjusted EPC contents+4
So it should never "return", it should branch, thus staying in Mmode..

If you have the correct version of the handler (318i, not 318h), GOTO_MMODE
should be setting x2 (NOT t2) to zero, and the spcl_handler should be
testing x2, (NOT t2).
That is the difference between the two versions

Abdulwadoodd:
> RVTEST_GOTO_MMODE executes an ECALL with x2=0, which should cause a trap.
ECALL must not delegate, else you can't ever get to Mmode(that is a test restriction)
That means it goes to the Mmode trap handler, regardless of which mode it
started in.

Agreed

> If you have the correct version of the handler (318i, not 318h), GOTO_MMODE
should be setting x2 (NOT t2) to zero, and the spcl_handler should be
testing x2, (NOT t2).

Yes, I'm using the correct version (318i), in which `RVTEST_GOTO_MMODE` set `x2` to zero.

**The problem is**: `x2` should not be used to track the call of the GOTO_MMODE macro, because `x2` is actually `sp`, it messes with the trap handler on `ecall` and gets overwritten before we make its comparison (to zero) to branch to `rtn2mmode`. 

Here is the sequence of events of what is actually happening:

1. We are in S/U mode, there is no delegation, and the RVTEST_GOTO_MMODE macro is called.
2. RVTEST_GOTO_MMODE macro sets `x2` to zero: sp=0.
3. The next instruction is `ecall` (by #define GOTO_M_OP), which causes the trap in Mmode.
4. The moment it jumps into the trap-handler (on label `trap_\__MODE__\()handler`), the first instruction it executes is `csrrw   sp, CSR_XSCRATCH, sp` which writes the value of sp (zero) to mscratch and reads the old value of mscratch (which is not happened to be zero) in sp. And here, the `sp` (or x2) is already overwritten.
5.  Later, after the execution of common_handler and common_entry, the program reaches to `spcl_\__MODE__\()2mmode_test`. 
    1. Here, it reads `mcause`  which contains 9 (exception number of ecall). Since it is in the 8..11 range so it **does not** branch to \__MODE__\()trapsig_ptr_upd.
    2. Then it checks if x2 (or sp) == 0. In this case, we are expecting it to be zero as it was set by RVTEST_GO_MMODE. But sp (or x2) has been overwritten by a non-zero address of mscratch (in Step 4). Therefore, it won't branch to `rtn2mmode` and falls through to \__MODE__\()trapsig_ptr_upd and proceed normally, which should not be the case.

If this interpretation is correct, `RVTEST_GOTO_MMODE` cannot work as intended until we replace x2 (or sp) with some other register (say t2) to track the GOTO_MMODE call, with the limitation that the tests cannot do t2==0 and ECALL.

allenjbaum:
Ooops - that is the register used in the existing ECALL tests; that's why
it was used.
That is rather... unfortunate.
I think we need to change that to t2, AND change the ECALL (and maybe
EBREAK test) to load some other register rather than x2.

allenjbaum:
I was trying to avoid changing ECALL for this fix, which was made because
someone used EBREAK to end the test,  which is what ultimately caused this
to begin with.
A  fix that avoids changing ECALL would be to retrieve x2 (which is sp)
from the trap_sv area ,
so inserting
  LREG    t2, trap_sv_off+7*REGWIDTH(sp)  // get test x2 (which is sp,
which has been save in the trap_sv area
before
   beqz   x2, rtn2mmode
 and changing ithat to be
   beqz   t2, rtn2mmode
should fix this problem. While we are at, we should restore as many
registers as we can, so adding the code from resto_<MODE>rtn
        LREG    t1, trap_sv_off+1*REGWIDTH(sp)
  //    LREG    t2, trap_sv_off+2*REGWIDTH(sp) <- this hold return address,
don't restore
        LREG    t3, trap_sv_off+3*REGWIDTH(sp)
        LREG    t4, trap_sv_off+4*REGWIDTH(sp)
        LREG    t5, trap_sv_off+5*REGWIDTH(sp)
        LREG    t6, trap_sv_off+6*REGWIDTH(sp)
        LREG    sp, trap_sv_off+7*REGWIDTH(sp)      // restore temporaries
just before the return jump restores all registers except the one holding
the return address
I'm attaching version j which has these fixes.

Abdulwadoodd:
> A  fix that avoids changing ECALL would be to retrieve x2 (which is sp)
from the trap_sv area ,
so inserting
  LREG    t2, trap_sv_off+7*REGWIDTH(sp)  // get test x2 (which is sp,
which has been save in the trap_sv area
before
   beqz   x2, rtn2mmode
 and changing ithat to be
   beqz   t2, rtn2mmode
should fix this problem.

This indeed has fixed the original problem. 

> //    LREG    t2, trap_sv_off+2*REGWIDTH(sp) <- this hold return address,
don't restore

Alright, it means `t2` cannot be restored and altered by a trap caused by GOTO_MMODE

> LREG    sp, trap_sv_off+7*REGWIDTH(sp) 

`sp` (which is x2) is also altered by the GOTO_MMODE (when it set x2 to zero) and its value (sp=0) is also updated into the save area by `trap_\__MODE__\()handler`. So, apparently, restoring `sp` doesn't actually restore the original value that may be set by the test, rather sp will always have zero after RVTEST_GOTO_MMODE.

So, RVTEST_GOTO_MMODE alters not only `t2` but `sp` as well. I suspect that altering the `sp`  is not intentional here, or is it?

allenjbaum:
Good catch- but it is OK.
It is important that x2 is restored to the value that it had when a  trap
occurs during a test. X2 is saved at trap entry (as SP),  and restored at
trap exit . So, it isn’t changed,
and that the important part.
The GOTO_MMODE macro explicitly changes x2==sp, and that value is restored
when it returns in Mmode.
so that  works, nit the fact that the macro does that isn’t noted in the
comments- and it should be.
If you can add that to the comments before you push the changes, that
should be good enough.

Abdulwadoodd:
The latest stable trap handler is pushed in the PR. In addition to the description, this PR adds:

- Infrastructure to support traps from S/U mode with Virtualization enabled.
- Added vmem segment in addition code, data, and signature segment in save area.
- Fixed #313 and  #332

allenjbaum:
Besides more accurate comments, this 
 - fixes redundant warnings when using the LI macro for RV32 (but they'll still be there for RV64, but only if there is an actual error as well), and 
 - changes the way GOTO_MMODE works, without using self-modifying code, and remaining compatible with existing ECALL and EBREAK tests, and handling the changing in virtual memory mapping. 
 It now has the side effect of clearing x2 (sp), and overwriting  t2 which will contain the address of the macro instantiation

allenjbaum:
Thanks so much! This is a big deal (for me, anyway).

UmerShahidengr:
All thanks to you @allenjbaum ,it was all your effort.

PR #319, opened 2023-03-07
Zfh and Zfinx Test are added

ptprasanna:
## Description

- Zfh and Zfinx Test are added
- Added half width into the NAN function to accomodate Zfh, as the base for Zfh is F which is also falls under the value as word
- Introduced two new flags (ZFINX and ZDINX) to aid the compilation of floating ops into integer regs

### Related Issues

NA

- [ ] Ratified

### List Extensions

- Zfinx
- Zfh

### Reference Model Used

- [ ] SAIL

### Mandatory Checklist:

Ran reports are placed here

https://gitlab.com/ptprasanna/actreports.git


pawks:
A quick glance at the generated signatures for Zfh tells me that all the results of the operations are cNaN values, which means the test is not loading the values correctly. Yet, the coverage reports are reporting 100% which means the implementation for coverage collection is also incorrect. Further, I doubt the RV32Zfinx tests will work for RV64 too. It is possible to run the same tests under RV64 by using macros appropriately.

ptprasanna:
> A quick glance at the generated signatures for Zfh tells me that all the results of the operations are cNaN values, which means the test is not loading the values correctly. Yet, the coverage reports are reporting 100% which means the implementation for coverage collection is also incorrect. Further, I doubt the RV32Zfinx tests will work for RV64 too. It is possible to run the same tests under RV64 by using macros appropriately.

possibly a macro update, as there are inconsistencies between ctg and archtest repos. But the tests looks to be ok.

pawks:
The entire floating point register should be loaded and stored for all the operations in the general case. Performing the memory operations using half-word operations does not sufficiently control the inputs to ensure compatibility.

allenjbaum:
I'm not quite sure what you're referring to here; you can't avoid loading
the entire register.
In both the Zfh case, and the Zfinx case, it always loads the entire reg,
though it only fetches a halfword from memory, right?
Do you mean you should be fetching the entire register width from memory,
using FLS  and LS (or FLD/LD) rather than  FLH and LH so you have more
control with what gets put into the upper bits?

pawks:
> Do you mean you should be fetching the entire register width from memory, using FLS and LS (or FLD/LD) rather than FLH and LH so you have more control with what gets put into the upper bits?

Yes. I was referring to this. It is necessary for the general test cases to load the entire registers in case of `Zfh` instruction. This will drastically reduce the number of coverpoints we need to test for specifically related to the NaN boxing behavior.

PR #345, opened 2023-05-04
Fix Compilation errors for privilege and Zifenci tests.

### LINKED ISSUE #343: Test that traps throws compilation error using latest GNU Toolchain
Privilege tests (and c.ebreak) throws error and cannot be compiled using the latest version of the toolchain. I'm using a pre-build nightly release (64-bit-elf-ubuntu) available [here](https://github.com/riscv-collab/riscv-gnu-toolchain/releases/tag/2023.04.29). 
### Reason
Whenever there is a trap, the trap-handler executes `fence.i` instruction [here](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/riscv-test-suite/env/arch_test.h#L903) (which is defined [here](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/riscv-test-suite/env/arch_test.h#L142)). The latest version of the toolchain expects `zifencei` defined as part of ISA string in `march` flag if we want to execute fence.i . And it is not defined in any of the tests under RVTEST_ISA (from where RISCOF picks up ISA string for march), even in Fencei.S test ([link](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/riscv-test-suite/rv32i_m/Zifencei/src/Fencei.S#L13)). Hence, toolchain throws the error on execution of fence instruction without using zifencei in `march` flag.

Abdulwadoodd:
### Related Issue

This PR fixes #343.

### Description
- The default definition of `RVMODEL_FENCEI` is set to `nop` instruction. This stops the compilation when running tests (using the latest GNU toolchain) on the implementation that does not support `Zifencei`. Implementation may change the definition of `RVMODEL_FENCEI` in the `model_test.h`. 
- The ISA string is updated for `Zifencei` tests (RV32 and RV64) which is required by `march` of the latest toolchain.

### DIFF
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -138,8 +138,19 @@
   #define EXCPT_CAUSE_MSK ((1<<4)-1)
 #endif
 
+//==========================================================================================
+// By default, it is defined as nop for the implementation that does not support Zifencei
+// Implementations that support Zifencei may use the fence.i instruction.
+// This only gets executed if xTVEC is not writable to point to the trap trampoline, 
+// and if it isn't writable, the model better have the zifencei extension implemented.
+//==========================================================================================
+
 #ifndef   RVMODEL_FENCEI
-  #define RVMODEL_FENCEI fence.i                                // make sure ifetches get new code
+  #ifndef ZIFENCE
+       #define RVMODEL_FENCEI nop
+  #else                            
+       #define RVMODEL_FENCEI fence.i 
+    #endif
 #endif
 
 #ifndef UNROLLSZ
@@ -900,7 +911,7 @@ init_\__MODE__\()tramp: /**** copy trampoline at mtvec tgt; t4->t2->t1  t3=end o
         bne     t3, t2, overwt_tt_\__MODE__\()loop      // haven't reached end of save area,  loop
 //----------------------------------------------------------------------
   endcopy_\__MODE__\()tramp:                    // vector table not writeable, restore
-        RVMODEL_FENCEI                          // make sure ifetches get new code
+        RVMODEL_FENCEI                          // By default it is defined as nop. See the definition
         csrr    t1, CSR_XSCRATCH                // reload trapreg_sv from scratch
         sw      t2, trampend_addr(t1)           // save copy progress
         beq     t3,t2, rvtest_\__MODE__\()prolog_done //full loop, don't exit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## [3.6.4] - 2023-05-04
+- In Zifencei test, updated the ISA string `zifencei` for `march` flag of toolchain.
+- Set the default definition of `RVMODEL_FENCEI` to `nop` in trap-handler.
+
 ## [3.6.3] - 2023-02-28
 - Split LI() macro for RV32 and RV64 to eliminate warning messages.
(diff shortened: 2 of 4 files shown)


pawks (approved):

riscv-test-suite/env/arch_test.h
```diff
@@ -138,8 +138,15 @@
...
+//==========================================================================================
+// By default, it is defined as nop for the implementation that does not support Zifencei
+// Implementations that support Zifencei may use the fence.i instruction.
+// This only gets executed if xTVEC is not writable to point to the trap trampoline, 
+// and if it isn't writable, the model better have the zifencei extension implemented.
+//==========================================================================================
+
 #ifndef   RVMODEL_FENCEI
-  #define RVMODEL_FENCEI fence.i                                // make sure ifetches get new code
+  #define RVMODEL_FENCEI nop
```
allenjbaum, line 149: The problem with this fix is that when an implementation supports Zifencei, and doesn't allow writing xTVEC, then tests may fail , because they execute a nop, and the test spec doesn't say anything about requiring DUTs to define  RVMODEL_FENCEI to be fence.i under this (or any other) circumstance.
What I think I'd like to see is (something like):
#ifndef   RVMODEL_FENCEI
  #ifndef ZIFENCE
       #define RVMODEL_FENCEI nop
    #else                            
       #define RVMODEL_FENCEI fence.i  
    #endif
#endif

pawks: I think its okay to have the default to be a `nop` as that is what the base spec supports. If the model implements non-coherent I&D streams and requires synchronization using fence.i, the model can define the RVMODEL_FENCEI macro as is. Using ZIFENCE under a `ifndef` in this case raises the question who defines that macro. Is it the implementation header file or riscof? I am comfortable with either of the solutions, but requires clear indication in the documentation.

allenjbaum: I agree that it is OK, in the sense that it is legal, but if they implement Zifencei, I don't want them to have to take the extra step of defining this (if they even notice that they need to and then have to go try to debug this, which won't be easy). Anything we can do to smooth the path is what we should be doing. The only reason that they should have to define this explicity is if they don't have coherent streams AND they don't implement Zifencei, in which case they can insert their own custom solution in place of that. Otherwise, they shouldn't even know about this. 
If it is custom, then it's the implementation header file; riscof will have no idea what their custom solution is.

Abdulwadoodd: Now there are two macros:
1. `RVMODEL_FENCEI`
2. `ZIFENCE`

> The only reason that they should have to define this explicity is if they don't have coherent streams AND they don't implement Zifencei

By "this" you mean `ZIFENCE` or `RVMODEL_FENCEI` or both? (which has to go into the documentation)
Can we do this, instead:
```
#ifndef RVMODEL_FENCEI
    #define RVMODEL_FENCEI nop
#else                            
    #define RVMODEL_FENCEI fence.i 
#endif
```
This does the same job as you are suggesting? (and it is working) but without using `ZIFENCE`

Abdulwadoodd: Please ignore my last comment.

> Otherwise, they shouldn't even know about this.

I understood why it is this way. Updating the commit.

Abdulwadoodd: Changes are upstreamed. Please review and merge

allenjbaum (approved):
This will do the trick

PR #352, opened 2023-05-16
Trap Handler and Test Macros Update

### LINKED ISSUE #336: Warning messages about missing CSRs from new macro code in arch_test.h
There's no explanation about how they should be set. 
Ie, if I create them and set them to empty, what does this do to the test sequence? Is this acceptable functionality?

riscv-test-suite/env/arch_test.h: Assembler messages:
riscv-test-suite/env/arch_test.h:749: Warning: RVMODEL_CLR_MSW_INT    not defined. Executing this will end test. Define an empty macro to suppress this warning
riscv-test-suite/env/arch_test.h:753: Warning: RVMODEL_CLR_MTIMER_INT not defined. Executing this will end test. Define an empty macro to suppress this warning
...

arch_test.h:
//==============================================================================
// Helper macro to set defaults for undefined interrupt set/clear
// macros. This is used to populated the interrupt vector table
//==============================================================================
//****************************************************************
#define RVTEST_DFLT_INT_HNDLR      j cleanup_epilogs
        //Mmode interrupts
#ifndef RVMODEL_SET_MSW_INT    
        .warning "RVMODEL_SET_MSW_INT    not defined. Executing this will end test. Define an empty macro to suppress this warning"
        #define  RVMODEL_SET_MSW_INT     RVTEST_DFLT_INT_HNDLR   
#endif
...

### LINKED ISSUE #337: Incorrect macro for register save in arch_test.h
Error: unrecognized opcode `sigupd( x1,x1)'

arch_test.h:
.macro RVTEST_SAVE_GPRS BASEREG REG_SV_ADDR     // optionally save GPRs
.option push
.option norvc
.set offset,0
  LA(     \BASEREG, \REG_SV_ADDR)               //this destroys basereg, but saves rest
  SIGUPD( \BASEREG, x1)
...
  
  Changing these to RVTEST_SIGUPD allowed compile to proceed

### LINKED ISSUE #338: LI macro throwing shift count out of range warnings for RV64
It appears that warning messages are being generated for shifts that are within the LI macro for the 64bit and beyond case, even they don't show up in the code. There is a warning for the very first instruction with a -1,  though the old and new code is exactly the same:

inst_0:
// rs2 == rd != rs1, rs1==x30, rs2==x31, rd==x31, 
// opcode: orn ; op1:x30; op2:x31; dest:x31; op1val:0xffffffffffffffff;  op2val:0x1
TEST_RR_OP(orn, x31, x30, x31, 0x0000000000000000, 0xffffffffffffffff, 0x1, x1, 0, x2)

riscv-test-suite/rv64i_m/B/src/orn-01.S:47: Warning: shift count out of range (-1 is not between 0 and 63)

Old generated code:
<inst_0>:
     fff00f13          	li	t5,-1
     00100f93          	li	t6,1
     41ff6fb3          	orn	t6,t5,t6
     01f0b023          	sd	t6,0(ra)
     
 New generated code:
 <inst_0>:
     fff00f13          	li	t5,-1
     00100f93          	li	t6,1
     41ff6fb3          	orn	t6,t5,t6
     01f0b023          	sd	t6,0(ra)

### LINKED ISSUE #339: Problem with nested ifndef inside #defined macro RVTEST_INIT_GPRS in arch_test.h
There is #ifndef RVTEST_E/#endif pair embedded inside the macro for RVTEST_INIT_GPRS that is intended to remove x16-x31 from initialization at the start of the test in the event rv32e architecture is selected.

This doesn't appear to work properly. 
No matter what the value is inside the ifndef (including changing it to a ifdef where I set the define just above) that code is not included. 
My dump file only shows the initialization of x1-x15.

Note that the macro for RVTEST_SAVE_GPRS uses a .macro assembler pragma rather than #define and it works properly, where x16-x31 are included.

### LINKED ISSUE #342: Trap and exit code includes non-RV32E compliant registers
The trap code in arch_test.h contains t3, 4, and 5 which are registers x28-x30, and do not exist for RV32E architectures

RVTEST_CODE_END uses t4 in abort_tests

Also, the init register code uses RVTEST_E, but the gpr save register code uses __RV32E__

### LINKED ISSUE #349: Getting misaligned exception with 64 bit CPU
Hi,

I got misaligned load exception inside exception handler for ebreak tests.

it was in this code in arch_test.h

...
/***************************************************/
/**** shared spcl int/excp dispatcher.          ****/
/**** JR to table in t3+t2, indexed by mcause   ****/
/***************************************************/
spcl_\__MODE__\()handler:               // case table branch to special handler code, depending on mcause
        LA(     t3, clrint_\__MODE__\()tbl)
        // load spcl int/excpt handler dispatch table 
        add     t3, t3, t2              // offset into the correct int/excpt dispatch table
        slli    t2, t5, 3               // index into 8b aligned dispatch entry and jump through it
        add     t3, t3, t2
        LREG    t3, 0(t3)  <<<<<<<<<<<<<< **this one got the exception**
spcl_\__MODE__\()dispatch:
        jr      t3                      // this defaults to resto_\__MODE__\()rtn

/**** this is the table of interrupt clearing routine pointers, which could include special handlers ****/
/**** They default to RVMODEL macros above, which are model supplied, then jump to rtn code          ****/
**?????????? <<< missing .align ??????**
clrint_\__MODE__\()tbl:                 //this code should only touch t2..t6
#ifdef rvtest_vtrap_routine  //  M/S/V/U
        .dword  resto_\__MODE__\()rtn           // int cause  0 is reserved, error
        .dword  \__MODE__\()clr_Ssw_int         // int cause  1  Smode SW int
        .dword  \__MODE__\()clr_Vsw_int         // int cause  2  Vmode SW int
        .dword  \__MODE__\()clr_Msw_int         // int cause  3  Mmode SW int

shouldn't the clrint_\___MODE__\()table have 8 bytes alignement enforcement with  .align ?

Abdulwadoodd:

## Description

This PR - co-authored by @allenjbaum :
- Resolves issues/bugs and adds requested features in the Trap handler. 
- Adds test macros for testing Virtual Memory in Sv32 mode. [(sample test)](https://github.com/Abdulwadoodd/riscv-arch-test/blob/vm_dev/riscv-test-suite/rv32i_m/vm/address_trans.S)

### Related Issues
- Resolves #338
- Resolves #349 
- Resolves #337 
- Resolves #339 
- Resolves #342 
- Resolves #336 

##

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## [3.7.0] - 2023-05-16
+- Updated the LI macro
+- Make Trap handler compatible for RV32E
+- Remove the warning messages [issue #336](https://github.com/riscv-non-isa/riscv-arch-test/issues/336)
+- Added Macros for testing Virtual Memory in Sv32 mode.
+
 ## [3.6.8] - 2023-06-22
 - Fix broken hyperlink in README
 
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -255,10 +255,11 @@
 #define MPP_SMODE  (1<<MPP_LSB)
 #define MPV_LSB    7    // bit pos of prev vmod mstatush.MPV in either mstatush or upper half of mstatus
 //define sizes
-#define tramp_sz        ((((XLEN + 3* NUM_SPECD_INTCAUSES + 6) * 4)+4) & -8) /* 16=common_xhandler to common_xentry, frc to dbvlwd size */
-#define ptr_sv_sz           (15*8)
-#define reg_sv_sz           ( 8*REGWIDTH)
-#define sv_area_sz      (tramp_sz + ptr_sv_sz +reg_sv_sz)
+#define actual_tramp_sz ((XLEN + 3* NUM_SPECD_INTCAUSES + 5) * 4)     // 5 is added ops before common entry pt
+#define tramp_sz        ((actual_tramp_sz+4) & -8)                     // round up to keep aligment
+#define ptr_sv_sz       (16*8)
+#define reg_sv_sz       ( 8*REGWIDTH)
+#define sv_area_sz      (tramp_sz + ptr_sv_sz + reg_sv_sz)           // force dblword alignment
 #define int_hndlr_tblsz (XLEN*2*WDBYTSZ)
 /*
 //#define sv_area_sz      (Msv_area_end-Mtramptbl_sv)           //sv_area start with aligned tramp_tbl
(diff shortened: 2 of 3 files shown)


UmerShahidengr:
The latest changes include the following changes:
(prolog)
* After moving the trampoline, store the final load address (trampoline+tramp_sz) into a new save area entry- that's the common entry point. 
(handler) 
* In the original trampoline, replace the auipc+immediate and instead load the entrypoint address from the save area
The pointer to the save area is already available, and when executed from the original trampoline, it will just branch to the next instruction - essentially a fall through.  
* This changes the size of the saved trampoline, so change tramp_sz definition to match
(save area); 
it's an explicit calculation because the assembler can't deal with label1-label2
* insert a new entry in the save area, and update all the #defines of offsets in header
* The definition of tramp_sz must be unrounded, because the address you want to save is exactly the common entry point - not one past.
So, define an actual_tramp_sz for the copy, and leave the rounded one for all other uses.

UmerShahidengr:
These changes are good to go now, @pawks can you please review and merge these changes?

pawks:
The changes look good. I've tested it for RV32I and RV64I. @Abdulwadoodd Please rebase/merge the main branch to resolve the conflicts and update the changelog entry accordingly.

Abdulwadoodd:
Thanks for the review @pawks. Conflicts are resolved now.

pawks (approved):
Looks good to me!

Abdulwadoodd:
@allenjbaum @pawks, Is there anything required for this PR to be merged?

allenjbaum:
I think this is ready to going. I'd just mrege it, but I'm not sure what the correct merge/squash & merge/rebase&merge option should be.

PR #357, opened 2023-05-31
Add tests for Atomic Memory Operation(AMO) instructions

ahadali-10x:

## Description

This PR
- Adds test macro for testing atomic memory operation instructions
- Adds tests for testing atomic memory operation instructions

- [x] Ratified
- [ ] Unratified

### List Extensions

This PR affects RISCV A extension

### Reference Model Used

- [x] SAIL
- [x] Spike

### Mandatory Checklist:

  - [x] All tests are compliant with the test-format spec present in this repo.
  - [x] Ran the new tests on RISCOF with SAIL/Spike as a reference model successfully?
  - [x] Link to Google-Drive folder containing the new coverage reports ([See this](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/CONTRIBUTION.md#uploading-test-stats) for more info): https://drive.google.com/drive/folders/1gL8FRx9japhWxGXgg4CWNqg-EHgaKGl-
  - [ ] Link to PR in RISCV-ISAC from which the reports were generated: < SPECIFY HERE > 

### Optional Checklist:

  - [x] RISCV-V CTG PR link if tests were generated using it: https://github.com/riscv-software-src/riscv-ctg/pull/69
  - [ ] Were the tests hand-written/modified?
  - [ ] Have you run these on any hard DUT model? Please specify the name and provide the link if possible in the description


pawks:
Please also provide the cgf files and reports in the google drive.

riscv-test-suite/env/test_macros.h
```diff
@@ -417,6 +417,16 @@ RVTEST_SIGUPD_F(swreg,destreg,flagreg)
     RVTEST_SIGUPD_FID(swreg,destreg,flagreg)	;\
     RVMODEL_IO_ASSERT_GPR_EQ(testreg, destreg, correctval)
 
+#define TEST_AMO_OP(inst, destreg, reg1, reg2, label, val1, val2, swreg, offset) ;\
+      LI(reg1, MASK_XLEN(val2))			;\
+      LA(reg2, label)                                 ;\
+      SREG reg1, (reg2)                               ;\
+      LI(reg1, MASK_XLEN(val1))			;\
+      inst destreg, reg1, (reg2)			;\
+      LA(reg2, label)                                 ;\
```
allenjbaum, line 426: Why is this instruction here? It is totally redundant unless reg1==reg2, though that might be an interesting corner case to test

allenjbaum (changes requested):
The macro is missing important coverage. 
First:  like store tests, you should be doing an AMO directly into the signature area. You don't need to read back the result then -- but you do need to store the result read from the AMO. You should also use the autoincrement feature of RVTEST_SIGUPD :

//(set swreg to point  signature area label and use it over and over)
//note that the macro shouldn't have an offset; SIGUPD keeps track of it automatically
// note that the macro shouldn't have a label; this uses the signature pointer as the label

LI(reg1, MASK_XLEN(val1))
SREG(reg, offset(swreg))            //init memory location in sig area; 
LI(reg1,MASK_XLEN(val2))
inst destreg, reg1, swreg            // dest<- (swreg), reg1 op (swreg)->swreg //perform the AMO
addi swreg, swreg, REGWIDTH // make sure SIGUPD writes into the following word
RVTEST_SIGUPD(swreg ,destreg) // write read value to  following sig entry & post increment the pointer

The signature should contain reg1 op reg2, followed by reg1 if everything was correct.
At the end, swreg is incremented by REGWIDTH, and so is offset, effectively ah increment of 2*REGWIDTH

allenjbaum (changes requested):
You've removed the store of the register state (the SIGUPD) so you're still missing the coverage. You need to store both the original value (which you're first storing into the signature), and the final value (which stays in the signature.) That mean storing the result register, which holds the initial value, into the signature also, like this:

#define TEST_AMO_OP(inst, destreg, reg1, tmpptr, val1, val2, swreg, offset)      ;\
      LI(reg1, MASK_XLEN(val2))                       /* ld value to init mem */              ;\
      RVTEST_SIGUPD(swreg , reg1,  offset)    /* store init val, update offset */ ;\
      LI(reg1, MASK_XLEN(val1)                         /* ld value to alter mem */           ;\
      ADDI(tmpptr, swreg, offset-XLEN             /* adj offset to pre-inc val */       ;\
      inst destreg, reg1, (reg2)                            /* perform atomic op */               ;\
      RVTEST_SIGUPD(swreg , destreg,  offset) /* store orig val, update offset */


You want this:

riscv-test-suite/env/test_macros.h
```diff
@@ -419,10 +419,12 @@ RVTEST_SIGUPD_F(swreg,destreg,flagreg)
 
 #define TEST_AMO_OP(inst, destreg, reg1, reg2, val1, val2, swreg, offset) ;\
       LI(reg1, MASK_XLEN(val2))			;\
-      addi reg2, swreg, offset                        ;\
-      SREG reg1, (reg2)                               ;\
+      RVTEST_SIGUPD(swreg, reg1, offset)              ;\
       LI(reg1, MASK_XLEN(val1))			;\
-      inst destreg, reg1, (reg2)
+      LI(reg2, MASK_XLEN(offset))                     ;\
+      add reg2, reg2, swreg                           ;\
```
allenjbaum, line 425: I think this is buggy. reg2 should have the original pointer value which is swreg+offset. But, first, offset must be in the range -2048..+2047, or the SIGUPD macro won't work, so you don't need the LI() macro+add, but you can use just an addi. Secondly: SIGUPD updates offset, so this is calculating the address of the signature value after the one you stored. Also: if you're adding macros  it should have reasonable parameter names, have comments,  and should use the minimal number of parameters and registers, e.g.:#define TEST_AMO_OP(inst, valreg, origptr, origval, updval, sigptr, offset) ;\ 
LI(valreg, origval)
addi origptr, sigptr, offset       /* calc where orig AMO will be stored */
RVTEST_SIGUPD(sigptr, valreg, offset    /* write orig AMO src */
LI(rvaleg,updval)
Inst valreg, valreg, (origptr)       /* orig->rwreg,  updated  val->orig mem
RVTEST_SIGUPD(sigptr, valreg)  /* write orig AMO val */

ahadali-10x: Fixed

allenjbaum:
I apologize - I got the ordering wrong, since sigupd post increments, but
first checks to see whether the offset is out of range and adjusts both
sigptr and offset if so.
So, you can't depend on sigptr+offset to work before SIGUPD (since offset
could be out of range) or after (because it has been post-incremented)
So, that line should be      addi origptr, sigptr, offset-REGWIDTH to
account for the post-increment

stnolting:
Just a question: what about the A extension's LR/SC operations? Will they also be part of this PR or will there (maybe) be another test set specifically for these instructions (maybe something like this: https://github.com/riscv-software-src/riscv-tests/blob/d4eaa5bd6674b51d3b9b24913713c4638e99cdd9/isa/rv64ua/lrsc.S)?

ahadali-10x:
> I apologize - I got the ordering wrong, since sigupd post increments, but first checks to see whether the offset is out of range and adjusts both sigptr and offset if so. So, you can't depend on sigptr+offset to work before SIGUPD (since offset could be out of range) or after (because it has been post-incremented) So, that line should be addi origptr, sigptr, offset-REGWIDTH to account for the post-increment

For the current tests, the offset is within the range, sigptr+offset will work.

ahadali-10x:
> Just a question: what about the A extension's LR/SC operations? Will they also be part of this PR or will there (maybe) be another test set specifically for these instructions (maybe something like this: https://github.com/riscv-software-src/riscv-tests/blob/d4eaa5bd6674b51d3b9b24913713c4638e99cdd9/isa/rv64ua/lrsc.S)?

No, those will not be included in this PR.

allenjbaum:
They may be for these specific tests - but if they going to be merged into
the repo, they may be reused, and it needs to be corrected now instead of
found later.
Or worse - not found later.

ahadali-10x:
> I apologize - I got the ordering wrong, since sigupd post increments, but first checks to see whether the offset is out of range and adjusts both sigptr and offset if so. So, you can't depend on sigptr+offset to work before SIGUPD (since offset could be out of range) or after (because it has been post-incremented) So, that line should be addi origptr, sigptr, offset-REGWIDTH to account for the post-increment

` addi origptr, sigptr, offset-REGWIDTH ` This will be the case when the offset is out of range but what if the offset is in the range then it will point to the wrong address (For the original source)

allenjbaum:
Good question.
In both cases, the original SIGUPD macro increments the offset by REGWIDTH
as the last thing it does. That incremented offset might be out of
range, or not.
So base+offset always points to the next element in the signature. The
-REGWIDTH is a correction to that, so it always points to the last item
that was stored,
and  - if it had been incremented to be out of range - , brings it back
into range.
The next SIGUPD will detect the out of range in the CHK_OFFSET function,
manually update base to be base+offset, and sets offset to zero.
(it's a bit more nuanced than that; it updates by the maximum it can, and
subtracts the maximum it can from offset, to cover the case where offset
was manual set to be>=2048)

ahadali-10x:
> So base+offset always points to the next element in the signature.

In the first SIGUPD call the offset is passed, so the src value will be stored at the base+offset address and the same address is stored in the rs1 register. In the second call, the offset is not passed because SIGUPD has incremented it in the previous call, so there is no need to do it manually.

allenjbaum:
No - first even if offset is explicitly passed, it is post incremented at the end of SIGUPD
Secondly, SIGUPD doesn't store anything in rs1; I think you meant that an explicit ADDI instruction does that

But that doesn't work for a subtle reason. Base+offset is indeed the address you want, but offset could be out of range (>2044 or 2040) because of the postincrement
The next SIGUPD will correct that before it does the store by adjusting the base up, and the offset down, so that's not a problem

What you need is a way to get the original value of base+offset into a register with it being guaranteed that offset is within range of an ADDI. It might be out of range of the first SIGUPD - so you can't calculate it there.
It also might be out of range for the second SIGUPD because of the postincrement in the first one. 
Call that one offset'
But, if it is, you know exactly how much out of range it will be: offset' = offset+REGWIDTH, or offset = offset'-REGWIDTH.
And you know that offset'-REGWIDTH will always be in range.

Abdulwadoodd:
> Please also provide the cgf files and reports in the google drive.

Hi @pawks, the tests are generated using CTG using. [Here](https://github.com/riscv-software-src/riscv-ctg/pull/69) is the PR link on CTG. 
The coverage artifacts are available in the [G-drive](https://drive.google.com/drive/folders/1gL8FRx9japhWxGXgg4CWNqg-EHgaKGl-).

ahadali-10x:
> Secondly, SIGUPD doesn't store anything in rs1; I think you meant that an explicit ADDI instruction does that

Yes, I meant that

>But that doesn't work for a subtle reason. Base+offset is indeed the address you want, but offset could be out of range (>2044 or 2040) because of the post-increment
The next SIGUPD will correct that before it does the store by adjusting the base up, and the offset down, so that's not a problem
What you need is a way to get the original value of base+offset into a register with it being guaranteed that offset is within range of an ADDI. It might be out of range of the first SIGUPD - so you can't calculate it there.
It also might be out of range for the second SIGUPD because of the postincrement in the first one.
Call that one offset'
But, if it is, you know exactly how much out of range it will be: offset' = offset+REGWIDTH, or offset = offset'-REGWIDTH.
And you know that offset'-REGWIDTH will always be in range.

I don't think that out-of-range offset is a problem. Because it is the limitation of the SIGUPD and it also handles it. `CHK_OFFSET` adjust it in such a way that` newbase + newoffset = oldbase + oldoffset`. So eventually the data is stored at the required location whether the offset is in the range or out of range.

allenjbaum:
No, because you'll be using that out-of-range offset in a load or store op,
or in an addi to compute the AMO address,
and the assembler will flag an error because the immediate, or offset , is
out of range.

ahadali-10x:
What about if we use the `CHK_OFFSET` before computing the AMO address it can solve this problem. It will adjust the offset accordingly.

ahadali-10x:
One more thing the range of the offset should be [0:4095] as its size is 12 bits.

allenjbaum:
no- the range of offset is -2048..+2047. It must match the offset in a
Ld/St instruction, since that is how it is used

ahadali-10x:
> no- the range of offset is -2048..+2047. It must match the offset in a
Ld/St instruction, since that is how it is used

Ok, I understand. What about adding `CHK_OFFSET ` before computing the AMO address?

UmerShahidengr:
There is one confusion in this discussion, 
In the macro definition, offset is passed as an argument which has been used in `addi origptr, sigptr, offset` instruction. This parameter is getting confused with the `offset` used in `RVTEST_SIGUP()` macro to keep track of the signature offset value.
Doing `addi origptr, sigptr, offset-REGWIDTH` will result into a wrong answer.

allenjbaum:
Hmm,yes,there is aliasing going on.
Some test macros have offset as an unnamed optional last parameter, e.g.
    #define RVTEST_SIGUPD(_BR,_R,...) ;\ <---"..." indicates optional
         .if NARG(__VA_ARGS__) == 1 ;\         <--- test to see if optional
parameter is present
             .set offset,_ARG1(__VA_OPT__(__VA_ARGS__,0)) ;\ <--  if so,set
offset to that optional parameter
          .endif <-- otherwise,just leave it alone
But others do not (see tests_store, which looks very similar)
This requires that every test using them has an explicit offset parameter,
and every macro in that test which has an optional offset parameter uses it
explicitly.
I think that is the case here.
We should probably fix that to be more general for all test macros,but it
is too late without rewriting a lot of tests, because offset needs to be
the last parameter for that to work.

UmerShahidengr:
Both approaches are good enough to use the signature file. RVTEST_SIGUPD() supports both approaches, and it depends on the user how he uses that macro (either provide offset himself or let the macro handle the offset itself). In all previous cases and even in this case, macro definitions are used by riscv-ctg to generate the tests, and the set of commands and coverpoint definitions have already been merged in ctg repo. CTG is generating the tests according to these macro definitions, changing macro here (for example, eliminating offset as argument to remove aliasing) will make macro arguments mismatch, and tests will stop working.
We can develop some policy for the future and document it properly to keep all future macros consistent but for this case, since ctg has already merged the changes and generated tests are using this macro definition which is working and offset is getting handled correctly, so there is no need to hold it back.

allenjbaum:
Yes, in the future we need to ensure that any use of "offset" in a test
macro be defined as a optional, last parameter.
I don't know what happens if "offset" is both a parameter name and a hidden
variable.
I don't know whether you can perform a .set offset, offset+REGWIDTH on
named parameters,
and whether that change persists when you exit the macro. This is a pass by
name vs pass by value issue, and I don't know what the assembler macros do.

ahadali-10x:
@allenjbaum Let me know if anything needs to be changed. Otherwise, this PR can be merged.

UmerShahidengr:
@allenjbaum when this PR is going to be merged?

allenjbaum:
I thought I left a comment on the TEST_AMO_OP code - but now I can't find it,,
I re-added it, but I'm duplicating it here make sure. This code is still broken.
In line 478, the signature address offset gets post incremented by REGWIDTH
Line 480 is supposed to calculate the pre-incremented address.
To do that, you need calculate base+(offset-REGWIDTH ), NOT just base+offset, which points past it.

Also: this requires that offset be an explicit parameter, and for ease of code generation, I want it to be optional, just as it is in SIGUPD.
To do that, you need to 
-change line 478 to be
     RVTEST_SIGUPD(sigptr, reg2) /*Write original AMO src */ ;
-change line 476 to be
    #define TEST_AMO_OP(inst, destreg, origptr, reg2, origval, updval, sigptr, ...) ;
-and finally add
  .if NARG(VA_ARGS) == 1 ;
     .set offset,_ARG1(VA_OPT(VA_ARGS,0)) ;
  .endif ;
immediately afterwards

ahadali-10x:
@allenjbaum Kindly review the changes and let me know if anything needs to be changed.

allenjbaum (approved):
This looks good now; let's ship it.

allenjbaum:
Well, what is there is good, but it doesn't meet the SOW.
1: No tests for unaligned atomics (which are supposed to trap, & therefore set RVTEST_MTRAP_ROUTINE)
There are actually two different trap causes: address misalign and access fault. Access fault has higher priority, and this can be tested by trying to store at an address that doesn't allow reads or writes (either a malformed address, or PMP or PTE permissions). If we're not counting on S or U modes being implemented, then we can only test this by using a physical address whose upper bits are non-zero. If we don't know the size of the PA, just set the PA to -1  The test should account for the value being intialized by explicitly writing each unaligned half, and adjusting  the signature address before to point to the unaligned address and after to be the aligned value.

2. There is also no coverage of LR/SC. While we can't test what happens with an interfering access, we can test what happens in all the other cases mentioned in the spec: too many instructions between LR and SC, load/store between LR and SC (both non-interfering and interfering, where interfering is accessing any byte in the implementation-defined reservation-set size, including when an unaligned address overlaps), LRs or SCs to different addresses between otherwise well formed LR/SC pairs, trapping between LR and SC (e.g. ecall or bkpt) etc.

PR #405, opened 2023-11-01
typo mistake resolved

Ammarkhan561:
Dear all,

I've noticed a typo mistake in the code, and I was expecting it to be addressed in the recent PRs. As it hasn't been updated, I'm submitting a separate PR to rectify this issue. Thank you!!


UmerShahidengr:
Effectively, that's not a typo. 2*sv_area_sz(sp) is reserved for Vmode (Its Hypervisor Mode save area), so using it for Umode will break a few things within trap handler, but there is a bug in GOTO_Lower_Mode macro as it is not going in Umode correctly.

UmerShahidengr:
@Ammarkhan561 I have updated the GOTO_Lower_mode macro, so this PR can be closed.

Issue #403, opened 2023-10-31
privilege spec 1.11 (2019) is generating mstatush CSR in trap handler which is not defined in that version.


animesh211:
Hi.
I am using default privilege version in riscof which is using version 1.11 (2019). With the latest version (3.8.0) I am seeing mstatush being present in Trap handler (which is not defined in version 1.11), due to which tests having exceptions are going to mstatush csr in trap handler, then taking illegal exception and not ending.
Can you please check on this once...

allenjbaum:
Ah, yes; that should only be accessed for RV32 && H-exstension implemented.

UmerShahidengr:
We can close this issue, as it has been resolved in PR #406

allenjbaum:
closing, as above

PR #406, opened 2023-11-01
Trap Handler Updates (Issue#403 & Goto_Lower_Mode macro updates)

UmerShahidengr:
## Description

Updated trap handler fixes two issues, one is to avoid using mstatush when using with Priv Arch 1.11 (as mstatush) is not defined in Priv 1.11. There are 2 places that reference mstatush, added the condition rvtest_vtrap_routine, which means H extension is present, and that isn't possible for priv 1.11. So, only 2 lines of code are changed. Second, there was a bug identified while looking at GOTO_Lower_mode macro, the save area was not being used properly when going from Mmode to Umode.

### Related Issues

Issue #403 , PR #405

### DIFF
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -849,7 +849,7 @@ RVMODEL_DATA_END        /* model specific stuff */
         /**** set MEPC to mret+4; requires relocating the pc   ****/
 .if     (\LMODE\() == Vmode)     // get trapsig_ptr & init val up 2 save areas (M<-S<-V)
         LREG    T1, code_bgn_off + 2*sv_area_sz(sp)
-.elseif (\LMODE\() == Smode)     // get trapsig_ptr & init val up 1 save areas (M<-S)
+.elseif (\LMODE\() == Smode || \LMODE\() == Umode)     // get trapsig_ptr & init val up 1 save areas (M<-S)
         LREG    T1, code_bgn_off + 1*sv_area_sz(sp)
 .else                            // get trapsig ptr & init val for this Mmode, (M)
         LREG    T1, code_bgn_off + 0*sv_area_sz(sp)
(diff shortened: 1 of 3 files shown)


riscv-test-suite/env/arch_test.h
```diff
@@ -1299,10 +1299,10 @@ common_\__MODE__\()excpt_handler:
         addi    T4, T4, 1
         andi    T4, T4, 4               
 // extract GVA into bit 1
-      #if (XLEN==32)
+      #if (XLEN==32 && rvtest_vtrap_routine)
```
allenjbaum, line 1302: I overlooked that this will use the equivalent bits in mstatus - but that they may be nonzero, which is wrong.
This needs to revert to the earilier code, and enclose lines 1302..1309 in a #if rvtest_vtrap_routine /. #endif pair

riscv-test-suite/env/arch_test.h
```diff
@@ -1659,7 +1659,7 @@ excpt_\__MODE__\()hndlr_tbl:            // handler code should only touch T2..T6
 rtn2mmode:
         addi    T4,T5, -CAUSE_MACHINE_ECALL
         beqz    T4, rtn_fm_mmode        /* shortcut if called from Mmode        */
-#if (XLEN==32)
+#if (XLEN==32 && rvtest_vtrap_routine)
```
allenjbaum, line 1662: I overlooked that this will use the equivalent bits in mstatus - but that they may be nonzero, which is wrong.
This needs to revert to the earlier code, and enclose lines 1662..1671 in a #if rvtest_vtrap_routine / #endif pair

allenjbaum (changes requested):
See comments; the two references to mstatush missed a condition and need to be separated from the RV32 case with a separate #if/#endif pair

allenjbaum (approved):
This fixes the non-existent mstatush issue for RV64, and ensure that the right code is generated even if H-extension doesn't exist

PR #381, opened 2023-08-30
Add read-only CSR MCONFIGPTR test

dansmathers:
<FOR DOC UPDATES FILL ONLY DESCRIPTION AND RELATED ISSUES SECTION AND REMOVE THE OTHERS>

## Description

mconfigptr is a read-only CSR defined in priv spec 1.12.  The test attempts a read/write (illegal instruction exception expected)/read.  The read data value is implementation dependent so the read data value is ignored.

### Related Issues

Requires sail update for signatures to pass: https://github.com/riscv/sail-riscv/pull/293

### Ratified/Unratified Extensions

- [X ] Ratified
- [ ] Unratified

### List Extensions

MCONFIGPTR exists in the priv spec 1.12 and used by the discovery task group.

### Reference Model Used

- [X] SAIL
- [X] Spike
- [ ] Other - < SPECIFY HERE >

### Mandatory Checklist:

  - [X] All tests are compliant with the test-format spec present in this repo ?
  - [X] Ran the new tests on RISCOF with SAIL/Spike as reference model successfully ?
  - [ ] Ran the new tests on RISCOF in [coverage mode](https://riscof.readthedocs.io/en/stable/commands.html#coverage)
  - [ ] Link to Google-Drive folder containing the new coverage reports ([See this](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/CONTRIBUTION.md#uploading-test-stats) for more info): < SPECIFY HERE >
  - [ ] Link to PR in RISCV-ISAC from which the reports were generated : < SPECIFY HERE > 
  - [ ] Changelog entry created with a minor patch

### Optional Checklist:

  - [ ] RISCV-V CTG PR link if tests were generated using it : < SPECIFY HERE >
  - [ ] Were the tests hand-written/modified ?
  - [ ] Have you run these on any hard DUT model ? Please specify name and provide link if possible in the description
  - [ ] If you have modified arch\_test.h Please provide a detailed description of the changes in the Description section above.

PR #409, opened 2023-11-15
Add menvcfg support (Zicbom_Zicboz_Zicsr_Ssdtso_Sstc)

cmuellner:

## Description

The {m,s,h}envcfg CSRs have been defined a long time ago. They contain bits that are defined by a the following ISA extensions: `Zicbom_Zicboz_Zicsr_Ssdtso_Sstc`.

This patch introduces a test which attempts to write all of the defined bits in `menvcfg` (similar to #381).  After the tests, the original value of the CSR will be restored to avoid side-effects with other tests.

Future work is to introduce the following additional tests:
* write tests for the bits in `senvcfg`
* write tests for the bits in `henvcfg`
* test that {`m`,`s`,`h`}`envcfg` CSRs are only accessible in the relevant modes

### Related Issues

NA.

- [x] Ratified
- [x] Unratified

Ssdtso is unratified.
The register itself and the other mentioned extensions are ratified.

### List Extensions

`Zicbom_Zicboz_Zicsr_Ssdtso_Sstc`.
Ssdtso draft2 can be found here: https://lists.riscv.org/g/tech-arch-review/message/183

### Reference Model Used

- [ ] SAIL
- [ ] Spike

### Mandatory Checklist:

### Optional Checklist:


allenjbaum (changes requested):
This test shows that specific bits can be set from Mmode, but doesn't test if they can also be cleared, and doesn't test that accessing it at all from S mode or U mode will trap properly.

There is no reason you shouldn't also be testing the SENVCFG and HENVCFG while you're doing it (you won't be able to run HENVCFG test on Sail yet, but you could on Spike, I think). It's more or less all the same tests, but wrapped with GOTO_LOWER_MODE and GOTO_MMODE macros, which you will need to the Mmode tests as well. You could even write the test once and specialize it for the modes you care about (see how I did this for the trap handler in arch_test.h)

allenjbaum (changes requested):
The macro definitions at the top are unused and/or inicomplete; I don't think this will compile without error. 

This test shows that specific bits can be set from Mmode, but doesn't test if they can also be cleared, (Actually, it doesn't really show that it can be set if it starts out set, so the "ori" should be "xori" .)
I'm unsure how throughly these need to be tested (i.e. do you only want to also test with CSRR/CSRC, and and/or also the immediate forms (for the lower 5 bits)
It also doesn't test that accessing it from S mode or U mode will trap properly. 

This doesn't test that the CSR is accessible or not in privilege modes lower than M, so those need to be added

There is no reason you shouldn't also be testing the SENVCFG and HENVCFG while you're doing this (you won't be able to run HENVCFG test on Sail yet, but you could on Spike, I think). It's just all the same tests, but wrapped with GOTO_LOWER_MODE and GOTO_MMODE macros, which you will need to the Mmode tests anyway. You could even write the test once and specialize it for the modes you care about (see how I did this for the trap handler in arch_test.h)

cmuellner:
Thanks for the comments and the help!
I've updated the PR to cover all three CSRs.
Further I've use the mentioned macros to switch the privilege levels.

However, I've not found out how to capture and acknowledge the illegal instruction exception.
Is there some code that can be used to get inspiration from?

allenjbaum:
If you define RVTEST_<x>TRAP_ROUTINE for x=M,S,V it will install trap
handlers for the appropriate priv levels.
The trap handlers will store all appropriate state and return to the next
double word aligned instruction, so you should follow any op that you
expect to trap with a couple of noops.
You also need to reserve enough space after the regular signature for the
trap signature, which is usually 4 words (interrupts, hypervisor exceptions
are different).
Note that GOTO_LOWER_MODE only works from Mmode, so you should use
GOTO_MMODE prior to using it.
You can get some inspiration from the misalign test in the priv folder.

cmuellner:
Thanks for the explanation! I was expecting some sort of call-back into the test file, where I would have to define a function that would take care of writing the exception cause into the signature. After reading through the trap handler code I understood that such call-backs are not necessary, since the default handlers capture all required state already.

I've reworked the PR as follows:
* New macros `RVTEST_CSR_{READ,WRITE}_ILLEGAL` that implement trap-able CSR read/write code (i.e. the necessary traps to ensure the trap will return within the macro's sequence).
* Split the three test files into one per run-mode and the illegal accesses further into read and write tests.
* The test files that don't expect illegal accesses don't disable the trap handler.
* Added the trap signature region at the end (gated by `rvtest_mtrap_routine`).

ved-rivos:
@cmuellner - I am finding that this code that attempts "relocation" in GOTO_LOWER_MODE does not lead to the right `mepc` value. This seems to be working for you. Was there anything specific you did that you could point out? 

```
        /**** set MEPC to mret+4; requires relocating the pc   ****/
        LREG    T1, code_bgn_off + 1*sv_area_sz(sp)
        LREG    T4, code_bgn_off(sp)
        sub   T1, T1,T4                /* calc addr delta between this mode (M) and lower mode code */
```
The T1 gets a value of 0 in my tests and T4 is rightly the start of where code is located. This then leads to a -ve offset being computed leading to a bad `mepc`. I could not figure out what is intended to go into `code_bgn_off + 1*sv_area_sz(sp)`

cmuellner:
> @cmuellner - I am finding that this code that attempts "relocation" in GOTO_LOWER_MODE does not lead to the right `mepc` value. This seems to be working for you. Was there anything specific you did that you could point out?
> 
> ```
>         /**** set MEPC to mret+4; requires relocating the pc   ****/
>         LREG    T1, code_bgn_off + 1*sv_area_sz(sp)
>         LREG    T4, code_bgn_off(sp)
>         sub     T1, T1,T4                /* calc addr delta between this mode (M) and lower mode code */
> ```
> 
> The T1 gets a value of 0 in my tests and T4 is rightly the start of where code is located. This then leads to a -ve offset being computed leading to a bad `mepc`. I could not figure out what is intended to go into `code_bgn_off + 1*sv_area_sz(sp)`

This "relocation" was introduced by @Abdulwadoodd as part of PR #318. Probably he knows what is going on.
Also tagging @UmerShahidengr @allenjbaum and @neelgala, which also participated in PR #318 and might have some insights.

ved-rivos:
I think I found the issue - this macro needs to be `rvtest_strap_routine` defined for the test to be able to use S mode.

cmuellner:
> I think I found the issue - this macro needs to be `rvtest_strap_routine` defined for the test to be able to use S mode.

Oh, I was not aware of this. I thought this was only needed for the trap handler that stores the exception reason in the signature.

But the whole thing is a bit unclear because why would a transition from M-mode to S-mode need S-mode traps? And what is needed for a transition from M-mode to U-mode?

Anyway, thanks for sharing this (presumably quite time-consuming) finding!

allenjbaum (dismissed):
These tests seem find, but there is no reason to have separate tests for the same source/dest pairs, e.g. henvcfg_u, henvcfg_read_illegal_u and henvcfg_write_illegal_u  should be just a single test. Also, only necessary to have a single nop after tests that (might) trap.

allenjbaum (approved):
approved after pulling on newest encoding.h updates

jordancarlin:
Right now these tests will only be run if an implementation supports all of Zicbom, Zicboz, Ssdtso, and Sstc, but really if any one of those extensions is supported then some of the fields should be tested.

allenjbaum:
Would this be just changes to the RVTEST_CASE and RVTEST_ISA macros so if any of the extensions are implemented, all tests should be run (there may be more traps than necessary, but that's all right)?
I know that multiple TEST_CASEs will OR the conditions; I don't know if that works for the ISA macro as well.

cmuellner:
Would it be sufficient to split the test cases into four files? E.g. `riscv-test-suite/rv32i_m/privilege/src/henvcfg_m.S` would become:
* `riscv-test-suite/rv32i_m/privilege/src/henvcfg_m_zicbom.S`
* `riscv-test-suite/rv32i_m/privilege/src/henvcfg_m_zicboz.S`
* `riscv-test-suite/rv32i_m/privilege/src/henvcfg_m_ssdtso.S`
* `riscv-test-suite/rv32i_m/privilege/src/henvcfg_m_sstc.S`

Just to be sure: are the new test macros ok (naming, implementation)?

The *envcfg CSR got a range of additional fields, which should be included here as well (that's a copy-and-paste exercise once everything else is settled).

allenjbaum:
Yes, splitting them up would be fine. Whether it would be necessary is a
separate question.
I think we would prefer that we have fewer tests with more test cases (more
more efficient for testing, as you spend quite a bit more time setting up
and tearing down the test for one bit than actually performing a test),
which is why I suggest combining them.
If we are losing coverage by combining them - that would argue for
splitting them up.
But, you could have a single test that just sequences through all the bits
in the CSR (in a loop, even), regardless of whether an extension is
implemented or not
(assuming that the result is deterministic). That would be a really short,
single test that would extend to future additional bits as well.
A test will either succeed or trap, and the correct result depends on
whether a particular extension is implemented or not.
The test doesn't care which result it is, it only cares that the
(identically configured) reference model did the same thing.

The downside of this approach is that we would have to dynamically
configure the signature check,
we can't just select one out of 2^N possible values (where N is the # of
extensions represented by CSR bits) when N is larger than a handful.

jordancarlin:
> But, you could have a single test that just sequences through all the bits in the CSR (in a loop, even), regardless of whether an extension is implemented or not (assuming that the result is deterministic). That would be a really short, single test that would extend to future additional bits as well. A test will either succeed or trap, and the correct result depends on whether a particular extension is implemented or not. The test doesn't care which result it is, it only cares that the (identically configured) reference model did the same thing.

That's the approach cvw-arch-verif takes. It attempts to write walking ones and walking zeros to most CSRs. Certain bits will be read-only 0 or cause a trap depending on the implemented extensions, but as long as the reference model matches the DUT, that should be fine.

PR #417, opened 2023-12-17
Add Zicfiss extension tests

ved-rivos:

## Description

> Implements tests for Zicfiss extension.
> Tests `sspush`, `sspopchk`, `ssamoswap`, `ssrdp`, `c_sspush`, and `c_sspopchk` instructions in S and U mode
> Specification: https://github.com/riscv/riscv-cfi/blob/main/cfi_backward.adoc

### Related Issues

> N/A

- [X] Ratified
- [ ] Unratified

### List Extensions

> Zicfiss: https://github.com/riscv/riscv-cfi/blob/main/cfi_backward.adoc

### Reference Model Used

- [X] SAIL
- [X] Spike

### Mandatory Checklist:

### Optional Checklist:

  - [X] RISCV-V CTG PR link if tests were generated using it : https://github.com/riscv-software-src/riscv-ctg/pull/94


riscv-test-suite/env/test_macros.h
```diff
@@ -1081,6 +1081,437 @@ ADDI(swreg, swreg, RVMODEL_CBZ_BLOCKSIZE)
...
+                                                           ;\
+  LI(x1, immval)                                           ;\
+  mv x5, x1                                                ;\
+                                                           ;\
+/* Four tests                                            */;\
+/* 1. Push using x1; popchk using x1                     */;\
+/* 1. Push using x1; popchk using x5                     */;\
+/* 1. Push using x5; popchk using x1                     */;\
+/* 1. Push using x5; popchk using x5                     */;\
+/* After each test ensure ssp back to initial value      */;\
```
allenjbaum, line 1232: I don't see where the check that "ssp is back to the initial value" is.
I see that you are comparing before/after, and adding the difference to x10,
but x10 (nor the difference in x11 )isn't savedni nthe signature

ved-rivos:
```
  mv swreg, x15                                            ;\
  RVTEST_GOTO_MMODE                                        ;\
  DISABLE_ZICFISS()                                        ;\
  RVTEST_SIGUPD(swreg, x10, offset) 
```
The last statement in this macro saves `x10` to the signature area.

allenjbaum: Ah, actually line 1318 for this macro, 1399 for the next.
 So, you sum up all the deltas, and save that at the end. That can hide a lot of sins. 
That's like testing lots of different add operation, combining all the results, and saving just that as the signature - which works, but not well (since compensating bugs can hide failures). And if there is a bug, you have no clue which invocation caused it. You need to replace "add x10, x10, x11" everywhere with "RVTEST_SIGUPD(swreg, x11) instead of summing them all up (you can't store the actual value because it might be loaded differently between reference and DUT). You would also need to make sure the explicit offset is handled correctly by changing the second ".if offset==0" to be ".if offset<6" I think. And. allocate more signature space, of course.

ved-rivos: Sorry I missed this comment. I have updated the test macro to retain full history by replacing `add x10, x10, x11` with
     `seqz x11, x11; or x10, x10, x11; sllw x10, x10, 1`

jordancarlin:
I am closing the remaining ACT3 PRs. We will refer to these when creating the ACT4 tests for these extensions. Thanks for contributing!

PR #431, opened 2024-01-27
Add tests for Smrnmi extension

ved-rivos:

## Description

> Add test for Smrnmi extension

### Related Issues

> N/A

- [X] Ratified
- [ ] Unratified

### List Extensions

> https://github.com/riscv/riscv-isa-manual/blob/main/src/rnmi.adoc

### Reference Model Used

- [X] SAIL
- [ ] Spike

### Mandatory Checklist:

### Optional Checklist:

  - [X] Were the tests hand-written/modified ?


allenjbaum (commented):
In general, this doesn't follow the structure of any other tests I've seen (e.g. doesn't use SIGUPD to update the signature, etc and doesn't use the standard trap handler (which make sense because this defines a new priv level for handler, though we could augment it to do that reasonably easy, especially if MPRV is ignored)). It isn't clear from the spec how custom the encodings of cause are, or if they are exactly the same as all the other priv levels

ved-rivos:
I have updated the tests to use the SIGUPD macros and to use the common exception trap handler.

allenjbaum:
I'll have to look at the ecall and ebreak tests; they were written a
very long time ago.
There really aren't any cases that can't use SIGUPD, sequential or not, but
that may require either that explicit offsets be used in the macro (which
CTG automatically inserts),
and/or that RVTEST_BASEUPD is invoked (especially if traps can occur
multiple times within loops).

Don't you still have the issue that there is no defined way to get an
actual RNMI handler, because the entry point is not architecturally defined?
I don't think it can be the contents of any existing CSR,

I'd like to see comments that explain the conditions that cause an RNMI
exception (but no an interrupt) to occur. Interrupt testing requires
resources that haven;t been identified.

And: the current trap handler saves EPC, CAUSE, TVAL, STATUS (+ an
interrupt vector if it was an interrupt)
- but there is no trap handler that saves the RNMI versions of those
registers (yet), only those for the modes it knows about.
For now, you can write your own handler, but eventually we need to add that
to the standard trap handler and have it installed (if variable
rvtest_ntrap_routine is defined), along with it's own prolog/epilog/stack.
It's a bit tricky if you are mixing different priv levels in the text,
because each priv level must share the pointer to the trap signature, and
each level may have a different VM mapping.
The standard trap handler takes care of all that (and will for RNMI when
the new priv mode is added to the handler).

*coverage*:
Can the rnmi handler be invoked directly from S/HS/VS/U/VU modes as well as
Mmode (assuming no delegation at any level, and any other conditions are
met)?
If so, for coverage, we need tests for each of those + tests for every
possible architectural cause .
And we need to define a standard way to invoke the trap handler, given that
the trap vector is completely implementation dependent.
This will need to be part of the BOOTROM image that gets loaded, but we
should define a standard way to do that (which will likely include linker
scripts).

But the standard handler assumes that interrupts and exceptions target the
same address, and are differentiated by the MSB of the xcause CSR.
As far as I can tell, that is not a requirement of this extension (i.e.
they may target different entrypoints) so we need to take that into account
for coverage
(which means we need to be able to tell which entrypoint was used; that
requires duplication of the trampoline, and saving more vector offset

ved-rivos:
You are right that we cannot have an ACT for the NMI trap handler itself because there is no architecturally defined way to set it up. Since such handlers may be embedded in the platform in a manner where they cannot be modified it may not be possible to test for exceptions in the NMI handler. I needed that code to test the Sail model but should not have been part of ACT. I have removed that.

allenjbaum:
Well, my reading of the spec is you can cause an exception, but not an
interrupt, though I'm not completely clear on the details.
What needs to be provided to the model is still the  platform-defined
interrupt and vector addresses.
Eventually we will have an event generator that models can use to stimulate
external events (with predictable timing) such as interrupts, and even
snoops.
The individual platforms will need to provide a shim to interface those
events to their specific models, just as the need shims to start and stop
tests, and to extract signatures.

allenjbaum:
I don't think it outside the scope of ACTs; they already expect to be
provided with platform defined parameters; these addresses are just another
one (or two).
More problematic is how to generate interrupts (and in rare cases,
exceptions), and ACTs have a provision for that in the model supplied
macros each implementation must include to run a test.

For interrupts, specifically, the intent is to define a standard memory
mapped event generator that any test can use to generate events
such as interrupts (which could be presented to the DUT as wires or as a
write), or memory reads and write (which might be presented to the model as
a snoop).
The writes would be synchronized using instret signals from DUT.
It would be the DUT's responsibility to convert outputs from the test (DUT
MMIO writes) into the event generator and outputs from the events generator
(wires and write) into DUT inputs

allenjbaum:
I agree with part of this. I do expect the NMI handler would be to a fixed
address. It is also possible that the handler is in ROM.
This configuration exists for normal interrupts and traps as well (although
we can determine what the address is by reading xTVEC CSR).
Arch-test does deal with this simply by testing what it can test, and just
giving up if it can't.
We still have tests, though; we don't avoid writing tests if there is a
possibility of being able to use them.

In theory, we can still test an RTL core and require (in order to certify
compatibility)
that either the vector address points to memory that is writable, (and
discoverable as a command-line or YAML variable),
or that it writable (via a "callback" backdoor, which may be something as
simple as an MMIO write)
We can't enforce that while testing real silicon - but we can for the
purposes of  compatibility .

NMI may represent error conditions, but if RTL can't assert an error
condition in some way - then it can't be tested by the vendor either.
If the vendor asserts that it supports NMI, we would require a callback to
assert the NMI as well -
(just as we expect that we expect that the RTL model will have
external-interrupt input wires that the proposed event generator will talk
to)

That is not to say that, even if those callbacks and addresses are
available, or if NMI can be asserted by the test, we can test  it right now.
We would need additions to the trap handler to ensure that.

ved-rivos:
> NMI may represent error conditions, but if RTL can't assert an error condition in some way - then it can't be tested by the vendor either. If the vendor asserts that it supports NMI, we would require a callback to assert the NMI as well 

Thats not always true - especially for products that are not soft IPs. NMIs may be triggered in the platform by sources such as thermal diodes when they detect a catastrophic thermal excursion impending. Other reasons for a NMI getting triggered may be things like the DDR command/data parity errors to request the NMI handler which runs out of ROM and special SRAM and hence not affected by a DDR failure to be able to do a failover to a mirror DIMM. Many cores have tons of DFT features - including ability to inject hardware errors that trigger NMI - using DFT hooks. But these DFT hooks are not exposed to general software - including M-mode software and using these facilities usually needs silicon vendor level debug authorization.

allenjbaum:
Hmm - this might be a certification vs compatibility difference.
Compatibility generally runs on RTL models, vs. running on HW.
We can still run on real HW (it's been done) but we can't test those
features in the cases you describe - which are legitimate cases, obviously.
Still - that does not prevent compatibility tests to be written that can
have access to what is effectively a wire into the core.
We probably can't test that all xCAUSE CSR values of RNMI are properly
handled (since those values are custom,
but we can test the basic functionality of one of them. Not perfect
coverage, but we can still get basic functionality.

We can't do that yet. We need the same event generator device
functionality that we will use for other asynchronous event testing
(but which we don't have resources to implement yet)

allenjbaum (approved):
This looks like it tests everything reasonable to test given the current infrastructure

PR #432, opened 2024-01-29
Add Zawrs extension tests

ved-rivos:

## Description

> Add Zawrs extension test.

### Related Issues

> NA

- [x] Ratified
- [ ] Unratified

### List Extensions

> https://github.com/riscv/riscv-zawrs

### Reference Model Used

- [X] SAIL
- [ ] Spike

### Mandatory Checklist:

### Optional Checklist:

  - [X] Were the tests hand-written/modified ?


allenjbaum (commented):
I would add a bit more coverage to ensure this is doing the right thing, e.g. read the INSTRET CSR before and after and see that it incremented by the proper amount.
You could be tricky and, in the timeout cases, check mtime or cycles and see that they at least incremented (the timeout should be provided to the test in terms of cycles (probably) and the test should check that it incremented by more than one, for example.

Silabs-ArjanB:
> You could be tricky and, in the timeout cases, check mtime or cycles and see that they at least incremented (the timeout should be provided to the test in terms of cycles (probably) and the test should check that it incremented by more than one, for example.

The Zawrs specifies its wrs.sto referring to an _implementation-define short timeout_. You cannot assume that _cycles_ is incrementing during that timeout (it might for example be clock gated as is allowed for WFI implementations) and also _mtime_ might not increase.

ved-rivos:
That is right. Adding such checks to the ACT will go beyond what architecture compatibility requires.

allenjbaum:
Well, not really. We expect DUT to provide a number in terms of
instructions retired or cycles or mtime, guaranteed to be larger than the
timeout,
and we can check that no interrupt has occurred when we inject an interrupt
after that time. The test pass criteria is a function of that number that
is provided to the test,
but that condition is true for a lot of tests with implementation defined
parameters.

jordancarlin:
#1745 adds Zawrs tests to the ACT4 framework. They cover everything this PR covers (plus many additional things). Thanks for the initial work on this. Closing this one.

PR #439, opened 2024-02-26
Atomic extension LR/SC tests and coverpoints

maarij-10xe:
- [x] Ratified
- [ ] Unratified

### Reference Model Used
- [x] SAIL
- [ ] Spike

### Mandatory Checklist:
  - [x] Ran the new tests on RISCOF in [coverage mode](https://drive.google.com/drive/folders/14LSkJ1vH9ygq1qCN95nIsEv80CUjoIpg?usp=sharing)
  - [x] Link to Google-Drive folder containing the new [coverage reports](https://drive.google.com/drive/folders/136KrrsVS7snWDmYWznggU60gc9lmiuRO?usp=sharing)
  - [x] Link to PR in RISCV-ISAC from which the reports were generated : [RISCV-ISAC PR#80](https://github.com/riscv-software-src/riscv-isac/pull/80)

### Optional Checklist:


UmerShahidengr:
@allenjbaum please review this PR when you are free. It is completed from our side but it has been stalled for a few weeks

davidharrishmc:
A few thoughts reviewing the PR visually.

I don't see these tests writing to signature memory.  How do we know if they failed or succeeded?

There are no lr.w and sc.w instructions tested on RV64, even though the ISA should support these.

Many of the tests try to write a 0 value to memory to prove that memory changed.  0 is special number and might have been written by an incorrect implementation.  Would it be safer to write a random number rather than 0?

allenjbaum (changes requested):
As with all test that involve storing to  memory, the write to memory should either be directly to the signature area, or should be re-read from whereever it is stored and written to the signature region. Storing directly to the signature area is the preferred method. 
In addition, the store conditional updates a destination register, and that must be stored into the signature as well. The RVTEST_SIGUPD macro should be used (unless you are storing non contiguous addresses)
These tests only load/store from the data region, which isn't acceptable. Storing a changing value (rather than zero)  is preferred (e.g. an incrementing value is fine).

A constrained loop must " exclude loads, stores, backward jumps, taken backward branches, JALR, FENCE, and SYSTEM instructions" and the compressed extension equivalents. Because we can't predict whether an implementation will pass or fail with an unconstrained loop, we are forced to only test with those constraints .These tests will need to be rewritten to exclude those and also be 16 instructions long ( I think those 16 instructions include both the LR and the SC, and any backwards branch for retry - and we shouldn't test anything longer than that because implementations are allowed to succeed.  Also, we shouldn't expect that SCs to a different address or length (in the same reservation set) will fail - they are allowed to succeed, so can't test those either.


Note that after each loop, the  SC should attempt to store into a different signature word, as should the Rd register.  I think this means that if the loop contains a store to something outside the reservation set, we don't know if will pass or fail, so we can't test that - but we can test something that is to the same reservation set and see that it fails.
Let's meet and talk about these tests

UmerShahidengr:
These tests need rewriting, will be assigned to somebody soon.

jamesbeyond:
ACT repository have been updated `act4` as the default branch,  ACT4 have the tests for Atomic extension LR/SC covered already. So this PR can be closed. Thanks for your contribution.

Issue #448, opened 2024-03-27
Zfh's 'flh-align-01' test is executed even if Zfh is not enabled


stnolting:
I am using the latest version of this repository for a custom RISCOF port. The test fails because the `Zfh/src/flh-align-01.S` test is executed although the `Zfh` ISA extension is not enabled.

* My test run of RISCOF: https://github.com/stnolting/neorv32-riscof/actions/runs/8446548698/job/23135504941#step:8:234
* The port's ISA config: https://github.com/stnolting/neorv32-riscof/blob/main/plugin-neorv32/neorv32_isa.yaml#L3

I am not sure, but maybe there is a typo in the `flh-align` test's regex list:

https://github.com/riscv-non-isa/riscv-arch-test/blob/8a0cdceca9f0b91b81905eb8497f6586bf8d1c6b/riscv-test-suite/rv32i_m/Zfh/src/flh-align-01.S#L32

which (I think) should also include `F` and `Zfh` as in the remaining `Zfh` test cases:

https://github.com/riscv-non-isa/riscv-arch-test/blob/8a0cdceca9f0b91b81905eb8497f6586bf8d1c6b/riscv-test-suite/rv32i_m/Zfh/src/flt_b1-01.S#L32

davidharrishmc:
flh_align-01.S contains 

RVTEST_ISA("RV32IF_Zicsr_Zfh,RV64IF_Zicsr_Zfh")

When I build the tests with riscof, this prevents running this test on a system without Zfh.

If you are building the tests with a different process that examines RVTEST_CASE rather than RVTEST_ISA, maybe the RVTEST_CASE should be more complete.  It should be harmless to make RVTEST_CASE more comprehensive.

I'd suggest you make a PR with the change.  If it's hard for you to make this PR, I can make it for you.

davidharrishmc:
The mistaken assertion is that not all tests have the same RISCV_ISA and RVTEST_CASE parameters. flh is an outlier, which is causing the problem.

flh-align-01 has

```
RVTEST_ISA("RV32IF_Zicsr_Zfh,RV64IF_Zicsr_Zfh")
...
RVTEST_CASE(0,"//check ISA:=regex(.*I.*);def TEST_CASE_1=True;",flh-align)
```

while other test cases such as flt_b1-01.S has

```
RVTEST_ISA("RV32IF_Zicsr_Zfh,RV64IF_Zicsr_Zfh")
...
[...]

PR #451, opened 2024-04-09
[flh-align-01] fix ISA config in RVTEST_CASE macro

stnolting:
Fixing #448

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+## [3.8.13] - 2024-03-26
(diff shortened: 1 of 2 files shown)


allenjbaum (approved):
This fixes an outlier in Zfh tests

stnolting:
> Oops - this won't go through unless you also add something a new entry in the changelog file

Done.

allenjbaum (approved):
Fixes the change log entry. Note that when we get the dev branch process going, maintainers will be responsible for the changelog, using committ messages for all staged PRs for that release.

PR #464, opened 2024-05-17
Cleanup

Abdulwadoodd:

## Description

- fli.* tests got selected by the RISCOF even if `zfa` extension is not included in the ISA string in yaml configuration file. RVTEST_CASE macro is updated to check for `zfa` extension in the ISA string to select the test.
- If DUT is configured for RV32, RISCOF tries to select and throws an error for most of `rv64i_m/D_Zfa` and `rv64i_m/F_zfa` tests that says `tests selected without enabling the appropriate extensions`. These tests are for RV64 - hence RVTEST_CASE macro is updated to check for RV64 in ISA string to select these tests.
- Empty in one of the macros inside test_macros.h results in scary assembler warnings. so removed those spaces.

:warning: It is important to note that the fixes in RVTEST_CASE macro for Zfa tests are not pulled in the corresponding cgf files. If the tests are generated from these GCF files again, they will have the same issues again. The best is to fix those cgf files or there should be an issue with open tracking (but I'm not sure where, either in riscv-arch-test or riscv-ctg?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## [3.9] - 2024-05-17
+- Cleanup: Fix RVTEST_CASE macros for Zfa tests.
+- Fix warning assembler warning message from test_macros.h
+
 ## [3.8.20] - 2024-05-08
 - Updated the Zcmop extension
 - Add Zimop extension.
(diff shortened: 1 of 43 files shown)


allenjbaum (commented):
This fixes an issue we discussed. Thanks for not regenerating all the tests! We can see exactly the exact same change in every test.

allenjbaum (approved):
As I said earlier, this fixes an issue we discussed. We can see exactly the exact same change in every test, and a fix that removes a warning. Thanks for not regenerating all the tests!

Abdulwadoodd:
Thanks Allen, for the review and merge!

PR #469, opened 2024-06-09
Updated Trap Handler

UmerShahidengr:
This PR is coauthored by @allenjbaum, trap handler file (arch_tests.h) has the following changes:
- converted one of the CANARY words to a delta instret count (there is a variable that will enable that)
- converted CODE/DATA/SIG_BEGIN/END to include all the little incidental code, so the tests template improves
- minor bug fixes to the trap handler in the cases of traps delegated to S-mode with virtualization enabled.
- rewrote the identity map macro. The macro now takes 3 parameters: BASE, LVLS, PERMS and generates a page of PTE entries with the most significant PPN counting from 0..511 (or 1023 if RV32). The macro takes into account the 2 different PTE formats, based on  XLEN. Whereever the macro is used, it defaults BASE to 0 (so VA==PA) and uses ALL_PERMS


allenjbaum (approved):
This passed Sail vs. Spike, so it should be good to go.

allenjbaum (dismissed):
This passed the Sail vs. Spike tests, so it should be good to go

jamesbeyond (approved):
Reviewed in the ACT SIG meeting, good to go

PR #473, opened 2024-06-11
Fixed reversed order of Zicboz and Zicsr in macros for cbo.zero

davidharrishmc:

## Description

### Related Issues

- [x] Ratified
- [ ] Unratified

### List Extensions

### Reference Model Used

- [ ] SAIL
- [x] Spike

### Mandatory Checklist:

### Optional Checklist:

### DIFF
--- a/riscv-test-suite/rv32i_m/CMO/src/cbo.zero-01.S
+++ b/riscv-test-suite/rv32i_m/CMO/src/cbo.zero-01.S
@@ -20,7 +20,7 @@
 // 
 #include "model_test.h"
 #include "arch_test.h"
-RVTEST_ISA("RV32IZicsr_Zicboz")
+RVTEST_ISA("RV32IZicboz_Zicsr")
 
 .section .text.init
 .globl rvtest_entry_point
@@ -30,7 +30,7 @@ RVTEST_CODE_BEGIN
 
 #ifdef TEST_CASE_1
 
-RVTEST_CASE(0,"//check ISA:=regex(.*32.*I.*Zicsr.*Zicboz.*);def TEST_CASE_1=True;",cbozero)
+RVTEST_CASE(0,"//check ISA:=regex(.*32.*I.*Zicboz.*Zicsr.*);def TEST_CASE_1=True;",cbozero)
 
 RVTEST_SIGBASE(x2,signature_x2_1)
 
--- a/riscv-test-suite/rv64i_m/CMO/src/cbo.zero-01.S
+++ b/riscv-test-suite/rv64i_m/CMO/src/cbo.zero-01.S
@@ -20,7 +20,7 @@
 // 
 #include "model_test.h"
 #include "arch_test.h"
-RVTEST_ISA("RV64IZicboz")
+RVTEST_ISA("RV64IZicboz_Zicsr")
 
 .section .text.init
 .globl rvtest_entry_point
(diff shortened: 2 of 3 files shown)


davidharrishmc:
Note that Sail does not yet support cbo.zero, so Sail throws an illegal instruction exception when attempting to simulate it during riscof compilation.  Consequently I have been unable to full test this.  However, it should fix the problem Tim Hutt raised about the commit.

https://github.com/riscv-non-isa/riscv-arch-test/commit/e3c5471f58573ea99cd282197251e7c453ba1696#r143001732

riscv-test-suite/rv32i_m/CMO/src/cbo.zero-01.S
```diff
@@ -20,7 +20,7 @@
 // 
 #include "model_test.h"
 #include "arch_test.h"
-RVTEST_ISA("RV32IZicsr_Zicboz")
+RVTEST_ISA("RV32IZicboz_Zicsr")
```
Timmmm, line 23: This is inconsistent with the other file. I couldn't work out exactly what this macro does but they should probably be consistent at least.

Timmmm (changes requested):
Thanks!

I have a PR to add cbo.zero support to Sail here if you want to test it:

https://github.com/riscv/sail-riscv/pull/455

However I'd say it's not necessary for you to test it - I made the same changes locally and it selects the test now & passes. (I would have made a PR but you were faster than our PR approval process!)

riscv-test-suite/rv32i_m/CMO/src/cbo.zero-01.S
UmerShahidengr: This macro is used to select the test from the test pool, both combinations `RV32IZicsr_Zicboz` and `RV32IZicboz_Zicsr` are doing the same thing, so this order does not matter for the case of `Zicboz`. I would recommend updating the RVTEST_ISA string for RV64 test as well. It will be good to go after updating the same line in RV64 variant of the test.

UmerShahidengr: I have done the modifications, now this one is good to go.

UmerShahidengr (approved):

UmerShahidengr:
@jamesbeyond please have a look at it, this one is good to go, so you may merge this one

Timmmm (approved):
Thanks!

riscv-test-suite/rv64i_m/CMO/src/cbo.zero-01.S
davidharrishmc, line 1: I added Zicsr to the RV64 cbo.zero RVTEST_ISA string for consistency with the other tests per @UmerShahidengr 's comment.

I'm not the author of cbo.  It appears that the cbo tests don't use CSRs and that the Zicsr would likely be unnecessary.  I believe somebody from my team had experimented with that months ago and presumably removed the Zicsr in just this case, but I agree that it is better to be consistent.  It's hard to envision a system with caches but no CSRs, so Zicsr is likely harmless, and if the tests are ever generalized to non-machine mode, Zicsr will be nessary to support menvcfg and the other modes.

@liweiwei90 was the original author of PR #226 with cbo.zero tests.  I will defer to the author if they decide to remove Zicsr from cbo.zero at a later point. If so, I think it should come out of RVTEST_ISA and RVTEST_CASE for both rv32i_m and rv64i_m.

jamesbeyond (approved):
LGTM

PR #476, opened 2024-07-01
WIP of Sdtrig proof of concept test.

rtwfroody:
WIP


riscv-test-suite/rv32i_m/Sdtrig/src/mcontrol6_read.S
```diff
@@ -0,0 +1,89 @@
...
+RVMODEL_BOOT
+RVTEST_CODE_BEGIN
+
+#ifdef TEST_CASE_1
+  RVTEST_CASE(1,"//check ISA:=regex(.*32.*);check ISA:=regex(.*Sdtrig.*);def rvtest_mtrap_routine=True;def TEST_CASE_1=True;",svadu_sv32)
+
+  RVTEST_SIGBASE(x1, signature_x1_0)
+
+  csrw  CSR_TSELECT, zero
+  li    t1, (CSR_TDATA1_TYPE_MCONTROL6 | \
```
allenjbaum, line 30: please use the LI macro (instead of the li pseudo op). This forces the code generation to be a known size, independent of toolchain.

allenjbaum:
Scary size of defines.h. If this test can cause a trap, you need to #define
rvtest_mtrap_routine.
This will instantiate the trap handler that tests use. The default trap
handler "simply" (hah) records trap state and returns the next doubleword.
If you need a special handler for these tests, there are hooks to do that.
There are macros already defined that set up identity page tables as well

rtwfroody:
Thanks for the early feedback, Allen. I'm still learning my way around this test framework.

> Scary size of defines.h.

This is auto-generated from the same source as the debug spec is generated. So it contains more than we need, but eliminates any possibility of errors when generating these constants.

allenjbaum:
I understand about the .h file. I'm also very impressed that it was
autogenerated.

What scares  me is (just scrolling through it very fast) is seeing the
word "option[al]" in there,
which means you need to account for th[os]e optional behavior[s] in both
the test and in Sail, and you need to define a variable that you can pass
to Sail that indicates the option that was chosen (defining it in the
riscv-config YAML file is a start), and probably test them in combination
as well.

rtwfroody:
> What scares me is (just scrolling through it very fast) is seeing the word "option[al]" in there, which means you need to account for th[os]e optional behavior[s] in both the test and in Sail, and you need to define a variable that you can pass to Sail that indicates the option that was chosen (defining it in the riscv-config YAML file is a start), and probably test them in combination as well.

That's fair. That language is all in the spec. In hindsight maybe the spec should not be so flexible so it's easier to test. I'm afraid that ship has sailed.

allenjbaum:
Including a list of all the possible options, and giving them names, in a
single table will go a long way to avoid future problems (and questions
from people complaining they don't understand why some implementation does
X when they expected Y)

rtwfroody:
This needs https://github.com/riscv-software-src/riscv-config/pull/183 applied to riscv-config for riscof to actually run the test.

Issue #458, opened 2024-04-23
Zfh tests missing float-to-float conversions


davidharrishmc:
rv32i_m/Zfh is missing float-to-float conversions:

fcvt.h.s
fcvt.h.d
fcvt.s.h
fcvt.d.h

UmerShahidengr:
PR #572 has resolved this issue, closing this one as #572 has been merged

PR #483, opened 2024-07-30
add fcvt testcases

Pagerd:
## Description

- Add floating-to-float conversions in Zfh tests

### Related Issues

> [#458 ](https://github.com/riscv-non-isa/riscv-arch-test/issues/458)

- [✓] Ratified

### List Extensions

- Zfh

### Reference Model Used

- [✓] SAIL

### Mandatory Checklist:

  - [X] Ran the new tests on RISCOF in [coverage mode](https://riscof.readthedocs.io/en/stable/commands.html#coverage)
  - [X] Link to Google-Drive folder containing the new coverage reports): < [coverage reports ](https://drive.google.com/drive/folders/1FeRqbMKqqz739rmwALl0Wzap2DYWER5-) >

### Optional Checklist:


allenjbaum:
This needs to be run in coverage mode, and have the coverpoint file pointed
to. Coverpoints that have comments are highly appreciated.

Pagerd:
update the coverage reports

allenjbaum:
I'm not seeing a pointer to the .cgf files, as requested. Am I just missing it somewhere?

Pagerd:
Sorry about missing .cgf files, I have uploaded a new compressed package in google drive

allenjbaum:
We have a little problem here. You did include the tests, the test results, you've added the coverpoints, but       you didn't include the coverage report.
IF you had, you may have noticed that there is a lot of missing coverage, because you just copied a particular integer coverage file - and these ops all involve floating point numbers.     There is no coverage for floating point values at all, in either integer or float point formats.
You need to go back and look at how other floating point tests have defined their cover points, and how integer tests have, and craft coverpoints that match the opcodes you're testing, e.g.      for int->float conversions, you want integer value  around powers of 2 +/- offsets that will cause different roundings that bump the exponent up or down (including to infinity or to zero).    (and you want to test the ops with all the different rounding modes)
   for float-> int, you want to to something similar: test floats that are near min/max integer values (including around +/-zero) +/-small offsets, and a few others (Naans, infinity,)
The cgf file you chose is incomplete (e.g. it tests max positive/negative value, 0, 1... but never -1, which is an enormous coverage miss. Also not your fault, but all those coverpoints need to be rewritten/augmented.That cgf also includes reference to rs3, which don't even exist in the operations you're testing, so you'll never get full coverage, another reason you'll need to be more careful crafting the coverpoints.
The FP tests check for FP values in categories such as ibm_b23, and can be found reasonably well documented in the ISAC repoThere should be an equivalent for the Int dataset so the definitions don't have to get repeated over&over- also not your problem.

allenjbaum:
Oh, also, this won't get merged into main, it must be merged into the dev branch

Pagerd:
I understand, pr will be temporarily closed until i resolve these issues.

PR #487, opened 2024-09-03
Draft : Added test cases for load address match trigger in sdtrig extension

anuani21:
## Description

    -  Handwritten test cases for mcontrol6 trigger using tcontrol register.
    -  Test cases for mcontrol6 address  match trigger for load operation with different set of match values(0,1,2,3,4,5).  
    
### Related Issues
NA

- Unratified

### List Extensions

 - Sdtrig

### Reference Model Used

-  Spike

### Mandatory Checklist:

Ran reports are placed here

https://gitlab.com/ptprasanna/actreports/-/tree/main/Sdtrig/RV32/Load_Address_match_trigger?ref_type=heads


riscv-test-suite/rv32i_m/sdtrig/src/load_match0.S
```diff
@@ -0,0 +1,134 @@
...
+  RVTEST_SIGUPD(x1, t0)
+  
+    # Read tinfo and append it to the signature
+   csrr t6, tinfo
+   RVTEST_SIGUPD(x1, t6)
+  
+    li a1, (CSR_TCONTROL_MTE_ENABLED  << 3)
+    csrw tcontrol, a1
+    csrr a2,tcontrol
+    RVTEST_SIGUPD(x1,a2)
```
rtwfroody, line 54: tcontrol isn't required to be implemented.

There are several possible solutions to this, but just assuming it is implemented doesn't seem right.

riscv-test-suite/rv32i_m/sdtrig/src/load_match0.S
```diff
@@ -0,0 +1,134 @@
...
+    RVTEST_SIGUPD(x1,t2)
+    csrr a3,tdata1
+    RVTEST_SIGUPD(x1,a3)
+   
+    
+    # Set the value of tdata2 to the address to match
+    la t2, 0x80001234
+    csrw tdata2, t2
+    csrr t6,tdata2
+    RVTEST_SIGUPD(x1, t6)
```
rtwfroody, line 77: Section 5.7 of the Debug Spec says:
As a result, a debugger can write any supported trigger as follows:
1. Write 0 to tdata1. (This will result in containing a non-zero value, since the register is WARL.)
2. Write desired values to tdata2 and tdata3.
3. Write desired value to tdata1.

riscv-test-suite/rv32i_m/sdtrig/src/load_match0.S
```diff
@@ -0,0 +1,134 @@
...
+    
+    # Code that accesses the matched data value (this should trigger the break)
+        # Execute an instruction at the matching address
```
rtwfroody, line 80: Does this comment apply here?

riscv-test-suite/rv32i_m/sdtrig/src/load_match0.S
```diff
@@ -0,0 +1,134 @@
...
+        la t4,0x80001234
+	lw t4, 0(t4)       # Load to trigger match (this should trigger the break)
+
```
rtwfroody, line 83: This falls through to the trap handler if the trigger didn't fire. Not the end of the world because mepc will have the wrong value, but it's unusual. Deserves a comment at least.

rtwfroody (changes requested):
I've never written anything for this framework, so some of my comments might stem from lack of understanding. Is there a document that shows how test cases should adapt to different hardware configurations?

allenjbaum:
I'll have to look at this, but there's a lot that doesn't seem right.
The normal reason for writing MPP is generally to return to a different
priv mode.
This is much trickier than it appears, because if VM is enabled, you'll
return to the address in MEPC which will be in a different translation mode
than Mmode (usually),
The standard trap handler deals with all the complexity setup by the test
if you define Mtraproutine and Straproutine.

It sets up mtvec/stvec to point to the handler (if possible; xTEVC is WARL
so you don't know if the value you want to write  is legal)
- but if it isn't, the handler will swap out the code that xTVEC points to,
and will swap it back in when it returns from the test)
It also sets up identity mapped page tables (it assumes 1:1) but there are
macros for easily defining different ones (e.g. offset from 1:1)
The  test will need to initialize the code begin, data begin, and signature
begin if different than bare mode

Finally, there are predefined macros that will let tests transition from
M->lower mode and lower mode->Mmode seamlessly (assuming you've set the
begin/end variables correctly).
It also takes care of ensuring the signature values are independent of
where the code is actually placed by the linker scripts.
The trap handler normally returns to the next doubleword offset following
the instruction that traps so code can take different paths
depending on whether it trapped or not, so if you don't care, always put
two noops after an instruction that might trap (whether it is supposed or
not)

riscv-test-suite/rv32i_m/sdtrig/src/execute_match0.S
```diff
@@ -0,0 +1,137 @@
...
+  and t2,t2,t1     
+
+  bnez t2, implement_mie
+  beqz t2 , implement_tcontrol
+
+implement_mie:                                   
+    li t1,(1<<3)
+    csrw mstatus, t1        
+    csrr t4, mstatus        
+    RVTEST_SIGUPD(x1, t4)
```
rtwfroody, line 32: Since these tests test triggers, but not explicitly test native re-entrant issues, I think it's better not to add this to the signature. Then these tests can compare DUTs and references that have different ways of enabling triggers.

But I might be misunderstanding the philosophy behind these tests.

riscv-test-suite/rv32i_m/sdtrig/src/execute_match0.S
```diff
@@ -0,0 +1,137 @@
...
+
+resume_code: 
+   # Write 0 to tselect, read back and append it to the signature
+  csrw tselect, zero
+  csrr t0, tselect
+  RVTEST_SIGUPD(x1, t0)
+  
+  # Read tinfo and append it to the signature
+  csrr t6, tinfo
+  RVTEST_SIGUPD(x1, t6)
```
rtwfroody, line 50: For similar reasons, do we want tinfo in the signature? I think we only care that the specific trigger type being tested is supported.

riscv-test-suite/rv32i_m/sdtrig/src/execute_match0.S
```diff
@@ -0,0 +1,137 @@
...
+
+  csrw tdata1, zero
+  csrr a3, tdata1
+  RVTEST_SIGUPD(x1, a3)
```
rtwfroody, line 54: This is another one. The spec only requires that the trigger is now disabled, but there are many encodings that could represent a disabled register.

riscv-test-suite/rv32i_m/sdtrig/src/execute_match0.S
```diff
@@ -0,0 +1,137 @@
...
+RVTEST_DATA_BEGIN
+  .align 12
+   page_4k:
+  .fill   4096/REGWIDTH, REGWIDTH, 0
+RVTEST_DATA_END
+
+.align 12
+rvtest_Sroot_pg_tbl:
+  .fill   4096/REGWIDTH, REGWIDTH, 0
+.section .data
```
rtwfroody, line 112: Do you need these?

riscv-test-suite/rv32i_m/sdtrig/src/execute_match0.S
```diff
@@ -0,0 +1,137 @@
...
+
+#ifdef rvtest_mtrap_routine
+mtrap_sigptr:
+  .fill 128*4, 4, 0xdeadbeef
+#endif
+
+#ifdef rvtest_gpr_save
+gpr_save:
+  .fill 32*(XLEN/32), 4, 0xdeadbeef
+#endif
```
rtwfroody, line 131: I think you can also delete these.

riscv-test-suite/rv32i_m/sdtrig/src/execute_match4.S
```diff
@@ -0,0 +1,113 @@
...
+  
+  # Read tinfo 
+  csrr t6, tinfo
+
+   # Set the value of tdata
+   csrw tdata1, zero
+   csrr a3, tdata1
+ 
+  # Set the address for tdata2 where the trigger should occur
+   la t2, 0x80008000
```
rtwfroody, line 55: I think this matches when executing an instruction where bit 15 of the address is set. I don't think you're guaranteed what address this program is placed at. Probably you need to not use a fixed address for your test here, but instead pick an address in your code.

E.g. (untested)
```
la t2, break_here
li t3, 0xffff
and t4, t2, t3
li t3, 0xffff0000
or t4, t3, t4
csrw tdata2, t4
...
j before_break
...
before_break: nop
break_here: nop
```

rtwfroody (changes requested):
I don't understand how the GT/LT execute tests work. You're using 0x80001234 for the compare value for both. But the code that sets the breakpoints must be either before or after that address. Wouldn't one of these two scenarios fire before tdata1/tdata2 are added to the signature?

Also, there's a *lot* of repeated code here. Refactoring things so there's a single macro/function that just takes the desired tdata1 and tdata2 values would make each test tiny (call the function to set up, then execute whatever code to test triggering/not triggering).

rtwfroody (approved):
This works for me.

pdonahue-ventana (approved):
I think that this looks like an OK set of tests for basic Sdtrig functionality. It's difficult to test everything because pretty much everything is optional so the possibilities are endless. But I do appreciate that, for instance, the tests that check match=2 (greater than or equal) will execute properly whether match=2 is supported or not (and the signature will indicate whether it's supported which is what really matters).

anuani21:
@rtwfroody or @pdonahue-ventana,

I had missed to update the icount test cases  in the PR for review. Can you please review it and tell me the comments?

riscv-test-suite/rv32i_m/sdtrig/src/test_icount_eq1_s_u_mode.S
```diff
@@ -0,0 +1,78 @@
...
+
+    la t0, user_mode                
+    csrw mepc, t0
+    mret                    #upon mret switch to switch to user mode
+    
+user_mode:           
+    csrr t1, mstatus               
+    li t0, (0 << 11) 
+    and t1, t1, t0              
+    csrw mstatus, t1
```
rtwfroody: This whole thing is just equivalent to `csrr mstatus, zero`. Why are you doing that? In user mode mstatus isn't even writable.

riscv-test-suite/rv32i_m/sdtrig/src/test_icount_eq1_s_u_mode.S
```diff
@@ -0,0 +1,78 @@
...
+    csrr t1, mepc
+    csrr t6,mepc
+    csrr t2, mtval
+    RVTEST_SIGUPD(x1, t0)
+    RVTEST_SIGUPD(x1, t6)
+    RVTEST_SIGUPD(x1, t2)
+
+    la t0, user_mode                
+    csrw mepc, t0
+    mret                    #upon mret switch to switch to user mode
```
rtwfroody: So when we're in user mode starting at user_mode, execute one instruction, and then the trigger happens. Doesn't the trigger send us right back to trap_handler? Why doesn't this repeat forever?

riscv-test-suite/rv32i_m/sdtrig/src/test_icount_eq1_s_u_mode.S
```diff
@@ -0,0 +1,78 @@
...
+    
+user_mode:           
+    csrr t1, mstatus               
+    li t0, (0 << 11) 
+    and t1, t1, t0              
+    csrw mstatus, t1              
+    
+    # Execute one instruction in U-mode
+    li a0, 42
```
rtwfroody: The instructions in user mode should be something like:
```
addi a0, zero, 1
addi a0, a0, 1
addi a0, a0, 1
addi a0, a0, 1
```

That way you can tell by the value of `a0` how many instructions were executed before the trigger hit. (Also `a0` would have to be zero'd out before returning to user mode.)

rtwfroody (changes requested):

allenjbaum:
You do realize there is already a standard trap handler that is
automatically loaded by defining rvtest_mtrap_routine in you test (and
rvtest_strap routine if it traps into smode)
It stores xepc, xcause, xtval (for the appropriate "x") and returns to the
next aligned doubleword access.
There are also standard goto_Mmode and goto_lower_mode macros already
defined. You should use them.
Installing the standard trap handler will update the trap vector to point
to standard trap handler,
or (if it isn't writable, which is legal) saves and overwrites the existing
handler entry point. All registers and the existing handler are
restored when the test halts.

You should use those, because it will handle configurations that you aren't
considering with this simple test.

riscv-test-suite/rv64i_m/sdtrig/src/icount/test_icount_eq1_s_u_mode.S
```diff
@@ -0,0 +1,92 @@
...
+  RVTEST_SIGBASE(x1, signature_x1_0)
+   
+  CHECK_MISA_AND_SETUP_TRIGGERS()
+   
+   li t2, (CSR_TDATA1_TYPE_ICOUNT << 60 |\
+           CSR_TDATA1_DMODE_BOTH << 59  |\
+           CSR_ICOUNT_ACTION_BREAKPOINT << 0 |\
+           (1 << 10) |\
+           CSR_ICOUNT_M |\
+           CSR_ICOUNT_S |\
```
rtwfroody: You're configuring this trigger to fire in M and S Mode, but then below there's some comment about executing one instruction in U-Mode. How does this work?

jamesbeyond:

jjscheel:
@jamesbeyond, @UmerShahidengr, @jordancarlin, there was significant work that went into this PR.  sdtrig will be part of what's needed for Certification.  I'd recommend that HMC look at the testcases for the new work.

davidharrishmc:
This looks sensible.  It has mcontrol6 tests for loads, stores, and executes, and one test for icount.
However, it does not look like there was a systematic effort to capture and exercise all of the normative rules
related to Sdtrig.

I would put this at about 25% of the test-writing work needed to get to certifying triggers, and of course 0% of the work
for writing a test plan, coverpoints, normative rules, etc.

Some things that seem to be remaining include:
* etrigger and itrigger
* data value matching
* chaining
* exercising all of the match fields
* exercising the size fields
* exercising privilege levels
* Operate from S-mode via a SBI
* Exercising DUTs that have more than one trigger register
* Handle DUTs that don’t implement all of the triggers

Tenstorrent owns Sdtrig.  Shall we put this on hold pending their delivery?

David

jjscheel:
> I would put this at about 25% of the test-writing work needed to get to certifying triggers, and of course 0% of the work
for writing a test plan, coverpoints, normative rules, etc.

It was intentionally known to be short of coverage.  But, what was done had work not only from testers but the specification writers to ensure it was not a drive-by set of tests.  Thus, I'd like to see it provide some value.

jordancarlin:
I am closing the remaining ACT3 PRs. We will refer to these when creating the ACT4 tests for these extensions. Thanks for contributing!

PR #513, opened 2024-10-08
Performance fix for PyYAML comments

Timmmm:
PyYAML has really really bad performance when adding comments, so I changed it to only add them for the first 100 entries. I also removed the l+i offset which doesn't make sense because the type is a dictionary.


Timmmm:
This is a reapplication of https://github.com/riscv-software-src/riscv-isac/pull/89

Please could someone review it so I don't have to recreate it again?

UmerShahidengr (approved):
Thank you @Timmmm , it is good to go.

Timmmm:
Thanks!

PR #517, opened 2024-10-11
fix ext_specific_vars bug

Pagerd:
## Description

fixed CTG issue when try to generate 32H tests

when try to generate 32h tests like `fcvt.d.h` ,ctg will display KeyError: 'inxFlag'
this pr will fix this bug


UmerShahidengr (approved):
A simple bug fix. Thank you @Pagerd. It is good to go

PR #521, opened 2024-10-14
Update Sail installation method in CI

jordancarlin:
Use the new binary release of the Sail compiler instead of installing it using Opam. This matches the sail-riscv repository CI and reduces the CI time from 19 minutes to 13 minutes.


UmerShahidengr (approved):
Looks good to go.
Thanks @jordancarlin

PR #522, opened 2024-10-14
Run Multiple CI jobs in parallel

jordancarlin:
@UmerShahidengr As discussed in #518, this splits CI into multiple jobs that run in parallel. This offers a further speedup from 13 minutes down to 10 minutes.


UmerShahidengr (approved):
Great job. Thank you @jordancarlin

PR #526, opened 2024-10-16
Cache Spike and Sail in CI

jordancarlin:
@UmerShahidengr More CI enhancements. 

This caches the builds of Spike and Sail and only rebuilds them if there is a newer version since the last time the CI was run. If no newer version exists, it restores the cached version and entirely skips building that tool. When it can restore from the cache (which should hopefully be most of the time) CI is down from 10 minutes to less than 4 minutes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -94,18 +94,41 @@ jobs:
           path: ${{ github.workspace }}/spike
           key: ${{ steps.cache-spike-restore.outputs.cache-primary-key }}
 
+      - name: Get Latest Sail Commit
+        run: |
+              SAIL_HASH=$(git ls-remote https://github.com/riscv/sail-riscv.git HEAD | awk '{ print $1}')
+              echo "SAIL_HASH=$SAIL_HASH" >> "$GITHUB_ENV"
+
+      - name: Restore cached Sail
+        id: cache-sail-restore
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ github.workspace }}/sail
+          key: sail-${{ env.SAIL_HASH }}-RV${{ matrix.xlen }}
+
       - name: Install Sail
+        if: steps.cache-sail-restore.outputs.cache-hit != 'true'
         run: |
           sudo mkdir -p /usr/local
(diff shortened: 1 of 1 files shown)


jordancarlin:
@UmerShahidengr Additionally, if we use the `[jobs] ` parameter in the config files for riscof to run multiple tests at once we can get the CI down to less than 3 minutes.  Not sure if you want to modify the config files in the riscof-plugins directory since they are templates. If not, we could also probably do it with sed or just leave it as one job if you'd prefer.

jordancarlin:
Closed and reopened PR to force CI to run a second time and use the cache

UmerShahidengr:
Wonderful job @jordancarlin. 
Regarding your comment on [jobs], its not a bad thing to do, since we are also looking forward to cover the floating point tests (RV32/64F/D), and all other remaining tests, and those tests take a lot of time, so it will be better to do more parallelization.

UmerShahidengr:
The eventual shape of this CI will be to run all tests available in the repo, and check coverage of all tests against the existing coverpoints using riscv-isac. For checking coverage, we will be a file (a sh file) which will contain all coverage commands for all extensions available in this repo. Since, checking coverage can take a lot of time thats why I have not added it in the CI yet. We can work together to improve this CI more, and add a new section of checking the coverage in parallel to these runs.

UmerShahidengr (approved):
Good to go. Great work @jordancarlin

jordancarlin:
> Wonderful job @jordancarlin. Regarding your comment on [jobs], its not a bad thing to do, since we are also looking forward to cover the floating point tests (RV32/64F/D), and all other remaining tests, and those tests take a lot of time, so it will be better to do more parallelization.

Sounds good. Would you rather I change the templates in riscof-plugins or should we leave those alone and just modify them for CI?

jordancarlin:
> The eventual shape of this CI will be to run all tests available in the repo, and check coverage of all tests against the existing coverpoints using riscv-isac. For checking coverage, we will be using a file (a sh file) which will contain all coverage commands for all extensions available in this repo. Since, checking coverage can take a lot of time thats why I have not added it in the CI yet. We can work together to improve this CI more, and add a new section of checking the coverage in parallel to these runs.

Makes sense. We probably want to expand the matrix to launch multiple different jobs that are configured with different isa strings so that tests for each extension can run in parallel.

PR #536, opened 2024-10-19
Prepare for testing multiple ISAs in CI

jordancarlin:
Update matrix structure and add `sed` commands to change the extensions being tested. Going forward all that needs to be done to test different sets of extensions is add an entry to the matrix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,38 +1,36 @@
-# This is a basic workflow to help you get started with Actions
-
 name: CI test
 
-# Controls when the action will run. 
+# Controls when the action will run.
 on:
-
-  # Triggers the workflow on pull request events but only for the main & dev branch
   pull_request:
     branches: [ '**'  ]
-  
-  # Triggers the workflow on push events
   push:
     branches: [ '**'  ]
-
-  # Allows you to run this workflow Actions manually
   workflow_dispatch:
-
-   # Triggers the action 2am every day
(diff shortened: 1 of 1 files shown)


MuhammadHammad001:
The following dummy WORKFLOW that I have created for a scenario in which one of the RV32 tests causes the RISCOF run to fail but the job will still continue and upload the artifacts. For RV64, I have made the job to succeed, hence, in both cases, uploading the artifacts successfully.
https://github.com/MuhammadHammad001/riscv-arch-test/actions/runs/11416510012

MuhammadHammad001:
Furthermore, I have made the CI to mark the job still as a failure in case the riscof run fails, @jordancarlin  please let me know if you want this to be changed!

UmerShahidengr:
I doubt this approach will work. We are trying to improve the CI to run multiple extensions and different architecture configurations, like e.g, we want to run misaligned_load tests (in privilege folder) twice, once with misaligned support enabled on both Spike and Sail, and once with misaligned support disabled for both, similarly there will be alot of other tests in future which will require to be run twice based on different configuration options. Such options cannot be passed via CI so we require multiple files in which we will change the configuration flags and manage those files. CI will take those files and run it.
Your work, although is brilliant, but will greatly complicate the CI, my recommendation would be to keep CI simple, and execute a bunch of files in the CI, and maintain those files for longer run to enable/disable the extension and/or to add more sail/spike configuration options running ACTs

PR #537, opened 2024-10-19
Update the CI to remove dependency from artifacts upload

MuhammadHammad001:

## Description
This PR adds updates:
  - Remove the dependency from the artifacts upload step to make sure they are always uploaded even in case of a failure, job cancellation if the riscof_work folder is created.


MuhammadHammad001:
@jordancarlin , @UmerShahidengr

MuhammadHammad001:
The following dummy WORKFLOW that I have created for a scenario in which one of the RV32 tests causes the RISCOF run to fail but the job will still continue and upload the artifacts. For RV64, I have made the job to succeed, hence, in both cases, uploading the artifacts successfully.
https://github.com/MuhammadHammad001/riscv-arch-test/actions/runs/11416510012

Furthermore, I have made the CI to mark the job still as a failure in case the riscof run fails, @jordancarlin  please let me know if you want this to be changed!

.github/workflows/test.yml
```diff
@@ -137,30 +137,33 @@ jobs:
...
      # Upload the riscof_work for rv${{ matrix.xlen }} folder if size check passes
       - name: Upload the riscof_work for rv${{ matrix.xlen }} folder
-        # Proceed with upload only if the size is within limit  
-        if: ${{ env.upload == 'true' }}
+        # Proceed with upload only if the size is within limit
+        if: always() && env.upload == 'true'
         uses: actions/upload-artifact@v4
         with:
           name: riscof-test-report-rv${{ matrix.xlen }}
           path: /home/runner/work/riscv-arch-test/riscv-arch-test/riscof-plugins/rv${{ matrix.xlen }}/riscof_work/
```
jordancarlin, line 167: `/home/runner/work/riscv-arch-test/riscv-arch-test/riscof-plugins/rv${{ matrix.xlen }}/riscof_work/` can become `${{ github.workspace }}/riscof-plugins/rv${{ matrix.xlen }}/riscof_work/`

jordancarlin (commented):
Looks good to make sure it uploads even if riscof fails. That's probably when we want the logs the most.

It would probably be a good idea to upload the HTML report that riscof generates even if the rest of the folder is too large. Once some of the larger extensions are added (like F) it will definitely be over 1 GB, and we still want to be able to review all of the tests that were run and their status.

.github/workflows/test.yml
```diff
@@ -137,30 +137,52 @@ jobs:
...
+      - name: Upload the riscof_work folder for rv${{ matrix.xlen }} 
+        if: always() && env.upload_folder == 'true'
         uses: actions/upload-artifact@v4
         with:
-          name: riscof-test-report-rv${{ matrix.xlen }}
-          path: /home/runner/work/riscv-arch-test/riscv-arch-test/riscof-plugins/rv${{ matrix.xlen }}/riscof_work/
-          if-no-files-found: warn
-          retention-days: 3
+          name: riscof-work-folder-rv${{ matrix.xlen }}
+          path: ${{ github.workspace }}/riscof-plugins/rv${{ matrix.xlen }}/riscof_work/
```
jordancarlin, line 176: You could simplify this into one step by having the check step save the bottom half of the path and then use that as the upload path. That way we don’t need 2 separate upload steps. And then the if condition could be if that variable was not empty. Fine as it is, but this simplifies things.

UmerShahidengr: I agree, this logic is complicating the things

UmerShahidengr:
Good work @MuhammadHammad001 and @jordancarlin .
I think the script require a minor optimization, once it is done, it will be good to go

.github/workflows/test.yml
MuhammadHammad001: Seems logical, I have updated the step.

.github/workflows/test.yml
```diff
@@ -137,30 +137,40 @@ jobs:
...
+        if: always()
         id: check_size
         run: |
-          folder_size=$(du -sm /home/runner/work/riscv-arch-test/riscv-arch-test/riscof-plugins/rv${{ matrix.xlen }}/riscof_work | cut -f1)
-          echo "Folder size: ${folder_size} MB"
-          if [ "$folder_size" -gt 1000 ]; then
-            echo "Size exceeds 1 GB. Skipping upload."
-            echo "upload=false" >> $GITHUB_ENV  # Set an environment variable to skip upload
+          work_folder="${{ github.workspace }}/riscof-plugins/rv${{ matrix.xlen }}/riscof_work/"
+          report_file="${{ github.workspace }}/riscof-plugins/rv${{ matrix.xlen }}/riscof_work/report.html"
```
jordancarlin, line 146: Could just be $work_folder/report.html so that if things ever move in the future it’s simpler, but fine either way.

jordancarlin (approved):
Looks much better! I left one minor comment, but not terribly important.

.github/workflows/test.yml
MuhammadHammad001: Done.

UmerShahidengr (approved):

PR #529, opened 2024-10-17
changes to point correct env_dir

JAYANTH-IITM:
changes in ctg to point at correct env_dir path .

PR #539, opened 2024-10-21
[CTG] Update env_dir

trdthg:
## Description

#529 I think that a fixed env_dir either `./` or `~/riscv-arch-test/riscv-test-suite/env` is not suitable for everyone anyway.

Why not set it as a new cli param?

> I personally prefer to place `riscv-arch-test` in the current directory(test_dir) via a symbolic link, which would be `./riscv-arch-test/riscv-test-suite/env`. Considering our different for the default path, I just casually wrote a `./` here.


UmerShahidengr:
This will change the script to run CTG, all scripts will now require a new command line parameter otherwise existing scripts will stop working, I don't recommend such changes which will affect all users without proper discussion and documentation. This will require changes in the documentation and announcement for everyone that a new CLI parameter has been added.

trdthg:
what about keep the default value to `~/riscv-arch-test/riscv-test-suite/env`
At least all scripts do not need to be modified

trdthg:
> keep the default value to `~/riscv-arch-test/riscv-test-suite/env`

done

UmerShahidengr:
@trdthg since the CTG is now part of riscv-arch-tests, I dont find any reason to keep this variable dynamic, it can be set as `../riscv-test-suite/env` and nobody will change it. It was supposed to be required when ctg was a separate repo and people had to give the parth of riscv-arch-tests before generating the tests.
Correct me if I am wrong, it seems like this change is not required in codebase.

trdthg:
> since the CTG is now part of riscv-arch-tests, I dont find any reason to keep this variable dynamic, it can be set as ../riscv-test-suite/env

Agree

However, there is another issue: Python uses the current working directory as `.` when calculating paths.

To obtain the correct file path for `riscv-arch-tests`, you can use the following calculation method
```txt
current_dir = os.path.dirname(os.path.abspath(__file__))
env_dir = os.path.join(current_dir, "../../riscv-test-suite/env")
```

But it is important to remind users to use `pip install --editable path_to_ctg`. Because `riscv-arch-tests` is not a Python package itself, if you directly use `pip install path_to_ctg`, the calculated path will also be wrong.

UmerShahidengr:
@trdthg I have added the command `pip install --editable .` in ReadMe file to install ctg and isac. Do you mean it require any update?
If yes, then you may make the corresponding changes in Readme and CI to install ctg & isac properly.

trdthg:
> I have added the command pip install --editable . in ReadMe file

Thank you, no need to update

I just want to let users know that if they **don't** do that, error will occur

UmerShahidengr (approved):
Simple fix. Can be merged

PR #541, opened 2024-10-22
[CI] Prepare for testing multiple ISAs

jordancarlin:
@UmerShahidengr New approach for testing multiple ISAs in CI. Instead of hardcoding the ISA strings, the matrix is built dynamically using any .yaml files in the `.github/isa_templates` directory. This allows for any of the configuration parameters to be modified and makes it easy to add new configurations to CI: just add a new configuration file to this directory.

For now I moved all of the templates that were in this directory into an `old` sub-directory (several of them are failing at the moment) and copied the two standard templates we had been using for CI into the directory.


jordancarlin:
@UmerShahidengr As more and more extensions are being added, this is something that we might want to revisit so that all of the tests can be run in CI. Right now, only a small subset are run, and it would take too long to run all of them (and some are incompatible, like F and Zfinx). This would allow multiple jobs to be started with different configurations so that various groups of tests can be run in parallel.

UmerShahidengr:
I fully agree to your points @jordancarlin and it is in my to-do list. I still dont know what can be the best way to optimise the CI, and run all tests on all possible configurations in minimum possible time.
May be we can discuss this one in the upcoming SIG-ARCH meeting.
CC @allenjbaum

Issue #523, opened 2024-10-14
[ISAC] When flen less than iflen,isac will generate error string


Pagerd:
In isac, the code for generating sanitise is as follows

```
sanitise_cvpt = lambda rm,x,iflen,flen,c,inxFlg: x + ' fcsr == '+hex(rm<<5) + ' and rm_val == 7 ' \
                + ('' if iflen == flen or inxFlg else ''.join([' and rs'+str(x)+'_nan_prefix == 0x' \
                + 'f'*int((flen-iflen)/4) for x in range(1,c+1)]))

sanitise_norm = lambda rm,x,iflen,flen,c,inxFlg: x + ' fcsr == 0'\
                + ('' if iflen == flen or inxFlg else ''.join([' and rs'+str(x)+'_nan_prefix == 0x' \
                + 'f'*int((flen-iflen)/4) for x in range(1,c+1)]))
```

when `flen` < `iflen`,it will generating Strings like `rs0_nan_prefix == 0x `  it will cause ctg to report an error when generating test cases

jamesbeyond:
PR #544 got merged, fixed the bug

PR #544, opened 2024-10-24
[ISAC] fix isac string bug

Pagerd:
## Description

Fix a bug when  flen < iflen , isac will generate wrong string

### Related Issues

[#523](https://github.com/riscv-non-isa/riscv-arch-test/issues/523)


UmerShahidengr:
Seems like a straight forward fix. Just as a precaution,
@anuani21 , @JAYANTH-IITM , this update was made by you earlier while adding inx flag in ISAC. is there any chance this change will break the Zfinx, Zdinx and other coverage commands? I will merge this one after your approval since it may affect your work on Zfinx.

jamesbeyond (approved):
LGTM

PR #547, opened 2024-11-04
[ISAC] Fix undefined rs1_val, rs2_val

trdthg:
## Description

seems like forgotten after a refactor, It should be very obvious


UmerShahidengr (approved):
Nice fix. well done @trdthg

PR #548, opened 2024-11-04
[CTG] replace load_yamls with load_yaml

trdthg:
## Description

seems not been updated synchronously when the function, variable were modified

### DIFF
--- a/riscv-ctg/riscv_ctg/utils.py
+++ b/riscv-ctg/riscv_ctg/utils.py
@@ -17,7 +17,13 @@ yaml.default_flow_style = False
 yaml.allow_unicode = True
 
 def load_yaml(foo):
+    """
+    Loads the template file or file list and returns the data in a dictionary.
+    """    
     try:
+        if isinstance(foo, list):
+            with combineReader(foo) as fp:
+                return dict(yaml.load(fp))
         with open(foo, "r") as file:
             return dict(yaml.load(file))
     except ruamel.yaml.constructor.DuplicateKeyError as msg:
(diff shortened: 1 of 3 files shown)


UmerShahidengr:
@trdthg we would like to keep the naming convention of `load_yaml` to be consistent everywhere. Having two functions with names `load_yaml` and `load_yamls` will create massive confusion.
I can see there are two instances of `load_yamls` used which can be replaced with `load_yaml` instead of adding more callings of `load_yamls`.
My recommendation would be to removed `load_yamls` and replace it with `load_yaml` everywhere to keep everything consistent.

riscv-ctg/riscv_ctg/utils.py
```diff
@@ -16,8 +16,14 @@
 yaml.default_flow_style = False
 yaml.allow_unicode = True
 
-def load_yaml(foo):
+def load_yaml(foo: str | list[str]):
```
trdthg: I really want to write a type hint here to keep it clear, although the current code does not use type hints

trdthg:
update

riscv-ctg/riscv_ctg/utils.py
trdthg: finally deleted, because `|` is not supported in python 3.6

riscv-ctg/riscv_ctg/utils.py
```diff
@@ -60,9 +66,6 @@ def get_instr_list():
 
     return instr_lst
 
-def load_yamls(foo):
```
UmerShahidengr, line 63: Thank you for removing this function

UmerShahidengr (approved):
LGTM

PR #549, opened 2024-11-04
[ISAC]: Replaces the use of round_trip_dump

trdthg:
## Description

`round_trip_dump` function has been deprecated since 0.18.0. And we didn't restrict it's version in `requirements.txt`

https://github.com/riscv-non-isa/riscv-arch-test/blob/3843c736e427a2b52a0d06e6220b073afa4be401/requirements.txt#L37

I suggest switching directly to a usage that can be used in all these versions


UmerShahidengr (approved):
Nice effort. This will fix issues of ISAC in python version 3.12.
Thank you @trdthg

PR #556, opened 2024-11-08
Resolved the bug in ISAC -> local variable 'yaml' referenced before assignment

UmerShahidengr:
PR #549 triggered an error in ISAC in util.py i.e, local variable 'yaml' referenced before assignment. I have given its fix

PR #462, opened 2024-04-29
Physical Memory Protection (32/64) Tests and Covergroups

MuhammadHammad001:
<FOR DOC UPDATES FILL ONLY DESCRIPTION AND RELATED ISSUES SECTION AND REMOVE THE OTHERS>

## Description

>This PR adds the arch tests for the PMP for 32 bit and 64 bit architecture and also the Covergroups have been added as well for the coverage analysis. This implements the Physical Memory Protection (PMP) section from the following [Test Plan](https://docs.google.com/spreadsheets/d/1p13gic7BD6aq7n_dHrqti4QmlGpxY7FkF17RVbG4DC0/edit#gid=0). The assembly arch tests have been developed by [Mr. Umer Shahid](https://github.com/UmerShahidengr) and the coverpoints are developed by [Muhammad Hammad Bashir](https://github.com/MuhammadHammad001).

### Related Issues

> The links to the PRs that need to be merged before this PR:
> - [PR #110 -- Support for Translator and Macros in RISCOF](https://github.com/riscv-software-src/riscof/pull/110)
> - [PR #80 -- Support for Privileged Architecture in RISC-V ISAC](https://github.com/riscv-software-src/riscv-isac/pull/80)
### Ratified/Unratified Extensions

- [x] Ratified
- [ ] Unratified

### List Extensions

> Privileged Architecture -- Physical Memory Protection

### Reference Model Used

- [x] SAIL
- [ ] Spike
- [ ] Other - < SPECIFY HERE >

### Mandatory Checklist:

  - [x] All tests are compliant with the test-format spec present in this repo ?
  - [x] Ran the new tests on RISCOF with SAIL/Spike as reference model successfully ?
  - [x] Ran the new tests on RISCOF in [coverage mode](https://riscof.readthedocs.io/en/stable/commands.html#coverage)
  - [x] Link to Google-Drive folder containing the new coverage reports available at: < [Coverage Reports PMP](https://drive.google.com/drive/folders/1jJ7yJ9oaWcGO-rH0jh_qVnutv4gM7bdU) >
  - [x] Link to PR in RISCV-ISAC from which the reports were generated : < [PR #80 -- Support for Privileged Architecture in RISC-V ISAC](https://github.com/riscv-software-src/riscv-isac/pull/80)>
  - [x] Changelog entry created with a minor patch

### Optional Checklist:

  - [ ] RISCV-V CTG PR link if tests were generated using it : < SPECIFY HERE >
  - [x] Were the tests hand-written
[...]

PR #557, opened 2024-11-08
Add riscof_work and __pycache__ to .gitignore

jordancarlin:
PRs #462 and #556 both ended up with these files being included in the PR at one point. Neither of them should be part of the repo. This adds them to the .gitignore to avoid issues in the future.


UmerShahidengr (approved):
LGTM

PR #563, opened 2024-11-14
[ISAC] - PTE Permissions Check Function and prev. instr. csr track func Updated

MuhammadHammad001:

## Description
Function to check permissions of a PTE has been updated to take lower and upper case characters for optimization.


MuhammadHammad001:
This commit adds the support for the old function to be renamed to `old_csr_val` since it just tracks the previous state CSR values. Additionally, I have also made this function global so it can be used in val_comb and op_comb in addition to csr_comb which will enable us to write coverpoints which are cross with the other instructions.

allenjbaum (approved):
This makes coverpoint definition much more concise

PR #565, opened 2024-11-19
Update RV64I sraw test cases

trdthg:
## Description

- Add sraw to unsgn_rs2 group, rs2 is shift, which should be unsign
- The sraw testcases were very old with a covarage of 98.20% (109/111), this patch regenerate the cases with riscv_ctg (without `-r`), making covarageto 100%.

- [x] Ratified
- [ ] Unratified

### List Extensions

### Reference Model Used

- [x] SAIL
- [x] Spike

### Mandatory Checklist:

  - [x] Link to Google-Drive folder containing the new coverage reports ([See this](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/CONTRIBUTION.md#uploading-test-stats) for more info): < https://drive.google.com/drive/folders/1DzS75_cLWSi--lD02cv06JEMYv-Bba9u >

### Optional Checklist:


allenjbaum:
This looks a bit strange; there is 1/4 the number of tests, yet you get
better coverage.
Can you explain that?  (I'm not sure what the -r option does)

trdthg:
I thought this is strange too.

My ctg command is below:

```sh
riscv_ctg -v debug -d ./tests/ -cf ./coverage/dataset.cgf -cf ./coverage/i/rv64i.cgf -bi rv64i -p2
```

I checked the 35 generated tests:

- op1, op2, rd are basically gradually decreasing, from x31 to x0, covering all cases of coverage
- The first 7 and the last 2 cover all of `val_comb.rfmt_base_shift`

It does seem to be covered, after all, 111 is the sum of these coverage points

The last generation was 4 years ago, and I don't know exactly what happened.

---

> (I'm not sure what the -r option does)

`-r, --randomize                 Randomize Outputs.`

	> If randomization is enabled we use the `MinConflictsSolver` solver to find solutions.

https://github.com/riscv-non-isa/riscv-arch-test/blob/cd94912fed2aab75d7d5f115b441da0813fdce8d/riscv-ctg/riscv_ctg/generator.py#L493-L496

I haven't tested the behavior of `MinConflictsSolver`, ~and I suspect it may lead to generated tests being difficult to reproduce~

After some testing, I found that the tests ctg generated each time is different. This is completely different from what I expected. I also had another person try it, and the results were also different.

allenjbaum:
I don't know how ctg selects a seed, but it doesn't seem to be user
selectable and appears to be random, as far as I know.

UmerShahidengr:
Multiple coverpoints are getting hit with a single testcase, which is reducing the test size to 25%, can we do it for all existing tests?

UmerShahidengr:
It is a very optimal approach of generating the tests and completing the coverage.

UmerShahidengr (approved):
Approved

PR #558, opened 2024-11-11
[ISAC] - Interrupt Coverage Support

MuhammadHammad001:
<FOR DOC UPDATES FILL ONLY DESCRIPTION AND RELATED ISSUES SECTION AND REMOVE THE OTHERS>

## Description

> Adds support for interrupt track for coverage

- Added the track for the interrupts in the mcause (delegation cases not solved yet)
- Updated the architectural state for the MIP register and now depends upon the hart update in the Sail Log

### Related Issues

> Please list all the issues related to this PR. Use NA if no issues exist

### Ratified/Unratified Extensions

- [x] Ratified
- [ ] Unratified

PR #567, opened 2024-11-21
[ACT] ACTs and coverage points for machine mode software and timer interrupts

alifaraz-10xe:

## Description

This pull request introduces tests and covergroups to verify machine mode software and timer interrupts. The tests and covergroups were developed following the guidelines outlined in the [Test Plan](https://docs.google.com/spreadsheets/d/1rwjSTbxKiXTDydAdFAE_o7mez_hcFj9a-9d3dXRVc4s/edit?usp=sharing)

### Related Issues

This pull request relies on the changes in the following PRs for full functionality and compatibility:
1. [Interrupt Coverage Support](https://github.com/riscv-non-isa/riscv-arch-test/pull/558) 
2. [Print MTI value upon hart update for machine timer interrupt visibility](https://github.com/riscv/sail-riscv/pull/620) 

- [x] Ratified
- [ ] Unratified

### List Extensions

N/A

### Reference Model Used

- [x] SAIL
- [ ] Spike

### DUT Model Used

- [ ] SAIL
- [x] Spike

### Mandatory Checklist:

  - [x] Link to Google-Drive folder containing the new coverage reports [available here](https://drive.google.com/file/d/1bgrGzx1_Zpoo4MCtk-YZ-WU2SIZjZzzm/view?usp=drive_link)
  - [x]  Link to PR in RISCV-ISAC from which the reports were generated [available here](https://github.com/riscv-non-isa/riscv-arch-test/pull/558)

### Optional Checklist:


allenjbaum (changes requested):
In general, equations with constants (e.g. mip>>3 & 0x1!= 0x1) are not satisfactory. What would be satisfactory would be ((mip & MIP_MSIP)==0), where MIP_MSIP is #defined -   and you don't even have to define it yourself because it is already defined in encoding.h which is #included. Note that this applies of almost CSR fields . The comments are quite appreciated, on the other hand.

coverage/header_file.yaml
```diff
@@ -932,4 +932,9 @@ PMP_helper_Coverpoints:
...
  NA4_REGION_ADDRESS_MATCH: (rs1_val + imm_val == (pmpaddr1 << 2))
  NA4_PRIORITY_REGION_MATCH: (rs1_val + imm_val == (pmpaddr3 << 2))
- NA4_PRIORITY_2_REGION_MATCH: (rs1_val + imm_val == (pmpaddr1 << 2))
\ No newline at end of file
+ NA4_PRIORITY_2_REGION_MATCH: (rs1_val + imm_val == (pmpaddr1 << 2))
+ 
+Interrupt_Macros:
+ MSIP_BASE: 0x02000000
+ MTIMECMP_BASE: 0x02004000
+ MTIME_BASE: 0x0200bff8
```
allenjbaum: Where did these default values come from? There is a model defined variables, which should be named RVMODEL_MSIP_BASE, etc to be consistent with all other model defined variables.

riscof-plugins/rv32/sail_cSim/env/model_test.h
```diff
@@ -45,13 +45,23 @@ RVMODEL_DATA_SECTION			;\
 //RVTEST_IO_ASSERT_DFPR_EQ
 #define RVMODEL_IO_ASSERT_DFPR_EQ(_D, _R, _I)
 
-#define RVMODEL_SET_MSW_INT
+#define RVMODEL_SET_MSW_INT           ;\
+    li   t1, 1                        ;\
+    li   t2, 0x2000000                ;\
```
allenjbaum: 1. using constants is prohibited; use the value you defined above: MSIP_MBASE
2. use of li is prohibited in tests; use LI(rd, mnemonic_value) instead.
These apply to ALL li ops in all tests, not just this one

riscv-test-suite/env/arch_test.h
```diff
@@ -1508,10 +1514,18 @@ common_\__MODE__\()int_handler:         // T1 has sig ptr, T5 has mcause, sp has
...
         sll     T3, T3, T2              // create mask 1<<xcause **NOTE**: that MSB is ignored in shift amt
+        csrrc   T4, CSR_XSTATUS, T3     // read then attempt to clear mstatus??
+        srli    T4, T4, 3               // Move MIE to LSB
+        andi    T4, T4, 1               // Mask MIE     
+sv_\__MODE__\()status:                  // note: clear has no effect on MxSTATUS
+        SREG    T4, 2*REGWIDTH(T1)      // save 3rd sig value, (xstatus)
         csrrc   T4, CSR_XIE, T3         // read, then attempt to clear int enable bit??
-        csrrc   T4, CSR_XIP, T3         // read, then attempt to clear int pend bit
+sv_\__MODE__\()ie:                      // note: clear has no effect on MxIE
+        SREG    T4, 3*REGWIDTH(T1)      // save 4th sig value, (xie)
```
allenjbaum: XIE should not be saved, as an interrupt doesn't affect its value.  The comments before label common_\_MODE__()handler give number&names of the variables that interrupts save in the interrupt signature:
Vect+mode, xcause, xip, IntID - and only if hypervisor is implemented, tvalt2 and tinst.

Basically, do not modify arch_test.h without consulting me first

riscv-test-suite/env/encoding.h
```diff
@@ -128,6 +128,26 @@
 #define MIP_HEIP            (1 << IRQ_H_EXT)
 #define MIP_MEIP            (1 << IRQ_M_EXT)
 
+#define MIE_SSIP            (1 << IRQ_S_SOFT)
```
allenjbaum: Hmm - these should already be defined in the master copy of encoding.h We need to make sure that the latest copy is automatically pulled in (not a problem for this PR)

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
```diff
@@ -0,0 +1,179 @@
...
+#ifdef TEST_CASE_1
+RVTEST_CASE(1,"//check ISA:=regex(.*32.*); check ISA:=regex(.*I.*Zicsr.*); def rvtest_mtrap_routine=True; def TEST_CASE_1=True; ",CHK_M_INTR_PRIORITY)
+RVTEST_SIGBASE( x3,signature_x3_1)
+
+	.attribute unaligned_access, 0
+	.attribute stack_align, 16
+  	.align	2   
+  	.option norvc
+  	
+#ifndef MSIP_BASE
```
allenjbaum: I'm not sure we actually want to default this; we should force models to declare it and raise an error if they don't (and they're testing interrupts...)

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
```diff
@@ -0,0 +1,179 @@
...
+    	// Set mtime and mtimecmp to trigger a timer interrupt
+    	// Demonstrates timer interrupt generation mechanism
+    	SREG t0,   0(a1)
+       	nop
+       	SREG zero, 4(a1)		
+       	nop
+       	SREG t0,   0(a2)
+       	nop		
+       	SREG zero, 4(a2)		
+       	nop
```
allenjbaum: 1. you have a noop after each write. Why? These MMIO addresses should all be written in order.
There shouldn't be an interrupt until the last one is written - but it may not be immediately (the next instruction).
2. Also note: there may be no guarantee that an interrupt will occur upon completion of the op that set the MIP_MIE bit. The test must take that into account

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
```diff
@@ -0,0 +1,179 @@
...
+    	// Initial MIP State Capture:
+    	// Read Machine Interrupt Pending (MIP) register before interrupt handling
+    	// Verifies initial interrupt pending states
+    	csrrs a4, CSR_MIP, x0		
+	RVTEST_SIGUPD(x3, a4)
+    	      	       	
+    	// Enable Local Interrupts:
+    	// Locally enable both software and timer interrupts
+    	// Prepares processor to handle pending interrupts
+    	LI(a1,    MIE_MSIP)
```
allenjbaum: Note that you could simply LI(a1, MIE_MSIP+MIE_MTIP) instead of loading 2 consts and ORing them

allenjbaum: Note that you're enabling 2 interrupts simultaneously, and both will be taken, one after the other. So there will be two trap signatures, and I believe they will be back to back - no instructions will be executed between the xret and entry to the handler

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-TIMER-INTR.S
```diff
@@ -0,0 +1,159 @@
...
+	// Verify interrupt behavior when MTIE is disabled and no timer interrupt is pending
+	LI(t0,    MIE_MTIP)
+	csrrc x0, CSR_MIE, t0
+	csrrs a4, CSR_MIP, x0
+	RVTEST_SIGUPD(x3,  a4)
+
+	// Test Case 02: Enable MTIE with mtime < mtimecmp
+	// Confirm interrupt state when MTIE is enabled but no timer interrupt is pending
+	csrrs x0, CSR_MIE, t0
+	csrrs a4, CSR_MIP, x0
```
allenjbaum: note that MIP is already set from the previous test, so the interrupt might occur before you set it again, which will mess up what you're trying to test for

coverage/rv32_m_interrupts.cgf
```diff
@@ -0,0 +1,159 @@
...
+# This coverpoint checks MIP register write protection
+# Confirms Machine Interrupt Pending (MIP) register remains read-only
+CHK_MIP_PERMISSONS:
+  config: 
+    - check ISA:=regex(.*32.*); check ISA:=regex(.*I.*Zicsr.*); def rvtest_mtrap_routine=True;
+  mnemonics: 
+      csrrw: 0
+  val_comb: 
+      # Validate MIP bits across different register segments
+      '(((mip >> 3)  &  0x1) != 1)' : 0 # Ensures Machine Software Interrupt Pending (MSIP) bit cannot be modified by write operations, maintaining read-only protection.
```
allenjbaum: Do not use constants in the coverage equations; use only #defined variables (most of which are in already in encoding.h). And for this don't shift, and, cmp with 1, just and with a mask and compare to zero

allenjbaum: This applies pervasively to all your constants (except zero...)

coverage/header_file.yaml
alifaraz-10xe: These are standard RISC-V CLINT (Core-Local Interrupt) base addresses, which also used in the Sail model. MSIP_BASE and others are now replaced with the RVMODEL_MSIP_BASE etc in all ACTs.

riscof-plugins/rv32/sail_cSim/env/model_test.h
alifaraz-10xe: Yes, it is replaced in all plugins and proper macro is used now.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
```diff
@@ -64,73 +64,73 @@ RVTEST_SIGBASE( x3,signature_x3_1)
...
 	RVTEST_SIGUPD(x3, a4)
-    	      	       	
+
     	// Enable Local Interrupts:
     	// Locally enable both software and timer interrupts
     	// Prepares processor to handle pending interrupts
-    	LI(a1,    MIE_MSIP)		
-    	LI(a2,    MIE_MTIP)		
+    	LI(a1,    MIE_MSIP)
+    	LI(a2,    MIE_MTIP)
```
allenjbaum: Why aren't you simply LI(a1, MIE_MSIP+MTIP)?
Not sure if this is guaranteed to make  both interrupts to occur simultaneously. (probably though).
I don't think the trap handler will clear all interrupts, just the highest priority one.
That means you'll get another interrupt as soon as you return from the trap handler. 
That can break this code, unless you have extra noops after setting MIE. 
It also requires you to allocate space for 2 interrupt signatures.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
```diff
@@ -64,73 +64,73 @@ RVTEST_SIGBASE( x3,signature_x3_1)
...
+       	nop
+
     	// Generate Software Interrupt:
     	// Write to memory-mapped Software Interrupt Pending (MSIP) register
     	// Creates a pending software interrupt alongside timer interrupt
-    	LI(a3,   0x1)		
+    	LI(a3,   0x1)
        	LI(t0,   RVMODEL_MSIP_BASE)
-       	SREG a3, 0(t0)		
+       	SREG a3, 0(t0)
```
allenjbaum: This may cause an immediate interrupt, because MIE_MSIP is unitialized. 
You need to clear it before you set MSTATUS_MIE.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-SOFTWARE-INTR.S
```diff
@@ -50,10 +50,10 @@ RVTEST_SIGBASE( x3,signature_x3_1)
 
 	.attribute unaligned_access, 0
 	.attribute stack_align, 16
-  	.align	2   
+  	.align	2
   	.option norvc
 
-main:	
+main:
```
allenjbaum: This sets MSTATUS_MIE before clearing MIP_MSIP. The value of MIP_MSIP is not guaranteed to be zero atreset

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-TIMER-INTR.S
```diff
@@ -55,12 +55,12 @@ RVTEST_SIGBASE( x3,signature_x3_1)
...
 	.attribute stack_align, 16
-  	.align	2   
+  	.align	2
   	.option norvc
-  	
+
 #define TIMER_MAX_VALUE         0xFFFF	// Maximum Value for the memory-mapped machine timer-compare register.
 
-main:	
+main:
```
allenjbaum: This needs to initialize MIE.MTIP to zero before setting MSTATUS.MIE
Note that there will be a delay between setting MTIME and MTIMECMP and the interrupt appearing in MIP that could be longer than the amount of time between them and the setting/clring the MIE.MTIP and MSTATUS.MIE bits. You need to insert a delay after writing MTIME and MTIMECMP 
9and the delay amount should be provided by the DUT as the #cycles from writing those register until the MIP.MTIP bit gets set.)

allenjbaum: Also: you're testing when mtime is < mtimecmp and when it is equal. - I on't see how you can guarantee that since mtime is constantly being updated at an unknown rate. You have to is write MTIMECMP with the value read from MTIME (when MIE and MIP.MTIP are enabled - and you need to do it before test 3 AND before test 4

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-MSIP-FUNCTIONALITY.S
```diff
@@ -51,52 +51,52 @@ RVTEST_SIGBASE( x3,signature_x3_1)
...
 	// Verify Software Interrupt Pending:
 	// Read MIP register to confirm software interrupt bit is set
-	csrrs a4, CSR_MIP, x0		
+	csrrs a4, CSR_MIP, x0
 	RVTEST_SIGUPD(x3,  a4)
 
 	// Clear Software Interrupt:
 	// Reset MSIP control register
 	// Validate interrupt clearing mechanism
-	SREG zero, 0(x5)
```
allenjbaum: The trap handler will clear the MSIP bit before returning, else it will just keep reinterrupting.
And it already saves it into the signature

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-MTIME-LENGTH.S
```diff
@@ -66,7 +66,7 @@ main:
 	nop
 	SREG t1,   0(a2)			// Lower half of mtime = max (0xFFFFFFFF)
 	nop
-	
+
```
allenjbaum: You aren't testing that all 64 bits are writable, just that the lower 32bits, and not testing that MTIME could count and overflow the low 32b or that it could wrap around from -1 to zero.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-MTIME-LENGTH.S
```diff
@@ -7,7 +7,7 @@
 // Developed by: Syed Ali Faraz Jaffery
 // ---------------------------------------------------------------------------------------------
 // Test Description:
-// This test verifies the 64-bit precision of machine timer (mtime) and 
+// This test verifies the 64-bit precision of machine timer (mtime) and
```
allenjbaum, line 10: As before, the > condition is easy to check, the = condition is harder, since it is always counting. By the time you check the MIP bit, it could have already counted

riscv-test-suite/rv32i_m/m_interrupts_32/src/DIS-GLB-MIE.S
```diff
@@ -50,7 +50,7 @@ RVTEST_SIGBASE( x3,signature_x3_1)
 
 	.attribute unaligned_access, 0
 	.attribute stack_align, 16
-  	.align	2   
+  	.align	2
```
allenjbaum: you also need to initialize MIP.MSIP before all this as it's not initialized at reset.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-MTIP-FUNCTIONALITY.S
```diff
@@ -67,7 +67,7 @@ main:
 	// Prevent maskable interrupts during test execution
 	// Ensures a controlled test environment
 	LI(a5,    MSTATUS_MIE)
-	csrrc x0, CSR_MSTATUS, a5	 
+	csrrc x0, CSR_MSTATUS, a5
```
allenjbaum: also need to initialize MIP.MTIP.  I'm not going to add this comment anymore, all the tests need to do this (or I'm missed where it is initialized)

allenjbaum (changes requested):
Almost all tests need to initialize MIP.MSIP/MTIP before doing anything else.
The MAX_VALE should be renamed TEST_VALUE, since it isn't a max value at all.
I think you could combine a lot of the 32 and 64b tests, except for the cases that need to write all 64bits of MTIME and MTiMECMP (which I would do with a macro to and read them differently depending on XLEN)

riscv-test-suite/env/encoding.h
UmerShahidengr: I have synced up the latest copy of encoding.h with this one

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
UmerShahidengr: This test is checking Interrupt priority resolution mechanism i.e,  software interrupt precedence over timer interrupts. So, this method is correct for checking that test case I think

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
UmerShahidengr: Agree. this change is required @Ali-Faraz-10xe , can you please do it immediately so that it gets approved?

riscv-test-suite/env/arch_test.h
```diff
@@ -1520,16 +1529,35 @@ resto_\__MODE__\()rtn:                  // restore and return
...
         andi    T2, T5, INT_CAUSE_MSK   // clr INT & unarched arched bits (**NOTE expand if future extns use them)
         sll     T3, T3, T2              // create mask 1<<xcause **NOTE**: that MSB is ignored in shift amt
-        csrrc   T4, CSR_XIE, T3         // read, then attempt to clear int enable bit??
-        csrrc   T4, CSR_XIP, T3         // read, then attempt to clear int pend bit
-sv_\__MODE__\()ip:                      // note: clear has no effect on MxIP
-        SREG    T4, 2*REGWIDTH(T1)      // save 3rd sig value, (xip)
+        csrrc   T2, CSR_MIDELEG, x0     // read machine interrupt delegation
+        nop	                // if mideleg is not used and trap occur
+        and     T2, T2, T3              // mask caused interrupt delegation bit
+        srl     T2, T2, T3              // Move interrupt delegation to LSB
```
allenjbaum: Why do you move it to LSB, clr all but LSB, then beq, rather than move it MSB and bge?

riscv-test-suite/env/arch_test.h
```diff
@@ -1520,16 +1529,35 @@ resto_\__MODE__\()rtn:                  // restore and return
...
- //**FIXME** - make sure this is kept up-to-date with fast int extension and others
+        LI(     T3, 1)
         andi    T2, T5, INT_CAUSE_MSK   // clr INT & unarched arched bits (**NOTE expand if future extns use them)
         sll     T3, T3, T2              // create mask 1<<xcause **NOTE**: that MSB is ignored in shift amt
-        csrrc   T4, CSR_XIE, T3         // read, then attempt to clear int enable bit??
-        csrrc   T4, CSR_XIP, T3         // read, then attempt to clear int pend bit
-sv_\__MODE__\()ip:                      // note: clear has no effect on MxIP
-        SREG    T4, 2*REGWIDTH(T1)      // save 3rd sig value, (xip)
+        csrrc   T2, CSR_MIDELEG, x0     // read machine interrupt delegation
+        nop	                // if mideleg is not used and trap occur
```
allenjbaum: You need to document the reason for this. This is setting up the trap handler to take an exception whenever it interrupts into Smode (because it can't read the mideleg CSR that caused it to go to Smode?) 
The problem here is that the trap signature for an illegal instruction won't store mideleg - so the test can never tell whether mideleg was set or not. It would be best to have the test (not the trap handler) explicitly store mideleg whenever it changes it, and avoid the extra trap.

riscv-test-suite/env/arch_test.h
```diff
@@ -1520,16 +1529,35 @@ resto_\__MODE__\()rtn:                  // restore and return
...
+        srl     T2, T2, T3              // Move interrupt delegation to LSB
+        LI     (T4, 1)
+        beq     T2, T4, sv_\__MODE__\()status  // if interrupt delegation bit is set then read and store SIE and SPIE in signature
+        LI(     T2, 3)                  // if interrupt delegation bit is not set then read and store MIE and MPIE in signature
+sv_\__MODE__\()status:                  // note: clear has no effect on XxSTATUS
+        csrrc   T4, CSR_XSTATUS, x0     // read xstatus
+        srl     T4, T4, T2              // move xIE to LSB
+        andi    T4, T4, 0b10001         // mask xIE and xPIE
+        SREG    T4, 2*REGWIDTH(T1)      // save 3rd sig value, (xstatus.XIE and xstatus.XPIE)
+        nop
```
allenjbaum: Why is there a nop here? This shouldn't ever trap. Also: you're storing individual bits from Xstatus, XIE, and XIP; you should collect the 3 bits and store them in the first signature entry, (label  sv_\__MODE__\()vect, the first word of the signature entry; there are lots of free bits in that entry. This avoids changing the size of the interrupt trap signature entry. That size is pre-calculated, is used to update the trap signature pointer, and is stored in this entry.  Changing the size introduces a bug because the pointer is pre-incremented with it, and the next trap will overwrite the extra word of this entry.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
Ali-Faraz-10xe: After each write, noops are added for a reason. If any implementation doesn't support/implement these memory-mapped registers, then there will be a series of exceptions because of each write, and the test will take a long time to fail/end (as the test will enter again and again into the trap handler when resolving one exception). So noops will help to fail the test in this case since different signatures will be saved in the reference model results and DUT result, hence the comparison will fail.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
alifaraz-10xe: Thanks for highlighting this important aspect. I have now changed the test flow by enabling the global interrupt bit (MSTATUS_MIE) after both interrupts are in pending state (MIP_MSIP = 1, MIP_MTIP = 1) and both are enabled locally (MIE_MSIE = 1, MIE_MTIE = 1). This way we guarantee that both interrupts occur simultaneously. The space for 2 interrupt signatures is already allocated. Also, I am going to follow the suggested way of enabling register bits like 'LI(a1, MIE_MSIP+MIE_MTIP)' in all tests. Thanks :)

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
allenjbaum: Not, exactly. IF you jump to an Mmode test, and any of these (architecturally mandated, vendor provided) addresses don't exist, it will trap, write an exception trap signature, and return, skipping the next instruction. It won't trap over and over, it will (in this case of 4 sequential writes failing) trap at most twice. Even if it traps only once, that will be enough to cause a signature miscompare.  IF it traps too many times (more than you expect and have allocated in the signature region), then it will safely abort the test because the handler checks to see if you've overrun that trap signature area, so its safe even if it keeps writing those addresses over and over in different test cases.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
allenjbaum: Ah, OK. but: 
First, I don't think there is an architecturally defined value of what value a write to MSIP_BASE will cause an interrupt.  That's why there is an RVMODEL_SET_MSW_INT macro. Use it.
Second: You can get away without using RVMODEL_SET_MTIMER_INT - but to be consistent, you should use that also. If makes it much easier to understand the code.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-SOFTWARE-INTR.S
alifaraz-10xe: You are right that the value of MIP_MSIP is not guaranteed to be zero at reset. So now I've added RVMODEL_CLR_MSW_INT in the all software interrupt tests, and RVMODEL_CLR_MTIMER_INT in the all timer interrupt tests. These macros are placed right before the start of actual test

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-SOFTWARE-INTR.S
```diff
@@ -0,0 +1,156 @@
...
+		.align	2
+		.option norvc
+
+		main:
+			// Clear Software & Hardware Interrupts at the start, if there is any
+			RVMODEL_CLR_MEXT_INT
+			RVMODEL_CLR_SEXT_INT
+			RVMODEL_CLR_MSW_INT
+			RVMODEL_CLR_SSW_INT
+
```
alifaraz-10xe: @UmerShahidengr this new change causes the coverage to drop to 25% for this test. Upon further investigation, I see there are no software interrupt being generated in both SAIL and DUT.

allenjbaum: Ah, I think we need to define these in the default plug-ins for spike and Sail, but if you have your own DUT you'll have to define it in your rvmodel_test.h file.

alifaraz-10xe: Issue fixed! You have deleted one code line by mistake which results in coverage to drop down.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
alifaraz-10xe: That's correct. No instructions is executing between the xret and entry to the handler

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-MSIP-FUNCTIONALITY.S
alifaraz-10xe: The trap handler is clearing the MSIP bit before returning, hence test is ending properly!

riscv-test-suite/rv32i_m/m_interrupts_32/src/DIS-GLB-MIE.S
alifaraz-10xe: Done!

riscv-test-suite/env/arch_test.h
UmerShahidengr: It has been done already, so I am resolving this one

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
UmerShahidengr: It will be removed from all tests 
CC: @Ali-Faraz-10xe

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-TIMER-INTR.S
UmerShahidengr: RVMODEL_CLR_MTIMER or _MSW_INT has been added on top of every test file, so this comment has also been resolved

coverage/rv32_m_interrupts.cgf
UmerShahidengr: Already done. So marking this one as resolved

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-TIMER-INTR.S
UmerShahidengr: @allenjbaum, This is not something that can be done as fixed delay, do we need to define a new RVMODEL macro for this purpose?

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-MTIME-LENGTH.S
UmerShahidengr: This test has to be rewritten, it is not complete.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
alifaraz-10xe: I've removed these macros from all tests.

UmerShahidengr (dismissed):
All requested changes have been done

coverage/m_interrupts/rv32_m_interrupts.cgf
```diff
@@ -0,0 +1,159 @@
...
+    # Timer interrupt scenario validation
+     '(call_type == "interrupt") and ((mstatus & ${MSTATUS_MPP}) == ${MSTATUS_MPP_M}) and ((mstatus & ${MSTATUS_MPIE}) == ${MSTATUS_MPIE}) and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mcause & ${INTERRUPT_MASK}) == ${CAUSE_M_TIMER_INTR}) and ((mie & ${MIE_MTIE}) == ${MIE_MTIE}) and ((mip & ${MIP_MSIP}) == ${MIP_MSIP})': 0 # Validates timer interrupt handling, verifying machine status register configuration(MPP, MPIE, MIE), interrupt cause, and timer interrupt enable/pending states.
+
+# This coverpoint checks MIP register write protection
+# Confirms Machine Interrupt Pending (MIP) register remains read-only
```
allenjbaum, line 19: I'm not sure that these coverpoints will tell you anything. This checks to see that mip.[MSIP,MTIP,MEIP] was ever !=1 . But, it will never be 1 because that tests bit0 only, and it needs to test the appropriate bit posiition.  In any case, don't you need to knwo that a write of that specific bit was attempted? This coverpoint be hit even if you don't actually write the CSR bit to 1, but write some other CSR or some other bit.

Also, you need to define a function that checks whether a bit of a CSR has value 1 or 0, so you don't have to write (csrname & ${fieldmask} == ${fieldmask};  e.g. csrbit1(mip,fieldname) and csrbit0(mip,fieldname) for ==0

coverage/m_interrupts/rv32_m_interrupts.cgf
```diff
@@ -0,0 +1,159 @@
...
+  mnemonics:
+      csrrc: 0
+      csrrs: 0
+      sw   : 0
+  val_comb:
+      # Verify MTIP set and clear mechanisms
+      '(call_type != "interrupt") and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mip & ${MIP_MTIP}) == ${MIP_MTIP})' : 0 # Checks timer interrupt pending state when no interrupt is being serviced.
+      '(call_type != "interrupt") and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mip & ${MIP_MTIP}) == 0)' : 0 # Confirms timer interrupt pending bit's behavior under non-interrupt execution.
+
+# This coverpoint used to validate the 32-bit precision of machine timer (mtime) and timer compare (mtimecmp) register if MTIP hit for the 4 times
```
allenjbaum, line 78: mtime & mtimecmp are each is logically 64 bit registers; you need to also check mtimeh and mtimecmph.
I don't understand  hitting mtip 4 times though - explain.

riscof-plugins/rv32/sail_cSim/env/model_test.h
```diff
@@ -52,13 +56,32 @@ RVMODEL_DATA_SECTION			;\
...
+    nop	                                ;
 
-#define RVMODEL_CLEAR_MSW_INT
+#define RVMODEL_CLR_MSW_INT             ;\
+    LI(  T2,  RVMODEL_MSIP_BASE)        ;\
+    SREG x0,  0(T2)	                    ;\
+    nop	                                ;
 
-#define RVMODEL_CLEAR_MTIMER_INT
+#define RVMODEL_SET_MTIMER_INT          ;\
```
allenjbaum, line 77: I don't think this will always work if you don't also update MTIME_BASE+4 and MTIMECMP_BASE+4.
Also, you disabled the clr mtime_int routine, and I think you need that because you don't know what the reset value will be.

riscv-test-suite/env/arch_test.h
```diff
@@ -1239,21 +1248,44 @@ rvtest_\__MODE__\()endtest:                     // target may be too far away, s
...
 sv_\__MODE__\()vect:                            // **FIXME?: breaks if tramp crosses pg && MMU enabled
-        sub     T6, T6, T3                      // cvt spreader-addr to vector offset fm top of tramptable 
+        sub     T6, T6, T3                      // cvt spreader-addr to vector offset fm top of tramptable
         slli    T6, T6, 4                       // make room for 4 bits; vector is 10b max  **FIXME: broken for SV64!)
         or      T6, T6, T2                      // insert entry size into bits 5:2
         addi    T6, T6, \__MODE__\()MODE_SIG    // insert mode# into 1:0
+
+        csrrc   T4, CSR_XSTATUS, x0             // read xstatus
+        srl     T4, T4, T2                      // move xIE to LSB
+        andi    T4, T4, 0b10001                 // mask xIE and xPIE
```
allenjbaum, line 1292: 1. use mnemonics, not constants
2. this doesn't work because xstatus.yIE  and yPIE are at different bit positions depending on mode y that it trapped from
3. Why not save yourself the trouble and just look the lower 8 bits, and don't mask it beyond that?
   (you'll have to move xIP[cause]and and sIE[cause] above that then.)

riscv-test-suite/env/arch_test.h
```diff
@@ -1239,21 +1248,44 @@ rvtest_\__MODE__\()endtest:                     // target may be too far away, s
...
+        sll     T4, T4, 16                      // Shift xIE and xPIE by 16
+        or      T6, T6, T4                      // insert xIE & xPIE into bits 16 & bit 20
+
+        andi    T2, T5, INT_CAUSE_MSK           // clr INT & unarched arched bits (**NOTE expand if future extns use them)
```
allenjbaum, line 1296: The defined causes have been increase, so NUM_SPECED_EXCPT_CAUSES should be 19, and EXCPT_CAUSE_MASK should be ((1<<5)-1) on line 163

riscv-test-suite/env/arch_test.h
```diff
@@ -1239,21 +1248,44 @@ rvtest_\__MODE__\()endtest:                     // target may be too far away, s
...
+
+        csrrc   T4, CSR_XIE, x0
+        srl     T4, T4, T2
+        andi    T4, T4, 1                       // Extracting XxIE bit into LSB
+        
+        sll     T4, T4, 17                      // Shift xxIE by 17
```
allenjbaum, line 1302: make this bit 24 after chgs above

riscv-test-suite/env/arch_test.h
```diff
@@ -1239,21 +1248,44 @@ rvtest_\__MODE__\()endtest:                     // target may be too far away, s
...
+        or      T6, T6, T4                      // insert xxIE into bits 17
+        
+        csrrc   T4, CSR_XIP, x0
+        srl     T4, T4, T2
+        andi    T4, T4, 1                       // Extracting XxIP into LSB
+        
+        sll     T4, T4, 18                      // Shift xxIP by 18
```
allenjbaum, line 1309: make this bit 25 after chgs above. The go fix the diagram above to match

riscv-test-suite/env/arch_test.h
```diff
@@ -2047,7 +2081,7 @@ CANARY
 .endm
 
 // Tests allocate normal signature space here, then define
-// the mtrap_sigptr: label to separate normal and trap
+// the mtrap_sigptr: label to separate normal and t
```
allenjbaum, line 2121: restore comment and add "signatures" afterward

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-SOFTWARE-INTR.S
```diff
@@ -0,0 +1,144 @@
...
+	#ifdef TEST_CASE_1
+		RVTEST_CASE(1,"//check ISA:=regex(.*32.*); check ISA:=regex(.*I.*Zicsr.*); def rvtest_mtrap_routine=True; def TEST_CASE_1=True; ",CHK_M_SOFTWARE_INTR)
+		RVTEST_SIGBASE( x3,signature_x3_1)
+
+		.attribute unaligned_access, 0
+		.attribute stack_align, 16
+		.align	2
+		.option norvc
+
+		main:
```
allenjbaum, line 52: This is probably true for all tests: I think you want to insert noops after any CSR write that might enable a trap to be taken, regardless of its supposed to or not. If you don't do that, the trap handler will cause instructions to be skipped when the trap is taken. ( won't comment on this again...)

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
```diff
@@ -0,0 +1,166 @@
...
+			// Prepares processor to handle pending interrupts
+			LI(a1,    MIE_MSIE + MIE_MTIE)
+			csrrs x0, CSR_MIE, a1
+			nop
+
+			// Enable Global Interrupts:
+			// Set the Machine Status Register (mstatus) Interrupt Enable (MIE) bit
+			// Allows maskable interrupts to be recognized
+			LI(t0,    MSTATUS_MIE)
+			csrrs x0, CSR_MSTATUS, t0
```
allenjbaum, line 114: This will trap... twice (on the mret of the first trap, it will immediately trap again with the other interrupt).
The trap handler will therefor advance the return address twice, and therefore at least two instructions after the instruction that causes the interrupt (the CSRRS) will be skipped. So, you need at least 2 noops after that instruction for this to work properly. I would put noops after every CSR write that might possibly cause a trap to be recognized, and maybe more.

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
```diff
@@ -0,0 +1,166 @@
...
+// ---------------------------------------------------------------------------------------------
+// Test Description:
+// This test validates the interrupt priority mechanism in RISC-V Machine Mode,
+// specifically demonstrating that software interrupts take precedence over timer interrupts
+// when both are simultaneously pending.
+// -----------
+// Test Methodology:
+// - Enable global machine-mode interrupts
+// - Disable interrupt delegation to ensure interrupts are handled in machine mode
+// - Simultaneously generate a software interrupt and a timer interrupt
```
allenjbaum, line 17: In general: SW and Timer interrupts are not guaranteed to see their respective interrupt pending bits immediately.
The MTIP/MSIP bits make take an unknown, but bounded time to be set after writing to the MMIO mtime/mtimecmp/swi_base addresses.

So, for this to work as expected, you must read and verify that the IP bits have been set before writing the global mstatus.mie bit (and loop polling for it if not), 
Or you can put a J .+0 infinite loop immediately after the setting of mstatus.mie. Then, when the trap is taken, it will skip [out of] the infinite loop upon returning, (so the infinite loop may never loop at all if the interrupt happens immediately, or it may happen after some arbitrary number of loops) This needs to be fixed in EVERY test. You can do something tricky by doing both, initializing a link register with an AUIPC, and using JAL .+0 as the infinite loop, followed by a store of (a relocated version of) the link register into the signature  after a noop or two. This will let you determine if it trapped immediately or not after polling showed that the IP bits were set
That is a very good thing because the spec does require the test to interrupt immediately after mtip, mie, and mtip are all set. You need to worry about this bounded delay in every test, so always check the IP bit before enabling the interrupt (unless you only use the infinite loop, in which case you don't care)

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-TIMER-INTR.S
```diff
@@ -0,0 +1,152 @@
...
+			// Verify interrupt behavior when MTIE is disabled and no timer interrupt is pending
+			LI(t0,    MIE_MTIE)
+			csrrc x0, CSR_MIE, t0
+			csrrs a4, CSR_MIP, x0
+			RVTEST_SIGUPD(x3,  a4)
+
+			// Test Case 02: Enable MTIE with mtime < mtimecmp
+			// Confirm interrupt state when MTIE is enabled but no timer interrupt is pending
+			csrrs x0, CSR_MIE, t0
+			csrrs a4, CSR_MIP, x0
```
allenjbaum, line 90: The spec says: MTIP is read-only in the mip register, and is cleared by writing to the memory-mapped
machine-mode timer compare register.MSIP is read-only in mip, and is written by accesses to memory-mapped control registers, which are used by remote harts to provide machine-level interprocessor interrupts, so this doesn't work. You should define and use the model specific RVMODEL_CLR_xTIMER_INT and RVMODEL_CLR_xSW_INT macros (where x is the mode) For setting them, use the SET versions of these (and do this in every tests)

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-TIMER-INTR.S
```diff
@@ -0,0 +1,152 @@
...
+			RVTEST_SIGUPD(x3,  a4)
+
+			// Test Case 03: Disable MTIE when mtime = mtimecmp
+			// Check program execution with pending timer interrupt but MTIE disabled
+			csrrc x0, CSR_MIE, t0
+			LREG  t1,  0(a0)
+			SREG  t1,  0(a1)
+			csrrs a4, CSR_MIP, x0
```
allenjbaum, line 98: note that there is no guarantee that MTIME won't increment before you set MIP, so it won't be equal (unlikely in fact)

coverage/m_interrupts/rv32_m_interrupts.cgf
```diff
@@ -0,0 +1,159 @@
...
+  config:
+    - check ISA:=regex(.*32.*); check ISA:=regex(.*I.*Zicsr.*); def rvtest_mtrap_routine=True;
+  mnemonics:
+      csrrc: 0
+      csrrs: 0
+      sw   : 0
+  val_comb:
+      # Verify MSIP set and clear mechanisms
+      '(call_type != "interrupt") and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mip & ${MIP_MSIP}) == ${MIP_MSIP})' : 0 # Checks software interrupt pending mechanism when writing 1 to memory-mapped MSIP control register without immediate interrupt service.
+      '(call_type != "interrupt") and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mip & ${MIP_MSIP}) == 0)' : 0 # Validates software interrupt clearing mechanism when writing 0 to memory-mapped MSIP control register.
```
allenjbaum, line 43: This one will get  hit right out of reset. This needs to be conditioned on the old value of MIP_MSIP

coverage/m_interrupts/rv32_m_interrupts.cgf
```diff
@@ -0,0 +1,159 @@
...
+  config:
+    - check ISA:=regex(.*32.*); check ISA:=regex(.*I.*Zicsr.*); def rvtest_mtrap_routine=True;
+  mnemonics:
+      csrrc: 0
+      csrrs: 0
+      sw   : 0
+  val_comb:
+      # Verify MTIP set and clear mechanisms
+      '(call_type != "interrupt") and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mip & ${MIP_MTIP}) == ${MIP_MTIP})' : 0 # Verifies Machine Timer Interrupt Pending (MTIP) bit setting mechanism during non-interrupt execution.
+      '(call_type != "interrupt") and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mip & ${MIP_MTIP}) == 0)' : 0 # Checks MTIP bit clearing mechanism when no interrupt is being processed.
```
allenjbaum, line 120: I think this will generate false positives unless you can check for edges rather than levels, i.e. mip=MTIP and old_mip!=MTIP

coverage/m_interrupts/rv32_m_interrupts.cgf
```diff
@@ -0,0 +1,159 @@
...
+# Confirms no interrupts are taken when global interrupt is disabled
+DIS_GLB_MIE:
+  config:
+    - check ISA:=regex(.*32.*); check ISA:=regex(.*I.*Zicsr.*); def rvtest_mtrap_routine=True;
+  mnemonics:
+      csrrs: 0
+      csrrc: 0
+      sw   : 0
+  val_comb:
+      '(call_type != "interrupt") and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mie & ${MIE_MSIE}) == ${MIE_MSIE}) and ((mip & ${MIP_MSIP}) == ${MIP_MSIP})' : 0 # Confirms that no software interrupt is taken when global machine interrupt is disabled, despite local interrupt enable being active.
```
allenjbaum, line 132: Anytime you are checking for call_type != "interrupt" you'll get false positives; but edge checking in all of them.

coverage/m_interrupts/rv32_m_interrupts.cgf
```diff
@@ -0,0 +1,159 @@
...
+  config:
+    - check ISA:=regex(.*32.*); check ISA:=regex(.*I.*Zicsr.*); def rvtest_mtrap_routine=True;
+  mnemonics:
+      csrrc: 0
+      csrrs: 0
+      sw   : 0
+      mret : 0
+  val_comb:
+      # Validate software interrupt handling
+      '(call_type == "interrupt") and ((mstatus & ${MSTATUS_MPP}) == ${MSTATUS_MPP_M}) and ((mstatus & ${MSTATUS_MPIE}) == ${MSTATUS_MPIE}) and ((mstatus & ${MSTATUS_MIE}) == 0) and ((mcause & ${INTERRUPT_MASK}) == ${CAUSE_M_SOFT_INTR}) and ((mie & ${MIE_MSIE}) == ${MIE_MSIE}) and ((mip & ${MIP_MSIP}) == ${MIP_MSIP})': 0 # Validates software interrupt processing when global machine interrupt is enabled, verifying all required interrupt conditions.
```
allenjbaum, line 146: Is this testing only trapping from Mmode? Don't we also want to test from Umode and Smode? Pretty tricky because you wo't necessarily have access to Mmode registers.

riscof-plugins/rv32/sail_cSim/env/model_test.h
```diff
@@ -52,13 +56,32 @@ RVMODEL_DATA_SECTION			;\
 //RVTEST_IO_ASSERT_DFPR_EQ
 #define RVMODEL_IO_ASSERT_DFPR_EQ(_D, _R, _I)
 
-#define RVMODEL_SET_MSW_INT
+#define RVMODEL_SET_MSW_INT             ;\
```
allenjbaum, line 66: The RVMODEL_Set/Clr_M/H/Sxxx_INT macros need to be redefined so that the aren't placed inline, but are called as simple subroutines with a JAL, and return to the test with a JR. The should be instantiated (not defined) in  model_test.h as they are #defined, but with the name as a label at the start (instead of #define), and a JR at the end with a fixed link register (that tests will have to accomodate).

riscv-test-suite/rv32i_m/m_interrupts_32/src/CHK-M-INTR-PRIORITY.S
allenjbaum: Actually, the skipping behavior might be exceptions only; need to check on that & see if we want to maintain skipping behavior

allenjbaum (changes requested):
This is the first batch of overall changes that are needed. I'll come back and do another review when these are fixed (I haven't made it all the way through yet.

jamesbeyond:

PR #570, opened 2024-11-26
[CTG] Replace set with OrderedSet

### LINKED ISSUE #566: CTG tests are not reproducible
After some testing, I found that the tests ctg (without `-r` option) generated each time is different. This is completely different from what I expected. I also had another person try it, and the results were also different.

I think Tim have already found this issue a few month ago: https://github.com/riscv-software-src/riscv-ctg/issues/115

I'm trying to fix this, but simply appending a line `random.seed(10)` after all `import random` statements does not solve the problem. I will check it again.

Update: The reason is the use of `set` in the code, and I am replacing them with `ordered_set`

---
if you don't know what `-r` option does:

`-r, --randomize                 Randomize Outputs.`

> If randomization is enabled we use the `MinConflictsSolver` solver to find solutions.

https://github.com/riscv-non-isa/riscv-arch-test/blob/cd94912fed2aab75d7d5f115b441da0813fdce8d/riscv-ctg/riscv_ctg/generator.py#L493-L496

I haven't tested the behavior of `MinConflictsSolver`, which will use random inside

_Originally posted by @trdthg in https://github.com/riscv-non-isa/riscv-arch-test/issues/565#issuecomment-2488203492_

---

Update: In summary, two steps are needed to solve the problem here
- ~random: <>~
- set: #570

trdthg:
## Description

resolve #566 

- Add new dependency `ordered-set`
	
  - I tried to implement an ordered-set myself, but I found it unnecessary; the ordered-set library is streamlined enough (only one file) and is fully compatible with the current usage of set (for example, overloading various operators)
- Replace `set` with `OrderedSet`
  
  - riscv-ctg use set to remove duplicates, but this will cause the unstable ordering of the for loop, and finally the generated tests will be different each time.

- Several very small if-else optimizations
  ```diff
  -if op in cgf:
  -     op_conds[op] = set(cgf[op])
  -else:
  -     op_conds[op] = set([])
  +op_conds[op] = OrderedSet(cgf.get(op, []))
  ```


UmerShahidengr (approved):
Great Fix. I didn't know there is such library exist in Python which will fix this issue that easily.
Well done @trdthg

PR #578, opened 2024-12-12
[ISAC] Add CMO Support

trdthg:

## Description

Add decoder support for CBO extension ('cbo.clean', 'cbo.flush', 'cbo.inval', 'cbo.zero', 'prefetch.i','prefetch.r','prefetch.w')

I achieved 100% test coverage during testing, so I think the decoder is fine.

One thing to note is that the encoding of the `prefetch` instruction is the same as `ori`, the only difference is that the `rd` of `prefetch` must be 0, so it is handled in `arithi_ops`


UmerShahidengr (approved):
Looks good to me

Issue #559, opened 2024-11-12
fcvt.d.s tests only test non-NaN-boxed input values


jakubadler:
In [fcvt.d.s_b1-01.S](https://github.com/riscv-non-isa/riscv-arch-test/blob/dev/riscv-test-suite/rv32i_m/D/src/fcvt.d.s_b1-01.S) (and all other fcvt.d.s tests), all the test data use 64 for the `__width__` argument of `NAN_BOXED` even thought the argument is supposed to be float32. I assume it should be 32 in most cases. The way it is now, no NaN-boxing takes place and thus only the one case of the input operand not being correctly NaN-boxed is tested.

UmerShahidengr:
PR #581 has solved this issue, so closing this on. Thank you for raising the issue, and thank you, @Pagerd, for resolving it.

PR #581, opened 2024-12-18
[ACT]fix fcvt_d_s testcases

Pagerd:
### Description
change fcvt.d.s NaN-boxed input
### Related Issues
https://github.com/riscv-non-isa/riscv-arch-test/issues/559

- [X] D

### Reference Model Used
- [X] SAIL

### Mandatory Checklist:
- [X] Ran the new tests on RISCOF in [coverage mode](https://riscof.readthedocs.io/en/stable/commands.html#coverage)
- [X] Link to Google-Drive folder containing the new coverage reports): < [coverage reports ](https://drive.google.com/drive/folders/1BO2CSCInwqirH70-gprP9pkdnjWXuvRj) >


allenjbaum (commented):
I'd much rather see this data in hex rather than decimal. I have no confidence in what it is doing by looking at this.

Pagerd:
> I'd much rather see this data in hex rather than decimal. I have no confidence in what it is doing by looking at this.

All D extension test cases are tested using decimal data, including the modified `fcvt.d.s` test case, so I used the same method to modify it. Is there a problem with this method? Many test cases are tested in this way

allenjbaum:
IEEE FP is a binary format, not a decimal format.
So yes, there is a problem for anyone who reviews the tests  (me).
And, yes, that should have been followed for the earlier tests as well.

UmerShahidengr:
This one is a simple fix, we can merge this one, and start an effort to convert all NAN_BOXED data in hex rather than decimal. I will create an issue on this.

UmerShahidengr (approved):
Looks good to go

PR #370, opened 2023-07-28
Add ratified Zacas extension

ved-rivos:
<FOR DOC UPDATES FILL ONLY DESCRIPTION AND RELATED ISSUES SECTION AND REMOVE THE OTHERS>

## Description
This patch introduces a coverage model and the generated test cases using riscv_ctg. The tests have been generated using the following command:

`riscv_ctg.py --cgf coverage/dataset.cgf --cgf coverage/zicond.cgf -bi rv32i -d rv32tests/`
`riscv_ctg.py --cgf coverage/dataset.cgf --cgf coverage/zicond.cgf -bi rv64i -d rv64tests/`

The generated tests are placed in 
`rv32i_m/Zacas/src` 
`rv64i_m/Zacas/src` 

### Ratified/Unratified Extensions

- [X] Ratified
- [ ] Unratified

### List Extensions

Zacas specification: https://github.com/riscv/riscv-zacas

### Reference Model Used

- [ X] SAIL
- [ X] Spike
- [  ] Other - < SPECIFY HERE >

### Mandatory Checklist:

  - [ x] All tests are compliant with the test-format spec present in this repo ?
  - [  ] Ran the new tests on RISCOF with SAIL/Spike as reference model successfully ?
  - [  ] Ran the new tests on RISCOF in [coverage mode](https://riscof.readthedocs.io/en/stable/commands.html#coverage)
  - [  ] Link to Google-Drive folder containing the new coverage reports ([See this](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/CONTRIBUTION.md#uploading-test-stats) for more info): < SPECIFY HERE >
  - [  ] Link to PR in RISCV-ISAC from which the reports were generated : < SPECIFY HERE > 
  - [ X] Changelog entry created with a minor patch

### Optional Checklist:

  - [x ] RISCV-V CTG PR link if tests were generated using it : https://github.com/riscv-software-src/riscv-ctg/pull/73
  - [  ] Were the tests hand-written/modified ?
  - [  ] Have you run these on any hard DUT model ? Please specify name and provide link if possible in the description
  - [ ] If you have modified arch\_test.h Please provide a detailed description of the changes in the Description section above.

PR #583, opened 2024-12-20
[ISAC] Add Zacas support

trdthg:
## Description

#370 is merged, but seems didn't run coverage mode test?

sail has a zacas pr, but the problem is that gcc doesn't has zacas support yet. I'll keep watching it.


UmerShahidengr (approved):
simple fix. Thank you @trdthg

PR #596, opened 2025-01-13
[ISAC] Add Zimop decoder

trdthg:
## Description

It seems that the support for zimop/zicfiss in gcc is still under review

https://gcc.gnu.org/pipermail/gcc-patches/2024-November/668918.html


jamesbeyond:
ACT repository have been updated `act4` as the default branch, ISAC is no longer used in act4. So we have to close this PR.

jamesbeyond:
ACT repository have been updated `act4` as the default branch, ISAC is no longer used in act4, So we have to close this PR

PR #597, opened 2025-01-13
Fix ISA strings for Zcf, Zcd, and RV64 sra

### LINKED ISSUE #590: Zcf/Zcd RVTEST_CASE macros inconsistent
The new Zcf and Zcd tests introduced in #497 and #587 are not consistent in their use of the RVTEST_CASE isa regex macro.

For rv32 and rv64 Zcd, c.fsdsp uses `check ISA:=regex(.*I.*D.*C.*)` while all of the other tests use `check ISA:=regex(.*I.*F.*D.*C.*)`. D implies F so this one isn't a big deal, but they should be consistent.

More significantly, for rv32 Zcf, c.fswsp uses `check ISA:=regex(.*I.*F.*C.*)` while all of the other tests use `check ISA:=regex(.*I.*F.*Zcf.*)`. Either approach is acceptable, but using `Zcf` requires modifying the ISA string that is passed to RISCOF when selecting tests. 

All of the Zcf and Zcd tests should use the same approach for this, either using the `Zcf` and `Zcd` extensions or the `C` and `F`/`D` extensions.

jordancarlin:
## Description

Fix inconsistent ISA strings in new Zcf and Zcd tests. Also restore `RV64` to sra ISA regex to avoid the following error:
```
ERROR | Error in test: /home/harris/cvw/addins/riscv-arch-test/riscv-test-suite/rv64i_m/I/src/sra-01.S
Test Selected without the relevant extensions being available on DUT.
```

### Related Issues

@UmerShahidengr fixes #590

### DIFF
--- a/riscv-test-suite/rv32i_m/D_Zcd/src/c.fld-01.S
+++ b/riscv-test-suite/rv32i_m/D_Zcd/src/c.fld-01.S
@@ -19,7 +19,7 @@
 // 
 #include "model_test.h"
 #include "arch_test.h"
-RVTEST_ISA("RV32IFDC")
+RVTEST_ISA("RV32IFD_Zcd")
 
 .section .text.init
 .globl rvtest_entry_point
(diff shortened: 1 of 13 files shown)


UmerShahidengr:
Thank you @jordancarlin for the quick followup

UmerShahidengr (approved):
Quick fix for the ISA string variations. Approved

PR #605, opened 2025-01-22
Added Tests and Coverage for Exception Handling(Illegal instruction, environment calls and misaligned address exceptions) and Trap State Restoration Exceptions

maria-rehman:

## Description

>Includes tests and coverage of illegal instructions, environment calls, address misaligned exceptions, trap entry/exit state handling, and validation of mcause, scause, mtval, and stval values.

### Related Issues

> N/A

- [x] Ratified

### List Extensions

> RV32I: Base integer instruction set
> S: Supervisor mode extension
> U: User mode Extension
> F: Single-precision floating-point extension
> D: Double-Precision Floating-Point
> Zicsr: Control and status register instructions

### Reference Model Used

- [x] SAIL

### Mandatory Checklist:

 > All tests are compliant with the test-format spec present in this repo
 > Ran the new tests on RISCOF with SAIL/Spike as reference model successfully
> Ran the new tests on RISCOF in coverage mode
> Link to Google-Drive folder containing the new coverage reports: > https://drive.google.com/drive/folders/1XwiTjZUGTCR7KKmygZ9NpDvy9ydp-lGH?usp=drive_link

### Optional Checklist:

- [x]  Tests are hand-written
- [x]  Ran these tests on DUT model(Spike)
- [x]  Modified test_macros.h file, added one macro to off mstatus.FS bits (RVTEST_FP_DISABLE)


riscv-test-suite/rv32i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,228 @@
...
+	LI (a4, 0xceed)
+	RVTEST_SIGUPD(x13,a4)
+#endif
+
+// In M- mode execute the ecall instruction,  MXLEN-1 bit of mcause will become zero and the 
+// the exception code will be 11(b) for the environment call from M mode, mepc register will 
+// get the address of the ecall instruction verify by checking both registers values
+
+	// machine ecall  
+	ecall
```
allenjbaum, line 39: Unless the instruction immediately previous to ecall is a beq x,x,.+4, this must ensure that x2!=0 because the trap handler will treat that as a parameter to the handler.
(if X2==0, this is GOTO_MMODE, !=0, it's the ECALL test, anything is is a test failure)

riscv-test-suite/rv32i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,228 @@
...
+// When a trap occurs, mtvec will hold the address of the respective trap handler. 
+// This test verifies whether the system returns to the correct state after handling the trap. 
+// It checks the value of mcause before entering the trap handler and after handling the trap 
+// to ensure it reflects the expected values for different exceptions. Additionally, 
+// it confirms that mtval contains the correct values for various exceptions.
+	
+	LI (a4, CSR_MEDELEG)  
+	csrw medeleg, a4
+	
+	ecall               // This exception will be handled in the machine mode
```
allenjbaum, line 126: Unless the instruction immediately previous to ecall is a beq x,x,.+4, this must ensure that x2!=0 because the trap handler will treat that as a parameter to the handler.
(if X2==0, this is GOTO_MMODE, !=0, it's the ECALL test, anything is is a test failure)

riscv-test-suite/rv32i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,228 @@
...
+// This test checks that if an extension is turned off (e.g., FS bits [13 and 14] in mstatus 
+// are 00), trying to access it will cause an illegal instruction error. 
+
+
+ RVTEST_FP_DISABLE()  // this will set the FS bit in mstatsus to 00 means off
+   // Load immediate values into integer registers
+    li t0, 0x3FC00000    //  Hex representation of 1.5 in IEEE-754 single-precision
+    li t1, 0x40200000    // Hex representation of 2.5 in IEEE-754 single-precision
+
+    // Move these values into floating-point registers
```
allenjbaum, line 144: This test should also try a csrrd x2, FCSR  (and FRM and FFLAGS to ensure that they trap also

riscv-test-suite/rv32i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,228 @@
...
+    /* When we try to jump to the misaligned address then instead of instruction misaligned adddress it gives the illegal instruction exception */
+    
+ /*   
+// This test will check that in machine mode, a load address misaligned exception will be raised 
+// when we try to attempt to load data from a misaligned memory address.For this we will check 
+// the log file that either it is giving the expected exception at that point and value of the 
+// mcause exception code will be  "4" in U mode
+*/
+
+	LI (a4, 0x0008)
```
allenjbaum, line 172: This is already tested in the misalign_lw_01.S test quite extensively. Also note that it very well might not trap, depending on misalign granularity

riscv-test-suite/rv32i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,228 @@
...
+	LI (a4, 0x0008) 
+   	LA(x17,rvtest_data)
+	lw x5,3(x17)
+	nop  
+    	RVTEST_SIGUPD(x13, a4)    // this signature will give information that test for loas address misaligned has successfullty returned from trap
+    
+       
+// This test will check that in machine mode, a store address misaligned exception is raised when we try 
+// to attempt to store data to a misaligned memory address. For this we will check the log file 
+// that either it is giving the expected exception at that point and value of the mcause is "6" in M mode
```
allenjbaum, line 181: Same comment: this appears to be redundant with (much more extensive) tests that already exist.

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+RVMODEL_BOOT
+RVTEST_CODE_BEGIN
+
+#ifdef TEST_CASE_1
+    RVTEST_CASE(1,"//check ISA:=regex(.*32.*); check ISA:=regex(.*I.*S.*Zicsr.*); def rvtest_mtrap_routine=True; def rvtest_strap_routine=True; def TEST_CASE_1=True ", cp_S_ecall,cp_S_excp_illegal, cp_S_load_addr_misaligned, cp_S_store_addr_misaligned) 
+
+RVTEST_SIGBASE( x13,signature_x13_1)
+
+main:
+#ifdef rvtest_mtrap_routine	// Verification of existance of rvtest_mtrap_routine
```
allenjbaum, line 29: you should fiix the second comment  mtrap->strap

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+	nop
+	nop
+
+// In S- mode execute the ecall instruction,  MXLEN-1 bit of mcause will become zero and the 
+// the exception code will be "9" for the environment call from "S" mode, mepc register will 
+// get the address of the ecall instruction and then verify by checking signatures
+// If S mode is implemented then will be used
+
+	// supervisor ecall 
+	ecall
```
allenjbaum, line 53: You must ensure x2 !=0 or the trap handler will not do what you expect

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+	RVTEST_SIGUPD(x13,a4)
+
+// If we try to access the non existinfg CSR and the instruction is csrr  
+// 0xfff in S mode so it will cause the illegal instruction and will update the mcause 
+// with the exception code 2, mepc will point to the address of the illegal 
+// instruction, mtval will contain the binary representation of illegal instruction
+
+	
+	LI (a4, 3)     
+	csrrc t1,0xFFF,a4  // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 67: There is no longer a requirement that unimplemented CSRs will trap.
In fact
Without an explicet spec requirement, and/or an explicit configuration option that the CSR is implemented, you must not try to touch it because you don't know what is going to happen. All such accesses must be surrounded by an #ifdef CRS_XXX_IMPLEMENTED ... #endif conditional

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+	                                // will cause illegal instruction exception
+	nop
+	
+	LI (a4, 2)																
+	csrw satp, a4           // writing satp exceptions will occur here
+	csrr t3,satp
+	nop
+	nop
+	
+	sfence.vma              // using sfence.vma but no faults occur
```
allenjbaum, line 128: First, in general, unless alignment is known, you should put a single noop after any instruction that can possibly trap. It's good practice to do this even if you think it will not trap (e.g. if there's a bug in the HW, but you still want the test behavior to be consistent).  
(Also, You can change 2 nops to just 1; I've been convinced it isn't strictly necessary)
Note that if you're writing compressed op tests, then there is a possibility that an unaligned 4B op will trap, and that specific case will only pad with 2B, so it should be followed by a C.NOP
Secondly, in the comment above you say Smode execution of fence.vma will trap if TVM is set, and the comment here is that it doesn't.  One of these comments needs to be fixed 
(by the way, your comments are great - not everyone does it well, but this is the best I've seen)

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+// Additionally, it confirms that stval contains the correct values for various exceptions.
+	
+	LI (a4, CSR_MEDELEG)  
+	csrw medeleg, a4
+	
+	RVTEST_GOTO_LOWER_MODE Smode
+	nop
+	nop
+	
+	ecall               // This exception will be handled in the Supervisor mode
```
allenjbaum, line 156: standard issue with ecall: x2 must !=0. The trap handler does store both its own mode and the handler mode into the trap signature.

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+// This test checks that if an extension is turned off (e.g., FS bits [13 and 14] in mstatus 
+// are 00), trying to access it will cause an illegal instruction error. 
+
+
+ RVTEST_FP_DISABLE()  // this will set the FS bit in mstatsus to 00 means off
+   // Load immediate values into integer registers
+    li t0, 0x3FC00000    //  Hex representation of 1.5 in IEEE-754 single-precision
+    li t1, 0x40200000    // Hex representation of 2.5 in IEEE-754 single-precision
+
+    // Move these values into floating-point registers
```
allenjbaum, line 175: why do you do this twice (with 2 different values)?

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+    fadd.s f2, f0, f1    // f2 = f0 + f1      // will give illegal instruction exception 
+    nop
+    nop
+    
+    LI (a4, 0x0007)   
+    RVTEST_SIGUPD(x13, a4)    // this signature will give information that test for FS extension off has successfullty completed
+   
+    
+    
+    /* When we try to jump to the misaligned address then instead of instruction misaligned adddress it gives the illegal instruction exception */
```
allenjbaum, line 194: This test already exists in the repo as misalign_ld/st - not needed.

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+    
+    
+// This test will check that in Supervisor mode, a load address misaligned exception will be raised
```
allenjbaum, line 197: note that this may or may not trap, depending on the capabilities of the core

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+       
+// This test will check that in Supervisor mode, a store address misaligned exception is raised when we try 
+// to attempt to store data to a misaligned memory address. For this we will check the log file 
+// that either it is giving the expected exception at that point and value of the mcause is "6" in M mode
+	
+	
+	
+	LA(x17,rvtest_data)
+	sw t1,0(x17)    // will not give any exception
+	sw t2,3(x17)    // will cause the exception
```
allenjbaum, line 217: no guarantee that this will trap,. This should replace the first nop with li -1, and the 2nd nop with a SIGUPD.

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+	nop
+	nop
+
+// In U- mode execute the ecall instruction,  MXLEN-1 bit of mcause will become zero and the 
+// the exception code will be "8" for the environment call from "U" mode, mepc register will 
+// get the address of the ecall instruction and then verify by checking signatures
+
+
+	// user ecall 
+	ecall
```
allenjbaum, line 53: usuall ecall issue: needs x2!=0

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+	RVTEST_SIGUPD(x13,a4)
+
+// If we try to access the non existinfg CSR and the instruction is csrr  
+// 0xfff in U mode so it will cause the illegal instruction and will update the mcause 
+// with the exception code 2, mepc will point to the address of the illegal 
+// instruction, mtval will contain the binary representation of illegal instruction
+
+	
+	LI (a4, 3)     
+	csrrc t1,0xFFF,a4  // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 67: no guarantee this will cause an exception. FFF is a custom readonly register, so there is no guarnatee what it will do if you try to execute this. Note that the you are in Umode at this time, so if you simply want to generate an illegal op exception, any legal mmode CSR will be safer and will guarantee the trap. But if the purpose is to show that a non-existent CSR access will trap - it's not possible unless Ssstrict extension is implemented (I think). At some point there may be a way to configure Sail to determine which CSRs will result in a trap if accessed, but even when that comes, you'll need to define it with an RVMODEL_CSR_TRAP variable (and skip the test if it isn't defined), so each implementation will have a slightly different test (but will have the same signature)

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+
+	
+	LI (a4, 0x100)  
+	csrw medeleg, a4
+	
+	RVTEST_GOTO_LOWER_MODE Umode
+	nop
+	nop
+	
+	ecall               // This exception will be handled in the Supervisor mode
```
allenjbaum, line 155: make sure to set x2 correctly (!=0) Also: this is a case where you can never get back to Mmode, so you need to install a special  illegal_op handler (which should uninstall itself) to get back

allenjbaum (changes requested):
This has a misuse of ECALL in the test. IT should always set x2 to 0x1111_1111 to be handled by the trap handler properly. There also a misunderstanding of how unimplemented CSR are not handled in the spec (not guarnateed to trap), how misaligned occurs (may or may not trap), coments that don't seem to match the code.
The same procedue is used for all the test; it would make it more readable if you created a test macro for all the boilerplate to (greatly) reduce test size

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+// instruction mcause will have the exception code 2 like if try to access the mepc and 
+// mstatus register then will cause illegal instruction
+
+	// try to access the in appropriate csr 
+	LI (t0, 0x1111)
+	csrrs t1, mstatus, t0 // Read mstatus into t1 and set the bits as indicated in t0
+	nop
+	nop
+	
+	LI (a4, 0x0003) // this signature will give information that test for accessing another mode csr has successfully returned from the trap handler
```
allenjbaum, line 85: If the trap handler wasn't ever entered, this would make it seem like it did.
The proper way to do this would be to preload a4 with some value X before the possible trap, load it immediately after the CSR write with a different value Y, and then store it. If the trap is entirely skipped, you'll see Y If it does trap and properly return you should see X.

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+	RVTEST_SIGUPD(x13,a4)
+
+// This test will check if try to write to the read onlty register then 
+// it will also cause the illegal instruction and mcasue will have exception 
+// code 2 and the mepc will have the address of the illegal instruction .
+// e.g storing any value in the mvendorid register which is read only.
+	
+	// Try to write the supervisor mode  "read only CSR" then exception in the S mode will occur
+	LI (a4, 3)
+	csrrw t1, 0xDC0, a4      // this will give an illegal instruction as trying to write the read only register of supervisor mode
```
allenjbaum, line 95: if CSR 0xDC0 has a name, it should be used. If it doesn't have a name, then it shouldn't be used in any test. Ditto for 0xFFF, unless that is known to always exist and always cause a trap.

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+// This test will check that when M mode has delegated the illegal instruction exception to S mode, 
+// and S mode executes an illegal instruction then trap is taken in S mode, So after delegation 
+// we will check that wheather it is being handled by expected handler if delegation is not done 
+// then trap will be handled in the M mode. When a trap occurs, stvec will hold the address of the 
+// respective trap handler. This test also check whether the system returns to the correct state 
+// after handling the trap. It checks the value of scause before entering the trap handler and 
+// after handling the trap to ensure it shows the expected values for different exceptions. 
+// Additionally, it confirms that stval contains the correct values for various exceptions.
+	
+	LI (a4, CSR_MEDELEG)
```
allenjbaum, line 149: This is the wrong immediate; this will put the address of MEDELEG into the mdeleg CSR. You want to put the mask bit for the  ECALL from Smode, which is 1<<CAUSE_SUPERVISOR_ECALL

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
allenjbaum: Not strictly necessary to have 2 diffferent values, but it doesn't hurt, so resolving

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
allenjbaum: Also: this is extremely dangerous. Once you've delegated ECALL to Smode, you can't goto Mmode or clear MEDELEG without installing a special illegal opcode handler and executing an illegal opcode.
The special handler will clear medeleg, then jump to the goto_mmode handler.
There is an array of pointers at excpt_\__MODE__\()hndlr_tbl, and the offset CAUSE_SUPERVISOR_ECALL*8 must be filled in with the address of this special handler, which will clear MEDELEG, restore the pointer to its default value of CAUSE_SUPERVISOR_ECALL*2+1, and branch to rtn2mmode:. Note that if you're executing this in Umode, you need to do the same thing, except with value CAUSE_USER_ECALL
We probably should define a macro to set and clear this handler because I'm sure it will come up again...

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+	nop
+	nop
+	
+	LI (a4, 0x0006)   
+	RVTEST_SIGUPD(x13, a4)    // this signature will give information that test for trap handler delegation 
+				  // for ecall has successfullty returned from trap handler
+	
+	
+ 
+// This test checks that if an extension is turned off (e.g., FS bits [13 and 14] in mstatus
```
allenjbaum, line 166: The comment isn't quite correct. It isn't any extension, this is specifically FP extensions.
 I believe that includes F, as well as D, Zfh,  BF26, Zfa  (though they all depend on F). 
You may want to test it with ops other than just FADD, (especially CSR reads of FCSR, FRM, and FFLAGS)

riscv-test-suite/rv32i_m/test/S_merged/S_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,264 @@
...
+	
+	
+	
+	LA(x17,rvtest_data)
+	sw t1,0(x17)    // will not give any exception
+	sw t2,3(x17)    // will cause the exception
+	nop
+	nop
+	
+        LI (a4, 0x0009)
```
allenjbaum, line 221: This can be deleted since it is handled as above

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+// In this access to the csr without appropriate privillage mode will cause the illegal 
+// instruction mcause will have the exception code 2 like accesing the sstatus register 
+// then will cause illegal instruction, because Csr of S mode are not accesible in U mode
+
+	csrrs t0, sstatus, x0
+           // try to write random value to sstatus
+	nop 
+	nop
+	
+	LI (a4, 0x0003) // this signature will give information that test for accessing another mode csr has successfully returned from the trap handler
```
allenjbaum, line 84: I think this would be a more robust test (**here, and in all other instances!**) if the LI () was moved above the CSR access, and the 2 nops after was replaced with a different LI() value.

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+	                                // will cause illegal instruction exception
+	nop
+	// the satp register and the sfence.vma instruction are not accessible regardless 
+	// of the mstatus.TVM value in U mode
+	
+	// Now use the satp and sfence.vma and it will generate illegal exception 
+	
+	LI (a4, 2)																
+	csrw satp, a4           // writing satp exceptions will occur here
+	csrr t3,satp
```
allenjbaum, line 126: these should be separated by at least one instruction, that instruction should reload or increment t4 by 1

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+	nop
+	nop
```
allenjbaum, line 128: these nops should be replace with an increment of t4 by 2

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+	
+	sfence.vma              // using sfence.vma but no faults occur
+	nop
```
allenjbaum, line 131: This noop should be replaced by an increment of t4 by 4, and the following LI() should be removed.
What this will do is produce a unique value in t4, where the lower 3 bits indicate which (if any) of the preceding ops failed to trap.

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+	nop
+	
+// Traps never transition from a more-privileged mode to a less-privileged mode.
+// but if in M mode, if delegate the exceptions to handle in S mode then 
+//  U mode exceptions will be handled by S-handler. Here below 
+// delegated the environment call in U mode to be handled in S mode
+
+	
+	LI (a4, 0x100)
```
allenjbaum, line 148: please replace constant with a mnemonic, e.g. 1<<CAUSE_USER_ECALL

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+ 
+// This test checks that if an extension is turned off (e.g., FS bits [13 and 14] in mstatus 
+// are 00), trying to access it will cause an illegal instruction error. 
+
+ 	RVTEST_FP_DISABLE()  // this will set the FS bit in mstatsus to 00 means off
+    // Load immediate values into integer registers
+    li t0, 0x3FC00000    //  Hex representation of 1.5 in IEEE-754 single-precision
+    li t1, 0x40200000    // Hex representation of 2.5 in IEEE-754 single-precision
+
+    // Move these values into floating-point registers
```
allenjbaum, line 173: This should move the LI above each op that is supposed to trap, replace each noop pair with a increment by 1<<op# (this should be done everywhere you have a series of ops that trap . This ensures that if a trap is not taken in self_test mode,  it will be caught .

riscv-test-suite/rv32i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,261 @@
...
+    
+    LI (a4, 0x0007)   
+    RVTEST_SIGUPD(x13, a4)    // this signature will give information that test for FS extension off has successfullty completed
+   
+    
+    
+    /* When we try to jump to the misaligned address then instead of instruction misaligned adddress it gives the illegal instruction exception */
+    
+    
+// This test will check that in user mode, a load address misaligned exception will be raised
```
allenjbaum, line 195: Why do you expect an exception to be raised? It won't if there is HW support of misaligned accesses.
Note that regardless of wehther it does or not, you don't need to change the test, as it will compare with whatever Sail does. But: you have a vunch of ops that might or might not trap. You should remove the last LI(), remove the nops, and insert addi a4,a4,1<<op# after each op that might trap (even if you think it won't!), as mentioned above. This should be done in every test, so I won't mention i t again.

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_illegal.S
```diff
@@ -0,0 +1,202 @@
...
+	LI (a4, 0xbeed)
+	RVTEST_SIGUPD(x13,a4)
+#endif
+
+	ALL_MEM_PMP          		  // set the PMP permissions for the whole memory
+	csrw satp, zero  		  // write satp with all zeros (bare mode)
+
+	// illegal instruction exception in M mode
+	LI (t1, 0xdead)
+	csrw 0xFFF, t1      // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 63: NOT guaranteed to trap! Use a different method for causing an illegal op trap. Also: don't you have several tests that do this already? You don't need another.

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_illegal.S
```diff
@@ -0,0 +1,202 @@
...
+	RVTEST_SIGUPD(x13,a4)
+	
+	RVTEST_GOTO_LOWER_MODE Smode    // goto S mode
+	nop
+	nop
+	
+	
+	// illegal instruction exception in S mode
+	LI (a4, 0xabcd)     
+	csrrc t1,0xFFF,a4  // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 78: as above: this doesn't work

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_illegal.S
```diff
@@ -0,0 +1,202 @@
...
+	nop
+
+	
+	RVTEST_GOTO_LOWER_MODE Umode
+	nop
+	nop
+
+	// illegal instruction exception in U mode
+	LI (a4, 0xbeef)
+	csrrs t1,0xFFF,a4  // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 98: as above; why is this even necessary at this point?

allenjbaum, line 98: As above - not guaranteed to trap. Does this need to be a CSR access, or will execution of an allzero opcode work just as well?

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_illegal.S
```diff
@@ -0,0 +1,202 @@
...
+	LI (a4, 0x0003)
+	RVTEST_SIGUPD(x13, a4)
+	nop
+	nop
+
+	RVTEST_GOTO_MMODE
+	nop
+	nop
+
+	LI (a4, 0x4)
```
allenjbaum, line 111: USe a mnemonic, not a constant here, e.g. 1<<CAUSE_ILL_OP or whatever

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_illegal.S
```diff
@@ -0,0 +1,202 @@
...
+	csrw medeleg, a4     // delegate illegal instruction exception 
+
+	// illegal instruction exception in M mode
+	LI (t1, 0xdead)
+	csrw 0xFFF, t1      // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 116: as above

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_illegal.S
```diff
@@ -0,0 +1,202 @@
...
+	nop
+
+	
+	RVTEST_GOTO_LOWER_MODE Umode
+	nop
+	nop
+
+	// illegal instruction exception in U mode
+	LI (a4, 0xbeef)
+	csrrs t1,0xFFF,a4  // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 152: pick a different illegal op

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_ld_sw.S
```diff
@@ -0,0 +1,290 @@
...
+#endif
+
+	ALL_MEM_PMP          		  // set the PMP permissions for the whole memory
+	csrw satp, zero  		  // write satp with all zeros (bare mode)
+	
+
+	LI (a4, 0xaaaa) 
+   	LA(x17,rvtest_data)
+	lw x5,3(x17)
+	nop
```
allenjbaum, line 66: replace with different LI() value

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_ld_sw.S
```diff
@@ -0,0 +1,290 @@
...
+    nop
+    nop
+    	
+    RVTEST_GOTO_LOWER_MODE Umode
+    nop
+    nop
+    	
+    LI (a4, 0xcccc) 
+   	LA(x17,rvtest_data)
+	lw x5,1(x17)          // exception
```
allenjbaum, line 138: SIGUPD after load tests should store the register that was loaded

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_ld_sw.S
```diff
@@ -0,0 +1,290 @@
+/* 
+/*
+        Note: Trap here in this file is taken load store address misaligned traps
```
allenjbaum, line 3: There is no guarantee that any of these will trap, so you have to ensure that the noop after every load/store loads a value different than the LI() before the load/store. If the op that traps is a load, then the way to do that is for SIGUPD to use the register that was loaded into. Also, make it clear that you're looking at different modes, and different widths being loaded or stored.

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_ld_sw.S
```diff
@@ -0,0 +1,290 @@
...
+
+	LI (a4, 0xaaaa) 
+   	LA(x17,rvtest_data)
+	lw x5,3(x17)
+	nop  
+    RVTEST_SIGUPD(x13, a4)    // this signature will give information that test for loas address misaligned has successfullty returned from trap
+    
+    LI (a4, 0xaaaa) 
+   	LA(x17,rvtest_data)
+	lh x5,3(x17)
```
allenjbaum, line 71: Use lh, x4... instead of x5, so SIGUPD gets the data that was loaded, to make sure it loaded the right thing (or didn't load anything, in the case a trap was taken.

riscv-test-suite/rv32i_m/test/traps/ms_cause_ms_tval_values/trap_entry_test_ld_sw.S
```diff
@@ -0,0 +1,290 @@
...
+    LI (a4, 0xcccc)   
+    RVTEST_SIGUPD(x13, a4)     // this signature will give information that test for store address misaligned has successfullty returned from trap
+
+    	
+    RVTEST_GOTO_MMODE
+    nop
+    nop
+    	
+	                       // Delegate and then check the same valuse for scause and stval//
+	LI (a4, 0x50)
```
allenjbaum, line 173: replace constant with nmemonic

riscv-test-suite/rv32i_m/test/undefined_opcodes/opcode_comb_not_covered.S
```diff
@@ -0,0 +1,110 @@
...
+	add x0,x0,x0
+	
+	
+	.word 0x00000013     // nop
+	nop
+	
+	// Here below are the some opcode combinations that are not defined for any instruction 
+	// so when we use such opcodes they give the illegal instruction exception
+	
+	// "110_0111" //
```
allenjbaum, line 56: Undefined opcodes are not guaranteed to trap unless Zsstrict is implemented. 
All zero ops are guaranteed, but not sure what else is.

riscv-test-suite/rv64i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,229 @@
...
+	RVTEST_SIGUPD(x13,a4)
+	
+
+// If we try to access the non existinfg CSR and the instruction is csrw 
+// 0xFFF, t1 so it will cause the illegal instruction and will update the mcause 
+// with the exception code 2, mepc will point to the address of the illegal 
+// instruction, mtval will contain the binary representation of illegal instruction
+	
+	LI (t1, 0xdead)
+	csrw 0xFFF, t1      // accessing non existing CSR will generate the illegal instruction exception
```
allenjbaum, line 54: no, this will not necessarily generate an illgal int exception !

riscv-test-suite/rv64i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,229 @@
...
+	LI (a4, 0x0006)   
+	RVTEST_SIGUPD(x13, a4)    // this signature will give information that test for trap handler delegation for ecall has successfullty completed
+	
+	
+// M mode 
+// This test checks that if an extension is turned off (e.g., FS bits [13 and 14] in mstatus 
+// are 00), trying to access it will cause an illegal instruction error. 
+
+
+ RVTEST_FP_DISABLE()  // this will set the FS bit in mstatsus to 00 means off
```
allenjbaum, line 140: I'm not sure this work if the F extension isn't implemented (and you're not filtering the test based on that).
How would you test this if it isn't? You can't depend on any particular extension being implemented, or, if it is, that you can disable it (except FP)

riscv-test-suite/rv64i_m/test/M_merged/M_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,229 @@
...
+    
+    LI (a4, 0x0007)   
+    RVTEST_SIGUPD(x13, a4)    // this signature will give information that test for FS extension off has successfullty completed
+   
+    
+    
+    /* When we try to jump to the misaligned address then instead of instruction misaligned adddress it gives the illegal instruction exception */
+    
+ /*   
+// This test will check that in machine mode, a load address misaligned exception will be raised
```
allenjbaum, line 167: haven't you done this in several other tests already?

riscv-test-suite/rv64i_m/test/U_merged/U_excp_ecall_illegal_misaligned.S
```diff
@@ -0,0 +1,262 @@
+/* // These tests are for the exceptions in U mode
```
allenjbaum, line 1: I wonder if thse M/U/S mode tests should just loop 3 times, with different GO_TO_LOWER mode each pass through the loop (and a BASEUPD each time through). You'd got the number of tests by 3x.

riscv-test-suite/env/test_macros.h
```diff
@@ -407,6 +407,18 @@ Mend_PMP:                                    ;\
...
  csrs mstatus, a0				;\
  csrwi fcsr, 0
+ 
+ 
+// this macro is to disable Floating point 
+// I have added this 
+
+#define RVTEST_FP_DISABLE()			;\
+ csrr a0, mstatus ;\
+ csrr a0, mtvec ;\
```
allenjbaum, line 451: why is this instruction here?

allenjbaum (changes requested):
There are persistent problems in all the tests that need fixing.

It is not the case that misaligned accesses will trap.
  They may or may not depending on whether HW misalignment is handled in HW,
  and whether it crosses specific address boundaries. 
  This may not actually affect the tests too much - as long as whatever happens is 
  duplicated by the reference model. But we do have an entire set of much more 
  comprehensive misaligned tests (except they don't check [v]Umode or [v]Smode 
  - I would instead ask that those tests be upgraded as I suggested, by looping 
  - through in each mode, changing modes at each loop with BASEUPD in between
  
It is no longer the case that accesses to undefined CSRs will trap
  Accesses to unimplemented CSRs are RSVD unless Zsstrict is implemented;
 then they will trap;  otherwise, they're unpredictable, so you can't use them in a test

It is no longer the case that FP ops will trap if mstatus.FS=0
  This is only the case if F-extension is supported. 
   If it isn't then those ops are undefined, RSVD unless Zsstrict is implemented;
  then they will trap;  otherwise, they're unpredictable, so you can't use them in a test

It is not the case that  undefined opcodes will trap (with few exceptions) .
  Execution of undefined opcode are RSVD unless Zsstrict is implemented;
 then they will trap;  otherwise, they're unpredictable, so you can't use them in a test
you can't use them in a test. An allzeros opcode is guaranteed. 

Tests use 2 noops after ops that trap, and none after ops that aren't supposed to
   Only one noop is needed, but they should all be replaced with a addI of 1<<op#.
  A series of ops will then set different bits depending on whether it didn't trap (or did) 
  and a single SIGUPD will enable SELF_TEST mode to catch cases where a trap didn't 
  happen the should have. 
  Note that this should be done even if you expect that the op will not trap

Tests delegate ECALL to Smode
  When this happens, special measures are required in advance to get back to Mmode: 
   Installing a special illegal op handler that resets mdeleg & then uninstalls itself
   Tests must execute an illegal op that goes to this handler before GOTO_MMODE

Constants are used instead of mnemonics
  Specifically for values of mdeleg masks, but there are probably others

Many tests cases appear to be duplicated or redundant
  e.g. there are many tests that write CSR 0xFFF to cause an illegal op trap. 
  There only needs to be one of them per test suite, not one per test case

There are a lot of tests that are identical except for the mode they are execute from
  The number of tests can be cut down by just executing the loop 3 times, 
  with different GOTO_LOWER modes each time (so, keep loop count in a reg, 
  and branch to different ones depending on loop count, 
  with a BASE_UPD macro between each to make sure offset is updated correctly

  SIGUPD after load should update the register that was (supposed to be) loaded
  Explicit use of ECALL must guarantee that x2 is nonzero (and that the previous op is not a beq .+4), else it will alias to GOTO_MMODE or SELFCHK_MISCOMPARE

Almost all the comments I added in

jamesbeyond:

PR #606, opened 2025-01-30
Build Sail with cmake

jordancarlin:
The RISC-V sail model has transitioned to using cmake as its build system. This updates the CI and README to use cmake to build the model. As part of the switch to cmake, the simulator was also renamed to `riscv_sim_rv64d` (the d has been added to indicate if was compiled with flen=64). This updates the riscof plugins to use the new name of the binary.


jordancarlin:
@UmerShahidengr All CI runs will start failing until this is merged because Sail will fail to compile with the existing CI script

UmerShahidengr:
@jordancarlin thank you for this contribution

UmerShahidengr (approved):
Looks good to go

PR #611, opened 2025-02-13
Remove docker image after use

Barabas5532:

## Description

Docker docs recommends this flag for short running foreground processes

https://docs.docker.com/reference/cli/docker/container/run/#rm

Try running `docker ps -a | grep incoresemi | wc -l` to check how many containers are left over. I have thousands on my machine because riscof creates a new container for each test case.

### Related Issues

https://gitlab.com/incoresemi/riscof-plugins/-/issues/20 I already reported it here. Not sure what is the relationship between this repo and incoresemi. Looks like this repo is more active.


Barabas5532:
Looks like this is vendored dependency and not intended for public use. I though you had adopted the riscof plugins since this is an official riscv organisation and there have been changes made to the code.

PR #623, opened 2025-03-24
Speedup riscv_ctg

trdthg:
## Description

When I run riscv_ctg, some test cases run very slowly, for example:

```
self.val_vars: ['rs1_val', 'rs2_val', 'rs2_hi_val', 'imm_val', 'ea_align']
req_val_comb: imm_val == 0
```

It takes a few minutes, some spend even tens of minutes

So what I'd like to do is, for all the vars we're not interested in, just take the first one


UmerShahidengr (approved):
Simple fix for avoiding unnecessary redundancy

PR #634, opened 2025-04-05
Virtual Memory SV39 Tests

Zain2050:

## Description

> Tests have been added to verify the functionality of the `SV39` virtual memory extension. `Core-V Wally` has been used as the `dut` for these tests.

### Related Issues

> NA

- [x] Ratified
- [ ] Unratified

### List Extensions

> SV39

### Reference Model Used

- [x] SAIL
- [ ] Spike

### Mandatory Checklist:

### Optional Checklist:

  - [x] Have you run these on any hard DUT model ? Please specify name and provide link if possible in the description
  - [x] If you have modified arch_test.h Please provide a detailed description of the changes in the Description section above.


allenjbaum:
You checked the box that said you updated arch_test.h and that the changes were listed above - but they weren't listed above

It is important that the trap handler is extremely well documented; it is pretty fragile.
You've added a function to the trap handler which (I think) forces an instruction access exception  (only) to return to the address in x4 if x3 is some (random?) value. 
This function needs  comments to explain that this is for testing (something) only, why it is needed at all, and the circumstances under which a test should enable it. I really can't tell why this is needed at all, though I might be able to guess.
SKIP_EPC doesn't reflect what it is doing (I originally though it was skipping the skipping behavior of EPC); please change to something more representative, e.g. RVTEST_SPCLCASE_FRC_IADDR_TRAPRTN. 
 The random value should be #defined, and named SPCL_IADDR_VAL to match other similar usecases

Zain2050:
Well, that isn't written by me. Sir Umer Shahid had made those changes to the trap handler.
@UmerShahidengr, would you explain your changes please?

allenjbaum:
That's OK, someone should be documenting it - preferably the author....

allenjbaum (changes requested):
on line 1422, don't use a constant, use a variable name, e.g. CAUSE_FETCH_ACCESS in this case (0xC might be wrong in this case; that is CAUSE_FETCH_PAGE_FAULT

riscv-test-suite/env/test_macros.h
```diff
@@ -232,17 +232,33 @@ Mend_PMP:                                    ;\
...
-        LA(_TR1, rvtest_slvl1_pg_tbl)                             ;\
-    .endif                                                        ;\
-    PTE_SETUP_COMMON(_PAR, _PR, _TR0, _TR1, _VAR, level)
+#define PTE_SETUP_SV39(_PAR, _PR, _TR0, _TR1, VA, level)                                     \
+    srli _PAR, _PAR, 12                 /* Shift PA right by 12 to get PPN */               ;\
+    slli _PAR, _PAR, 10                 /* Shift left by 10 to align PPN in PTE format */   ;\
+    or _PAR, _PAR, _PR                  /* Combine PPN with permissions */                  ;\
+    .if (level==2)                      /* Level 2 (1GB superpage) */                       ;\
+        LA(_TR1, rvtest_Sroot_pg_tbl)   /* Load root page table address */                  ;\
+        LI(_TR0, ((VA>>30)&0x1FF)<<3)   /* Calculate index for LEVEL2 (bits 38:30) */       ;\
```
allenjbaum, line 241: We should have a field extract macro e.g. 
   EXTRACT(_VAL_,_POS_,_BITWIDTH)  ((_VAL_>>_POS_)&((1<<_BITWIDTH_)-1))
and define RV64_LVL0_PPNPOS, RV64_LVL1_PPNPOS, RV64_LVL2_PPNPOS, RV64PPNWIDTH)
It could probably be used in a lot of places. I think we've defined XLEN_WIDTH somewhere.

riscv-test-suite/env/test_macros.h
```diff
@@ -232,17 +232,33 @@ Mend_PMP:                                    ;\
...
+        LI(_TR0, ((VA>>12)&0x1FF)<<3)   /* Calculate index for LEVEL0 (bits 20:12) */       ;\
+    .endif                                                                                  ;\
+    add _TR1, _TR1, _TR0                /* Add index to page table base */                  ;\
+    SREG _PAR, 0(_TR1)                  /* Store PTE at calculated address */               ;\
+
+// More Robust version of PTE_SETUP_SV39 to setup a PTE for a PA using VA
+// in a single line.
+// args: PA_LBL: Label of Physical Address, PERMS: permissions in hex
+// args: VA: Virtual Address in hex, level: Level to store at (0, 1, or 2)
+#define PTE_SETUP_RV39_New(PA_LBL, PERMS, VA, level)                                         \
```
allenjbaum, line 258: IS using a fixed reg# for PA_LBL and  and Perms a problem? The old macro had the register parameters expicit in the parameter definitions instead of explicit in the code .  Could it instead reuse TR0 or TR1, or have them supplied as explicit parameters to the new macro?

allenjbaum (changes requested):
in 1422, use defined variable CAUSE_FETCH_ACCESS instead of 0xC (which is incorrect, I think, because that is CAUSE_FETCH_PAGE_FAULT instead of 0x1

Zain2050:
I have changed 0xC to CAUSE_FETCH_PAGE_FAULT. 
Well, we actually need to check for **Fetch page fault** instead of **Fetch access fault**, as this fault occurs when we are trying to execute pages with X or V unset. I think the comment caused this confusion. I have improved the comment as well.

riscv-test-suite/env/test_macros.h
Zain2050: In the previous macro we actually have to store physical address and premissions in registers and then pass those registers. This updated macro sets up a PTE in one line. 
It has been changed to match the PTE_SETUP_SV32 macro which is already present. TR0 and TR1 can't be reused as they store page table address and offset. 
Same as the SV32 tests, we have made sure that a0 and a1 (fixed registers) are not being used anywhere else.
But if you want, I can change the macro to accept registers as parameters.

riscv-test-suite/env/test_macros.h
Zain2050: Yes, this is a good idea. I'll add it.

riscv-test-suite/env/test_macros.h
allenjbaum: Once _PAR is calculated (first 3 lines), then _PR is dead, and can be used instead of one of the temp registers (BUT only assuming that that _PR Xreg  will not be reused outside the macro!, as opposed to be reloaded) 
You should add comments as to which of the registers must be preserved by the macro somewhere if you make that change

riscv-test-suite/rv64i_m/vm_sv39/src/vm_canonical_S_mode.S
```diff
@@ -0,0 +1,251 @@
...
+
+#include "arch_test.h"
+
+RVTEST_ISA("RV64I_Zicsr")
+
+# Test code region
+.section .text.init
+.globl rvtest_entry_point
+rvtest_entry_point:
+RVMODEL_BOOT												// This test supports max 255 words for RVMODEL_BOOT
```
allenjbaum, line 32: good comments, but they're scolled off my screen for no good reason., which makes it harder to review. Could you move them in closer?

riscv-test-suite/env/test_macros.h
allenjbaum: The optimization  of changing code to reduce the number of registers just isn't important enough to waste time on. But: replacing a0 and a1 with passed in register numbers: I still want to see that done.

allenjbaum (approved):
This isn't perfect, but it works, and has been checked

riscv-test-suite/rv64i_m/vm_sv39/src/vm_canonical_S_mode.S
allenjbaum: We will do this in the next iteration

PR #637, opened 2025-04-16
Updates for latest RISC-V Sail Model and Initial Coverage Support in github workflow

MuhammadHammad001:
This PR adds support for:
1. Initial coverage in github workflow
2. Update the workflow to use latest RISC-V Sail Model


MuhammadHammad001:
Hi @UmerShahidengr , can you please review this PR? I would recommend not to squash these commits as most of the commits are not related and would help the future developer.

MuhammadHammad001:
Additionally, I will soon create a PR for the RISCOF to update the template as well.

jordancarlin:
It looks like the sail json config file you are storing in the repo is the same as the default one in the simulator. Since this configuration file is still new, I'm expecting it to continue evolving rapidly for the next little while. To keep this from getting out of date, it might make more sense to use the sail flag that prints out the default config to generate this at runtime (`riscv_sim_rv64d --print-default-config`). You can then apply the changes to pmp to that version. That way other new features are handled properly.

As things settle down, it will eventually have sense to have riscv-config/udb generate this json config to match the implementation, but we're not there yet.

MuhammadHammad001:
Good idea! I will implement this way.

MuhammadHammad001:
For now, I have updated the github workflow to run only I type tests as the current RISC_V Sail config uses MISALIGNED=True which fails the misaligned tests. In the next PR, I will add a flag in the riscof command something like `--user-sail-config` to use customized sail configs for different combinations and we can add all of them for different tests.

jordancarlin:
Looks like the generated JSONs from before still need to be removed.

> For now, I have updated the github workflow to run only I type tests as the current RISC_V Sail config uses MISALIGNED=True which fails the misaligned tests. In the next PR, I will add a flag in the riscof command something like `--user-sail-config` to use customized sail configs for different combinations and we can add all of them for different tests.

That was recently changed so that sail could pass all of the [riscv-tests](https://github.com/riscv-software-src/riscv-tests) with the default config. Would it be simpler to enable misaligned accesses in Spike for the default config so everything matches? Or pull this data from riscv-config and pass it to the models appropriately? Either way, definitely something that can be addressed in a future PR.

UmerShahidengr:
In my opinion, we should keep a control over the sail-config in CI, instead of relying on the latest copy of config file everything when CI will run. @MuhammadHammad001 keep a config file, and run the CI on that config script. We will keep updating config file with our needs.

MuhammadHammad001:
Okay sure, I will revert the commits.

UmerShahidengr:
@jordancarlin you have a good suggestion, but your method will not provide better control for us to keep checking the ACTs on different configs. We will switch over to the runtime config method once the sail model development settles down

MuhammadHammad001:
Closing this PR.

PR #641, opened 2025-04-28
Update CLI to show help instead of throwing an error

MuhammadHammad001:
This PR updates the RISC-V CTG to print options instead of throwing an error when running:

```
riscv_ctg
```
Previously, on running the above command:
image

Now:

image

### DIFF
--- a/riscv-ctg/riscv_ctg/main.py
+++ b/riscv-ctg/riscv_ctg/main.py
@@ -2,6 +2,7 @@
 """Console script for riscv_ctg."""
 
 import click,os
+import sys
 
 from riscv_ctg.log import logger
 from riscv_ctg.ctg import ctg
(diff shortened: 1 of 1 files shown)


MuhammadHammad001:
@UmerShahidengr @allenjbaum

allenjbaum:
My ignorance is whether this particular case is always because some command line option wasn't recognized, or there could be other reasons, in which case listing the CLI options isn't particularly useful

MuhammadHammad001:
To clarify: in this case, the crash only happens when the user does not provide any command-line arguments, leading to base_isa being None.

There are no other internal code paths where base_isa becomes None unexpectedly. Therefore, showing the help message is appropriate here, and the current behavior should improve the user experience without hiding any real errors.

allenjbaum (approved):
Pretty simple fix, solves a problem, removes some ambiguity (or adds reasonable defaults) when parameters are left out

PR #643, opened 2025-05-02
Made changes as per discussion in PR#427

UmerShahidengr:
Rectified the changes for CTG which are discussed in #427. Now the tests will be generated with regex RV32/RV64 in the RVTEST_CASE macro


allenjbaum (approved):
Fixes an existing problem; pretty simple fix

PR #644, opened 2025-05-02
Modified B-extension Tests

UmerShahidengr:
This is a replica of PR#427. Made the desired changes and kept the regex(32) and regex(64) checks intact


allenjbaum (approved):
This fixes an issue  where CTG didn't properly condition the tests case.
CTG has been fixed so that tests are executed if B mode is implemented, or if one of its submodes have been implemented. This adds the test cases that were missing before the CTG change

PR #645, opened 2025-05-02
Minor change in GOTO_LOWER_MODE macro

UmerShahidengr:
RVTEST_GOTO_LOWER_MODE macro to avoid using sp register as sp might already be in use to point the program stack (if any) in the test file. RVTEST_GOTO_LOWER_MODE was updating sp register.


allenjbaum (approved):
Simple substitution of Xregs to avoid overwriting sp

PR #648, opened 2025-05-08
Fix RVTEST_CASE macro for rv32i bseti-01 test

jordancarlin:
#644 added a second `RVTEST_CASE` macro to the rv32i `bseti-01` test with the number 0, which breaks riscof. This switches it to be number 1, like all of the other tests.


allenjbaum (approved):
This fixes a duplicate testscase name - simple fix.

jordancarlin:
Thanks for the quick merge.

PR #650, opened 2025-05-12
Update CI to use Latest Sail Config at Run time

MuhammadHammad001:
Since my last PR for the CI update, the Sail config has been continuously updated, causing the workflow to fail. To avoid continuous maintenance of the `sail_config.json`, we have decided to use the latest version at runtime.


MuhammadHammad001:
@UmerShahidengr

UmerShahidengr (approved):
Sail plugin has been updated with the latest changes in sail-model configurations.

Issue #649, opened 2025-05-09
ISA string issues with new Zdinx tests


jordancarlin:
The new Zdinx fma tests added by #642 have incorrect `RVTEST_ISA` and `RVTEST_CASE` strings, which causes riscof to break.

Some of the tests (like `fmadd.d_b1-01.S`) have the correct `RVTEST_ISA` macro (`"RV32I_Zfinx_Zdinx"`), but the incorrect `RVTEST_CASE` macro (`"//check ISA:=regex(.*I.*D.*)`). This should be checking for Zdinx, not D.

Others have both macros wrong. For example, `fmadd.d_b8-01.S` has `RVTEST_ISA("RV32IFD_Zicsr,RV64IFD_Zicsr")` and `RVTEST_CASE(0,"//check ISA:=regex(.*I.*D.*);def TEST_CASE_1=True;",fmadd.d_b8)`. Both of these should have Zdinx, not F/D.

These issues result in repeated warnings when compiling and cause riscof to fail:
```
/home/jcarlin/cvw-new/addins/riscv-arch-test/riscv-test-suite/rv32i_m/D/src/fmadd.d_b15/fmadd.d_b15-007.S: Assembler messages:
/home/jcarlin/cvw-new/addins/riscv-arch-test/riscv-test-suite/rv32i_m/D/src/fmadd.d_b15/fmadd.d_b15-007.S:5432: Warning: bignum truncated to 8 bytes
/home/jcarlin/cvw-new/addins/riscv-arch-test/riscv-test-suite/rv32i_m/D/src/fmadd.d_b15/fmadd.d_b15-007.S:5433: Warning: bignum truncated to 8 bytes
/home/jcarlin/cvw-new/addins/riscv-arch-test/riscv-test-suite/rv32i_m/D/src/fmadd.d_b15/fmadd.d_b15-007.S:5434: Warning: bignum truncated to 8 bytes
/home/jcarlin/cvw-new/addins/riscv-arch-test/riscv-test-suite/rv32i_m/D/src/fmadd.d_b15/fmadd.d_b15-007.S:5435: Warning: bignum truncated to 8 bytes
/home/jcarlin/cvw-new/addins/riscv-arch-test/riscv-test-suite/rv32i_m/D/src/fmadd.d_b15/fmadd.d_b15-007.S:5436: Warning: bignum truncated to 8 bytes
[...]

PR #651, opened 2025-05-13
[ACT]fix isa string bugs with Zdinx tests

Pagerd:
## Description

fix Zdinx tests isa string bugs 

### Related Issues
[#649](https://github.com/riscv-non-isa/riscv-arch-test/issues/649)

- [X] Ratified
- [ ] Unratified

### List Extensions

- [X] Zdinx


jordancarlin:
If the RVTEST_ISA macro specifically calls out rv32 only, then doesn't the RVTEST_CASE regex need to check for 32 as well? Right now I think an rv64izfinx_zdinx implementation would match the regex but fail the run the compiled tests since they specify rv32 in RVTEST_ISA.

allenjbaum:
Can someone explain why this email says 300 files changed, but Github says
580 files changed?
(and I can't actually see all the files on github;  it gives up at
fmadd.d_b15/fmadd.d_b15-018.S
<https://github.com/riscv-non-isa/riscv-arch-test/pull/651/files#diff-b63b4e4dee353fa8ace6ba99ceeec4d889c2155281a96982b011490d781bc870>
and gives the message " Oops, something went wrong. ")

Pagerd:
> If the RVTEST_ISA macro specifically calls out rv32 only, then doesn't the RVTEST_CASE regex need to check for 32 as well? Right now I think an rv64izfinx_zdinx implementation would match the regex but fail the run the compiled tests since they specify rv32 in RVTEST_ISA.

Yes you are right, restrictions should be added now

jordancarlin:
Any update on this? #649 is preventing us from updating to a newer version of riscv-arch-test.

Pagerd:
it should be fix now?

UmerShahidengr (approved):
Looks good to go. Thank you for the work

PR #652, opened 2025-05-26
Bump Sail version to 0.19 in README

nadime15:


README.md
```diff
@@ -212,7 +212,7 @@ First install the [Sail Compiler](https://github.com/rems-project/sail/). It is
 
 ```bash
 $ sudo apt-get install libgmp-dev pkg-config zlib1g-dev curl
-$ curl --location https://github.com/rems-project/sail/releases/download/0.18-linux-binary/sail.tar.gz | [sudo] tar xvz --directory=/path/to/install --strip-components=1
+$ curl --location https://github.com/rems-project/sail/releases/download/0.19-linux-binary/sail.tar.gz | [sudo] tar xvz --directory=/path/to/install --strip-components=1
```
jordancarlin, line 215: What about doing `https://github.com/rems-project/sail/releases/latest/download/sail.tar.gz` so that we don't need to keep updating this?

nadime15: Yeah even better, there are parts in the code that otherwise need to be replaced every single time.

jordancarlin:
Can be closed now that #663 was merged instead.

PR #658, opened 2025-06-05
Fix missing val_comb values after expand_cgf

trdthg:
## Description

@UmerShahidengr If I understand correctly, we are missing all the val_comb values when i < 100, `cgf[labels][label].yaml_add_eol_comment(comment, key=cp)` only added the comment.

When I use riscv-ctg for local generation, I want to generate a total of about 128 + x(<10) data, but the actual result is only 30+, which is too few, so I think it's a bug that was introduced by accident.


jamesbeyond:
ACT repository have been updated `act4` as the default branch, ISAC is no longer used in act4, So we have to close this PR

PR #660, opened 2025-06-09
Add ruff format check in pre-commit hook and CI

trdthg:
## Description

I believe that we will fully format the code anyway, as it is necessary for the maintainability of the project.

This PR adds a CI to check whether the PR is correctly formatted.
It also adds a pre-commit hook to help users perform format check locally

I created a basic formatted configuration and simply modified the default line-length 和 quote-style

The only thing to note is that the ruff format will break our manual alignment of the dictionary，ruff format also doesn't support many customization options, but this seems to be a common issue with formatting tools, and black (a more mature formatter) has the same problem https:github.com/psf/black/issues/4281

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -1,0 +1,16 @@
+name: Lint and format
+
+on:
+  push:
+    branches: [ '**' ]
+  workflow_dispatch:
+
+jobs:
+  format:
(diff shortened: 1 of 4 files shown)


jamesbeyond:
ACT repository have been updated `act4` as the default branch, ruff had completed in act4 branch, So I have to close this PR

Issue #647, opened 2025-05-07
Example config requests excessive memory from spike


Wren6991:
I'm migrating from the old makefile flow to the new python flow. Following the instructions in the readme, I tried testing spike against itself using the config in this repository (on the `main` branch). This line:

https://github.com/riscv-non-isa/riscv-arch-test/blob/b0179e138f18b85a8c79ea315c5cd256b65bab64/riscof-plugins/rv32/spike_simple/riscof_spike_simple.py#L161

Seems to request `1ull << 39` megabytes of memory from spike, per this output from `spike --help`:

```
Spike RISC-V ISA Simulator 1.1.1-dev

usage: spike [host options] <target program> [target options]
Host Options:
  -p<n>                 Simulate <n> processors [default 1]
  -m<n>                 Provide <n> MiB of target memory [default 2048]
  -m<a:m,b:n,...>       Provide memory regions of size m and n bytes
```

This causes spike to segfault. I'm not sure what the intent of that flag was, but things seem to run ok if it's removed.

allenjbaum:
...-m8796093022208 ...  - 8 Terabytes?
Separately: I object to using decimal values for memory size as opposed to hex values

Wren6991:
Thanks, looks like this was a mistake and is now fixed by #662.

PR #662, opened 2025-06-13
[ACT]Fix spike Example config requests excessive memory

Pagerd:
## Description

Now the spike sample plugin will require up to 8TB of memory. Deleting this line will not affect spike operation, and the previous plugin did not contain this line. I think it should be deleted.

### Related Issues

[#647](https://github.com/riscv-non-isa/riscv-arch-test/issues/647)


UmerShahidengr (approved):
Thank you @Pagerd. It was left there by mistake.

PR #663, opened 2025-06-13
Update to use latest Sail compiler release (0.19.1)

jordancarlin:
Always downloads the latest Sail compiler release so this doesn't need to be continually updated. Currently that is Sail 0.19.1.

This makes it possible to build the Sail model again and fixes CI.


UmerShahidengr (approved):
Simple fix. Thank you.

PR #666, opened 2025-06-17
Fix ISAC coverage drop in A, C, I, and M

sudo-apt-abdullah:
Modified the CGF file for ori to exclude cases where rd == x0. This prevents a coverage drop due to a special case: when rd == x0, the internaldecoder.py interprets it as a prefetch instead of ori. Regenerated the ori tests with the updated CGF using ctg.

For RV32M, coverage was missing cases related to op_comb. These were added after regenerating the tests with ctg.

Fixed a bug in riscv_isac/coverage.py that was causing coverage drops for A, C, I, and M instructions. Now, all are at 100% and here you can see the initial and final coverage before and after the changes made: https://docs.google.com/spreadsheets/d/1V1TRYfl7aYsMEXnGEAn7FwO558mKAwf9a3cCnG6dBow/edit?gid=0#gid=0


MuhammadHammad001:
Looks good to me...

UmerShahidengr (approved):
Looks good to me

PR #667, opened 2025-06-23
Fix ISAC coverage drop for PMP

sudo-apt-abdullah:
### CSR Regex Update

Modified instr_pattern_c_sail_csr_reg_val to support the new Sail log format, which now includes the CSR address:

Old format: CSR mstatus <- 0x00000000
New format: CSR mstatus (0x300) <- 0x0000000000000000

The updated regex captures the CSR address (e.g., (0x300)) into a new named group csr_address.
Updated csr_commit indexing logic in InstructionObject.py to correctly extract the CSR value based on the new format.

### CGF Updates for PMP

Replaced deprecated usage of "old" with "old_csr_val" in PMP-related CGF files to align with the latest convention.
Updated corresponding coverpoints to reflect this change.

### Header Enhancements

Added pmp_helper_coverpoints in the header file to support their use in PMP CGF files.


MuhammadHammad001:
Nice work @sudo-apt-abdullah, looks good to me.

UmerShahidengr (approved):
LGTM

PR #668, opened 2025-06-25
Updated the Reference Plugins to Include FLEN Flag.

sudo-apt-abdullah:
While running the riscof command for coverage, the Sail plugin generates a Makefile, but the isac command within that Makefile omits the -f64 flag. As a result, isac computes coverage for the F extension using FLEN = 32, whereas it should be 64 if the D extension is present. To address this, I’ve updated the reference plugin to include the -f64 flag in the isac command when D is supported. This ensures correct coverage computation for the extension using FLEN = 64.

Now, plugin will set FLEN to 64 if D is supported, else 32 and hence sail will not throw the warning "FLEN is set to 32. Commit values in the log will be terminated to 32 bits irrespective of their original size."


UmerShahidengr (approved):
LGTM

PR #669, opened 2025-06-26
Update test_macros.h

Marina-Bellido:
When trying to use this macro I run into the following compiler error:  
Error: expected comma after "corr"

Added commas after all of the corr variables in the function and it worked.


allenjbaum (approved):
That. the correct fix

PR #670, opened 2025-06-27
Fixed Coverage Holes for RV32F

sudo-apt-abdullah:
The coverage was initially dropped due to the incorrect operand values. Each test has an associated dataset at its end, but the dataset values were incorrect. 
op1val: 0x7f800001
op2val: 0xff800000
But the dataset incorrectly records:
NAN_BOXED(0x7f800001139095041,32,FLEN)
NAN_BOXED(0xff800000286578688,32,FLEN)
To correct this and achieve full coverage, I regenerated the tests with the correct boxing logic. Here's the difference in between the initial and final coverage can be seen: https://docs.google.com/spreadsheets/d/1V1TRYfl7aYsMEXnGEAn7FwO558mKAwf9a3cCnG6dBow/edit?gid=0#gid=0


allenjbaum:
I suspect it isn't possible to regenerate test with the same seeds and
expect to only see changes in boxing behavior.
I also don't understand what is going on, e.g. are you saying that the
tests could not meet 100% coverage,
or that the hit 100% cover, but that coverage had a lot of holes in it?

sudo-apt-abdullah:
Yes, I worked to bring the tests up to 100% coverage, which was initially very low because the val_comb coverpoints weren’t being hit due to incorrect dataset values.

At first, I regenerated the tests with a corrected dataset. However, in the latest commit, I simply fixed the dataset using a Python script, since the test cases themselves were already correct.

allenjbaum:
The reason I asked is that all the tests were completely replaced, and if
it were just the coverpoints being updated, then I wouldn't have expected
any changes to the .S files,

Can you make it a bit more clear why the tests themselves were regenerated,
since the test cases were already correct?

sudo-apt-abdullah:
Thank you for pointing that out, and I apologize for the confusion caused by GitHub showing the entire file as changed.

To clarify, the test cases themselves were not regenerated — they remain exactly the same. The only modification was at the end of each file, where the dataset values are defined. These were updated using a Python script to ensure that the val_comb coverpoints are now correctly hit. The original test logic was valid; the issue was solely with incorrect operand values in the dataset section, which led to poor coverage initially.

Unfortunately, due to line-ending normalization or whitespace changes (possibly introduced by tooling or script formatting), GitHub's diff view shows the entire file as changed, even though the actual update is limited to the dataset portion.

You can confirm this by comparing the beginning and middle sections of each file — only the dataset at the end has been updated.

UmerShahidengr:
I have checked the test files from @sudo-apt-abdullah system directly, and it looks fine to me. A python script is doing the same change in all tests, no other change was observed in any test. So I am merging this one.

UmerShahidengr (approved):
LGTM

PR #678, opened 2025-07-09
Update arch_test.h to support IAF testing

Marina-Bellido:
Adding logic into the  trap handler to support IAF testing. Approved solution in today's cvw-arch-verif meeting. Umer will add to this flag to generalize it and make it also applicable for PMP tests. It was also discussed in the meeting that using ra is fine since we need more registers than the 6 T defined at the top of the trap handler, plus ra is not being modified only read.


allenjbaum:
sounds good to me.

UmerShahidengr:
Hey @Marina-Bellido! Just a heads-up — the latest sv39 PR (#634) already includes the fix for this issue, covering both the Illegal Access Fault and Access Page Fault. So, this PR ends up being redundant. I'm closing it now, but huge thanks for jumping in and working on this — really appreciate your contribution and effort! Looking forward to more from you!

Marina-Bellido:
Hi @UmerShahidengr just saw the message. Thank you so much :)

Issue #589, opened 2024-12-31
NAN_BOXED data in F/D-ext tests requires coversion from decimal to hex


UmerShahidengr:
IEEE FP is a binary format, not a decimal format. All D extension test cases are tested using decimal data, including the modified fcvt.d.s test case that I just merged, #581 developer has used the same method to modify it as used previously by the other developers i.e, use decimal data. There is a problem for anyone who reviews or uses these tests, so all tests are needed to be updated to follow the hex format.

jordancarlin:
The tests have been regenerated for ACT4 and how use hex values.

PR #680, opened 2025-07-18
[CTG]Convert the generated test data from decimal to hexadecimal

Pagerd:
## Description

Changing the NAN_BOXED test data to hexadecimal will help with review, since IEEE FP is a binary format, not a decimal format.

### Related Issues

[#589](https://github.com/riscv-non-isa/riscv-arch-test/issues/589)


allenjbaum (commented):
I'm a little puzzled; I heartily approve of using hex instead of decimal. 
But what I'm not seeing here is any use of decimal; 
Shouldn't this also remove any use of  print(dval)?

Pagerd:
> I'm a little puzzled; I heartily approve of using hex instead of decimal. But what I'm not seeing here is any use of decimal; Shouldn't this also remove any use of print(dval)?

Sorry I got the wrong repository, the current fix should be correct

allenjbaum (approved):
Hex is much more appropriate than decimal.

PR #687, opened 2025-07-30
Modified trace argument in Sail Plugins

Zain2050:
## Description

Tests execution on reference started failing due to a recent change in Sail (https://github.com/riscv/sail-riscv/pull/1129).
`-v` is no longer used for trace. It is separated into different trace arguments & `--trace-all` enables all of them. 

CI will fail without this fix.


UmerShahidengr (approved):
Thank you for the fix

Issue #686, opened 2025-07-26
Why are vclmul and vclmulh checking for Zvkb and not Zvbc


nibrunieAtSi5:
https://github.com/riscv-non-isa/riscv-arch-test/blob/bf20861f03d8661b7d252b84c7ca09bf19cf6477/coverage/zvk/vclmul.cgf#L5

(Note that I may not correctly grasp how riscv-arch-test check for ISA support).
I am surprised  that the `vclmul.v[xv]` and `vclmulh.v[vx]` are checking for **Zvkb** in the ISA while they  are not part of that extension but are part of **Zvbc**:
```
vclmul.vv:
    config:
      - check ISA:=(.*I.*V.*Zvkb)
```

this is the case in both the config files and the assembly tests:
https://github.com/riscv-non-isa/riscv-arch-test/blob/bf20861f03d8661b7d252b84c7ca09bf19cf6477/riscv-test-suite/rv32i_m/Zvk/src/vclmul.vv-01.S#L14

(Spec source: https://github.com/riscv/riscv-isa-manual/blob/853e233ee35295a3e6784cdc791585d7d01596e1/src/vector-crypto.adoc#zvbc---vector-carryless-multiplication)

Pagerd:
this issue has been fixed

UmerShahidengr:
PR #688 fixed this issue, so closing this one.

PR #688, opened 2025-07-30
Fix vclmul &vclmulh ISA bug

Pagerd:
## Description

Fixed a small problem with isa. `vclmul.v[xv]` and `vclmulh.v[vx]` are from Zvbc, but the test case and cgf mistakenly set isa to Zvkb.

### Related Issues

[#686](https://github.com/riscv-non-isa/riscv-arch-test/issues/686)


allenjbaum (approved):
Pretty simple fix; LGTM

PR #689, opened 2025-08-01
PMP{M, S, U, F, Zaamo, Zalrsc, Zca, Zcb, Zcf, Zcd, Zicbo} tests by UET Team

sudo-apt-abdullah:
Test Plan for PMP: https://docs.google.com/spreadsheets/d/1jvJCSf9vdZ23PEaaYETLxfwZoyq_PzpyJzIB1OhdhcQ/edit?gid=2072825600#gid=2072825600


allenjbaum:
Will coverage files be coming also?

sudo-apt-abdullah:
> Will coverage files be coming also?

System Verilog Coverpoints for the testplan are present in cvw-arch-verif directory. These tests are giving 100% coverage and verified thorugh Sail and Wally logs.

> https://github.com/openhwgroup/cvw-arch-verif/blob/main/fcov/priv/PMPS_coverage.svh
> https://github.com/openhwgroup/cvw-arch-verif/blob/main/fcov/priv/PMPU_coverage.svh

allenjbaum:
Yes, but those coverage files aren't in the "files changed" list - so where
is that cvw-arch-verif directory?
a separate branch? It should be in the same branch as the tests

sudo-apt-abdullah:
> so where is that cvw-arch-verif directory?

I've provided the links to the coverage files in cvw-arch-verif - a separate repository for verification of Core V Wally.

> https://github.com/openhwgroup/cvw-arch-verif/blob/main/fcov/priv/PMPS_coverage.svh
> https://github.com/openhwgroup/cvw-arch-verif/blob/main/fcov/priv/PMPU_coverage.svh

allenjbaum:
I don't think we can merge this unless we have a copy of that in the
arch-test repo
(and even when we do... we don't have the means to measure coverage. This
is a problem.

allenjbaum (commented):
I see a lot orf inconsequential changes here; mostly formatting and moving around lines, and a bit of renaming.
The substantial change is to the TEST_CASE macro, where you split it into two cases: without and with Smode, where TEST_CASE_2 is defined when rvtest_strap_routine is defined..
But, but then, there is no TEST_CASE_2 in any of the tests.
So,  why split it (i.e. why define TESTS_CASE_2 at all?)

sudo-apt-abdullah:
> The substantial change is to the TEST_CASE macro, where you split it into two cases: without and with Smode, where TEST_CASE_2 is defined when rvtest_strap_routine is defined..
> But, but then, there is no TEST_CASE_2 in any of the tests.
> So, why split it (i.e. why define TESTS_CASE_2 at all?

Thanks for the review! We didn’t add a new TEST_CASE_2, both variants still set TEST_CASE_1=True and run the same test body. The split is only to handle the difference in trap setup: If S is found, Case 1 runs (defines both mtrap and strap routines), and if S is not found, Case 2 runs (defines only mtrap). This way the same test works for both S-mode and non-S-mode targets.

 Let me know if I'm overlooking something!

allenjbaum:
I understand the intent, but my search for the tag TEST_CASE_2 showed no
instances - so there are no tests using that tag.
If that is the case, then why define it? Or, are some tests mislabelled
with TEST_CAE_1 instead of TEST_CASE_2?
Or, was I searching the wrong folder? (that doesn't make sense, as it is
only defined in that folder)

sudo-apt-abdullah:
I think you are confusing **RVTEST_CASE(2,** with **TEST_CASE_2**.
TEST_CASE_2 is not defined anywhere and I'm not using it. The only purpose of RVTEST_CASE(2 is to enable strap_routine if S mode is supported. It still sets TEST_CASE_1=True.

allenjbaum:
OK, I understand what you're doing doing now. This all looks good now.

sudo-apt-abdullah:
Sir @UmerShahidengr, PR can be merged after a final review. @hamza-1821 and I removed the tests and putting them in draft which were causing the CI to fail either because of sail bug (Issue [#1239](https://github.com/riscv/sail-riscv/issues/1239)) or configurations of sail and spike.

UmerShahidengr (approved):
Looks good to go. Merging this one

PR #690, opened 2025-08-01
Adding new macros: RVTEST_SIGWRITE(_F)

Marina-Bellido:
This macro Stores register into signature region and increment the signature pointer. Currently, RVTEST_SIGUPD does not properly handle code that jumps over macros due to garbling the offset, so we created this new macros to handle the privilege tests. 

Notes:
 /* Do not mix RVTEST_SIGWRITE and RVTEST_SIGUPD in the same program */
 /* RVTEST_SIGWRITE(basereg, sigreg) stores sigreg at 0(basereg) and increments basereg by regwidth	 */

### DIFF
--- a/riscv-test-suite/env/test_macros.h
+++ b/riscv-test-suite/env/test_macros.h
@@ -593,6 +593,25 @@ Mend_PMP:                                    ;\
       .set offset,offset+(REGWIDTH)
 
 
+ /* Stores register into signature region and increment the signature pointer */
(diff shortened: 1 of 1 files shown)


allenjbaum:
Could you give an example of an existing test where that garbling happens?
I am guessing that SIGUPD followed by BASE_UPD probably does what you need,
but I haven't seen tests that need that, which is why I am asking

jamesbeyond (approved):
LGTM

PR #691, opened 2025-08-01
Solving rtn2mmode NOTE: (Ecall must NOT delegate when T2=0 or this fails).

Marina-Bellido:
Spcl handler created for returning from GOTO_SMODE                                                    
(Executed in S-mode. Enter w/ T1=ptr to Mregsave, T2=0)

Created to solve rtn2mmode NOTE: (Ecall must NOT delegate when T2=0 or this fails). rtn2smode solve the delegation failure and also uses CSRs that can be accessed in supervisor                                      

 NOTE: 
- Code needs to be updated when rvtest_vtrap_routine= True, currently commented out     
- There is a relocation calculation that is not currently working (commented out)

### DIFF
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -319,6 +319,7 @@
 #define MPP_LSB   11    //bit pos of LSB of the mstatus.MPP  field
 #define MPRV_LSB  17    //bit pos of LSB of the mstatus.MPRV field
 #define MPV_LSB    7    // bit pos of prev vmod mstatush.MPV in either mstatush or mstatus upper
+#define SPV_LSB    7    // bit pos of prev vmod hstatus.SPV 
 #define MPP_SMODE (1<<MPP_LSB)
 #define MPP_MMODE (3<<MPP_LSB)
 //define sizes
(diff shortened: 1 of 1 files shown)


allenjbaum:
The updated risc_arch has a different plan; tests that delegate ECALL must
use a different trapping opcode, and must set a1 to 0 before they use it
(a1 was already used for ECALL; if there is reason to use t0, that could
work also)
e.g.
    #define ALT_GOTO_M_CAUSE CAUSE_ILLEGAL_INSTRUCTION
    #define ALT_GOTO_M_OP    .word 0
or
   #define ALT_GOTO_M_CAUSE CAUSE_EBREAK
   #define ALT_GOTO_M_OP    EBREAK

In the Mmode handler, we check ECALL cause and a1, and stay in Mode.
But, the Mmode  handler also checks the ALT_GOTO_M_OP cause, and a1, and
will stay in Mmode for that case also.
This avoids trampolining through Smode (or HSmode or VSmode), and puts the
burden on the test writer to use the alternate op.
(There may also another way to do this; there is a table of special
handlers built in, and that table rntry could be modified so that any
cause  could be dealt with as a GOTO_MMODE.
But in any case, the test has to be cognizant that ECALL is delegated and
not use ECALL.

jamesbeyond:
@allenjbaum is working on updating the trap handler macro, so will closing this PR

PR #694, opened 2025-08-07
fix: remove problematic test cases that incorrectly check for C extension

zouguangxian:

## Description

Remove RVTEST_CASE(0,...) entries that check for C extension compatibility from misalignment test files. These tests contain only standard 32-bit instructions and no compressed instructions, making the C extension compatibility checks unnecessary and misleading.

- [x] Ratified
- [ ] Unratified

### List Extensions

- RV32I_Zicsr

### Reference Model Used

- [ ] SAIL
- [x] Spike

### Mandatory Checklist:

### Optional Checklist:


allenjbaum:
I'm not sure that we want to remove C extension tests.
What they show is that misaligned tests don't trap on halfword
aligned addresses

zouguangxian:
If the DUT has `ISA: RV32IMAC`, 
1, for these files,  the first check `RVTEST_CASE(0,"//check ISA:=regex(.*E.*C.*) ;def RVTEST_E = True;def TEST_CASE_1=True;"` will be true, thus macros is not [].
2, and when macro is not [], it will call prod_isa
```
if not macros == []:
            try:
                isa = prod_isa(ispec['ISA'],db[file]['isa'])
            except TestSelectError as e:
                logger.error("Error in test: "+str(file)+"\n"+str(e))
                continue

```
However, In prod_isa, it will throw `Test Selected without the relevant extensions being available on DUT.` due to `RV32E_Zicsr` is not a subset of `RV32IMAC`. 

Should it check Zicsr in `RVTEST_CASE(0` instead of removing C extension tests?

allenjbaum:
Then are we missing C extension tests for "misaligned" jump/branches and
load/store?
WE have tests without C that traps on misaligned; Tests with the C
extension should test whether misaligned branches don't trap
 - even for tests that don't use C.Bxx - the uncompressed ones can still
branch to an unaligned address and then an unaligned branch target should
be able to branch back to an aligned target,
al without executing any C-extension ops.
and I suspect there should be load/store test could do the same (e.g test
that LH and LH work if they don't cross a word boundary) without using a
C-ext op.

We do want to test that C-extension load/store/bxx ops also dont trap for
halfword aligned ops.

jamesbeyond:
ACT repository have been updated `act4` as the default branch, Those tests are no longer exist in act4. So we have to close this PR

Issue #482, opened 2024-07-25
Missing coverage for compressed illegal instruction 


Bill94l:
Hi, 
For some compressed instructions, the non-zero check for the RS/RD registers and the immediate value is missing. Here is the list of instructions with the associated missing checks:
- **C.ADDI4SPN** : **imm != 0**
- **C.ADDIW** : **rd != 0**
- **C.LII** : **imm != 0**
- **C.LWSP** : **rd != 0**
- **C.LDSP** : **rd != 0**
- **C.JR** : **rs != 0**

UmerShahidengr:
@Bill94l can you please elaborate why do you need these coverpoints? All these coverpoints seem redundant. If you look at any test of C.LWSP, rd is producing non-zero results without having such coverpoints.

jordancarlin:
These are all either hints or reserved. Reserved instructions can only be tested in the Ssstrict test. The other instructions are hints and are tested in ACT4.

PR #695, opened 2025-08-15
[ACT] Fix Missing coverage for compressed illegal instruction

Pagerd:
## Description

WIP, it will fix missing coverage for compressed illegal instruction

### Related Issues

[#482](https://github.com/riscv-non-isa/riscv-arch-test/issues/482)


UmerShahidengr:
Some explanation is needed by @Bill94l on why these coverpoints are needed, I feel like this work is redundant.

jamesbeyond:
ACT repository have been updated `act4` as the default branch, these test have been replaced in act4, So those fixes are no longer required

PR #696, opened 2025-09-01
Reduced time for Reference Makefile generation

Zain2050:
"Running tests on Reference" took a greater amount of time as compared to "Running tests on DUT". First I though that Sail is relatively slower than Spike. However, that's not the case. The Makefile for Spike was being generated instantly, while the Makefile for Sail was taking a lot of time (~40s for RV64I). 
The issue is in the Sail plugin. It was loading sail_isa.yaml, setting configurations and writing those into sail_config.json in **every iteration** (where it is iterating through every test in the directory). 
This was a problem in the code as neither the isa.yaml would change nor the configurations. Therefore, I have changed the code to generate the sail_config.json file **once** and then enter the loop for generating Makefile.

**This has greatly reduced the runtime for Reference**. In case of RV64I
**Previous:**
```
  INFO | Running Tests on Reference Model.
  Time elapsed in generating Sail Makefile:  0:00:39.930905
```
**Now:**
```
  INFO | Running Tests on Reference Model.
  Time elapsed in generating Sail Makefile:  0:00:00.826270
```

After that the plugin executes the Makefile. This change **saves us about 40 seconds for RV64I.**

Similarly, **about 32 seconds are being saved for RV32I.**

### DIFF
--- a/riscof-plugins/rv32/sail_cSim/riscof_sail_cSim.py
+++ b/riscof-plugins/rv32/sail_cSim/riscof_sail_cSim.py
@@ -88,6 +88,50 @@ class sail_cSim(pluginTemplate):
             os.remove(self.work_dir+ "/Makefile." + self.name[:-1])
         make = utils.makeUtil(makefilePath=os.path.join(self.work_dir, "Makefile." + self.name[:-1]))
         make.makeCommand = self.make + ' -j' + self.num_jobs
+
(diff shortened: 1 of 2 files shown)


Zain2050:
CI usually takes about 11 minutes, but got completed in 6 minutes for this PR.

UmerShahidengr (approved):
Nice work. Thank you @Zain2050

PR #698, opened 2025-09-26
(CI Fix) Disabled Svrsw60t59b extension from RV32 Sail plugin

### LINKED ISSUE #697: Error in make when test riscof example
ERROR | make: *** [/home/cat/riscv-arch-test/riscof-plugins/rv32/riscof_work/Makefile.Reference-sail_c_simulator:5: TARGET0] Error 1

which makes no Reference-sail_c_simulator.signature file.

Zain2050:
Closes #697. 

The cause of failure was: 
**Svrsw60t59b** is enabled but Sv39 is disabled, supporting Svrsw60t59b requires supporting Sv39.
Therefore, I have disabled it from the RV32 plugin file.
["mtval_has_illegal_instruction_bits"] has a default value of True now, therefore that line in RV64 is no more required.

### DIFF
--- a/riscof-plugins/rv32/sail_cSim/riscof_sail_cSim.py
+++ b/riscof-plugins/rv32/sail_cSim/riscof_sail_cSim.py
@@ -124,6 +124,7 @@ class sail_cSim(pluginTemplate):
         # Enabling extensions that are disabled by default
         sail_config["extensions"]["Sv32"]["supported"] = True
         sail_config["extensions"]["Zcf"]["supported"] = True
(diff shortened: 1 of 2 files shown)


jordancarlin (approved):

allenjbaum:
Maybe this is moot, but this dependency is the sort of configuration rule
that should be in riscv-config, and will be in UDB (eventually, if it isn't
already.)
It should be marked as a normative rule in the .adoc spec.

UmerShahidengr (approved):
Thank you @Zain2050

PR #701, opened 2025-10-07
fix the fsd-align test

sudo-apt-abdullah:
The `fsd` attempts to make a misaligned store. It is an unpriv test, so there is no trap handler, and it hangs on the first access. The issue happened because of an unintended change I made while fixing the incorrect dataset at the end of the F and D extension test files. It’s fixed now.


davidharrishmc:
Is there another PR fixing the dataset so this doesn’t recur?

sudo-apt-abdullah:
> Is there another PR fixing the dataset so this doesn’t recur?
I don't think so.

davidharrishmc:
That’s important.  Otherwise we will have this bug again next time somebody runs the test generator.

davidharrishmc:
More generally, it looks like this file was last autogenerated in 2022, and is now being changed manually.  That's a touchy practice to hand-edit generated files.  @UmerShahidengr is this necessary, or is it feasible to fix the dataset and autogenerate again?

sudo-apt-abdullah:
> That’s important. Otherwise we will have this bug again next time somebody runs the test generator.

I checked, and only fsd-align-01.S was getting stuck due to bad change.

> More generally, it looks like this file was last autogenerated in 2022, and is now being changed manually. That's a touchy practice to hand-edit generated files. @UmerShahidengr is this necessary, or is it feasible to fix the dataset and autogenerate again?

Basically, these tests were modified by python script. I wrote a script that took each test file, extracted the actual operand values used in each test case, and replaced the incorrect dataset entries with these corrected values.

UmerShahidengr (approved):
Looks good to go

PR #705, opened 2025-10-11
Some fixes and consistency in PMP tests

sudo-apt-abdullah:
- In pmpzicbo_prefetch test file, regex check was commente out and it was getting selected even for RV32E. 
- In pmpzca_legal_lxwr test, pmp checks for {c.ld, c.ldsp, c.sd, c.sdsp} for XLEN 64 was missing.
- In pmp{s,u}_csr_access, we're supposed to check the access on all PMP entries even if writable pmp entries are not 64 in S/U mode.


UmerShahidengr (approved):
Looks good to go

Issue #704, opened 2025-10-10
auipc hint test includes test binary addresses in its signature


Wren6991:
The `TEST_AUIPC` macro here subtracts program addresses from the values under test to ensure the output is position-independent:

https://github.com/riscv-non-isa/riscv-arch-test/blob/2510be0ac23020ed5137c5d4c1ae89135ce7f9fd/riscv-test-suite/env/test_macros.h#L937-L943

However the absolute address is left behind in `test_reg`. That means the last invocation of that macro here loads a non-position-independent value into `a2`:

https://github.com/riscv-non-isa/riscv-arch-test/blob/2510be0ac23020ed5137c5d4c1ae89135ce7f9fd/riscv-test-suite/rv32i_m/hints/src/auipc-01.S#L112

This is then written out in the test signature by `TEST_STORE_GPRS_AND_STATUS`. This causes spurious test failure when the two binaries have different addresses, which is quite likely to be the case if additional support code is required for running the tests on a real DUT.

Wren6991:
Possibly just need to clear that register:

```patch
diff --git a/riscv-test-suite/env/test_macros.h b/riscv-test-suite/env/test_macros.h
index 1272fbd4..a64c9d1e 100644
--- a/riscv-test-suite/env/test_macros.h
+++ b/riscv-test-suite/env/test_macros.h
@@ -940,6 +940,7 @@ ADDI(swreg, swreg, RVMODEL_CBZ_BLOCKSIZE)
       1:                               ;\
       inst destreg, imm                        ;\
       sub destreg, destreg, testreg    ;\
+      li testreg, 0 ;\
       )
 
 //Tests for instructions with register-immediate operand
```

UmerShahidengr:
Good spot. This will get fixed by #706 . Thank you @Pagerd

UmerShahidengr:
PR merged. Closing this one.

PR #706, opened 2025-10-14
Fix auipc hint test includes test binary addresses in its signature

Pagerd:
## Description

This pr fixed issue at [#704](https://github.com/riscv-non-isa/riscv-arch-test/issues/704)

### Related Issues

[#704](https://github.com/riscv-non-isa/riscv-arch-test/issues/704)


allenjbaum (changes requested):
This may be a nice-to-have... but is it necessary?
The testreg is always initialized before being used.
There is no register that is used without being initialized.
Leaving testreg  unchanged leaves an absolute offset, since LA will compile to 
AUIPC testreg, testreg 
ADDI   testreg, testreg, 4 (the offset of the following instruction)- which is followed by 
AUIPC destreg, imm   which has testreg subtracted from it (and should ==imm) which  is stored in the signature. 
Testreg is then a dead register: it has no impact on the test, and any further use will always be initialized before being reused , so no impact on other tests either. 
Leaving it unchanged may be useful for debug - so I'm not sure it is even a nice-to-have.

Wren6991:
> Testreg is then a dead register: it has no impact on the test, and any further use will always be initialized before being reused , so no impact on other tests either.

This is unfortunately untrue for the auipc hint test because after invoking this macro it writes _all_ GPRs to the test signature, including `testreg`:

https://github.com/riscv-non-isa/riscv-arch-test/blob/2510be0ac23020ed5137c5d4c1ae89135ce7f9fd/riscv-test-suite/rv32i_m/hints/src/auipc-01.S#L112-L115

This means the same test linked at different addresses will have a different signature and will be flagged as a test fail.

allenjbaum:
Hmm; we need to look at other hint tests then, not just this one.
This is an issue with any test that saves all the registers .

There is a RVTEST_SAVE_GPRS(_BR, _LBL, ...) macro  with a base reg and base
address parameters, with an optional mask that has a bit per Xreg.
It is rarely used - for debug only - because its too hard to track which
registers contain an address .
I don't see how swreg (the signature pointer) would match for different
load points.
Granted, that's usually just one register, and you can avoid that (so you
don't save ALL the registers in the signature) using the above macro and a
mask.

The real fix in this case is to make the load points always be the same,
and the SAIL team intends to enable that.
But, this is the easiest short-term fix. A better one is to use a single
temp register for all auipc hint tests, and avoid saving just that (and
swreg) with a mask and the above macro.

Wren6991:
> I don't see how swreg (the signature pointer) would match for different
load points.

I was also puzzled by this. It looks like that macro just doesn't store registers `x0`, `x3`, `x30`, `x31` and it's used with a base of `x30`. https://github.com/riscv-non-isa/riscv-arch-test/blob/2510be0ac23020ed5137c5d4c1ae89135ce7f9fd/riscv-test-suite/env/test_macros.h#L418-L457

The `TEST_AUIPC` macro is used with `x3` as a signature base and that's not covered by the `TEST_STORE_GPRS_AND_STATUS` macro either.

> The real fix in this case is to make the load points always be the same,
and the SAIL team intends to enable that.

Surely the binary layout is still different if you have DUT-specific code in your `RVMODEL_BOOT` for example?

allenjbaum:
No; the test itself is a fixed size. RVMODEL_BOOT  (and any RVMODEL macro
which isn't fixed size) is inserted at an address outside the test itself
- specifically after the test itself so if it's loaded at the same start
point, then all test locations will match. There will be branch offsets
inside that differ, only.
(which has its own set of issues, obviously, but can be addressed in
several ways, e.g. limiting test size to avoid exceeding immediate sizes,
among other things.

Wren6991:
The example linker scripts for RV32 (spike and sail) put `.text.init` at the start of the binary, so that's where I'd expect `RVMODEL_BOOT` to end up. This part is still fine as there's a 4k alignment before the remainder of `.text` and you're unlikely to need more than 4k, so fair enough.

However I could still get a different binary layout within `.text` because I put the test signature in its own section to make it easier to scrape it out of the testbench memory. That can give different data relaxations (`la` relaxing to either `auipc + addi` or `auipc + c.addi`) which means different instruction sizes.

This is all quite subtle; is it really less work than ensuring test signatures don't leak information about binary layout?

allenjbaum:
The signature may be at a different address, but we've been careful to ensure that any (physical) addresses stored into a signature are actually offsets from known address, not absolute (e.g. offsets from code_begin, data_begin, or even signature_begin) for exactly this kind of reason. 
Virtual addresses don't matter - those are the same regardless of how tests are loaded.  
We may have missed some cases, and those will need to be fixed. AUIPC hint tests are one of the ones that got missed, obviously, because it saved something that isn't the output of an instruction, but just the opposite.

allenjbaum:
The trap handler, prolog and epilog, and all RVMODEL macros that are not a
single op are in the .text section... but at the end of it,, not the
beginning.
The first thing the new trap handler does it branch to a point after the
CODE_END and insert the RVMODEL_BOOT code, then branch back, for example.
I'm not sure how .text.init deals with that, but it doesn't seem to break
anything.

As a rule, relaxation is supposed to be forbidden inside tests for exactly
that reason (except in C-extension tests and the like,
of course which explictly enable it for short stretches).
IT's OK to have relaxation on in prolog, epilog, and inside RVMODEL macros.
but generally we don't.

Wren6991:
> The trap handler, prolog and epilog, and all RVMODEL macros that are not a
single op are in the .text section... but at the end of it,, not the
beginning.

This appears untrue, at least for the example `riscof-plugins` folders on the current `dev` branch. Is there somewhere else I should be looking? Reading through the code in this repository, tests start by setting the output section to `.text.init` and then invoking the `RVMODEL_BOOT` and `RVTEST_CODE_BEGIN` macros:

https://github.com/riscv-non-isa/riscv-arch-test/blob/241b1fe294567751bd3ae6f338b1dbfc40c41a20/riscv-test-suite/rv32i_m/I/src/add-01.S#L22-L26

The `RVMODEL_BOOT` macro is empty for the example targets in `riscof-plugins`:

https://github.com/riscv-non-isa/riscv-arch-test/blob/241b1fe294567751bd3ae6f338b1dbfc40c41a20/riscof-plugins/rv32/spike_simple/env/model_test.h#L35

The `RVTEST_CODE_BEGIN` macro redundantly sets the section to `.text.init` and also does not pop the section (`.option pop` just pops options):

https://github.com/riscv-non-isa/riscv-arch-test/blob/241b1fe294567751bd3ae6f338b1dbfc40c41a20/riscv-test-suite/env/arch_test.h#L1923-L1939

Based on the above I'd expect `RVMODEL_BOOT`, `RVTEST_CODE_BEGIN` and then the actual test body to all be emitted into the `.text.init` section at the start of the binary with ~no intermediate padding. If I disassemble one of the test binaries then this is exactly what I see. I think part of the reason the example spike and sail plugins don't miscompare is because they both have empty `RVMODEL_BOOT` macros.

> As a rule, relaxation is supposed to be forbidden inside tests for exactly
that reason (except in C-extension tests and the like,
of course which explictly enable it for short stretches).

I don't see any `.option norelax` or `-mno-relax` anywhere in this repository. Doesn't binutils add `R_RISCV_RELAX` relocs by default for `la`?

allenjbaum:
Ah, yes - there is a new version that should be merged soon that (I think)
fixes most of these issues; you found these issues at the same time I was
fixing them.
The RVMODEL_BOOT macro either does (or will) get moved into the
RVTEST_CODE_END macro, which is instantiated after the test ends.
Spike and Sail simply don't need to do anything; real implementations may
or may not either.

Note that it is a requirement of the authors of RVMODEL macros that are
instantiated inside tests to be defined to be branches to a subroutine
defined inside RVMODEL_BOOT to ensure that code offsets don't change
between reference models

The order of the sections shouldn't matter, as any code or data addresses
that are stored as signatures are relocated relative to the section they
are in.
It sounds like you've found some other issues that we need to fix,
regarding section definitions. We will need to look at those.

Hmm - the missing norelax may be bugs. I was using .option norvc, which may
have the wrong behavior.
Also, you will not that we don't use la anywhere; we use a custom LA()
macro that guarantees a fixed size (and we try hard to avoid needing LA at
all in the trap handler, at least)

UmerShahidengr (commented):
Good spot, and nice fix. Thank you

UmerShahidengr (approved):

allenjbaum (approved):
This fix should work

PR #708, opened 2025-10-24
Glossary updates

MikeOpenHWGroup:
A few minor edits to the Glossary to pipeclean the process.


allenjbaum:
For what it's worth, I think I know how to construct self-checking tests
that will handle arbitrary tail agnostic behavior.

ctp/src/intro.adoc
```diff
@@ -29,7 +29,7 @@ Parameters will eventually all have formal names in UDB when they are all identi
...
 
-* Non-deterministic Behaviors: behaviors that are unpredictable despite a full configuration.  For example, a tail-agnostic vector instruction might copy vector elements from the source, or might provide elements that are all 1s.  Either behavior is correct, and the hart might change behaviors from one instruction to the next, so the DUT and reference model cannot be guaranteed to exactly match.  Self-checking tests must accept all legal outcomes for non-deterministic behaviors.
+* Non-deterministic Behaviors: behaviors that are unpredictable despite a full configuration.  For example, a tail-agnostic vector instruction might copy vector elements from the source, or might provide elements that are all 1s.  Either behavior is correct, and the hart might change behaviors from one instruction to the next, so the DUT and reference model cannot be guaranteed to exactly match. footnote:[Ideally, in such cases a Reference Model would produce a list of all legal responses. However, there is no known open-source RISC-V Reference Model that does this.] Self-checking tests must accept all legal outcomes for non-deterministic behaviors.
```
allenjbaum, line 46: I think you also want to define "constrained nondeterministic, where the resulting behoving can be only one of a small set of behaviors (e.g. either return zero or the other known value, either trap with misaligned or page_fault trap cause, one of 7 possible behaviors for misaligned stores, etc.

MikeOpenHWGroup: Good point @allenjbaum.  @jordancarlin, what say you?  I can add this verbiage if we all agree.

allenjbaum:
A couple of points:
 - tail agnostic is allowed to be non-deterministic, it isn't required to
be. So, any particular implementation could be completely deterministic for
this case.
 - the concept of constrained-nondeterministic is especially important for
CSC; tail agnostic is an excellent example of that.

Tail agnostic values are limited - constrained - to just two possibilities:
either the destination element is unchanged, or it is set to all1.
When the constraints are tight (e.g. 2 values as above) certification tests
can fairly easily accommodate it.
  (or if a particular implementation isn't actually non-deterministic, we
could handle that even more easily)
FP unordered reductions (or integer reduction with saturation?) is actually
constrained... but with *very* loose constraints (e.g. =inf..-inf)
You can actually calculate what the limits are unordered FP reductions (so
tests with fixed values could actually easily check whether results stay
within the limits).
That would be easy to do in a test (though calculating those limits is not,
but would be outside the test).

The other interesting case is when the constrained-non-determinism depends
deterministically on some non-architectural parameter.
E.g. if misaligned stores
- are always split when going across some granularity boundary,
 - are always executed in increasing or decreasing address order
 - page-fault exceptions always (or never) have higher priority than
misaligned exceptions
       (the spec allows both, but it is not currently a profile-named
parameter/extension, nor is it clear from the spec wording if it must be a
fixed value)
IF tests are given those parameters, then we can have a single test with 7
sets of result values (which sounds easier than generating 7
spearate tests)

ctp/src/intro.adoc
allenjbaum: How about:  *Constrained Non-Deterministic Behaviors: behaviors that cannot be predicted based on either architectural configuration or defined parameters, but whose result are constrained to be a small number of possible values. The example of a tail-agnostic vector instruction is an one of those, where the number of legal results is limited to just two: either the previous value, or all ones. Misaligned stores may have a much larger set of legal results (at least 7), though may be predicted based on parameter values that can be are supplied by the implementor.

davidharrishmc:
@MikeOpenHWGroup thank you for starting this pipe cleaning process.

I've made PR #754, which I think supersedes this.  Please close #708 if you are satisfied.

754 splits nondeterministic behaviors into constrained and unconstrained nondeterministic behaviors.  It partially addresses Allen's comment.  

I'm not clear about Allen's comment of 7 misaligned store behaviors that can be predicted based on parameters.  If parameters are sufficient to completely define the behavior, I don't think that is an example of a nondeterministic behavior, but rather one that can be predicted by the reference model.

I also did not address @MikeOpenHWGroup 's comment about separating the Coverage Model from the Test Suite.  Presently the CTP envisions the test suite also containing a coverage model.  Mike seems to have started to separate these, but the edited text still says I_coverage.svh is part of the test suite, so the new language is not self-consistent.  It may be necessary to separate the two, but let's have a conversation about that.

jordancarlin:
> I also did not address @MikeOpenHWGroup 's comment about separating the Coverage Model from the Test Suite. Presently the CTP envisions the test suite also containing a coverage model. Mike seems to have started to separate these, but the edited text still says I_coverage.svh is part of the test suite, so the new language is not self-consistent. It may be necessary to separate the two, but let's have a conversation about that.

I agree with @MikeOpenHWGroup on this one. All of the coverage files/coverpoints make up the Coverage Model and the Test Suite has the assembly tests that cover everything specified in the Coverage Model. What about something like this:

> * Coverage Model: all of the covergroups and coverpoints used for a extension, combination of extensions, or group of extensions. For example, the RV32I coverage model contains the I_coverage.svh coverage file while the RV64ZfaD coverage model contains the ZfaD_coverage.svh coverage file.  The RVA23S64 coverage model contains all of the required and selected optional coverage models for the profile, incluidng RV64I, RV64ZfaD, ExceptionsS, etc.
> * Test Suite: A collection of test files to test a particular extension, combination of extensions, or group of extensions. Each test suite corresponds to and fully exercises a coverage model. For example, the RV32I test suite contains the I the rv32/I/add.S, addi.S, sub.S, and other tests files with tests that hit the coverpoints in the RV32I coverage model.  The RV64ZfaD test suite contains rv64/ZfaD/fround.d.S and other tests files with tests that hit the coverpoints in the RV64ZfaD coverage model.  The RVA23S64 test suite contains all of the required and selected optional test suites for the profile, incluidng RV64I, RV64ZfaD, ExceptionsS, etc.

allenjbaum:
There's a nuance in my statement. It could be predicted IF the
implementation supports those specific parameters consistently.
But, those parameters do not necessarily appear in any spec (e.g.:

   1.   - what granularity crossing will result in a split access
   2.  - if the access is split, will lower/higher addresses be
   (consistently) fetched first or lower/higher significance (so depending on
   big/little endian?)
           (that's 4 possibilities)
   3.  - if the access is split, can partial data be stored if only one
   half is not legal
   4.  - if both halves of a split access are illegal, then will the cause
   be reported as the first half fetched, or the highest priority of each half


Assumptions I have made is that
 #3 if true, will only store partial data if the first access is legal
 #4 is assumed to be first half fetched

IF so, the required parameters are granularity, relative priority of
misaligned vs. page fault, order of split access
If not, it implies that nothing is stored until both halves have passed
permission checks

MikeOpenHWGroup:
I would suggest anything to do with split accesses are a _verification_, not _certification_ problem.

allenjbaum:
Why Assuming we have parameters, it is no different than anything else that
is defined by parameters in the spec
(and there is at least one explicitly defined, and others that don't have
names,e.g for which boundary crossinging do you support misaligned) That's
a PMA definition.

jordancarlin:
@MikeOpenHWGroup can you update this PR if any of it is still relevant and close it if not?

MikeOpenHWGroup:
Thanks for the gentle reminder @jordancarlin.  This one popped off my mental stack and never made it back.

Closing.

PR #709, opened 2025-10-25
(CVW) Coverage files for Svade, Svadu, Svnapot & Svpbmt

Zain2050:


coverpoints/priv/Svade_coverage.svh
```diff
@@ -0,0 +1,126 @@
...
+// and limitations under the License.
+////////////////////////////////////////////////////////////////////////////////////////////////
+
+`define COVER_SVADE
+covergroup Svade_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include  "coverage/RISCV_coverage_standard_coverpoints.svh"
+
+    PTE_Abit_unset_s_i: coverpoint ins.current.pte_i[7:0] {
+        wildcard bins leaflvl_s = {8'b?0?01111};
```
davidharrishmc, line 17: do the RW bits matter for an I pte?

coverpoints/priv/Svade_coverage.svh
```diff
@@ -0,0 +1,126 @@
...
+    }
+    PTE_Abit_unset_u_i: coverpoint ins.current.pte_i[7:0] {
+        wildcard bins leaflvl_u = {8'b?0?11111};
+    }
+    PTE_Abit_unset_s_d: coverpoint ins.current.pte_d[7:0] {
+        wildcard bins leaflvl_s = {8'b?0?01111};
```
davidharrishmc, line 23: Does the x bit matter for data PTE?  Does the w bit matter for the access tests?

coverpoints/priv/Svade_coverage.svh
```diff
@@ -0,0 +1,126 @@
...
+    PTE_Dbit_unset_s_d: coverpoint ins.current.pte_d[7:0] {
+        wildcard bins leaflvl_s = {8'b01?0?111};
+    }
+    PTE_Dbit_unset_u_d: coverpoint ins.current.pte_d[7:0] {
+        wildcard bins leaflvl_u = {8'b01?1?111};
+    }
+
+    `ifdef XLEN64
+        PageType_i: coverpoint ins.current.page_type_i {
+            `ifdef SV48
```
davidharrishmc, line 37: Don't forget to add SV57 when moving on to that.

coverpoints/priv/Svade_coverage.svh
```diff
@@ -0,0 +1,126 @@
...
+            `endif
+            `ifdef SV39
+                bins sv39_giga = {2'b10} iff (ins.current.csr[12'h180][63:60] == 4'b1000);
+                bins sv39_mega = {2'b01} iff (ins.current.csr[12'h180][63:60] == 4'b1000);
+                bins sv39_kilo = {2'b00} iff (ins.current.csr[12'h180][63:60] == 4'b1000);
+            `endif
+        }
+    `else
+        PageType_i: coverpoint ins.current.page_type_i {
+            bins sv32_mega = {2'b01} iff (ins.current.csr[12'h180][31] == 1'b1);
```
davidharrishmc: I usually drop the == 1'b1 from this kind of Boolean expression

coverpoints/priv/Svade_coverage.svh
```diff
@@ -0,0 +1,126 @@
...
+        bins set = {1};
+    }
+    read_acc: coverpoint ins.current.read_access {
+        bins set = {1};
+    }
+    write_acc: coverpoint ins.current.write_access{
+        bins set = {1};
+    }
+    ins_page_fault: coverpoint  ins.current.csr[12'h342][31:0] {
+        bins ins_page_fault = {32'd12};
```
davidharrishmc: I have forgotten - does the current mcause value get updated while the access is taking place?  Will we see exec_acc and ins_page_fault concurrently on the first access that causes the fault?  Should we also be monitoring the ins.current.trap signal to confirm the trap is really happening, not stale information from an earlier trap?  I might be wrong - I haven't coded these kinds of coverpoints in a while.

coverpoints/priv/Svade_coverage.svh
Zain2050: No. We can use '?' here but this is just how other coverpoints are written. We setup a single PTE with RWX and use it for both data and instruction accesses.

coverpoints/priv/Svade_coverage.svh
```diff
@@ -0,0 +1,126 @@
+///////////////////////////////////////////
+//
+// RISC-V Architectural Functional Coverage Covergroups
+//
+// Copyright (C) 2024 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+//
+// Licensed under the Solderpad Hardware License v 2.1 (the “License”); you may not use this file
```
jordancarlin: Update license header to match the other coverpoints (it changed when we migrated repos). Applies to all files.

coverpoints/priv/Svade_coverage.svh
Zain2050: Yes, I'll do it when wally gets SV57. Currently page_type_ is a 2-bit signal and therefore I couldn't write coverpoint for a level 4 page.

coverpoints/priv/Svade_coverage.svh
Zain2050: I'll remove it.

coverpoints/priv/Svadu_coverage.svh
```diff
@@ -0,0 +1,139 @@
...
+    Abit_unset_exec_s:  cross PTE_Abit_unset_s_i, PageType_i, Svadu_enabled, jalr, priv_mode_s;
+    Abit_unset_exec_u:  cross PTE_Abit_unset_u_i, PageType_i, Svadu_enabled, jalr, priv_mode_u;
+    Abit_unset_read_s:  cross PTE_Abit_unset_s_d, PageType_d, Svadu_enabled, lw, priv_mode_s;
+    Abit_unset_read_u:  cross PTE_Abit_unset_u_d, PageType_d, Svadu_enabled, lw, priv_mode_u;
+    Abit_unset_write_s: cross PTE_Abit_unset_s_d, PageType_d, Svadu_enabled, sw, priv_mode_s;
+    Abit_unset_write_u: cross PTE_Abit_unset_u_d, PageType_d, Svadu_enabled, sw, priv_mode_u;
+    Dbit_unset_write_s: cross PTE_Dbit_unset_s_d, PageType_d, Svadu_enabled, sw, priv_mode_s;
+    Dbit_unset_write_u: cross PTE_Dbit_unset_u_d, PageType_d, Svadu_enabled, sw, priv_mode_u;
+
+    // Hardware Update of DA bits fail when pagetable/pte doesn't have write PMP permission
```
jordancarlin: We should probably move this to a separate Svadu_PMP suite. An implementation can support Svadu without supporting PMP.

coverpoints/priv/Svade_coverage.svh
Zain2050: Yes, mcause gets updated while the access is taking place. 
`ins.current.trap` is being used in some coverage files, not all of them. If you want, I could add it to these coverpoints.

davidharrishmc: I like the trap to make the coverpoint more robust.

coverpoints/priv/Svade_coverage.svh
```diff
@@ -0,0 +1,126 @@
...
+        Svade_enabled: coverpoint  ins.current.csr[12'h30A][61] {
+            bins ADUE_unset = {1'b0};
+        }
+    `else
+        Svade_enabled: coverpoint  ins.current.csr[12'h31A][29] {
+            bins ADUE_unset = {1'b0};
+        }
+    `endif
+
+    Abit_unset_exec_s:  cross PTE_Abit_unset_s_i, PageType_i, Svade_enabled, ins_page_fault, exec_acc, priv_mode_s;
```
davidharrishmc, line 102: The names of these coverpoints deviate from the names in the spreadsheet
https://docs.google.com/spreadsheets/d/1LGwD2e01j4SN9tg85nt6gPwpp7nXyXMF7hqbw3jvXko/edit?gid=1346750896#gid=1346750896

Maybe these names should change to things like cp_a_x_u in this file?

coverpoints/priv/Svadu_coverage.svh
```diff
@@ -0,0 +1,139 @@
...
+        Svadu_enabled: coverpoint  ins.current.csr[12'h31A][29] {
+            bins ADUE_set = {1'b1};
+        }
+    `endif
+
+    PMP_pagetable: coverpoint  ins.current.csr[12'h3A0][7:0] {      // PMP permission of page table
+        wildcard bins w_unset = {8'b??????01};
+    }
+
+    Abit_unset_exec_s:  cross PTE_Abit_unset_s_i, PageType_i, Svadu_enabled, jalr, priv_mode_s;
```
davidharrishmc: There is a lack of parallel structure between Svadu and Svade.  Svade used exec_access, while this Svadu uses jalr.  Could these be made the same, changing Svadu to exec_access as well, and changing trap to 0 instead of 1?

coverpoints/priv/Svadu_coverage.svh
```diff
@@ -0,0 +1,139 @@
...
+    Abit_unset_exec_u:  cross PTE_Abit_unset_u_i, PageType_i, Svadu_enabled, jalr, priv_mode_u;
+    Abit_unset_read_s:  cross PTE_Abit_unset_s_d, PageType_d, Svadu_enabled, lw, priv_mode_s;
+    Abit_unset_read_u:  cross PTE_Abit_unset_u_d, PageType_d, Svadu_enabled, lw, priv_mode_u;
+    Abit_unset_write_s: cross PTE_Abit_unset_s_d, PageType_d, Svadu_enabled, sw, priv_mode_s;
+    Abit_unset_write_u: cross PTE_Abit_unset_u_d, PageType_d, Svadu_enabled, sw, priv_mode_u;
+    Dbit_unset_write_s: cross PTE_Dbit_unset_s_d, PageType_d, Svadu_enabled, sw, priv_mode_s;
+    Dbit_unset_write_u: cross PTE_Dbit_unset_u_d, PageType_d, Svadu_enabled, sw, priv_mode_u;
+
+    // Hardware Update of DA bits fail when pagetable/pte doesn't have write PMP permission
+    Abit_unset_pmpW_unset_exec_s:  cross PTE_Abit_unset_s_i, PageType_i, Svadu_enabled, PMP_pagetable, jalr, ins_acc_fault, priv_mode_s;
```
davidharrishmc: Should these tests that depend on Svadu and PMP go into a new SvaduPMP file and spreadsheet tab?

coverpoints/priv/Svade_coverage.svh
Zain2050: Sure, I'll change it

coverpoints/priv/Svadu_coverage.svh
```diff
@@ -0,0 +1,139 @@
...
+        Svadu_enabled: coverpoint  ins.current.csr[12'h30A][61] {
+            bins ADUE_set = {1'b1};
+        }
+    `else
+        Svadu_enabled: coverpoint  ins.current.csr[12'h31A][29] {
+            bins ADUE_set = {1'b1};
+        }
+    `endif
+
+    PMP_pagetable: coverpoint  ins.current.csr[12'h3A0][7:0] {      // PMP permission of page table
```
davidharrishmc: We're getting to the point that there are some pretty big gaps between the coverpoints and the actual conditions that must be checked.  Here, we're just looking at pmpcfg0.  There's an assumption that pmpaddr0 is matching the region with the PTE.  Should that be checked in some way?

coverpoints/priv/Svadu_coverage.svh
Zain2050: Yes, we probably should.

coverpoints/priv/Svadu_coverage.svh
Zain2050: Well there's a reason for doing this. In Svadu, we're getting a fault during the page table walk and during the walk read_acc & write_acc are 0. Therefore, I'm using lw and sw to collect coverage here. Same thing was done in VM_PMP covergroup.

coverpoints/priv/Svadu_coverage.svh
Zain2050: Yes, this is a bit weird. This is just how Huda wrote in the VM_PMP covergroup. I think the reason is that the PTE can be at any physical addr and in the coverpoint we can't really specifiy the addr of the PTE. Therefore, not checking it in the coverpiont, only specifiying it in the testplan.

coverpoints/priv/Svade_coverage.svh
Zain2050: When you moved these coverpoints from the VM tab to a separate Svade tab you said that you're temporarily writing cp_a and cp_d because the coverpoint names were not present in the VM_tab. But the VM_coverage file had names like `Abit_unset_exec_s`. 
Should we replace these with `cp_a_x_u`?

coverpoints/priv/Svade_coverage_init.svh
```diff
@@ -0,0 +1,11 @@
+///////////////////////////////////////////
+//
+// RISC-V Architectural Functional Coverage Covergroups Initialization File
+//
+// Copyright (C) 2024 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
```
jordancarlin: The license for the `*_init.svh` files still needs to be updated.

coverpoints/priv/Svadu_coverage.svh
Zain2050: Done!

coverpoints/priv/Svade_coverage.svh
davidharrishmc: Any way you want is ok, so long as the name is consistent across the SystemVerilog and spreadsheet.

coverpoints/priv/Svnapot_coverage.svh
```diff
@@ -0,0 +1,114 @@
...
+    }
+    PTE_Svnapot_u_d: coverpoint ins.current.pte_d[63:0] {
+        wildcard bins leaflvl_u = {64'b1?????????????????????????????????????????????????1000_??11?11111};
+    }
+    PTE_Svnapot_reserved_enc_s_i: coverpoint ins.current.pte_i[63:0] {
+        wildcard bins illegal_xxx1 = {64'b1????????????????????????????????????????????????????1_??11?01111};
+        wildcard bins illegal_xx1x = {64'b1???????????????????????????????????????????????????1?_??11?01111};
+        wildcard bins illegal_x1xx = {64'b1??????????????????????????????????????????????????1??_??11?01111};
+        wildcard bins illegal_0xxx = {64'b1?????????????????????????????????????????????????0???_??11?01111};
+        wildcard bins illegal_PPNi = {64'b1????????????????????????????????????????1000?????????_??11?01111};
```
davidharrishmc: illegal_PPNi is associated with an NAPOT entry on a non-kilo page PTE.  I'm not following how the illegal_PPNi bin is enough to identify that it is non-kilo page.

coverpoints/priv/Svnapot_coverage.svh
```diff
@@ -0,0 +1,114 @@
...
+    }
+    load_page_fault: coverpoint  ins.current.csr[12'h342][31:0] {
+        bins load_page_fault = {32'd13} iff (ins.current.trap);
+    }
+    store_page_fault: coverpoint  ins.current.csr[12'h342][31:0] {
+        bins store_amo_page_fault = {32'd15} iff (ins.current.trap);
+    }
+
+    Svnapot_legal_enc_exec_s:  cross PTE_Svnapot_s_i, PageType_i, mode, exec_acc, priv_mode_s;
+    Svnapot_legal_enc_exec_u:  cross PTE_Svnapot_u_i, PageType_i, mode, exec_acc, priv_mode_u;
```
davidharrishmc: Tie coverpoint names to spreadsheet for all files.

coverpoints/priv/Svnapot_coverage.svh
```diff
@@ -0,0 +1,114 @@
...
+    Svnapot_legal_enc_read_s:  cross PTE_Svnapot_s_d, PageType_d, mode, read_acc, priv_mode_s;
+    Svnapot_legal_enc_read_u:  cross PTE_Svnapot_u_d, PageType_d, mode, read_acc, priv_mode_u;
+    Svnapot_legal_enc_write_s: cross PTE_Svnapot_s_d, PageType_d, mode, write_acc, priv_mode_s;
+    Svnapot_legal_enc_write_u: cross PTE_Svnapot_u_d, PageType_d, mode, write_acc, priv_mode_u;
+
+    Svnapot_illegal_enc_exec_s:  cross PTE_Svnapot_reserved_enc_s_i, PageType_i, mode, exec_acc, ins_page_fault, priv_mode_s;
```
davidharrishmc: These names should involve "reserved"

davidharrishmc: The PageType_i indicates kilo page.  I think you'll need additional coverpoints for i >= 1 (nonkilopage).

coverpoints/priv/Svnapot_coverage.svh
Zain2050: The spec doesn't mention anything about non-kilo pages neither the testplan. The spec just tells us that NAPOT encoding for pte.ppn[i] where i>=1 is reserved. I thought this meant encoding ppn[i>=1] for kilo pages. 
In case of that encoding for ppn[1], would we need to test it for mega page?

coverpoints/priv/Svpbmt_coverage.svh
```diff
@@ -0,0 +1,151 @@
...
+//
+////////////////////////////////////////////////////////////////////////////////////////////////
+
+`define COVER_SVPBMT
+covergroup Svpbmt_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include  "coverage/RISCV_coverage_standard_coverpoints.svh"
+
+    nonleaf_PTE_pbmt_i: coverpoint ins.current.pte_i[63:0] {
+        wildcard bins pbmt_1 = {64'b?01?????????????????????????????????????????????????????00?00001};
```
davidharrishmc: Why are the 0s required in the lsb's of the PTE?  Could we just look at valid?

Zain2050: These zeros (RWX=000) make sure that it is a non leaf PTE.

coverpoints/priv/Svnapot_coverage.svh
```diff
@@ -0,0 +1,114 @@
...
+covergroup Svnapot_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include  "coverage/RISCV_coverage_standard_coverpoints.svh"
+
+    PTE_Svnapot_s_i: coverpoint ins.current.pte_i[63:0] {
+        wildcard bins leaflvl_s = {64'b1?????????????????????????????????????????????????1000_??11?01111};
+    }
+    PTE_Svnapot_u_i: coverpoint ins.current.pte_i[63:0] {
+        wildcard bins leaflvl_u = {64'b1?????????????????????????????????????????????????1000_??11?11111};
+    }
```
jordancarlin: What about representing these coverpoints like this instead so we don't need to be explicit about all of the wildcard bits in the middle. Then just update the crosses to include both the napot coverpoint and the priv mode coverpoint. This applies to several of the coverpoints for Svnapot and Svpbmt.

```suggestion
    PTE_napot_i: coverpoint ins.current.pte_i[63] {
        bins napot = {1'b1};
    }
    PTE_s_i: coverpoint ins.current.pte_i[13:0] {
        wildcard bins leaflvl_s = {14'b1000_??11?01111};
    }
    PTE_u_i: coverpoint ins.current.pte_i[13:0] {
        wildcard bins leaflvl_u = {14'b1000_??11?11111};
    }
```

Zain2050: Good suggestion!

coverpoints/priv/Svnapot_coverage.svh
Zain2050: Should I do it for all the levels or just one? Like test for Svanpot on megapage only?

coverpoints/priv/Svnapot_coverage.svh
davidharrishmc: I'm interpreting ppn[i] with respect to Section 12.3.2.  i is the level of the page.

coverpoints/priv/Svnapot_coverage.svh
davidharrishmc: Might as well test on all levels unless it is hard.

coverpoints/priv/Svnapot_coverage.svh
Zain2050: Jordan, take a look.

coverpoints/priv/Svnapot_coverage.svh
Zain2050: Sure

coverpoints/priv/Svpbmt_coverage.svh
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    leaf_PTE_pbmt_s_i: coverpoint ins.current.pte_i[7:0] {
+        wildcard bins pbmt_1 = {8'b11?01111} iff (ins.current.pte_i[62:61] == 2'b01);
+        wildcard bins pbmt_2 = {8'b11?01111} iff (ins.current.pte_i[62:61] == 2'b10);
+    }
+    leaf_PTE_pbmt_u_i: coverpoint ins.current.pte_i[7:0] {
+        wildcard bins pbmt_1 = {8'b11?11111} iff (ins.current.pte_i[62:61] == 2'b01);
+        wildcard bins pbmt_2 = {8'b11?11111} iff (ins.current.pte_i[62:61] == 2'b10);
+    }
```
jordancarlin, line 34: Might be more clear like this and then add both to the cross, but up to you:

```suggestion
    pbmt_mode_i: coverpoint ins.current.pte_i[62:61] {
       bins nc = {2'b01);
       bins io = {2'b10);
    }
    leaf_PTE_s_i: coverpoint ins.current.pte_i[7:0] {
        wildcard bins pbmt = {8'b11?01111};
    }
    leaf_PTE_u_i: coverpoint ins.current.pte_i[7:0] {
        wildcard bins pbmt = {8'b11?11111};
    }
```

coverpoints/priv/Svnapot_coverage.svh
jordancarlin: Svnapot looks great. Left a suggestion for Svpbmt.

coverpoints/priv/Svpbmt_coverage.svh
Zain2050: I thought of taking this approach, but it would require` ignore bins` as some coverpoints are checking against all 3 values and some against just 1.

jordancarlin: Seems fine to leave it as is then.

coverpoints/priv/Svnapot_coverage.svh
Zain2050: Added coverpoints to test for faults when PTE.N=1 for level 1,2 & 3. If it looks good, then this PR can be merged.

PR #716, opened 2025-11-04
stop self loop with trap handler in sail and fix Priv covergroup selection

Noah-Fotenos:


framework/src/act/coverreport.py
```diff
@@ -71,7 +71,7 @@ def report_to_summary(report_path: Path, summary_path: Path) -> None:
         entries.append((name, metric_value, goal_value, bins_value, status_value))
 
     if not entries:
-        raise ValueError("No coverage entries found in report.")
+        raise ValueError("No coverage entries found in report " + report_path.as_posix())
```
jordancarlin:
```suggestion
        raise ValueError(f"No coverage entries found in report {report_path}")
```

jordancarlin (changes requested):

jordancarlin (approved):

PR #728, opened 2025-11-11
ExceptionsS + Reorganization of Exceptions Tests

Noah-Fotenos:


jordancarlin (changes requested):
Please review the files being changed in this PR. You are currently downgrading the `docs-resources` submodule and checking in a Questa transcript.

Also, I assume you should be moving `ExceptionsS.S` from the `tests-dev` directory, not adding a new copy of it.

PR #729, opened 2025-11-11
ExceptionsF

Marina-Bellido:
90% coverage in new framework


davidharrishmc:
Undecided if we want an ExceptionsF that also requires D.

jordancarlin:
Superseded by #811.

PR #731, opened 2025-11-11
ExceptionsInstr

Marina-Bellido:
Getting ExceptionsInstr to run in the new framework. Current coverage is 60%. Did not have a coverage for this file, so I created by just including the ExceptionsM_coverage (this is bc the test states at the top that its a portion of the Exceptions_coverage.) 

also edited Makefile to run for both rv32 and rv64 configs.


jordancarlin:
Coverage files should not be modified. The test needs to be moved to the directory corresponding to the appropriate coverpoints (Ssstrict I believe).

jordancarlin:
This test has been deleted. Closing.

PR #735, opened 2025-11-11
ZicsrU running full coverage

Noah-Fotenos:


tests/priv/ZicsrU/ZicsrU.S
```diff
@@ -0,0 +1,83 @@
...
+// Purpose: Functional coverage tests for CSRs in user mode
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+///////////////////////////////////////////
+
+// General notes:
+// Use csrrw/csrrs/csrrc t6, csr, rs1    when modifying a CSR to also check the old value.
+
+##### START_TEST_CONFIG #####
+# implemented_extensions: [I, Zicsr, Zihintpause]
```
jordancarlin, line 15: Is Zihintpause really needed? It’s a hint, so even if the implementation doesn’t support it, then it should just be a nop. I agree it should be in the march string, but probably not the required extensions.

PR #738, opened 2025-11-12
Updated VM_PMP ACTs to support any value of G

Zain2050:
### Description

These tests initially worked for DUTs having PMP grain <= 4. I have updated them to support any value of G. The ones for SV57 have been pushed to the SV57 PR.


UmerShahidengr (approved):
Thanks for the fix @Zain2050 . LGTM.

PR #739, opened 2025-11-12
ZicsrS running near complete coverage

Noah-Fotenos:
Covergroup              Metric      Goal     Bins     Status        
ZicsrS_scause_cg        100.00%     100      -        Covered       
ZicsrS_sstatus_cg       100.00%     100      -        Covered       
ZicsrS_sprivinst_cg     92.02%      100      -        Uncovered     

RV32 and RV64

Removed Siginitures from test generator to avoid unreconized opcode.


generators/tests/scripts/csrtests.py
jordancarlin, line 1: Can you leave the `SIGUPD` macros and just comment them out so we can add easily add them back later if we decide to do so?

Noah-Fotenos: done

PR #743, opened 2025-11-14
Update arch_test.h

UmerShahidengr:
Initialised the M-mode CSR registers as they are getting written in the signature file in HINT tests, whereas no traps are getting generated in HINT tests, and uninitialised values of these CSRs will cause the signature mismatch.

Issue #703, opened 2025-10-10
Hint tests cover UNSPECIFIED initial value of mstatus.mpp


Wren6991:
The hint tests write out the value of some CSRs including `mstatus` at the start of the test:

https://github.com/riscv-non-isa/riscv-arch-test/blob/2510be0ac23020ed5137c5d4c1ae89135ce7f9fd/riscv-test-suite/rv32i_m/hints/src/add-01.S#L28

There are no traps beforehand, so the test signature includes the reset value of `mstatus.mpp`. However the privileged spec doesn't require a specific reset value, and e.g. spike does not let you configure the reset value of this field to match the DUT. As it has no specified initial value it should probably either be initialised to a known one by the test, or masked in the output.

allenjbaum:
Hmm: is this something that the RVMODEL_BOOT code should be taking care of?
OR we could ensure that the setup code at label rvtest_entry_pt does this
before calling RVMODEL_BOOT by adding this code just before boot code is run
#ifndef    RVMODEL_MSTATUS_RESET
  #define RVMODEL_MSTATUS_RESET 0
#endif
    LI(T1, RVMODEL_MSTATUS_RESET)
    csrr   CSR_MSTATUS, T1

(and setting it to zero is usually safe; tests can still override it if
necessary).
 The only bits guaranteed to be clear are MIE, MPRV, and usually MBE (if
the hart isn't bigendian always)

Wren6991:
Thanks, initialising in `RVMODEL_BOOT` is a good workaround, I'll use that. I ended up using these lines because the assembler is invoked without `_zicsr` for some tests:

```
.option push                    ;\
.option arch, +zicsr            ;\
  csrw mstatus, zero            ;\
.option pop                     ;\
```

I think it's cleaner to fix it in the testcases because otherwise we would have to document that requirement on RVMODEL_BOOT, and it's not something that's actually necessary for booting the model.

allenjbaum:
Hmm, that doesn't always fix the problem. They might not support csrw
mstatus , and might not even implement mstatus.
Even if they do implement mstatus, they might have a different way of
clearing it or it might have a different format,
In all of the above cases  we should make them explicitly define and supply
the custom code that does that - and do it in RVMODEL_BOOT

#ifdef RVMODEL_CUSTOM_MSTATUS_CLR ;\
         RVMODEL_CUSTOM_MSTATUS_CLR ;\
#else ;\
.option push ;\
[...]

PR #744, opened 2025-11-14
Fix the bug caused by #743

UmerShahidengr:
CSR registers will only be initialised in RVTEST_INIT_GPRS when mtrap_routine is defined (i.e, when Zicsr is enabled)

### DIFF
--- a/riscv-test-suite/env/test_macros.h
+++ b/riscv-test-suite/env/test_macros.h
@@ -209,7 +209,7 @@ Mend_PMP:                                    ;\
 // More Robust version of PTE_SETUP_32 to setup a PTE for a PA using Va
 // in a single line.
 //args: PA: Label of Physical Address, PERMS: permissions in hex
-//args: VA: Virtual Address in hex, level: Level to store at
+//args: VA: Virtual Address in hex, level: Level to  at
 #define PTE_SETUP_RV32_New(PA_LBL, PERMS, VA, level)           ;\
     LA(a0, PA_LBL)                                             ;\
     LI(a1, PERMS)                                              ;\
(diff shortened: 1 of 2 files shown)


UmerShahidengr:
@allenjbaum please approve this one, it will fix the issue caused by #743 .
CC: @jordancarlin

jordancarlin:
I think this reintroduces the problem that #743 was trying to solve (#703) because `hints/src/add-01.S` does not define `mtrap_routine` but stores `mstatus`.

allenjbaum:
OK, I am/was confused, and I think I see what is going on.
There are 2 macros defined: INIT_GPRS in env/arch_test.h,
and TEST_STORE_GPRS_AND_STATUS in env/test_macros.h,
The latter stores CSRs into unconditionally; 
the former initialized them unconditionally (on mtrap_routine being defined)
The problem is that the latter is present (twice!) in the hint tests (well, at leasts in the add test).
I think the real fix is to change both of them to only touch CSRs conditionally in both macros.

riscv-test-suite/env/test_macros.h
```diff
@@ -463,7 +463,8 @@ Mend_PMP:                                    ;\
     RVTEST_SIGUPD(sigptr, x27)             /* Store x27 */ ;\
     RVTEST_SIGUPD(sigptr, x28)             /* Store x28 */ ;\
     RVTEST_SIGUPD(sigptr, x29)             /* Store x29 */ ;\
-    /* Store the CSR registers */ ;\
+	#ifdef rvtest_mtrap_routine			;\
+	/* Store the CSR registers */ ;\
     csrr a0, mepc                          /* Read mepc register */ ;\
     RVTEST_SIGUPD(sigptr, a0)              /* Store mepc */ ;\
```
allenjbaum, line 469: The problem with this code is that it also stores sigptr, which is an address, and whose value could be different between reference and implementation.  IF we believe reference and implementation will be loaded at the same addresses, then this is probably OK.

allenjbaum (approved):
LGTM

PR #745, opened 2025-11-17
[ACT4] Interrupts Tests

SadhviNarayanan:
Ports interrupt tests to new framework and adds InterruptsU.


allenjbaum:
Why is the description of this PR "Provide a detailed description of the
changes performed by the PR."

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+// Individual cross coverage for each exception type with delegation registers
+
+// Building block coverpoints (similar to ExceptionsM pattern)
+
+`define COVER_EXCEPTIONSH
+covergroup ExceptionsH_exceptions_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+
+    // Include standard RISCV coverpoints
+    `include "coverage/RISCV_coverage_standard_coverpoints.svh"
```
davidharrishmc: I don't see this file in the act4 repo yet.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    imm_bit_1: coverpoint ins.current.imm[1] {
+        bins one = {'1};
+    }
+
+    // ============================================================================
+    // PRIVILEGE MODE COVERPOINTS
+    // ============================================================================
+
+    // All 5 privilege modes (M/HS/VS/VU/U)
+    modes: coverpoint {ins.prev.mode, get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hstatus", "v")} {
```
davidharrishmc: status doesn't have a v field. Check with Neil's PR about the preferred idiom for mode with hypervisor.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins not_delegated = {0};
+    }
+
+    // All ECALL bits disabled (no delegation to M-mode)
+    medeleg_ecall_bits_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") & 32'h00000F00) {
+        bins all_ecall_bits_zero = {32'h00000000};
+    }
+
+    // ECALL delegation enabled (generic)
+    medeleg_ecall_enabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") & (1<<8)) {
```
davidharrishmc: I think you want [8] instead of (1<<8).  Currently this looks like it makes a 32-bit value using bitwise AND, but the bin is a 1-bit value.  Same below.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins delegated = {1};
+    }
+
+    // ECALL from VU-mode delegated to VS-mode
+    hedeleg_ecall_enabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hedeleg", "deleg") & (1<<8)) {
+        bins delegated = {1};
+    }
+
+    // EBREAK bits disabled (no delegation)
+    medeleg_ebreak_bits_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") >> 3) {
```
davidharrishmc: I think you want [3] rather than >> 3.  nd then a bin of 0 not 32'h0.  I don't know what "all_ebreak_bits_zero" means because there is only one breakpoint exception.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins all_ebreak_bits_zero = {32'h0};
+    }
+
+    // ============================================================================
+    // VSSTATUS CSR COVERPOINTS (VS-mode traps)
+    // ============================================================================
+
+    vsstatus_spp: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "vsstatus", "spp") {
```
davidharrishmc: There are a bunch of sample_after and sample_before statements that are tricky to understand and get right.  How about some comments explaining why each one is used where.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    cp_hedeleg_ecall: cross ecall, modes, medeleg_delegation, hedeleg_delegation;
+
+    // ECALL delegation crosses
+    cp_ecall_to_vs: cross 
+        ecall,
+        priv_mode_vu,
+        medeleg_ecall_u,
+        hedeleg_ecall_u,
+        vsstatus_spp,
+        vsstatus_spie_correct,
```
davidharrishmc: The bits to check are part of the test, not part of the coverpoint.  Remove vsstatus_spp, vsstatus_spie_correct, vsstatus_sie_after_trap,

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    cp_hedeleg_load_misaligned: cross loadops, adr_LSBs, modes, medeleg_delegation, hedeleg_delegation;
+    cp_hedeleg_load_access_fault: cross loadops, illegal_address, modes, medeleg_delegation, hedeleg_delegation;
+    cp_hedeleg_store_misaligned: cross storeops, adr_LSBs, modes, medeleg_delegation, hedeleg_delegation;
+    cp_hedeleg_store_access_fault: cross storeops, illegal_address, modes, medeleg_delegation, hedeleg_delegation;
+    cp_hedeleg_ecall: cross ecall, modes, medeleg_delegation, hedeleg_delegation;
+
+    // ECALL delegation crosses
+    cp_ecall_to_vs: cross 
+        ecall,
+        priv_mode_vu,
```
davidharrishmc: Let's change this to prove_mode_to_vs to test calls from both vu and vs.  I've changed the testplan accordingly.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    // ============================================================================
+    // SSTATUS CSR COVERPOINTS (HS-mode traps)
+    // ============================================================================
+
+    sstatus_sie_after_trap: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "sstatus", "sie") {
+        bins sie_disabled = {0};
+        bins sie_enabled = {1};
+    }
+
+    sstatus_spie_correct: coverpoint
```
davidharrishmc: I don't think this coverpoint is needed.  See where it was used.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    // ============================================================================
+    // VSSTATUS CSR COVERPOINTS (VS-mode traps)
+    // ============================================================================
+
+    vsstatus_spp: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "vsstatus", "spp") {
+        bins vu_mode = {0};
+        bins vs_mode = {1};
+    }
+
+    vsstatus_spie_correct: coverpoint
```
davidharrishmc: Also not needed.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins not_virtual = {0};
+        bins virtual = {1};
+    }
+
+    mstatus_gva: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mstatus", "gva") {
+        bins gva_invalid = {0};
+        bins gva_valid = {1};
+    }
+
+    mstatus_mpie_correct: coverpoint
```
davidharrishmc: probably not needed.  Check all coverpoints still needed after simplifying the main coverpoints later.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    mstatus_mie_before_trap: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "mie") {
+        bins mie_was_disabled = {0};
+        bins mie_was_enabled = {1};
+    }
+
+    // ============================================================================
+    // TRAP VERIFICATION COVERPOINTS
+    // ============================================================================
+
+    trap_to_vs: coverpoint {ins.current.mode, get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "hstatus", "v")} {
```
davidharrishmc: Check idiom for all of these.  No hstatus.v

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+
+    // ============================================================================
+    // TRAP VECTOR COVERPOINTS
+    // ============================================================================
+
+    vstvec_different_from_stvec: coverpoint 
+        (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vstvec") != 
+        get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "stvec")) {
+        bins different_handlers = {1};
+        bins same_handlers = {0};
```
davidharrishmc: Testplan only calls for different_handlers.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    }
+
+    uses_vstvec_handler: coverpoint
```
davidharrishmc: Unused coverpoint.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        trap_to_m;
+
+    // Trap vector crosses
+    cp_vstvec: cross 
+        ecall,
+        priv_mode_to_vs,
+        medeleg_ecall_enabled,
+        hedeleg_ecall_enabled,
+        vstvec_different_from_stvec,
+        not_using_stvec,
```
davidharrishmc: doesn't need not_using_stvec or trap_to_vs

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "stvec")) {
+        bins stvec_not_used = {1};
+        bins stvec_used = {0};
+    }
+
+    // ============================================================================
+    // COUNTER DELEGATION COVERPOINTS
+    // ============================================================================
+
+    h_counteren: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hcounteren", "counteren") & (1<<2)) {
```
davidharrishmc: Check for all of these [2] rather than & (1<<2) kinds of things.

davidharrishmc: There should be something about h/mcounteren_ir in the name because that is the field being checked.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins enabled = {1};
+    }
+
+    // ============================================================================
+    // HYPERVISOR INSTRUCTION COVERPOINTS
+    // ============================================================================
+
+    hypervisor_instr: coverpoint ins.current.insn {
+        wildcard bins hlv_w = {32'b011010000000?????100?????1110011};
+        wildcard bins hlvx_w = {32'b011010000011?????100?????1110011};
```
davidharrishmc: renamed to hlvx_wu

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        wildcard bins hfence_gvma = {32'b0110001??????????000000001110011};
+    }
+
+    hlv_instructions: coverpoint ins.current.insn {
+        wildcard bins hlv_b  = {32'b011000000000?????100?????1110011};
+        wildcard bins hlv_bu = {32'b011000000001?????100?????1110011};
+        wildcard bins hlv_h  = {32'b011001000000?????100?????1110011};
+        wildcard bins hlv_hu = {32'b011001000001?????100?????1110011};
+        wildcard bins hlv_w  = {32'b011010000000?????100?????1110011};
+        wildcard bins hlv_wu = {32'b011010000001?????100?????1110011};
```
davidharrishmc: This should be in the XLEN64 ifdef.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        wildcard bins hlvx_hu = {32'b011001000011?????100?????1110011};
+        wildcard bins hlvx_wu = {32'b011010000011?????100?????1110011};
+        `ifdef XLEN64
+            wildcard bins hlv_d = {32'b011011000000?????100?????1110011};
+        `endif
+    }
+
+    hlv_instructions_three: coverpoint ins.current.insn {
```
davidharrishmc: better name?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins stval = {12'h143};
+    }
+
+    csrr_vstval_read: coverpoint ins.current.insn {
+        wildcard bins vstval_read = {32'b001001000011_00000_010_?????_1110011};
+    }
+
+    `ifdef RV32
+    rv32_instreth: coverpoint ins.current.insn {
+        wildcard bins instreth_read = {32'b11001000001000000010?????1110011};
```
davidharrishmc: underscores make this more readable.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+
+    // ============================================================================
+    // HSTATUS BIT COVERPOINTS (Extended)
+    // ============================================================================
+
+    hstatus_vtw: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hstatus", "vtw")) {
+        bins one = {1};
+    }
+
+    hstatus_vtsr: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hstatus", "vtsr")) {
```
davidharrishmc: It's a bit confusing that vtsr means vtsr = 1, but vtvm means vtvm = {0/1} and vtvm_on means vtvm = 1.  Can the naming be more consistent?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+
+    mstatus_tw: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "tw") == 0) {
+        bins zero = {0};
+    }
+
+    // ============================================================================
+    // COUNTER ENABLE COVERPOINTS (Extended)
+    // ============================================================================
+
+    s_counteren: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "scounteren", "counteren") & (1<<2)) {
```
davidharrishmc: rewrite 1<<2.  Also, all of these coverpoints should mention _ir because that is the bitfield.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+
+    s_counteren_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "scounteren", "counteren") & (1<<2)) {
+        bins disabled = {0};
+    }
+
+    h_counteren_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hcounteren", "counteren") & (1<<2)) {
+        bins disabled = {0};
+    }
+
+    `ifdef RV32
```
davidharrishmc: why are these specific to rv32?  Seems like this field exists for any xlen.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins zero = {0};
+    }
+    `endif
+
+    // ============================================================================
+    // ADDRESS ALIGNMENT COVERPOINTS
+    // ============================================================================
+
+    address_lsbs: coverpoint (ins.current.rs1_val + ins.current.imm)[2:0] {
+        bins addr_000 = {3'b000};
```
davidharrishmc: No need for these 8 lines.  They are automatically populated if omitted.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        bins addr_101 = {3'b101};
+        bins addr_110 = {3'b110};
+        bins addr_111 = {3'b111};
+    }
+
+    // ============================================================================
+    // EXCEPTION CODE COVERPOINTS
+    // ============================================================================
+
+    virtual_instruction_exception: coverpoint ins.current.exception_code {
```
davidharrishmc: Can this use ins.current.csr rather than ins.current.exception_code to be consistent with the others?

davidharrishmc: What is the point of virtual_instruction_exception vs. virtual_instruction?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+
+    virtual_instruction: coverpoint (ins.current.csr[12'h342][31:0] == 32'd22) {
+        // Auto fill 0/1
+    }
+
+    // ============================================================================
+    // VIRTUAL INSTRUCTION TRIGGERS (VU MODE)
+    // ============================================================================
+
+    virtual_instr_triggers_vu: coverpoint ins.current.insn {
```
davidharrishmc: Doesn't seem to be used anywhere.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    }
+
+    // ============================================================================
+    // DELEGATION REGISTER COVERPOINTS
+    // ============================================================================
+
+    // Machine Exception Delegation Register (medeleg)
+    medeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+        bins zeros = {32'h00000000};
+        wildcard bins walking_pattern = {32'b1111_0000_1011_1111_1111_111?};
```
davidharrishmc: Instead of calling this walking_pattern, how about "ones".  Same with next coverpoint.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        trap_to_vs;
+
+    cp_ecall_to_hs: cross 
+        ecall,
+        priv_mode_to_hs,
+        medeleg_ecall_enabled_u_vu,
+        medeleg_ecall_enabled_vs,
+        hedeleg_ecall_disabled_u_vu,
+        hedeleg_ecall_disabled_vs,
+        hstatus_spvp,
```
davidharrishmc: None of these crosses from hstatus_spvp downward seem to be needed.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        hstatus_gva,
+        sstatus_sie_after_trap,
+        sstatus_spie_correct,
+        trap_to_hs;
+
+    cp_ecall_to_m: cross 
+        ecall,
+        priv_mode_to_m,
+        medeleg_ecall_bits_disabled,
+        mstatus_gva,
```
davidharrishmc: None of these from mstatus_gva downward are needed.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        mstatus_mpie_correct,
+        mstatus_mie_after_trap,
+        trap_to_m;
+
+    cp_ebreak_to_m: cross 
+        ebreak,
+        priv_mode_to_m,
+        medeleg_ebreak_bits_disabled,
+        mstatus_gva,
```
davidharrishmc: None from mstatus_gva downward are needed.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        medeleg_ecall_enabled,
+        hedeleg_ecall_enabled,
+        vstvec_different_from_stvec,
+        not_using_stvec,
+        trap_to_vs;
+
+    // Priority crosses
+    cp_priority: cross 
+        hlv_hsv_instr,
+        illegal_address,
```
davidharrishmc: This should be tested with both legal and illegal addresses.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        addr_misalignment,
+        priv_mode_to_m,
+        hstatus_hu;
+
+    // Virtual instruction exception crosses - VS mode
+    cp_virtual_instr_vs_instret: cross h_counteren, m_counteren, csrr, instret, priv_mode_vs, virtual_instruction_exception;
```
davidharrishmc: virtual_instruction_exception is not needed on any of these coverpoints because it is an outcome, not a stimulus

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    cp_virtual_instr_vs_execute_hypervisor: cross hypervisor_instr, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_read_vstval_hval: cross csrr, vstval_htval, virtual_instruction_exception, priv_mode_vs, virtual_instruction_exception;
```
davidharrishmc: virtual_instruction_exception appears twice and isn't needed either time.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    cp_virtual_instr_vs_mstatus_satp: cross csrr, satp, vsatp, mstatus_tvm, priv_mode_vs, virtual_instruction_exception;
```
davidharrishmc: satp and vsatp can't both be hit in this cross.  You need a new xsatp coverpoint with the union, rather than the intersection.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    stval: coverpoint ins.current.insn[31:20] {
+        bins stval = {12'h143};
+    }
+
+    csrr_vstval_read: coverpoint ins.current.insn {
+        wildcard bins vstval_read = {32'b001001000011_00000_010_?????_1110011};
+    }
+
+    `ifdef RV32
+    rv32_instreth: coverpoint ins.current.insn {
```
davidharrishmc: Indent within `ifdefs

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        wildcard bins instreth_read = {32'b11001000001000000010?????1110011};
+    }
+
+    rv32_hedelegh: coverpoint ins.current.insn {
```
davidharrishmc: Asymmetric with rv32_instreth, one being 32 bits, the other 12 bits.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: Seems like all of these should have _ir in the name to indicate instret field

davidharrishmc: naming convention is funny, can't tell if the bits are 0 or 1 from the name.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    cp_virtual_instr_vs_read_vstval_hval: cross csrr, vstval_htval, virtual_instruction_exception, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_mstatus_satp: cross csrr, satp, vsatp, mstatus_tvm, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_wfi: cross wfi, hstatus_vtw, mstatus_tw, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_sret: cross sret, hstatus_vtsr, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_s_vma_instr: cross sfence_sinval_vma, hstatus_vtvm_on, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_satp: cross csrr, satp, hstatus_vtvm_on, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_rv32_instreth_mcounter: cross rv32_instreth, rv32_mcounter, rv32_hcounter, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_rv32_hedelegh: cross csrr, rv32_hedelegh, priv_mode_vs, virtual_instruction_exception;
+
+    // Virtual instruction exception crosses - VU mode
```
davidharrishmc: again, virtual_instruction_exception is not needed in crosses

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+
+    s_counteren: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "scounteren", "counteren") & (1<<2)) {
+        bins enabled = {1};
+    }
+
+    s_counteren_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "scounteren", "counteren") & (1<<2)) {
+        bins disabled = {0};
+    }
+
+    h_counteren_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hcounteren", "counteren") & (1<<2)) {
```
davidharrishmc: This is the same as h_counteren

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    cp_virtual_instr_vs_wfi: cross wfi, hstatus_vtw, mstatus_tw, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_sret: cross sret, hstatus_vtsr, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_s_vma_instr: cross sfence_sinval_vma, hstatus_vtvm_on, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_satp: cross csrr, satp, hstatus_vtvm_on, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_rv32_instreth_mcounter: cross rv32_instreth, rv32_mcounter, rv32_hcounter, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_rv32_hedelegh: cross csrr, rv32_hedelegh, priv_mode_vs, virtual_instruction_exception;
+
+    // Virtual instruction exception crosses - VU mode
+    cp_virtual_instr_vu_instret_1: cross csrr, instret, h_counteren, m_counteren, s_counteren, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_instret_2: cross csrr, instret, h_counteren_disabled, s_counteren_disabled, m_counteren, priv_mode_vu, virtual_instruction_exception;
```
davidharrishmc: test plan calls for hcounteren[2] = 1 here, not 0

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: Or do it as in cp_virtual_instr_vu_satp and cp_virtual_instr_vu_vsatp

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        addr_misalignment,
+        priv_mode_to_m,
+        hstatus_hu;
+
+    // Virtual instruction exception crosses - VS mode
+    cp_virtual_instr_vs_instret: cross h_counteren, m_counteren, csrr, instret, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_execute_hypervisor: cross hypervisor_instr, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_read_vstval_hval: cross csrr, vstval_htval, virtual_instruction_exception, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_mstatus_satp: cross csrr, satp, vsatp, mstatus_tvm, priv_mode_vs, virtual_instruction_exception;
+    cp_virtual_instr_vs_wfi: cross wfi, hstatus_vtw, mstatus_tw, priv_mode_vs, virtual_instruction_exception;
```
davidharrishmc: hstatus_vtw should not be part of this cross.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+
+    // Virtual instruction exception crosses - VU mode
+    cp_virtual_instr_vu_instret_1: cross csrr, instret, h_counteren, m_counteren, s_counteren, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_instret_2: cross csrr, instret, h_counteren_disabled, s_counteren_disabled, m_counteren, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_execute_h: cross hypervisor_instr, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_read_vstval_hval: cross csrr, vstval_htval, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_read_stval: cross csrr, stval, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_satp: cross csrr, satp, mstatus_tvm, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_vsatp: cross csrr, vsatp, mstatus_tvm, priv_mode_vu, virtual_instruction_exception;
+    cp_virtual_instr_vu_wfi: cross wfi, hstatus_vtw, mstatus_tw, priv_mode_vu, virtual_instruction_exception;
```
davidharrishmc: hstatus_vtw should not be part of this cross.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+        priv_mode_all,
+        hstatus_hu;
+
+    cp_hfence_priv: cross 
+        hfence_instructions,
+        priv_mode_to_m,
+        mstatus_tvm,
+        hstatus_vtvm;
+
+    // GVA (Guest Virtual Address) crosses - GVA should be set for address-related exceptions
```
davidharrishmc: Let's delete all of these cp_gva coverpoints because they are captured in cp_hedeleg.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,887 @@
...
+    cp_gva_load_access_fault: cross loadops, illegal_address, modes, hstatus_gva_set;
+    cp_gva_load_page_fault: cross loadops, load_page_fault, modes, hstatus_gva_set;
+    cp_gva_load_guest_page_fault: cross loadops, load_guest_page_fault, modes, hstatus_gva_set;
+    cp_gva_store_misaligned: cross storeops, adr_LSBs, modes, hstatus_gva_set;
+    cp_gva_store_access_fault: cross storeops, illegal_address, modes, hstatus_gva_set;
+    cp_gva_store_page_fault: cross storeops, store_page_fault, modes, hstatus_gva_set;
+    cp_gva_store_guest_page_fault: cross storeops, store_guest_page_fault, modes, hstatus_gva_set;
+    cp_gva_breakpoint: cross ebreak, modes, hstatus_gva_set;
+
+    // GVA should be clear for non-address-related exceptions
```
davidharrishmc: Delete these too.

davidharrishmc (changes requested):
@SadhviNarayanan this is a massive PR.  Excellent material.  It took a long time to review carefully.  

I have not reviewed the Interrupts code, but presume it is the same as before but fixed to compile.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+// Written: sanarayanan@hmc.edu       17 Nov 2025
+//
+// Purpose: Functional coverage tests for User-mode Interrupts
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+///////////////////////////////////////////
+
+
+##### START_TEST_CONFIG #####
+# implemented_extensions: [I, S, Zicsr]
```
davidharrishmc: S should be U

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+# params:
+#   M_MODE_ENDIANNESS: dynamic
```
davidharrishmc: Why is this needed?

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+# MARCH: rv${XLEN}i_zicsr
+# CONFIG_DEPENDENT: true
```
davidharrishmc, line 15: I don't remember what this is and why it is needed.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+##### END_TEST_CONFIG #####
+
+#define rvtest_mtrap_routine
+#define rvtest_strap_routine
```
davidharrishmc: Not needed for U because it should never trap to S.

tests/priv/InterruptsU/InterruptsU.S
davidharrishmc: S->U

tests/priv/InterruptsU/InterruptsU.S
davidharrishmc: What is this and why is it needed here?

tests/priv/InterruptsU/InterruptsU.S
davidharrishmc, line 15: What is this and is it needed?

tests/priv/InterruptsU/InterruptsU.S
davidharrishmc: strap and strap shouldn't be needed in U

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+#define rvtest_dtrap_routine
+
+#define SIGUPD_COUNT 402
```
davidharrishmc, line 23: Fragile.  Is there a better way to do this?

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+
+
+# # Test code region
+# .section .text.init
+# .globl rvtest_entry_point
+# rvtest_entry_point:
+RVMODEL_BOOT
+RVTEST_BEGIN
+
+    ALL_MEM_PMP          		                  	                      	// set the PMP permissions for the whole memory
```
davidharrishmc: Too much indenting to comments makes them wrap funny.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+    la sp, scratch
+    //jal reset_mtimecmp
+    RVMODEL_CLR_MTIMER_INT
+
+
+
+    /////////////////////////////////
+    // cp_user_mti
+    // Undelegated MTI Interrupt
+    // Cross of mtvec.MODE={00/01}, mstatus.MIE = {0/1}, mstatus.SIE = {0/1}, mie.MTIE = 1, set MTIMECMP
```
davidharrishmc: SIE is no longer part of the coverpoint and should be removed from the comment.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+
+    ALL_MEM_PMP          		                  	                      	// set the PMP permissions for the whole memory
+	csrw satp, zero  		                                        		// write satp with all zeros (bare mode)
+    
+    //la sp, topofstack
+    la sp, scratch
+    //jal reset_mtimecmp
+    RVMODEL_CLR_MTIMER_INT
+
+
```
davidharrishmc, line 64: Missing a part to write mideleg=0

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+            andi t0, s0, 2 // Check bit 2
+            beqz t0, skip_mstatus_mie
+            csrrsi t6, mstatus, 8 // Set MIE
+            skip_mstatus_mie:
+
+            // 4. Set mie.MTIE = 1 (Bit 1 of s0 is unused for this test)
+            li t0, 0x080 // MTIE bit
+            csrrsi t6, mie, t0
+
+            // can we execute the test here? umode?
```
davidharrishmc: Yes, need to be in U-mode.  Why the questions in the comments.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+    /////////////////////////////////
+    // cp_user_mti
+    // Undelegated MTI Interrupt
+    // Cross of mtvec.MODE={00/01}, mstatus.MIE = {0/1}, mstatus.SIE = {0/1}, mie.MTIE = 1, set MTIMECMP
+    // Interrupt should trap to M-mode, regardless of MIE/SIE.
+    /////////////////////////////////
+
+    cp_user_mti:
+
+        li s0, 31 // s0 = 5'b11111 (Mie/Sie x MtvecMode x MtvecMode) -> 32 combinations
```
davidharrishmc: li s0, 3. 
Seems like only 2x2 combinations are needed: moved.MODE and mstatus.MIE.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+        
+        for_mti_cross:
+            // Clear MIE/SIE and MTVEC
+            li t0, 0x00000008 // MIE=1, SIE=1 bits
```
davidharrishmc: SIE should not be part of this.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+            csrrci t6, mstatus, t0
+            csrrci t6, mtvec, 3
+
+            // 1. Set MTVEC Mode (Bit 1 of s0)
```
davidharrishmc: This should be called bit 0.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+            andi t0, s0, 1 // Check bit 1
+            beqz t0, skip_mtvec_vectored
+            csrrsi t6, mtvec, 1 // Set MTVEC to vectored (01)
+            skip_mtvec_vectored:
+
+            // 2. Set MSTATUS.MIE (Bit 2 of s0) --> like a mask
```
davidharrishmc: called bit 1

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+            andi t0, s0, 2 // Check bit 2
+            beqz t0, skip_mstatus_mie
+            csrrsi t6, mstatus, 8 // Set MIE
+            skip_mstatus_mie:
+
+            // 4. Set mie.MTIE = 1 (Bit 1 of s0 is unused for this test)
```
davidharrishmc: Delete comment about s0.  Doesn't seem relevant.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,326 @@
...
+            // can we execute the test here? umode?
+            RVTEST_GOTO_LOWER_MODE Umode
+
+            // Trigger the interrupt
+            RVMODEL_CAUSE_MTIMER_INT_NOW
+            nop // Allow time for the trap to occur
+            nop
+            nop
+
+            // clear the interrupt
```
davidharrishmc: Can you clear the MTIMER_INT if you aren't back in M-mode now?
You won't be able to write the m CSRs in the next test if you aren't back in M-mode.

davidharrishmc (changes requested):
Reviewed through cp_user_mti.  Apply these comments to the later coverpoints, and try running the tests, and then let me know when they are all ready to review.

tests/priv/InterruptsU/InterruptsU.S
SadhviNarayanan: This is for tests that will have diferent results depending on the DUT it is running on. So, every priv test needs to be true as it is dependent on the DUT.

tests/priv/InterruptsU/InterruptsU.S
SadhviNarayanan: We don't particularly need this param for InterruptsU - I think I just ported it over from a different test and didn't change this. It wouldn't hurt to have it, but it not needed.

tests/priv/InterruptsU/InterruptsU.S
SadhviNarayanan: This might take some more time to find a less fragile approach. I can discuss it with Jordan, and when we have a better way I can update it!

tests/priv/InterruptsU/InterruptsU.S
jordancarlin: That's not quite true. Requiring `M_MODE_ENDIANNESS: dynamic` for the `InterruptsU` test means this test will not be selected for DUTs that only support little endian (which is a very common choice), so it definitely needs to be removed since endianness should not impact whether or not the interrupts tests run.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+
+
+   illegalops: coverpoint ins.current.insn {
+       bins zeros = {'0};
+       bins ones  = {'1};
+   }
+
+
+   hlv_hsv_instr: coverpoint ins.current.insn {
+       wildcard bins hlv_w = {32'b011010000000_?????100?????_1110011};
```
davidharrishmc: Some more underscores would make this more readable.  See csrr using underscores between fields.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+   }
+
+
+   csrr: coverpoint ins.current.insn {
+       wildcard bins csrr = {32'b????????????_00000_010_?????_1110011};
+   }
+
+
+   instret: coverpoint ins.current.insn[31:20] {
+       wildcard bins instret_read = {12'hc02};
```
davidharrishmc: No need for wildcard here.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+   // ============================================================================
+
+
+   illegal_address: coverpoint ins.current.imm + ins.current.rs1_val {
+       bins illegal = {`ACCESS_FAULT_ADDRESS};
+   }
+
+
+   address_legality: coverpoint ins.current.imm + ins.current.rs1_val {
+       bins legal = {[0:`ACCESS_FAULT_ADDRESS-1]};
```
davidharrishmc: I don't think this address is legal.  Let's just use true/false of imm+rs1_val == `ACCESS_FAULT_ADDRESS.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+
+   imm_bit_1: coverpoint ins.current.imm[1] {
+       bins one = {'1};
+   }
+
+
+   // ============================================================================
+   // PRIVILEGE MODE COVERPOINTS
+   // ============================================================================
+   mode_virt_cp: coverpoint mode_virt {
```
davidharrishmc: mode_virt isn't defined.  Can you add a v bit to the instruction data structure, similar to ins.prev.mode?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+       bins zero = {0};
+       bins one  = {1};
+   }
+
+
+   // All 5 privilege modes (M/HS/VS/VU/U)
+   modes: coverpoint {ins.prev.mode, mode_virt} {
```
davidharrishmc: If you switch the order {mode_virt, ins.prev.mode, then the V matches the use in the name.  Also, the bin can be written as M_mode = {3'b011}.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+   // Previous modes for delegation to HS (U, VS, VU)
+   priv_mode_to_hs: coverpoint {ins.prev.mode, mode_virt} {
+       bins U_mode  = {2'b00, 1'b0};
+       bins VS_mode = {2'b01, 1'b1};
+       bins VU_mode = {2'b00, 1'b1};
+   }
+
+
+   // Previous modes for delegation to M (all 5 modes)
+   priv_mode_to_m: coverpoint {ins.prev.mode, mode_virt} {
```
davidharrishmc: how does this differ from the "modes" coverpoint and why does V=? in M_mode (rather than 0)?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b1111_0000_1011_1111_1111_111?};
+   }
+
+
+   // Hypervisor Exception Delegation Register (hedeleg)
+   hedeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hedeleg", "deleg") {
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b1111_0000_1011_1111_1111_111?};
+       bins walk_1_bit0 = {32'h00000001}; // Instruction address misaligned
```
davidharrishmc: This bin should only exist if misaligned accesses are supported (ZICCLSM_SUPPORTED).  Use `ifdef

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+   medeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b1111_0000_1011_1111_1111_111?};
+   }
+
+
+   // Hypervisor Exception Delegation Register (hedeleg)
+   hedeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hedeleg", "deleg") {
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b1111_0000_1011_1111_1111_111?};
```
davidharrishmc: This turns out to be impossible per priv spec 21.2.2.  I've updated the testplan.  Please change this number to
0000_11?0_1?11_0001_1111_111?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+       bins M_mode  = {2'b11, 1'b?};
+       bins HS_mode = {2'b01, 1'b0};
+       bins U_mode  = {2'b00, 1'b0};
+       bins VS_mode = {2'b01, 1'b1};
+       bins VU_mode = {2'b00, 1'b1};
+   }
+
+
+   // Previous modes for VS (VU and VS)
+   prove_mode_to_vs: coverpoint {ins.prev.mode, mode_virt} {
```
davidharrishmc: prove -> priv

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,798 @@
...
+       bins VU_mode = {2'b00, 1'b1};
+       bins VS_mode = {2'b01, 1'b1};
```
davidharrishmc: VS_mode bin is unreachable per priv spec 21.2.2.  Deleted this from the testplan, and take it out of coverpoint.

davidharrishmc:
model_test may have a conflict with the act4 branch.  Check if you really want to be changing it.

I don't think this will compile until you have the v bit in the RVVI interface.
When it compiles and has no merge conflicts, it's ready to merge.

davidharrishmc (commented):
test

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+
+
+  /////////////////////////////////
+  // cp_user_mti
+  // Undelegated MTI Interrupt
+  // Cross of mtvec.MODE={00/01}, mstatus.MIE = {0/1}, mie.MTIE = 1, set MTIMECMP
+  // Interrupt should trap to M-mode, regardless of MIE.
+  /////////////////////////////////
+
+
```
davidharrishmc, line 48: Avoid excessive white space.  Each space should have purpose and consistency

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+  cp_user_mti:
+
+
+
+
+      li s0, 3 // -> 4 combinations
+    
+      for_mti_cross:
+          // Clear MIE and MTVEC
```
davidharrishmc, line 59: MIE -> mstatus.MIE would be clearer because there is also a mie register.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+          // 1. Set MTVEC Mode (Bit 0 of s0)
+          andi t0, s0, 1 // Check bit 0
+          beqz t0, skip_mtvec_vectored
+          csrrsi t6, mtvec, 1 // Set MTVEC to vectored (01)
+          skip_mtvec_vectored:
+
+
+
+
+          // 2. Set MSTATUS.MIE (Bit 1 of s0) --> like a mask
```
davidharrishmc, line 76: What is "like a mask" for?

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+
+
+
+          RVTEST_GOTO_MMODE
+          nop
+
+
+
+
+          RVMODEL_CLR_MSIP
```
davidharrishmc, line 189: RVMODEL_CLR_MSW_INT

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+
+          // Set MIE (Bit 1 of s2)
+          andi t0, s2, 2
+          beqz t0, case_TW_wfi
+          csrrsi t6, mstatus, 8
+
+
+
+
+      case_TW_wfi:
```
davidharrishmc, line 321: Earlier you had used a skip rather than a case label convention and had put the label right after the block being skipped.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+      case_end_wfi:      
+          // 1. Set mie.MTIE = 1
+          li t0, 0x080
+          csrrw t6, mie, t0
+
+
+
+
+          // 2. Set MTIMECMP to interrupt soon
+          RVMODEL_CAUSE_MTIMER_INT_SOON
```
davidharrishmc, line 340: Put this after going to U mode because it takes a while to go to U mode and the interrupt might fire before you get there.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+          wfi
+          nop
+
+
+
+
+          RVTEST_GOTO_MMODE
+        
+          // trap handler should interrupt + return back to the instruction WFI.
+          // if TW=1 and MIE=0/1, it shld still trap on pending interrupt
```
davidharrishmc, line 361: should

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+
+
+          // Clean up
+          RVMODEL_CLR_MTIMER_INT
+
+
+
+
+          addi s2, s2, -1
+          bge s2, zero, for_wfi
```
davidharrishmc, line 373: Earlier you used bgez.  Might as well be consistent.

tests/priv/InterruptsU/InterruptsU.S
davidharrishmc: Does this affect other files?

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+
+
+      // Force TW=1 for the whole test block
+      li t0, 0x200000 // TW bit
+      csrrs t6, mstatus, t0
+
+
+
+
+      for_wfi_timeout:
```
davidharrishmc, line 405: There are lots of inconsistencies in the coding. If you inherited them from InteruptsS, it is ok to live with them for now, but let's clean up when we unroll the loops.  For example, the skip vs case labels.  The order of bits in the inner loops is opposite here vs. cp_user_m*i.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+
+
+
+
+          li t0, 0x00000008
+          csrrc t6, mstatus, t0 // Clear MIE
```
davidharrishmc, line 411: This could be a csrrci to be consistent with 419.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+          andi t0, s2, 2
+          beqz t0, case_MTIE_timeout
+          csrrsi t6, mstatus, 8
+
+
+
+
+      case_MTIE_timeout:
+          // Set mie.MTIE (Bit 0 of s2)
+          andi t0, s2, 1
```
davidharrishmc, line 426: The other loops always clear the bit and then sometimes set it.  This loop either sets or clears.  Inconsistent style.

tests/priv/InterruptsU/InterruptsU.S
```diff
@@ -0,0 +1,508 @@
...
+          nop
+          wfi
+          nop
+
+
+
+
+          RVTEST_GOTO_MMODE
+        
+          // will trap to M-mode with cause=2 (illegal instruction) happen
```
davidharrishmc, line 459: This comment belongs before the GOTO_MMODE.

davidharrishmc (changes requested):
One or two possible bugs and some comments on style.  I know you inherited sloppy Interrupt code.  Looking good.  Does it run and are you getting some coverage?

jordancarlin:
@SadhviNarayanan is this PR still relevant? If not, can you please close it.

PR #749, opened 2025-11-19
Vector coverage

georgiatai:
Vector scalar coverage fully running (including sew=64). Have not tested running the whole suite (including Vf and Vls) and how long that takes.

Would like some feedback on the fix on RVTEST_SIGUPD macro. This change is made due to the misalignment of SEW64, but I'm not sure if this is the cleanest way to solve this issue without modifying to much of the general structure.


Makefile
```diff
@@ -107,10 +114,19 @@ lint:
...
+vector-tests: covergroupgen vector-testgen
+
+.PHONY: generate-makefiles-ref-vector
+generate-makefiles-ref-vector: # too many dependencies to track; always regenerate Makefile
+	$(MAKE) vector-tests
+	$(UV_RUN) act $(REF_CONFIG_FILES) --workdir $(WORKDIR_REF) --test-dir $(TESTDIR) --coverage
+
+.PHONY: vector-coverage
+vector-coverage: generate-makefiles-ref-vector Makefile
+	$(MAKE) -C $(WORKDIR_REF) coverage
```
jordancarlin: Why are these targets needed? Can't you run `make vector-tests` and then use the regular `make coverage` target?

generators/tests/scripts/vector_testgen_common.py
```diff
@@ -0,0 +1,2230 @@
+##################################
+# vector_testgen_common.py
+#
+# James Kaden Cassidy kacassidy@hmc.edu 25 Jun 2025
+# SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
```
jordancarlin: Update license for all files to match the new license being used in this repo.

generators/tests/testgen/src/testgen/templates/testgen_footer_vector1.S
jordancarlin: This was removed from the regular footer. Any reason to retain it in the vector version?

generators/tests/testgen/src/testgen/templates/testgen_footer_vector2.S
jordancarlin, line 1: Why do we need two footers for vector?

generators/tests/testgen/src/testgen/templates/testgen_header_vector.S
jordancarlin: These should be defined in the `rvtest_config.h` files that are specific to each config. Eventually that file will be autogenerated, but for now we are manually maintaining them. They are the in `configs` directory.

jordancarlin: Does it make more sense for this chunk to be in one of the `.h` header files? I'm thinking it might be possible to eliminate the need for separate `testgen_header.S` files for vector that way, which will make it easier to maintain.

tests-dev/priv/ZicsrV.S
```diff
@@ -0,0 +1,1384 @@
...
+
+// Define special purpose registers before including test_macros_vector.h
+#define DATA_BASE x4
+#define SIG_BASE x3
+#define VLENB_CACHE x5
+#define HELPER_GPR x6
+
+#include "test_macros_vector.h"
+
+RVTEST_ISA("RV32IMV_Zicsr, RV64IMV_Zicsr")
```
jordancarlin: Looks like this is still using the ACT 3.0 framework format?

tests-dev/priv/ZicsrV.S
```diff
@@ -0,0 +1,1384 @@
...
+    // attempt read/write/clear on all vector CSRs
+    // Signature: CSR values after execution
+    ///////////////////////////////
+    cp_csrrswc:
+        li    x2, -1             // Create a all 1s mask
+
+        // Testing CSR vtype: read-only, can only be updated by vset{i}vl{i}
+        csrr  x1, vtype	         // Read CSR
+        nop
+        RVTEST_SIGUPD(x3, x1)
```
jordancarlin: This is using the old version of `RVTEST_SIGUPD`. The new version takes 4 registers.

tests/env/signature.h
```diff
@@ -19,7 +19,15 @@
...
 //  _R - Register containing value to store/compare
-#ifdef SELFCHECK
+#ifdef SEW64
+  #define RVTEST_SIGUPD(_SIG_BASE, _LINK_REG, _TEMP_REG, _R)  \
+    LREG _TEMP_REG,0(_SIG_BASE)                            ;\
+    beq _TEMP_REG, _R, 1f                                  ;\
+    jal _LINK_REG, failedtest_##_LINK_REG##_##_TEMP_REG    ;\
+    nop                                                    ;\
+    1:                                                     ;\
+    addi _SIG_BASE, _SIG_BASE, 8
```
jordancarlin: Instead of all of this, can we just define something like `SIGWIDTH` as the max of SEW and XLEN and then use that in the `RVTEST_SIGUPD` macro?

jordancarlin (changes requested):

Makefile
georgiatai: The coverage target would generate normal test also (shown below), so I thought I'd keep them separate for now just for the ease of running purely vector.
```.PHONY: generate-makefiles-ref
generate-makefiles-ref: # too many dependencies to track; always regenerate Makefile
	$(MAKE) tests
	$(UV_RUN) act $(REF_CONFIG_FILES) --workdir $(WORKDIR_REF) --test-dir $(TESTDIR) --coverage

.PHONY: coverage
coverage: generate-makefiles-ref Makefile
	$(MAKE) -C $(WORKDIR_REF) coverage```

File: configs/ref/sail-rv32gcv/sail-rv32gcv.yaml [hunk 6]
```diff
@@ -0,0 +1,296 @@
+# yaml-language-server: $schema=../../../external/riscv-unified-db/spec/schemas/config_schema.json
+---
+$schema: config_schema.json#
+kind: architecture configuration
+type: fully configured
+name: sail-rv32gc
```
jordancarlin: Name needs to be updated. Have you run with this config yet? UDB usually throws an error for this.

riscv-arch-test
jordancarlin, line 1: You seem to have added `riscv-arch-test` as a submodule of `riscv-arch-test`? Not sure what happened, but this needs to be reverted.

configs/ref/sail-rv32gcv/rvtest_config.h
```diff
@@ -1,3 +1,7 @@
 #define ACCESS_FAULT_ADDRESS 0x00000000
 #define RVMODEL_PMP_GRAIN 4
 #define RVMODEL_NUM_PMPS 16
+
+#define SEWMIN 8
+#define ELEN 64
+#define MAXINDEXEEW 64
```
jordancarlin: Use consistent parameter names with what @jacassidy is adding to UDB in https://github.com/riscv-software-src/riscv-unified-db/pull/1245.

tests/env/utils.h
```diff
@@ -32,6 +32,16 @@
...
 
+# SEW signature update stride
+#ifndef SEW
+  #define SEW 0
+#endif
+#if (SEW > XLEN)                // max(SEW, XLEN)
+  #define SIG_STRIDE (SEW / 8)
+#else
+  #define SIG_STRIDE (XLEN / 8) 
+#endif
```
jordancarlin: I think this should work:

```suggestion
#define SIG_STRIDE (MAX(SEW, XLEN) / 8)
```

tests/env/utils.h
```diff
@@ -32,6 +32,16 @@
 #define REGWIDTH (XLEN>>3)      // in units of #bytes
 #define ALIGNSZ ((XLEN>>5)+2)   // log2(XLEN): 2,3,4 for XLEN 32,64,128
 
+# SEW signature update stride
```
jordancarlin: Should we modify this so that it only uses `SEW` to calculate `SIG_STRIDE` if it is a vector test? Maybe add a `#define RVTEST_VECTOR` to the beginning of all vector tests and then use an `#ifdef` here.

Makefile
jordancarlin: Discussed in person. Going to drop these targets.

tests/env/utils.h
jordancarlin: Let's move the `SIG_STRIDE` calculation to `signature.h` so that all signature related logic is in the same file.

tests/env/test_macros_vector.h
```diff
@@ -40,35 +18,115 @@
...
+  addi _SIG_BASE, _SIG_BASE, OFFSET
+
+
+/************************************* RVTEST_SIG_SETUP ************************************/
+/**** RVTEST_SIG_SETUP creates signature region to support self-checking tests          ****/
+/**** - Main signature region for results from test, initialized with correct values    ****/
+/****   for self-checking                                                               ****/
+/**** - Trap handler signature region                                                   ****/
+/*******************************************************************************************/
+.macro RVTEST_SIG_SETUP_V
```
jordancarlin, line 86: How does this differ from the regular `RVTEST_SIG_SETUP`? I don't see anything that warrants a separate version.

tests/env/test_macros_vector.h
```diff
@@ -7,29 +7,7 @@
...
-#ifndef DATA_BASE
-# error "DATA_BASE"
-#endif // DATA_BASE
-#ifndef VLENB_CACHE
-# error "VLENB_CACHE is not defined"
-#endif // VLENB_CACHE
-#ifndef HELPER_GPR
-# error "HELPER_GPR is not defined"
-#endif // HELPER_GPR
+#include "riscv_arch_test.h"
```
jordancarlin: Why can't this file be included in `riscv_arch_test.h` like all of the other headers? I don't see a reason to reverse the `#include` flow for vector.

configs/ref/sail-rv32gcv/sail-rv32gcv.yaml
```diff
@@ -0,0 +1,296 @@
+# yaml-language-server: $schema=../../../external/riscv-unified-db/spec/schemas/config_schema.json
+---
+$schema: config_schema.json#
+kind: architecture configuration
+type: fully configured
+name: sail-rv32gcv
+description: The rv32gc configuration of Core-V Wally;
```
jordancarlin: Update description. This is not a Wally config.

generators/tests/scripts/vector_testgen_common.py
```diff
@@ -0,0 +1,2317 @@
...
+        .replace("@SEW@", str(sew))
+        .replace("@TEST_DATA@", test_data)
+        .replace("@TEST_FILE_NAME@", f"{test}.S")
+        .replace("@SIGUPD_COUNT_FROM_TESTGEN@", str(500)) # TODO: change this to a dynamic value
+        .replace("@VLEN@", str(maxVLEN)) # TODO: make this configurable
+        .replace("@ELEN@", str(maxELEN)) # TODO: make this configurable
+        .replace("@SEWMIN@", str(minSEW_MIN)) # TODO: make this configurable
+        .replace("@CONFIG_DEPENDENT@", "false")  # TODO: Make this configurable for some tests (e.g. Zimop)
+        .replace("RVTEST_SIG_SETUP", "RVTEST_SIG_SETUP_V")
+        .replace("RVTEST_BEGIN", "RVTEST_BEGIN\nRVTEST_V_ENABLE(x5, x6)")
```
jordancarlin: Instead of replacing things in the template, just add `RVTEST_V_ENABLE(x5, x6)` to the `RVTEST_BEGIN` macro and put it inside of an `#ifdef` that checks for vector tests.

generators/tests/scripts/vector_testgen_common.py
```diff
@@ -0,0 +1,2317 @@
...
+        .replace('''#include "riscv_arch_test.h"''', f'''#define SEW {str(sew)}\n#include "test_macros_vector.h"''')
```
jordancarlin: Isn't `SEW` already defined in `rvtest_config.h`? Why does it need to be defined here too?

tests-dev/priv/ZicsrV.S
georgiatai: seems like all priv tests currently do not have rvtest_sigupd. should they be commented out/deleted in this case, before we get the loops working?

jordancarlin (changes requested):
Getting closer. Here's another round of feedback.

tests-dev/priv/ZicsrV.S
jordancarlin: Loops shouldn't be an issue because we are adding to the pointer directly now. I think if you update the macro calls to give the correct list of registers, it should work. This is in `tests-dev` though, so it won't be run for now and we can defer this if you want.

generators/tests/scripts/vector_testgen_common.py
georgiatai: SEW_MIN is defined in the config. This SEW is test specific, more so to determine the SIGUPD stride amount or related stuff

tests/env/test_macros_vector.h
georgiatai: There is an extra .align 3 before signature_base, which some tests, SEW of 64 in RV32 if i remember correctly, would fail without. If it is fine to add it into the normal `RVTEST_SIG_SETUP`, I can remove the _V version

generators/tests/scripts/vector_testgen_common.py
jordancarlin: Let's name it `RVTEST_SEW` or something like that.

tests/env/test_macros_vector.h
jordancarlin: You can add the `.align` to the regular `RVTEST_SIG_SETUP` macro. It's better to avoid proliferation of macros.

georgiatai:
Currently Vx and Vls are mostly runable. Vx specifically fails on vsadd.vi, vnclipu.wv, and vmv.x.s. Each of them has different issue, for example vnclipu.wv has something that is writing to the failure scratch which I am still trying to track down, and vmv.x.s has a trace failure. Vls is running but missing coverage, I am assuming it's because of the macro name change to fit with udb, so I will also work on that.

configs/ref/sail-rv32gcv/sail-rv32gcv.yaml
```diff
@@ -0,0 +1,296 @@
+# yaml-language-server: $schema=../../../external/riscv-unified-db/spec/schemas/config_schema.json
+---
+$schema: config_schema.json#
+kind: architecture configuration
+type: fully configured
+name: sail-rv32gcv
+description: The rv32gc configuration with the "V" (vector) for testing vector tests;
```
jordancarlin:
```suggestion
description: An rv32gcv configuration for testing vector tests
```

ctp/docs-resources
jordancarlin, line 1: You shouldn't be updating this submodule.

external/riscv-unified-db
jordancarlin, line 1: You shouldn't be updating this submodule.

generators/tests/scripts/vector_testgen_common.py
```diff
@@ -0,0 +1,2344 @@
...
+        .replace("@EXTENSION_LIST@", f"{['I'] + ext_parts_no_I}")
+        .replace("@MARCH@", march.lower())
+        .replace("@XLEN@", str(xlen))
+        .replace("@SEW@", str(sew))
+        .replace("@TEST_DATA@", test_data)
+        .replace("@TEST_FILE_NAME@", f"{test}.S")
+        .replace("@SIGUPD_COUNT_FROM_TESTGEN@", str(50000)) # TODO: change this to a dynamic value
+        .replace("@VLEN@", str(maxVLEN)) # TODO: make this configurable
+        .replace("@ELEN@", str(maxELEN)) # TODO: make this configurable
+        .replace("@SEWMIN@", str(minSEW_MIN)) # TODO: make this configurable
```
jordancarlin, line 1206: Please review your `.replace` commands. I think these ones can be dropped.

tests/env/signature.h
```diff
@@ -3,6 +3,25 @@
 # Jordan Carlin jcarlin@hmc.edu October 2025
 # SPDX-License-Identifier: Apache-2.0
 
+// RVTEST_SIGBASE(sigbase, label) initializes the sigbase register to the
+// address of the signature region.
+// _SIG_BASE - Base register for signature region
+// _TAG - Tag to identify signature region
+#define RVTEST_SIGBASE(_SIG_BASE,_TAG)      ;\
+  LA(_SIG_BASE,_TAG)
```
jordancarlin, line 11: This was dropped in favor of just using the `LA` macro directly.

tests/env/utils.h
jordancarlin: @georgiatai Did this not work? I see you changed this section but you still aren't using the `MAX` macro.

tests/env/test_macros_vector.h
```diff
@@ -7,29 +7,6 @@
 #define RISCV_TEST_SUITE_TEST_MACROS_VECTOR_H
 
 #include "model_test.h"
```
jordancarlin, line 9: This is already included in `riscv_arch_test.h`.

tests/env/test_macros_vector.h
jordancarlin: After you drop this, it seems like you can probably get rid of this file entirely. Move `RVTEST_V_ENABLE` to `utils.h`, `RVTEST_SIGUPD_V` to `signature.h`.

jordancarlin (changes requested):

jordancarlin:
Superseded by #775.

PR #765, opened 2025-12-03
Fix save area delta calculation with virtualization

youngar:

## Description

When calculating the save area offset, the code tries to determine the illegal flag combinations according to the table in the comments above. To do this, it is supposed to form a 3 bit number from the (vMPP,H,GVA) flags.  This code had several problems which are now fixed:

1. The H flag, which indicates if the hypervisor is enabled, was incorrectly stored in bit 0 instead of bit 1, overwriting the GVA flag.
3. Incorrectly using `#if` instead of `#ifdef` to check for the `rvtest_vtrap_routine` macro, causing the code to be permanently disabled.
4. Incorrect comment explaining the illegal bit patterns, although an earlier comment in the code had the correct patterns. The constant 0x5D is actually the correct constant to use.
5. Typographic error spelling the instruction `srli` as `slri`.

### Related Issues

N/A

- [ ] Ratified
- [ ] Unratified

### List Extensions

### Reference Model Used

- [ ] SAIL
- [ ] Spike

### Mandatory Checklist:

### Optional Checklist:

### DIFF
--- a/riscv-test-suite/env/arch_test.h
+++ b/riscv-test-suite/env/arch_test.h
@@ -1337,7 +1337,7 @@ common_\__MODE__\()excpt_handler:
   // can use T3, T6 because relocation will overwrite them
   //********************************************************************************
         
-        // create an index from these values: vMPP, x.GVA , H-ext
+        // create an index from these values: vMPP, H-ext, x.GVA
         // where vMPP = m.PRV ? svedMPP : m.MPP
         
   .ifc \__MODE__ ,  M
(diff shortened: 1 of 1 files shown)


allenjbaum:
Thanks for fixing this (and others)! We never got around to testing it because of missing support in reference models (much less missing tests....)

UmerShahidengr:
Thanks alot for the fix, this calculation has been integrated in the new trap handler which is trying a whole different logic to calculate the save area delta at various locations. Closing this one.

PR #769, opened 2025-12-07
[ACT 4] Update `RVMODEL_IO_WRITE_STR` macro and add coverpoints to debug message

jordancarlin:
- Add additional temporary registers to `RVMODEL_IO_WRITE_STR` macro
- Inline all printing code
- Add testcase number and coverpoint to debug messages in test data section

### DIFF
--- a/configs/duts/cvw/cvw-rv64gc/model_test.h
+++ b/configs/duts/cvw/cvw-rv64gc/model_test.h
@@ -1,15 +1,6 @@
 #ifndef _COMPLIANCE_MODEL_H
 #define _COMPLIANCE_MODEL_H
 
-# Expects a PC16550-compatible UART.
-# Change these addresses to match your memory map
-.EQU UART_ENABLED, 1                     # set to 0 to not print
-.EQU UART_BASE_ADDR, 0x10000000
-.EQU UART_THR, (UART_BASE_ADDR + 0)
-.EQU UART_RBR, (UART_BASE_ADDR + 0)
-.EQU UART_LCR, (UART_BASE_ADDR + 3)
-.EQU UART_LSR, (UART_BASE_ADDR + 5)
-
 #define RVMODEL_DATA_SECTION \
         .pushsection .tohost,"aw",@progbits;                \
         .align 8; .global tohost; tohost: .dword 0;         \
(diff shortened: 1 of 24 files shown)


configs/duts/cvw/cvw-rv64gc/model_test.h
davidharrishmc: What is UART_ENABLED?  Is this a bit that should be set after RVMODEL_IO_INIT is called, or something else?  I think Wally doesn't really need it.

davidharrishmc: Label 1 missing.

configs/duts/cvw/cvw-rv64gc/model_test.h
```diff
@@ -44,18 +44,29 @@
...
-  lbu a0, 0(x30)             ; /* Load byte */       \
-  beqz a0, 2f                ; /* Exit if null */    \
-  call uart_putc             ; /* Print char */      \
-  addi x30, x30, 1           ; /* Next char */       \
-  j 1b                       ; /* Loop */            \
-2:
+  lbu _R1, 0(_STR_PTR)        ;/* Load byte */        \
+  beqz _R1, 3f                ;/* Exit if null */     \
+2: /* uart_putc */           ;                      \
+  li _R2, UART_ENABLED   ;\
```
davidharrishmc: Do we need UART_ENABLED?

davidharrishmc: And does it need to be in every iteration of the loop, if we do have it?

davidharrishmc (changes requested):

configs/duts/cvw/cvw-rv64gc/model_test.h
jordancarlin: Carryover from earlier UART implementation (that I believe you wrote). Removed it because we always need to print the log messages.

configs/duts/cvw/cvw-rv64gc/model_test.h
jordancarlin: I'm surprised this didn't cause a compilation error. Fixed.

PR #775, opened 2025-12-10
[ACT 4] vector (integer and load/store) coverpoints and tests

georgiatai:
Vx and Vls mostly working. Exceptions including vnclipu.wv, vsadd.vi, and vmv.x.s. Have not excluded them from this PR. Vls are missing coverage due to macro name change.


jordancarlin (approved):

georgiatai:
Vx and Vls should be running but missing coverage now.

PR #780, opened 2025-12-11
ACT 4 RVI 20 Progress

jordancarlin:
- early fp support
- compressed fixes


configs/ref/sail-rvi20_64/rvtest_config.svh
```diff
@@ -0,0 +1,31 @@
...
+
+// PMP Grain (G)
+// Set G as needed (e.g., 0, 1, 2, ...)
+`define G 4
+
+// Uncomment below if G = 0
+// `define G_IS_0
+
+// PMP mode selection
+`define PMP_16     // Choose between PMP_16 or PMP_64 or None
```
davidharrishmc: Should this be NONE?  No machine features in RVI20.

configs/ref/sail-rvi20_32/rvtest_config.h
```diff
@@ -0,0 +1,3 @@
+#define ACCESS_FAULT_ADDRESS 0x00000000
```
davidharrishmc, line 1: I don't think we need any of this because there is no PMP or exceptions in RVI20

configs/ref/sail-rvi20_64/rvtest_config.svh
```diff
@@ -0,0 +1,31 @@
...
+// David_Harris@hmc.edu 7 September 2024
+// SPDX-License-Identifier: Apache-2.0
+
+// This file is needed in the config subdirectory for each config supporting coverage.
+// It defines which extensions are enabled for that config.
+
+// Define XLEN, used in covergroups
+`define XLEN64
+`define FLEN64
+`define VLEN512
```
davidharrishmc: Not needed.  Neither is G, ACCESS_FAULT, CLINT

configs/ref/sail-rvi20_32/rvtest_config.svh
```diff
@@ -0,0 +1,31 @@
...
+// David_Harris@hmc.edu 7 September 2024
+// SPDX-License-Identifier: Apache-2.0
+
+// This file is needed in the config subdirectory for each config supporting coverage.
+// It defines which extensions are enabled for that config.
+
+// Define XLEN, used in covergroups
+`define XLEN32
+`define FLEN64
+`define VLEN512
```
davidharrishmc: Don't think we need VLEN, G, PMP_16, ACCESS_FAULT_ADDRESS, or CLINT_BASE for RVI20.

configs/ref/sail-rvi20_32/sail-rvi20_32.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - { name: Zifencei, version: "2.0.0" }
+
+params:
+  # A params
+  MISALIGNED_AMO: false
+  LRSC_RESERVATION_STRATEGY: reserve exactly enough to cover the access # XLEN size reservation
+  LRSC_FAIL_ON_VA_SYNONYM: false
+  LRSC_MISALIGNED_BEHAVIOR: always raise access fault
+  LRSC_FAIL_ON_NON_EXACT_LRSC: false
+  MUTABLE_MISA_A: false
```
davidharrishmc, line 26: There is no MISA, so do we need these parameters?  Similar for others outside the profile...

configs/ref/sail-rvi20_32/sail.json
```diff
@@ -0,0 +1,375 @@
...
+  "base": {
+    "xlen": 32,
+    "E": false,
+    "writable_misa": false,
+    "writable_fiom": true,
+    "writable_hpm_counters": {
+      "len": 32,
+      "value": "0x0"
+    },
+    "xtval_nonzero": {
```
davidharrishmc, line 12: There are no faults.  How much of this is needed?

configs/ref/sail-rvi20_32/sail.json
```diff
@@ -0,0 +1,375 @@
...
+        "include_in_device_tree": true
+      }
+    ]
+  },
+  "platform": {
+    "vendorid": 0,
+    "archid": 0,
+    "impid": 0,
+    "hartid": 0,
+    "reset_vector": 4096,
```
davidharrishmc, line 121: Are we trying to start at 0x1000, or go directly to 0x8000_0000?

configs/ref/sail-rvi20_32/sail.json
```diff
@@ -0,0 +1,375 @@
...
+    "cache_block_size_exp": 6,
```
davidharrishmc, line 122: No cache operations, peripherals, etc.  Can we trim more?

Makefile
```diff
@@ -59,13 +60,13 @@ $(STAMP_DIR)/covergroupgen.stamp: generators/coverage/covergroupgen.py $(COVERGR
 	@touch $@
 
 .PHONY: testgen
-testgen:  $(STAMP_DIR)/testgen.stamp
+testgen: $(STAMP_DIR)/testgen.stamp
 $(STAMP_DIR)/testgen.stamp: $(TESTGEN_DEPS) Makefile | $(STAMP_DIR)
-	$(UV_RUN) testgen testplans -o tests --extensions M
+	$(UV_RUN) testgen testplans -o tests --extensions I,M,Zca,Zifencei # I,M,F,D,Zca,Zcf,Zcd,Zaamo,Zalrsc,Zifencei
```
davidharrishmc, line 67: How about adding all of the extensions that work to this list?

Makefile
```diff
@@ -4,7 +4,8 @@
 
 # Directories and files
 CONFIG_FILE ?= configs/duts/cvw/cvw-rv32gc/test_config.yaml configs/duts/cvw/cvw-rv64gc/test_config.yaml
```
davidharrishmc: Could you add comments explaining the difference between CONFIG_FILE and REF_CONFIG_FILES?  Should CONFIG_FILE also be plural because both seem to be a list?

allenjbaum:
A lot of these seem to be implementation specific defaults.
for priv CSRs and feature:  is there anything wrong with simply defining
them, knowing they won't be used?
and also: we know none of these tests should trap, but if they do (for
some reason which should be a test failure -  -what happens?
    Do we need to define something (e.g. a default trap vector in the
implementation supplied boot code)?
   Do we at least want to document what a DUT should do so that it doesn't
hang/infinite loop?

For reset vector:
 don't know that we want it to be the base (I assume) of DRAM, rather than
lowest non-debug page. I think 0x1000 is a good default, though the real
address for tests would be whatever the DUT defines.

configs/ref/sail-rvi20_64/rvtest_config.svh
jordancarlin: Dropped

configs/ref/sail-rvi20_64/rvtest_config.svh
jordancarlin: Dropped

configs/ref/sail-rvi20_32/rvtest_config.h
jordancarlin: Dropped

configs/ref/sail-rvi20_32/rvtest_config.svh
jordancarlin: `CLINT_BASE` is needed for Zicntr. All others removed.

configs/ref/sail-rvi20_32/sail-rvi20_32.yaml
jordancarlin: Right now they are needed. There is a large refactoring of how UDB handles config files that should be merged very soon, and once that is done I will update all of the configs to match the new requirements and only contain what they need. Right now there is no way to know other than with trial and error, but the new version will tell you exactly which parameters you are defining that are not part of the implemented extensions.

configs/ref/sail-rvi20_32/sail.json
jordancarlin: Sail requires all of the json keys to be have values. I changed these ones to be false, but they still need to be defined.

configs/ref/sail-rvi20_32/sail.json
jordancarlin: The elf is loaded at 0x8000_0000. This will have Sail reset to 0x1000 and then jump to 0x8000_0000. That's how most of the other simulators (Spike, etc.) handle it, so for now it makes sense to match them.

Makefile
jordancarlin: Done

Makefile
jordancarlin: Discovered an issue with the interaction between the list of extensions passed here and the config files. Will address that and update the list of extensions in a follow up PR.

jordancarlin:
> we know none of these tests should trap, but if they do (for some reason which should be a test failure - -what happens? Do we need to define something (e.g. a default trap vector in the implementation supplied boot code)? Do we at least want to document what a DUT should do so that it doesn't hang/infinite loop?

We've been thinking about that lately, and I'm leaning towards having a very basic trap handler that will get included if `rvtest_mtrap_routine` is not defined and Zicsr is supported. That way tests at least don't infinite loop.

davidharrishmc:
@allenjbaum the unpriv tests can run on a machine with no Zicsr, so there is no way to define a default trap vector.

davidharrishmc (commented):
Looks good

davidharrishmc (approved):
Approved.

PR #784, opened 2025-12-12
Generate `RISCV_coverage_config.svh` dynamically

jordancarlin:
This file was maintained by hand before and people frequently forgot to update it. Generate it instead.


davidharrishmc (approved):
approve

PR #786, opened 2025-12-12
Updated ExceptionsH to run with 0% coverage

SadhviNarayanan:
Updated ExceptionsH to run with 0% coverage through a nop test.
Coverpoints have been revised based on all the feedback that has been received.


coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,783 @@
+///////////////////////////////////////////
+//
+// RISC-V Architectural Functional Coverage Covergroups
+//
+// Written: Sadhvi Narayanan sanarayanan@hmc.edu 5 September 2025
+//
+// Copyright (C) 2024 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
+//
+// SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
```
jordancarlin: Old license header

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,783 @@
...
+   }
+
+   // Previous modes for delegation to M (all 5 modes)
+   priv_mode_to_m: coverpoint {ins.prev.mode_virt, ins.prev.mode} {
+       bins M_mode  = {3'b?11};
+       bins HS_mode = {3'b001};
+       bins U_mode  = {3'b000};
+       bins VS_mode = {3'b101};
+       bins VU_mode = {3'b100};
+   }
```
jordancarlin: Why do we need this in addition to `modes`?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,783 @@
...
+// Licensed under the Solderpad Hardware License v 2.1 (the “License”); you may not use this file
+// except in compliance with the License, or, at your option, the Apache License version 2.0. You
+// may obtain a copy of the License at
+//
+// https://solderpad.org/licenses/SHL-2.1/
+//
+// Unless required by applicable law or agreed to in writing, any work distributed under the
+// License is distributed on an “AS IS” BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+// either express or implied. See the License for the specific language governing permissions
+// and limitations under the License.
```
jordancarlin: Old license header

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,783 @@
...
+   // ECALL delegation disabled at HS-mode (VS mode)
+   hedeleg_ecall_disabled_vs: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hedeleg", "deleg")[10]) {
+       bins not_delegated = {0};
+   }
+
+
+   // All ECALL bits disabled (no delegation to M-mode)
+   medeleg_ecall_bits_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") & 32'h00000F00) {
+       bins all_ecall_bits_zero = {32'h00000000};
+   }
```
jordancarlin: Might be more obvious if you represent this by just selecting the relevant bits instead of anding it with 0x00000F00 and checking all 32 bits.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,783 @@
...
+       bins hu_disabled = {0};
+       bins hu_enabled = {1};
+   }
+
+
+   // ============================================================================
+   // TRAP VERIFICATION COVERPOINTS
+   // ============================================================================
+
+
```
jordancarlin: There's a lot of unnecessary whitespace that you could remove, but no strong preference one way or another.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,783 @@
...
+   hfence_instructions: coverpoint ins.current.insn {
+       wildcard bins sfence_vma = {32'b0001001??????????000000001110011};
+       wildcard bins hfence_vvma = {32'b0010001??????????000000001110011};
+       wildcard bins hfence_gvma = {32'b0110001??????????000000001110011};
+   }
+
+
+   wfi: coverpoint ins.current.insn {
+       wildcard bins wfi = {32'b00010000010100000000000001110011};
```
jordancarlin: No need for wildcard

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,783 @@
...
+   cp_virtual_instr_vs_instret: cross h_counteren_disabled_ir, m_counteren_enabled_ir, csrr, instret, priv_mode_vs;
+   cp_virtual_instr_vs_execute_hypervisor: cross hypervisor_instr, priv_mode_vs;
+   cp_virtual_instr_vs_read_vstval_hval: cross csrr, vstval_htval, priv_mode_vs;
+   cp_virtual_instr_vs_mstatus_satp: cross csrr, satp, mstatus_tvm, priv_mode_vs;
+   cp_virtual_instr_vs_mstatus_vsatp: cross csrr, vsatp, mstatus_tvm, priv_mode_vs;
+   cp_virtual_instr_vs_wfi: cross wfi, mstatus_tw, priv_mode_vs;
+   cp_virtual_instr_vs_sret: cross sret, hstatus_vtsr_enabled, priv_mode_vs;
+   cp_virtual_instr_vs_s_vma_instr: cross sfence_sinval_vma, hstatus_vtvm_enabled, priv_mode_vs;
+   cp_virtual_instr_vs_satp: cross csrr, satp, hstatus_vtvm_enabled, priv_mode_vs;
+   `ifdef RV32
```
jordancarlin: Global comment: Should use `XLEN32` and `XLEN64` like all of the over coverpoints, not `RV32` and `RV64`.

jordancarlin (changes requested):

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+   }
+
+
+   pc_bit_1: coverpoint ins.current.pc_rdata[1] {
+       bins zero = {0};
+   }
+
+
+   imm_bit_1: coverpoint ins.current.imm[1] {
+       bins one = {'1};
```
jordancarlin: Only one bit, so better to be explicit

```suggestion
       bins one = {1'b1};
```

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+   // All ECALL bits disabled (no delegation to M-mode)
+   medeleg_ecall_bits_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg")[11:8]) {
+       bins all_ecall_bits_zero = {4'b0000};
+   }
+
+
+   // ECALL delegation enabled (generic)
+   medeleg_ecall_enabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg")[8]) {
+       bins delegated = {1};
+   }
```
jordancarlin: Isn't this the same as `medeleg_ecall_enabled_u_vu`?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+       bins spvp_0 = {0};
+       bins spvp_1 = {1};
+   }
+
+
+   // Sample after trap to verify GVA bit was set correctly
+   hstatus_gva: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "hstatus", "gva") {
+       bins gva_invalid = {0};
+       bins gva_valid = {1};
+   }
```
jordancarlin: These seem like they are checking outcomes, which should be part of the test and not part of the coverpoints.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+   trap_to_hs: coverpoint {ins.prev.mode_virt, ins.prev.mode} {
+       bins trapped_to_hs = {3'b001};
+   }
+
+
+   trap_to_m: coverpoint ins.current.mode {
+       bins trapped_to_m = {2'b11};
+   }
+
```
jordancarlin: Unclear what these coverpoints are doing. They say `trap_to_*` but they are checking the priv mode at the start of the instruction.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+   stval: coverpoint ins.current.insn[31:20] {
+       bins stval = {12'h143};
+   }
+
+
+   rv32_hedelegh: coverpoint ins.current.insn[31:20] {
+       wildcard bins hedelegh_read = {12'h602};
+   }
```
jordancarlin: Should this be in an `ifdef XLEN32?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+
+   csrr_vstval_read: coverpoint ins.current.insn {
+       wildcard bins vstval_read = {32'b001001000011_00000_010_?????_1110011};
+   }
```
jordancarlin: Why does this one include the full csrr instruction and all of the other coverpoints are just checking the top 12 bits of the instruction? They should be consistent.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+
+   // ============================================================================
+   // HSTATUS BIT COVERPOINTS (Extended)
```
jordancarlin: Extended compared to what?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+   hstatus_gva_set: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "hstatus", "gva") {
+       bins gva_set = {1};
+   }
+
+
+   hstatus_gva_clear: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "hstatus", "gva") {
+       bins gva_clear = {0};
+   }
+
```
jordancarlin: `SAMPLE_AFTER` makes me concerned. Are these being used for checking or are they necessary to confirm we are testing the right thing?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+   rv32_scounter_enabled_ir: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "scounteren", "counteren")[2]) {
+       bins one = {1};
+   }
+
+
+   rv32_scounter_disabled_ir: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "scounteren", "counteren")[2]) {
+       bins zero = {0};
+   }
```
jordancarlin: ifdef XLEN32

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+
+   // ============================================================================
+   // PRIVILEGE MODE VS DELEGATION SETTINGS
+   // ============================================================================
+
+   vs_mode: coverpoint {ins.prev.mode, get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hstatus", "v")} {
+       bins VS_mode = {2'b01, 1'b1};
+   }
```
jordancarlin: Don't you already have a `priv_mode_vs` coverpoint? How is this different?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,748 @@
...
+
+   medeleg_settings: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+       bins all_zero = {32'h00000000};
+       bins all_one_except_m_hs_ecall = {32'hFFFFF7FF};
+   }
+
+
+   hedeleg_disabled: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hedeleg", "deleg") {
+       bins disabled = {32'h00000000};
+   }
```
jordancarlin: Why aren't these with the other medeleg and hedeleg coverpoints?

jordancarlin (changes requested):
More comments in slack

davidharrishmc:
Looks like I should do another close review after @SadhviNarayanan implements @jordancarlin review.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: There isn't a hstatus.V bit.  Is this old code? It should be replace with ins.prev.mode_virt.

jordancarlin: Yes. Already discussed with @SadhviNarayanan (along with lots of other things I didn't comment on here directly). Updated version coming soon.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,683 @@
...
+
+
+   // All ECALL bits disabled (no delegation to M-mode)
+   medeleg_ecall_bits_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg")[11:8]) {
+       bins all_ecall_bits_zero = {4'b0000};
+   }
+
+
+   // EBREAK delegation disabled in medeleg (bit 3)
+   medeleg_ebreak_bits_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg")[3]) {
```
jordancarlin: Bits is not adding anything to the name and just makes it longer. Comment applies to several other coverpoints.

```suggestion
   medeleg_ebreak_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg")[3]) {
```

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,683 @@
...
+
+
+   // ============================================================================
+   // DELEGATION REGISTER COVERPOINTS
+   // ============================================================================
+
+   // Machine Exception Delegation Register (medeleg)
+   medeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b1111_0000_1011_1111_1111_111?};
```
jordancarlin: This is not possible. Not all of these bits are writable.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,683 @@
...
+
+   // EBREAK delegation disabled in medeleg (bit 3)
+   medeleg_ebreak_bits_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg")[3]) {
+       bins ebreak_not_delegated = {32'h0};
+   }
+
+   // All medeleg bits 0 or 1 (except M/HS-mode ECALL bits 11, 9)
+   medeleg_all_bits_except_ecall: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+       bins all_zero = {32'h00000000};
+       bins all_one_except_m_hs_ecall = {32'hFFFFF5FF};
```
jordancarlin: This is not possible. Not all of these bits are writable.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,684 @@
...
+
+
+   ebreak: coverpoint ins.current.insn {
+       bins ebreak = {32'h00100073};
+   }
+
+
+   illegalops: coverpoint ins.current.insn {
+       bins zeros = {0};
+       bins ones  = {1};
```
jordancarlin: I assume you want this to be all ones, not a single one.
```suggestion
       bins ones  = {'1};
```

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,684 @@
...
+   cp_hedeleg_load_access_fault: cross loadops, illegal_address, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_store_access_fault: cross storeops, illegal_address, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_instr_misaligned_branch: cross branch, pc_bit_1, imm_bit_1, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_instr_misaligned_jal: cross jal, pc_bit_1, imm_bit_1, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_instr_misaligned_jalr: cross jalr, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_load_misaligned: cross loadops, adr_LSBs, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_store_misaligned: cross storeops, adr_LSBs, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_illegal_instruction: cross illegalops, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_breakpoint: cross ebreak, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
```
jordancarlin: Right now these crosses are much larger than what is on the spreadsheet. For example, instead of just a load access fault, you are doing load access faults with ALL of the load instructions. This is already a pretty big cross and I'm not convinced that these extra cases are useful. @davidharrishmc thoughts?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+       hedeleg_ecall_u_vs_enabled,
+       vstvec_different_from_stvec;
+
+
+   // Priority crosses
+   // checking for legal addresses is not as rigorous right now
+   cp_priority: cross
+       hlv_hsv_instr,
+       address_legality,
+       addr_misalignment,
```
jordancarlin: This is not possible with the way these coverpoints are currently defined. `address_legality` requires the address to be exactly equal to `ACCESS_FAULT_ADDRESS` while `addr_misalignment` requires the bottom bit to sweep through different values.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+       priv_modes_m_hs_vs_u_vu,
+       hstatus_hu;
+
+
+   // Virtual instruction exception crosses - VS mode
+   cp_virtual_instr_vs_instret: cross h_counteren_disabled_ir, m_counteren_enabled_ir, csrr, instret, priv_mode_vs;
+   cp_virtual_instr_vs_execute_hypervisor: cross hlvw_hlvxwu_hsvw_hfencevvma_hfencegvma_instr, priv_mode_vs;
+   cp_virtual_instr_vs_read_vstval_hval: cross csrr, vstval_htval, priv_mode_vs;
+   cp_virtual_instr_vs_mstatus_satp: cross csrr, satp, mstatus_tvm, priv_mode_vs;
+   cp_virtual_instr_vs_mstatus_vsatp: cross csrr, vsatp, mstatus_tvm, priv_mode_vs;
```
jordancarlin: Spreadsheet specifies these are with `mstatus.TVM == 0`, not both 0 and 1 (which is what you currently have).

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+   cp_virtual_instr_vs_wfi: cross wfi, mstatus_tw_disabled, priv_mode_vs;
```
jordancarlin: Missing `hstatus.VTW` = 1

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+   cp_virtual_instr_vs_sret: cross sret, hstatus_vtsr_enabled, priv_mode_vs;
+   cp_virtual_instr_vs_s_vma_instr: cross sfence_sinval_vma, hstatus_vtvm_enabled, priv_mode_vs;
+   cp_virtual_instr_vs_satp: cross csrr, satp, hstatus_vtvm_enabled, priv_mode_vs;
+   `ifdef XLEN32
+       cp_virtual_instr_vs_rv32_instreth_mcounter: cross csrr, instret, rv32_mcounter_enabled_ir, rv32_hcounter_disabled_ir, priv_mode_vs;
```
jordancarlin: Should be instreth, not instret.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+       cp_virtual_instr_vs_rv32_hedelegh: cross csrr, rv32_hedelegh, priv_mode_vs;
+   `endif
+
+
+   // Virtual instruction exception crosses - VU mode
+   cp_virtual_instr_vu_instret_1: cross csrr, instret, h_counteren_disabled_ir, m_counteren_enabled_ir, s_counteren_enabled_ir, priv_mode_vu;
```
jordancarlin: Global comment: the CSRs themselves are named mcounteren/scounteren/hcounteren, so let's name the coverpoints `mcounteren_enabled_ir` instead of `m_counteren_enabled_ir` (and the same for the other priv modes).

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+   cp_virtual_instr_vu_instret_2: cross csrr, instret, h_counteren_enabled_ir, s_counteren_disabled_ir, m_counteren_enabled_ir, priv_mode_vu;
+   cp_virtual_instr_vu_execute_h: cross hlvw_hlvxwu_hsvw_hfencevvma_hfencegvma_instr, priv_mode_vu;
+   cp_virtual_instr_vu_read_vstval_hval: cross csrr, vstval_htval, priv_mode_vu;
+   cp_virtual_instr_vu_read_stval: cross csrr, stval, priv_mode_vu;
+   cp_virtual_instr_vu_satp: cross csrr, satp, mstatus_tvm, priv_mode_vu;
+   cp_virtual_instr_vu_vsatp: cross csrr, vsatp, mstatus_tvm, priv_mode_vu;
```
jordancarlin: Spreadsheet specifies these are with mstatus.TVM == 0, not both 0 and 1 (which is what you currently have).

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+   cp_virtual_instr_vu_wfi: cross wfi, hstatus_vtw_enabled, mstatus_tw_disabled, priv_mode_vu;
```
jordancarlin: Spreadsheet doesn't mention hstatus_vtw_enabled. Check which one needs to be updated, but they should be consistent.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+   cp_virtual_instr_vu_sret: cross sret, priv_mode_vu;
+   cp_virtual_instr_vu_sfence_vma: cross sfence_vma, priv_mode_vu;
+   `ifdef XLEN32
+       cp_virtual_instr_vu_rv32_instreth_1: cross csrr, instret, rv32_hcounter_disabled_ir, rv32_scounter_enabled_ir, rv32_mcounter_enabled_ir, priv_mode_vu;
+       cp_virtual_instr_vu_rv32_instreth_2: cross csrr, instret, rv32_hcounter_enabled_ir, rv32_scounter_disabled_ir, rv32_mcounter_enabled_ir, priv_mode_vu;
```
jordancarlin: Should be instreth, not instret.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+   // HLV access fault crosses
+   cp_hlv_access_fault: cross hlv_instructions, illegal_address, priv_mode_m;
+
+
+   // HSV address misalignment crosses
+   cp_hsv_address_misaligned: cross hlvb_hlvh_hlvw_hlvd_instructions, address_lsbs, priv_mode_m;
+
+
+   // HSV access fault crosses
+   cp_hsv_access_fault: cross hlvb_hlvh_hlvw_hlvd_instructions, illegal_address, priv_mode_m;
```
jordancarlin: These should both be trying `hsv.{b, h, w, d}` instructions, not `hlv.*`.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+
+
+   // HTINST/XTINST crosses - transformed instruction encoding
+   // Execute if Zca not supported?
+   cp_xtinst_instr_misaligned_1: cross jal, pc_bit_1, imm_bit_1, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+   cp_xtinst_instr_misaligned_2: cross jalr, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
```
jordancarlin: Spreadsheet doesn't say to require attempting with both `jal` and `jalr`. Update one or the other, but they should be consistent.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+
+   // Execute if Zicclsm not supported?
+   cp_xtinst_load_misaligned: cross loadops, adr_LSBs, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+
+   cp_xtinst_load_access: cross loadops, illegal_address, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+
+   // Execute if Zicclsm not supported?
+   cp_xtinst_store_misaligned: cross storeops, adr_LSBs, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+
+   cp_xtinst_store_access: cross storeops, illegal_address, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
```
jordancarlin: Spreadsheet only says to do each of these with a single load/store op, you are doing it with all of them.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,692 @@
...
+   cp_xtinst_breakpoint: cross ebreak, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+   cp_xtinst_virtinstr: cross csrr, vstval, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+
+   // Execute if Zicclsm not supported?
+   cp_xtinst_load_misaligned: cross loadops, adr_LSBs, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+
+   cp_xtinst_load_access: cross loadops, illegal_address, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
+
+   // Execute if Zicclsm not supported?
+   cp_xtinst_store_misaligned: cross storeops, adr_LSBs, priv_mode_vs, medeleg_all_bits_except_ecall, hedeleg_disabled;
```
jordancarlin: Spreadsheet only to trigger misaligned load fault, you are sweeping through all 4 of the bottom address bits.

jordancarlin (changes requested):
One more round of feedback from me. Getting close.

coverpoints/priv/ExceptionsH_coverage.svh
SadhviNarayanan: For some reason, my latest push did not update, these were all already fixed.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: Per Table 43 of the privileged spec, only bits 14 and 16-19 are reserved in hypervisor mode, so all of the others should be writable.  I think this is correct.  @jordancarlin do you have specific bits you are worried about?

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: I think we can use the same "ones" number from line 177 above except also disable bits 11 and 9.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: Agreed that one way to generate each type of fault is sufficient.

coverpoints/priv/ExceptionsH_coverage.svh
jordancarlin: I think I'm looking at a different version of the spec so my table numbers are different.

- 24-31 are "designated for custom use".
- 18 and 19 are not reserved in the version I'm looking at. They are for software check and hardware error respectively. Software check is writable with Zicfiss or Zicfilp, so should probably be  a wildcard.
- From section 3.1.8, bits 11 and 16 are read-only zero.

davidharrishmc: Priv Spec 250508.
Ok to make 18 and 19 a wildcard. 
ok to make 24-31 wildcard
Agreed 11 and 16 should be0

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   }
+
+
+   // ============================================================================
+   // CROSS COVERAGE
+   // ============================================================================
+
+   // cp_hedeleg crosses
+   cp_hedeleg_instr_access_fault: cross jalr, illegal_address, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_load_access_fault: cross loadop, illegal_address, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
```
davidharrishmc: @jordancarlin had suggested only testing one load, one store etc. in cp_hedeleg.  Looks like you still have all of them, which is a lot of tests.  If you prefer this way, respond to the comment with your justification.  Otherwise, simplify loadop, storeop, etc. to a single bin.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       wildcard bins jal = {32'b????????????????????_?????_1101111};
+   }
+
+
+   jalr: coverpoint ins.current.insn {
+       wildcard bins jalr = {32'b????????????_?????_000_?????_1100111};
+   }
+
+
+   loadops: coverpoint ins.current.insn {
```
davidharrishmc: Not used; delete

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       `endif
+   }
+
+
+   loadop: coverpoint ins.current.insn {
+       wildcard bins lw  = {32'b????????????_?????_010_?????_0000011};
+   }
+
+
+   storeops: coverpoint ins.current.insn {
```
davidharrishmc: not used.  Delete

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   imm_bit_1: coverpoint ins.current.imm[1] {
+       bins one = {1};
+   }
+
+
+   // ============================================================================
+   // PRIVILEGE MODE COVERPOINTS
+   // ============================================================================
+
+   // All 5 privilege modes (M/HS/VS/VU/U)
```
davidharrishmc: Jordan's PR has been merged.  Pull it, delete these, and test it works together.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: Better make medeleg 1111_0000_1011_0111_1111_111?
because bit 11 is reserved and might not be writable.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: delete; no longer used.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: delete - no longer used.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   }
+
+
+   imm_bit_1: coverpoint ins.current.imm[1] {
+       bins one = {1};
+   }
+
+
+   // ============================================================================
+   // PRIVILEGE MODE COVERPOINTS
```
davidharrishmc: All of these coverpoints have been merged per Jordan's PR.  Pull that change, remove these, and check the coverpoints still work.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+
+
+   // ============================================================================
+   // DELEGATION REGISTER COVERPOINTS
+   // ============================================================================
+
+   // Machine Exception Delegation Register (medeleg)
+   medeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b????_????_1111_??10_1111_0111_1111_111?};
```
davidharrishmc: Better change ones to

????_????_1111_??00_1011_0111_1111_111?
because bits 17 and 14 is reserved, bits 18 and 19 might not be implemented

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   // ============================================================================
+   // CROSS COVERAGE
+   // ============================================================================
+
+   // cp_hedeleg crosses
+   cp_hedeleg_instr_access_fault: cross jalr, illegal_address, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_load_access_fault: cross loadop, illegal_address, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_store_access_fault: cross storeop, illegal_address, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_instr_misaligned_branch: cross branch, pc_bit_1, imm_bit_1, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_instr_misaligned_jal: cross jal, pc_bit_1, imm_bit_1, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
```
davidharrishmc: One misaligned instr coverpoint is sufficient.  jalr is probably easiest, so I would drop the other two coverpoints.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   cp_hedeleg_instr_misaligned_jalr: cross jalr, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_load_misaligned: cross loadop, adr_LSBs, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
```
davidharrishmc: One misaligned load and one misaligned store is sufficient.  Replace adr_LSBs with addr_misaligned.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   cp_hedeleg_store_misaligned: cross storeop, adr_LSBs, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
+   cp_hedeleg_illegal_instruction: cross illegalops, priv_modes_m_hs_vs_u_vu, medeleg_delegation, hedeleg_delegation;
```
davidharrishmc: One illegal op is sufficient.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+
+
+   ebreak: coverpoint ins.current.insn {
+       bins ebreak = {32'h00100073};
+   }
+
+
+   illegalops: coverpoint ins.current.insn {
+       bins zeros = {0};
+       bins ones  = {'1};
```
davidharrishmc: Drop ones and change to illegalop.  One illegal operation is enough for H.  Our goal is to test things that are different about H, not repeat all of the things we already tested in ExeptionsM.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       wildcard bins lw  = {32'b????????????_?????_010_?????_0000011};
+       wildcard bins lh  = {32'b????????????_?????_001_?????_0000011};
+       wildcard bins lhu = {32'b????????????_?????_101_?????_0000011};
+       wildcard bins lb  = {32'b????????????_?????_000_?????_0000011};
+       wildcard bins lbu = {32'b????????????_?????_100_?????_0000011};
+       `ifdef XLEN64
+           wildcard bins ld  = {32'b????????????_?????_011_?????_0000011};
+           wildcard bins lwu = {32'b????????????_?????_110_?????_0000011};
+       `endif
+   }
```
jordancarlin: Unused

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+
+
+   storeops: coverpoint ins.current.insn {
+       wildcard bins sb = {32'b????????????_?????_000_?????_0100011};
+       wildcard bins sh = {32'b????????????_?????_001_?????_0100011};
+       wildcard bins sw = {32'b????????????_?????_010_?????_0100011};
+       `ifdef XLEN64
+           wildcard bins sd = {32'b????????????_?????_011_?????_0100011};
+       `endif
+   }
```
jordancarlin: Unused

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   // ============================================================================
+   // FAULT CONDITION COVERPOINTS
+   // ============================================================================
+
+   illegal_address: coverpoint ins.current.imm + ins.current.rs1_val {
+       bins illegal = {`ACCESS_FAULT_ADDRESS};
+   }
+
+
+   address_legality: coverpoint ((ins.current.imm + ins.current.rs1_val) & ~32'h3) == (`ACCESS_FAULT_ADDRESS & ~32'h3) {
```
jordancarlin: Will 32 work when `imm`/`rs1_val` are 64 bits?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+
+
+   imm_bit_1: coverpoint ins.current.imm[1] {
+       bins one = {1};
+   }
+
+
+   // ============================================================================
+   // PRIVILEGE MODE COVERPOINTS
+   // ============================================================================
```
jordancarlin: Drop all of these now that #788 is merged.

coverpoints/priv/ExceptionsH_coverage.svh
jordancarlin: Shouldn't bits 14 and 17 be `0`?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+
+   // EBREAK delegation disabled in medeleg (bit 3)
+   medeleg_ebreak_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg")[3]) {
+       bins ebreak_not_delegated = {32'h0};
+   }
+
+   // All medeleg bits 0 or 1 (except M/HS-mode ECALL bits 11, 9)
+   medeleg_except_ecall: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+       bins all_zero = {32'h00000000};
+       wildcard bins all_one_except_m_hs_ecall = {32'b????_????_1111_??10_1111_0101_1111_111?};
```
jordancarlin: Shouldn't bits 14 and 17 be `0`?

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   medeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "medeleg", "deleg") {
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b????_????_1111_??10_1111_0111_1111_111?};
+   }
+
+
+   // Hypervisor Exception Delegation Register (hedeleg)
+   hedeleg_delegation: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hedeleg", "deleg") {
+       bins zeros = {32'h00000000};
+       wildcard bins ones = {32'b0000_11?0_1?11_0001_1111_111?};
```
jordancarlin: You say this is a 32 bit number but don't have 32 actual digits in the value.

jordancarlin: Shouldn't bits 14 and 17 be 0? They are reserved.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+
+   mstatus_tvm_disabled: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "tvm") {
+       bins tvm_disabled = {0};
+   }
+
+
+   mstatus_tvm: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "tvm") {
+       bins tvm_disabled = {0};
+       bins tvm_enabled = {0};
+   }
```
jordancarlin, line 394: Newly introduced typo? Not sure why this coverpoint changed from the previous version, but both bins should not have 0 as their value.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   mstatus_tw_disabled: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "tw") == 0) {
+       bins zero = {0};
+   }
+
+
+   // ============================================================================
+   // COUNTER ENABLE COVERPOINTS
+   // ============================================================================
+
+   scounteren_enabled_ir: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "scounteren", "counteren")[2]) {
```
jordancarlin, line 406: You have counteren helper coverpoints at different places in this file (here and up on line 306). Bring all of them together.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+        }
+   `endif
+
+
+   // ============================================================================
+   // ADDRESS ALIGNMENT COVERPOINTS
+   // ============================================================================
+
+   address_lsbs: coverpoint {ins.current.rs1_val + ins.current.imm}[2:0] {
+   }
```
jordancarlin: This is the same as the `adr_LSBs` coverpoint up on line 124

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   cp_hsv_address_misaligned: cross hsvb_hsvh_hsvw_instructions, address_lsbs, priv_mode_m;
+
+
+   // HSV access fault crosses
+   cp_hsv_access_fault: cross hsvb_hsvh_hsvw_instructions, illegal_address, priv_mode_m;
+
+
+   // HTINST/XTINST crosses - transformed instruction encoding
+   // Execute if Zca not supported?
+   cp_xtinst_instr_misaligned_1: cross jal, pc_bit_1, imm_bit_1, priv_mode_vs, medeleg_except_ecall, hedeleg_disabled;
```
jordancarlin: Why the `_1` suffix on the coverpoint name?

jordancarlin (changes requested):
A few more comments

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       address_legality,
+       addr_alignment,
+       priv_modes_m_hs_vs_u_vu,
+       hstatus_hu;
+
+
+   // Virtual instruction exception crosses - VS mode
+   cp_virtual_instr_vs_instret: cross hcounteren_disabled_ir, mcounteren_enabled_ir, csrr, instret, priv_mode_vs;
+   cp_virtual_instr_vs_execute_hypervisor: cross hlvw_hlvxwu_hsvw_hfencevvma_hfencegvma_instr, priv_mode_vs;
+   cp_virtual_instr_vs_read_vstval_hval: cross csrr, vstval_htval, priv_mode_vs;
```
davidharrishmc: hval -> htval

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       cp_virtual_instr_vs_rv32_instreth_mcounter: cross csrr, instreth, rv32_mcounter_enabled_ir, rv32_hcounter_disabled_ir, priv_mode_vs;
+       cp_virtual_instr_vs_rv32_hedelegh: cross csrr, rv32_hedelegh, priv_mode_vs;
+   `endif
+
+
+   // Virtual instruction exception crosses - VU mode
+   cp_virtual_instr_vu_instret_1: cross csrr, instret, hcounteren_disabled_ir, mcounteren_enabled_ir, scounteren_enabled_ir, priv_mode_vu;
+   cp_virtual_instr_vu_instret_2: cross csrr, instret, hcounteren_enabled_ir, scounteren_disabled_ir, mcounteren_enabled_ir, priv_mode_vu;
+   cp_virtual_instr_vu_execute_h: cross hlvw_hlvxwu_hsvw_hfencevvma_hfencegvma_instr, priv_mode_vu;
+   cp_virtual_instr_vu_read_vstval_hval: cross csrr, vstval_htval, priv_mode_vu;
```
davidharrishmc: hval -> htval

jordancarlin (approved):
LGTM. This one was a ton of work. Great job with it!

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   stval: coverpoint ins.current.insn[31:20] {
+       bins stval = {12'h143};
+   }
+
+   vstval: coverpoint ins.current.insn[31:20] {
+       bins vstval = {12'h243};
+   }
+
+   `ifdef XLEN32
+        rv32_hedelegh: coverpoint ins.current.insn[31:20] {
```
davidharrishmc: instreth doesn't have a rv32_ prefix.  I'd suggest dropping it everywhere in this file because it doesn't add useful info.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       bins disabled = {0};
+   }
+
+
+   hcounteren_enabled_ir: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "hcounteren", "counteren")[2]) {
+       bins enabled = {1};
+   }
+
+
+   `ifdef XLEN32
```
davidharrishmc: The following coverpoints are missing en from the register name.  For example, rv32_hcounter_disabled_ir should be hcounteren_disabled_ir.  Some of them overlap with the non-rv32 coverpoints defined above (e.g. hcounteren_enabled_ir) and duplicates should be removed here.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       wildcard bins hsv_b  = {32'b0110001??????????100000001110011};
+       wildcard bins hsv_h  = {32'b0110011??????????100000001110011};
+       wildcard bins hsv_w  = {32'b0110101??????????100000001110011};
+       `ifdef XLEN64
+           wildcard bins hsv_d = {2'b0110111??????????100000001110011};
+       `endif
+   }
+
+
+   hlv_hlvx_hsv_instr: coverpoint ins.current.insn {
```
davidharrishmc: Above, there was a coverpoint named hlvw_hlvxwu_hsvw_hfence...  Can you be consistent with w,wu

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       wildcard bins hlv_w = {32'b011010000000?????100?????1110011};
+       wildcard bins hlvx_w = {32'b011010000011?????100?????1110011};
```
davidharrishmc: should be wu

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+
+   cp_hfence_priv: cross
+       sfencevma_hfencevvma_hfencegvma_instr,
+       priv_modes_m_hs_vs_u_vu,
+       mstatus_tvm,
+       hstatus_vtvm;
+
+
+   // HLV address misalignment crosses
+   cp_hlv_address_misaligned: cross hlv_instructions, address_lsbs, priv_mode_m;
```
davidharrishmc: address_lsbs depends on imm, but hlv* doesn't take an immediate.  You'll need another flavor of address_lsbs without adding immediate.  Same for cp_hlv_access_fault with illegal_address

davidharrishmc: and the hsv coverpoints.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       wildcard bins hlvx_hu = {32'b011001000011?????100?????1110011};
+       wildcard bins hlvx_wu = {32'b011010000011?????100?????1110011};
+       `ifdef XLEN64
+           wildcard bins hlv_d = {32'b011011000000?????100?????1110011};
+           wildcard bins hlv_wu = {32'b011010000001?????100?????1110011};
+       `endif
+   }
+
+
+   hlvb_hlvh_hlvw_hlvd_instructions: coverpoint ins.current.insn {
```
davidharrishmc: This is not used.  Delete.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+       wildcard bins hlv_b  = {32'b011000000000?????100?????1110011};
+       wildcard bins hlv_h  = {32'b011001000000?????100?????1110011};
+       wildcard bins hlv_w  = {32'b011010000000?????100?????1110011};
+       `ifdef XLEN64
+           wildcard bins hlv_d = {32'b011011000000?????100?????1110011};
+       `endif
+   }
+
+
+   hsvb_hsvh_hsvw_instructions: coverpoint ins.current.insn {
```
davidharrishmc: Change to hsv_instructions to be consistent with hlv_instructions.

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   // HSV address misalignment crosses
+   cp_hsv_address_misaligned: cross hsvb_hsvh_hsvw_instructions, address_lsbs, priv_mode_m;
+
+
+   // HSV access fault crosses
+   cp_hsv_access_fault: cross hsvb_hsvh_hsvw_instructions, illegal_address, priv_mode_m;
+
+
+   // HTINST/XTINST crosses - transformed instruction encoding
+   // Execute if Zca not supported?
```
davidharrishmc: Yes, put an ifndef Zca around this coverpoint.

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: ????_????_1111_??00_1011_0101_1111_111?
to be consistent with medeleg_delegation

coverpoints/priv/ExceptionsH_coverage.svh
```diff
@@ -0,0 +1,726 @@
...
+   cp_xtinst_instr_misaligned_1: cross jal, pc_bit_1, imm_bit_1, priv_mode_vs, medeleg_except_ecall, hedeleg_disabled;
+
+   cp_xtinst_instr_access: cross jalr, illegal_address, priv_mode_vs, medeleg_except_ecall, hedeleg_disabled;
+   cp_xtinst_illegalinstr: cross illegalops, priv_mode_vs, medeleg_except_ecall, hedeleg_disabled;
+   cp_xtinst_breakpoint: cross ebreak, priv_mode_vs, medeleg_except_ecall, hedeleg_disabled;
+   cp_xtinst_virtinstr: cross csrr, vstval, priv_mode_vs, medeleg_except_ecall, hedeleg_disabled;
+
+   // Execute if Zicclsm not supported?
```
davidharrishmc: ifndef Zicclsm

coverpoints/priv/ExceptionsH_coverage.svh
davidharrishmc: It might be easier to write the test if this one also uses jalr like the next one.

davidharrishmc (changes requested):
Looking very good.  Here are some changes to catch a few bugs and make the code as consistent as possible.

coverpoints/priv/ExceptionsH_coverage.svh
jordancarlin: Testplan updated to always test this to ensure it doesn't trap with Zicclsm.

coverpoints/priv/ExceptionsH_coverage.svh
jordancarlin: Testplan updated to always test this to ensure it doesn't trap with Zca.

davidharrishmc (approved):
Deleted some unused exception coverpoints.  Ready to merge.

PR #787, opened 2025-12-12
Print error message and terminate program if Questa throws an error during coverage collection

jordancarlin:


davidharrishmc (approved):
Looks good.

PR #792, opened 2025-12-17
[ACT 4] Update self-checking floating point sigupd macro

jordancarlin:


tests/env/signature.h
```diff
@@ -84,30 +84,85 @@
...
-    csrr _LINK_REG, fcsr                                   ;\
-    LREG _TEMP_REG, SIG_STRIDE(_SIG_PTR)                   ;\
-    beq _TEMP_REG, _LINK_REG, 2f                           ;\
-    jal _LINK_REG, failedtest_##_LINK_REG##_##_TEMP_REG    ;\
-    2:                                                     ;\
-    addi _SIG_PTR, _SIG_PTR, 2*SIG_STRIDE
+// Floating point values are stored to memory and then loaded back into integer registers
+// for comparison, to avoid issues with NaN that arise from using feq. There is no way to
+// directly transfer a floating point value to an integer register without Zfa when FLEN > XLEN.
+#if FLEN > XLEN
```
davidharrishmc, line 90: Do you want to have a FLEN=128 XLEN=32 check to produce an error, so we don't get bitten when trying Q later?

davidharrishmc (approved):

PR #799, opened 2025-12-18
CTP Cleanup

davidharrishmc:
Phase 0 changes
Normative rules in CTP for more extensions
Generator for normative rules
Ssstrict improvements


ctp/src/i.adoc
```diff
@@ -18,14 +18,11 @@ include::{testplansdir}/I.adoc[]
...
 |cp_align_{byte/hword/word}|8/4/2|Alignment of naturally-aligned load/store operand within doubleword in memory
 |cp_custom_fence|3|Test that fence, fence rw,rw, and fence.tso all execute without trapping.
 |===
 
-<<t-I-normative-rules>> summarizes the normative rules in the RV{32/64}I specification and the unprivileged coverpoints that exercise them.  Privileged coverpoints related to integer instructions are described in the relevant parts of <<Privileged Test Plan>>. Note that the testplans always test the following fundamental capabilities, even if no normative rule is cited:
+<<t-I-normative-rules>> summarizes the normative rules in the RV{32/64}I specification and the unprivileged coverpoints that exercise them.  Privileged coverpoints related to integer instructions are described in the relevant parts of <<Privileged Test Plan>>. In particular, `ebreak` and `ecall` are only tested in the Privileged test plan because there behavior is UNSPECIFIED without an `mtvec` trap vector CSR. Reserved behaviors are checked in <<Sstrict>>. Note that the testplans always test the following fundamental capabilities, even if no normative rule is cited:
```
jordancarlin:
```suggestion
<<t-I-normative-rules>> summarizes the normative rules in the RV{32/64}I specification and the unprivileged coverpoints that exercise them.  Privileged coverpoints related to integer instructions are described in the relevant parts of <<Privileged Test Plan>>. In particular, `ebreak` and `ecall` are only tested in the Privileged test plan because there behavior is UNSPECIFIED without an `mtvec` trap vector CSR. Reserved behaviors are checked in <<Ssstrict>>. Note that the testplans always test the following fundamental capabilities, even if no normative rule is cited:
```

coverpoints/norm/yaml/Zifencei.yaml
```diff
@@ -7,4 +7,4 @@ normative_rule_definitions:
   - name: fence_i_op
     coverpoint: ["Zifencei_fence_i_cg/{cp_custom_fencei}"]
   - name: fence_i_rsv
-    coverpoint: ["*** see SsstrictS"]
+    coverpoint: ["NOT TESTED; implementations should ignore"]
```
jordancarlin: I think we should add testing reserved values of the fence.i fields to the standard unpriv fence.i test. The spec is explicit about what is supposed to happen and we need to certify that implementations follow that. We could either do exhaustive testing of this (I don't think the fields are that large) or just pick a few reserved values for each of the fields.

coverpoints/norm/yaml/I.yaml
```diff
@@ -128,6 +128,6 @@ normative_rule_definitions:
   - name: fence-tso_ordering_rw_rw_ok
     coverpoint: ["I_fence_cg/{cp_custom_fence}"]
   - name: fence_unused_flds_rsv
-    coverpoint: ["TODO"]
+    coverpoint: ["NOT TESTED; implementations should ignore"]
```
jordancarlin: Same comment as for fence.i

jordancarlin (changes requested):

coverpoints/norm/yaml/Zifencei.yaml
jordancarlin: The coverpoints was added, but you forgot to update the YAML.

generators/coverage/templates/cp_custom_fence.txt
```diff
@@ -1,7 +1,9 @@
...
     cp_custom_fence : coverpoint ins.current.insn  {
-        bins fence =  {32'h0ff0000f}; // iowr, iowr
-        bins fence_rw_rw = {32'h0330000f};
-        bins fence_tso_rw_rw = {32'h8330000f}; // fence.tso
+        bins fence            = {32'h0ff0000f}; // iowr, iowr
+        bins fence_rw_rw      = {32'h0330000f}; // rw, rw
+        bins fence_tso_rw_rw  = {32'h8330000f}; // fence.tso
+        bins fence_nonzerors1 = {32'h0331000f}; // nonzero rs should behave normally
+        bins fence_nonzerord  = {32'h0330008f}; // nonzero rd should behave normally
     }
```
jordancarlin, line 11: There are two other cases explicitly called out in the spec that are not yet tested here:

- Reserved (non 0 or 1) values for the `fm` field
- `fm = 1` with the predecessor/successor set not equal to `rw`/`rw`

Both of these should be added to the coverpoints and tests.

ctp/src/c.adoc
```diff
@@ -8,13 +8,42 @@ The C extension is subdivided into Zca, Zcf, and Zcd.  All configurations suppor
...
+Modified coverpoints for Zc* are summarized in <<t-Zc-modified-coverpoints>>.  As defined in <<t-unprivileged-coverpoints>>, the p versions of coverpoints work on registers x8-x15.
+
+[[t-Zc-modified-coverpoints]]
+.I Modified Coverpoints
+[cols="2, 1, 4" options=header]
+|===
+|Modified Coverpoint|Bins|Definition
+|cp_rs1_imm_edges_6bit|16*13|rs1 edges x Signed 6-bit immediates 0, 1, 2, 3, 4, 8 16, 30, 31, -32, -31, -2, -1
+|cp_rs1_imm_edges_6bit_n0|16*12|rs1 edges x cp_rs1_imm_edges_6bit excluding 0
+|cp_rs1_imm_edges_c|16*11|rs1 edges x unsigned 5 or 6-bit uimm 1, 2, 3, 4, 8, 14, 15, 16, 17, 30, 31. XLEN=64 adds 32, 33, 48, 62, 63
```
jordancarlin: cr instead of cp because these are crosses. Applies to several items in this list.

ctp/src/m.adoc
```diff
@@ -12,7 +12,7 @@ include::norm/M_norm_rules.adoc[]
 
 ==== Zmmul
 
-The Zmmul extension is a subset of the M extension that includes only the multiply instructions.  The coverpoints are listed in <<t-Zmmul-coverpoints>>.
+The Zmmul extension is a subset of the M extension that includes only the multiply instructions.  The coverpoints are listed in <<t-Zmmul-coverpoints>>. If a configuration supports both M and Zmmul, these multiply instructions are tested twice, but the runtime is negligible so multiplication is left in M to avoid confusion.
```
jordancarlin: M requires Zmmul, so any implementation that supports M must also support Zmmul (and UDB enforces that). This statement seems misleading right now.

jordancarlin (approved):

PR #801, opened 2025-12-20
[ACT4] Remove custom test templates

jordancarlin:
- Switch `fence`, `fence.i` and compressed hints to use Python generators
- Remove logic for custom test templates. They don't work well because of register management and string messages.


davidharrishmc (approved):
Looks good

PR #807, opened 2025-12-24
Hint and nop tests

davidharrishmc:
Added normative rules to coverpoint links for hints and nops
Added Ssstrict rules for RV32Zbs
Added fence hints
Old merges are showing up as commits but aren't changing anything
Updated documentation about hints

Fence hints are getting covered
nop is not getting covered.


generators/coverage/templates/cp_custom_fence.txt
```diff
@@ -7,5 +7,9 @@
         bins fence_nonzerors1 = {32'h0331000f}; // nonzero rs should behave as fence
         bins fence_nonzerord  = {32'h0330008f}; // nonzero rd should behave as fence
         bins fence_fm         = {32'h1330000f}; // reserved fm should behave as fence
+        bins fence_hint0      = {32'h0001000f}; // fence with rd = x0, rs1 != x0, fm = 0, pred = 0, succ=0 is a hint
+        bins fence_hint1      = {32'h0000008f}; // fence with rd != x0, rs1 = x0, fm = 0, pred = 0, succ=0 is a hint
```
jordancarlin: The table in https://docs.riscv.org/reference/isa/unpriv/rv32.html#2-10-hint-instructions says these are hints if pred=0 or succ=0. We should test both combinations instead of combining them into a single check. Applies to hint 0 and 1.

jordancarlin (approved):

PR #809, opened 2025-12-26
ZicsrM generation script

davidharrishmc:
Initial coverage of ZicsrM_mcause_cg with unrolled loops
Introduces CSR_WRITE macro that writes and reads back to signature
Eliminated reserved and custom values from cause codes tested


tests/env/signature.h
```diff
@@ -3,6 +3,13 @@
 # Jordan Carlin jcarlin@hmc.edu October 2025
 # SPDX-License-Identifier: Apache-2.0
 
+// Privileged tests prefer to use default register allocation for signature/self-check
+
+#define DEFAULT_SIG_PTR    x2
+#define DEFAULT_DATA_PTR   x3
+#define DEFAULT_TEMP_REG   x4
+#define DEFAULT_LINK_REG   x5
```
jordancarlin: These are already defined at the top of `arch_test.h`. Why are they defined here too now?

tests/env/signature.h
```diff
@@ -195,3 +188,24 @@
   #define CANARY \
       .word CANARY_VALUE
 #endif
+
+// CSR Macros
+
+// Read _CSR into _R and record/check the signature
+#define RVTEST_SIGUPD_CSR_RD(_CSR, _R, _SIG_PTR, _LINK_REG, _TEMP_REG, _STR_PTR) \
```
jordancarlin: All of the other sigupd macros use the following order:
```
  #define RVTEST_SIGUPD(_SIG_PTR, _LINK_REG, _TEMP_REG, _R, _STR_PTR)  \
```
This should probably be (mostly) consistent with that. So maybe something like this?
```
#define RVTEST_SIGUPD_CSR_RD(_SIG_PTR, _LINK_REG, _TEMP_REG, _CSR, _R, _STR_PTR) \
```
or 
```
#define RVTEST_SIGUPD_CSR_RD(_CSR, _SIG_PTR, _LINK_REG, _TEMP_REG, _R, _STR_PTR) \
```

tests/env/test_macros.h
```diff
@@ -1,4 +1,15 @@
-/* This function set up the Page table entry for Sv32 Translation scheme
+// Utility Macros
+
+// Place 1 in msb
+#define SET_MSB(_R) \
```
jordancarlin: Move to `utils.h`.

tests/priv/ZicsrM/ZicsrM_gen.py
```diff
@@ -0,0 +1,152 @@
...
+        # CONFIG_DEPENDENT: true
+        ##### END_TEST_CONFIG #####
+
+        #define TEST_FILE "{testname}"
+        #define SIGUPD_COUNT {sigupd_count}
+        #include "riscv_arch_test.h"
+
+        // Set up test environment and declare main entry point
+        RVTEST_BEGIN
+        .option norvc          # Do not automatically replace uncompressed instructions with compressed instructions
```
jordancarlin: Did it not end up working to move this into `RVTEST_BEGIN`?

davidharrishmc: Looks like it worked.  I had misunderstood your suggestion.

jordancarlin (approved):

PR #811, opened 2025-12-27
ExceptionsZicboU/S and ExceptionsF

Marina-Bellido:
ExceptionsF -> 90% coverage 
ExceptionsZicboU and S -> 100% coverage.

Working on fixing coverage for ExceptionsF.


coverpoints/priv/ExceptionsF_coverage.svh
```diff
@@ -63,23 +44,56 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
...
-        wildcard bins fsub         = {32'b00001_??_?????_?????_111_?????_1010011};
-        wildcard bins fmul         = {32'b00010_??_?????_?????_111_?????_1010011};
-        wildcard bins fdiv         = {32'b00011_??_?????_?????_111_?????_1010011};
-        wildcard bins fcvt_x_f     = {32'b11000_??_?????_?????_111_?????_1010011};
-        wildcard bins fcvt_f_x     = {32'b11010_??_?????_?????_111_?????_1010011};
-        wildcard bins fcvt_f_f     = {32'b01000_??_?????_?????_111_?????_1010011};
-        wildcard bins fmadd        = {32'b?????_??_?????_?????_111_?????_1000011};
-        wildcard bins fsqrt        = {32'b01011_??_00000_?????_111_?????_1010011};
-        wildcard bins fround       = {32'b01000_??_00100_?????_111_?????_1010011};
+        `ifdef F_SUPPORTED
```
jordancarlin, line 81: Why would ExceptionsF ever be run without F supported? Drop all of the ifdef F_SUPPORTED in both the coverpoints and the test.

coverpoints/priv/ExceptionsF_coverage.svh
```diff
@@ -63,23 +44,56 @@ covergroup ExceptionsF_exceptions_cg with function sample(ins_t ins);
...
+            wildcard bins fadd         = {32'b00000_??_?????_?????_111_?????_1010011};
+            wildcard bins fsub         = {32'b00001_??_?????_?????_111_?????_1010011};
+            wildcard bins fmul         = {32'b00010_??_?????_?????_111_?????_1010011};
+            wildcard bins fdiv         = {32'b00011_??_?????_?????_111_?????_1010011};
+            wildcard bins fcvt_x_f     = {32'b11000_??_?????_?????_111_?????_1010011};
+            wildcard bins fcvt_f_x     = {32'b11010_??_?????_?????_111_?????_1010011};
+            wildcard bins fmadd        = {32'b?????_??_?????_?????_111_?????_1000011};
+            wildcard bins fsqrt        = {32'b01011_??_00000_?????_111_?????_1010011};
+            `ifdef D_SUPPORTED
+                wildcard bins fcvt_f_f     = {32'b01000_??_?????_?????_111_?????_1010011};
```
jordancarlin, line 91: This is an interesting one. It should probably be tested if D or Zfhmin is supported, not just D.

coverpoints/priv/ExceptionsZicboS_coverage.svh
```diff
@@ -17,14 +17,20 @@ covergroup ExceptionsZicboS_exceptions_cg with function sample(ins_t ins);
 
     // building blocks for the main coverpoints
     cbo_inval: coverpoint ins.current.insn {
-        wildcard bins cbo_inval = {32'b000000000000_?????_010_00000_0001111};
+        `ifdef ZICBOM_SUPPORTED
```
jordancarlin, line 20: Doesn’t this ifdef need to go around the entire coverpoint, not just the bins? I think right now this will create a coverpoint with bins for all possible combinations of a 32 bit instruction when Zicbom is not supported.

This applies to several coverpoints in this file.

coverpoints/priv/ExceptionsZicboS_coverage.svh
jordancarlin, line 40: All of these envcfg coverpoints need ifdefs too.

jordancarlin: This is making me question if it makes sense to keep this all as one covergroup. Maybe we should split this into ExceptionsZicbomS and ExceptionsZicbozS (and the same for U). @davidharrishmc, any thoughts?

tests/priv/ExceptionsF/ExceptionsF.S
```diff
@@ -8,6 +8,17 @@
 // SPDX-License-Identifier: Apache-2.0
 ///////////////////////////////////////////
 
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Zicsr, Sm]
```
jordancarlin, line 12: Shouldn’t this require F?

tests/priv/ExceptionsF/ExceptionsF.S
```diff
@@ -8,6 +8,17 @@
...
+# MARCH: rv${XLEN}ifd_zicsr_zfa
+# CONFIG_DEPENDENT: true
+##### END_TEST_CONFIG #####
+
+//Zfa  D, F,    _zfa
```
jordancarlin, line 17: Delete this comment.

tests/priv/ExceptionsF/ExceptionsF.S
jordancarlin, line 89: Comment needs to be updated now that there are ifdefs.

tests/priv/ExceptionsF/ExceptionsF.S
```diff
@@ -106,50 +117,52 @@ fp_instructions:
     sw     t1, 12(t5)
 
     // Initialize floating point registers used
-    li t0, 1
-    fcvt.s.w f0, t0
+    #ifdef F_SUPPORTED
```
jordancarlin, line 120: F should always be supported if this test is run.

jordancarlin (changes requested):

coverpoints/priv/ExceptionsZicboS_coverage.svh
davidharrishmc: I'd prefer not to keep splitting if possible.  We are up to about 250 test suites already.

davidharrishmc:
Replaced with newer python-generated tests.

PR #812, opened 2025-12-28
[ACT4] `cp_imm_edges_jal` and `cp_offset` 100% coverage

jordancarlin:
- Move around some coverpoints for better organization
- Reenable `cp_imm_edges_jal` coverpoint for `jal`
- `c.j` and `c.jal` getting 100% coverage on `cp_imm_edges_c_jal` coverpoint
- Improve self-checking for branch instructions
- `cp_offset` and `cp_offset_lsbs` 100% coverage

This brings all tests that existed in the old cvw-arch-verif repo to 100% coverage. The only tests that are not at 100% are the new compressed hint and Zicclsm tests.


generators/testgen/src/testgen/coverpoints/cp_imm_edges_branch.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@add_coverpoint_generator("cp_imm_edges_branch")
+def make_cp_imm_edges_branch(instr_name: str, instr_type: str, coverpoint: str, test_data: TestData) -> list[str]:
+    """Generate tests for branch immediate edge values."""
+    test_data.add_testcase_string(coverpoint)
+    test_lines: list[str] = ["\n# Testcase cp_imm_edges_branch"]
+    params = generate_random_params(test_data, instr_type, exclude_regs=[0])
+    assert params.rs1 is not None and params.rs2 is not None
+    # TODO: Improve checking logic. Add `addi` instructions between branches and check counter at the end.
```
davidharrishmc: Should we do this before merging?  Some way of knowing the branches are actually executed? Same for everything with control transfer and no ra change.  But only when the distance is big enough that there is room to put in the addi comfortably.

davidharrishmc (changes requested):

generators/testgen/src/testgen/coverpoints/cp_imm_edges_branch.py
jordancarlin: Updated

davidharrishmc (approved):

PR #831, opened 2026-01-06
Replace `rvtest_dtrap_routine` with `rvtest_strap_routine`

jordancarlin:
@allenjbaum @UmerShahidengr I would appreciate one of you taking a look at this because I'm not as familiar with the ACT trap handler. I think the single occurrence of `rvtest_dtrap_routine` is a typo and should be replaced with `rvtest_strap_routine`.


UmerShahidengr (approved):
LGTM. It was a typo there

PR #836, opened 2026-01-08
[ACT4] objdump improvements

jordancarlin:


framework/src/act/makefile_gen.py
```diff
@@ -23,6 +23,8 @@
 .PHONY: compile
 """
 
+OBJDUMP_FLAGS = "-Sd -M no-aliases,numeric"
```
davidharrishmc: Did you want to beef up the flags the way you showed Sadhvi today?

davidharrishmc (changes requested):

framework/src/act/makefile_gen.py
jordancarlin: Done

davidharrishmc (approved):

PR #840, opened 2026-01-08
Updated InterruptsU.S based on the feedback received and cleaned it up

SadhviNarayanan:
Updated InterruptsU.S based on the feedback received and made the tests more consistent with each other.


davidharrishmc:
Let's not push your Makefile because other people will want to be compiling different groups of tests.  Just keep your change local for that.

Is InterruptsU running and getting 100% coverage?  How is InterruptsM?  I'd suggest getting InterruptsM hitting coverage first in its loop form, then writing a generator for testgen to automatically produce it (see the Sm generator in testgen), getting it up to 100% coverage, then moving on to do the same for InterruptsU.

PR #843, opened 2026-01-09
(ACT4) New Trap Handler

Zain2050:
## Description

New features in arch_test are:
* Supports hypervisor tests.
* Supports Self Checking.
* HS and VS Trap handlers.
* Has 2 new save areas for Hypervisor & Guest modes.
* New way to go to M-mode when s-ecall is delegated.
* RVTEST_GOTO_MMODE now checks x3=0.
* Hypervisor 16KB page table added
* New fields in save area: misa, hgatp, mpp, unused, rvmodel_sv.
* Using new TRAP_SIGUPD macro for writing trap signatures.
* Supports custom return address for Fetch guest page fault.


tests/env/utils.h
```diff
@@ -46,15 +46,20 @@
 # FLEN specific macros
 #define FREGWIDTH (FLEN>>3)      // in units of #bytes
 
-#if FLEN==32
+#if (ZDINX | ZFINX | ZHINX)
```
jordancarlin: Zdinx and Zhinx both require Zfinx, so I think this can be simplified to just `#ifdef ZFINX`.

tests/env/utils.h
jordancarlin: Should this TODO section be removed if you are handling Zfinx above now?

tests/env/encoding.h
jordancarlin, line 1: Do not hand-modify this file. Regenerate it from `riscv-opcodes` and if there are additional macros needed then put those in a different file.

tests/env/test_setup.h
```diff
@@ -100,15 +100,27 @@
     #ifdef rvtest_mtrap_routine
       #ifdef rvtest_strap_routine
         #ifdef rvtest_vtrap_routine
-          RVTEST_TRAP_EPILOG V  // actual v-mode prolog/epilog/handler code
+          RVTEST_TRAP_EPILOG V        // actual v-mode prolog/epilog/handler code
         #endif
-        RVTEST_TRAP_EPILOG S    // actual s-mode prolog/epilog/handler code
+        #ifdef rvtest_htrap_routine
+          RVTEST_TRAP_EPILOG H        // actual h-mode prolog/epilog/handler code
+        #endif
```
jordancarlin, line 110: What is the intent with `rvtest_htrap_routine`? When would it be defined instead of or in addition to `rvtest_vtrap_routine`?

tests/env/test_setup.h
```diff
@@ -119,17 +131,12 @@
     SREG    T1, -4(T4)            // save into last signature canary
     j       exit_cleanup          // skip around handlers, go to RVMODEL_HALT
 
-  // Instantiate trap handlers for each priv mode
-  INSTANTIATE_MODE_MACRO RVTEST_TRAP_HANDLER
-
-  // Include test failure handling code
-  RVTEST_FAILURE_CODE
+  // when the text starts, it jumps to this at end of test code to keep test code constant size
```
jordancarlin:
```suggestion
  // when the test starts, it jumps to this at end of test code to keep test code constant size
```

tests/env/test_setup.h
```diff
@@ -119,17 +131,12 @@
...
+  rvtest_entry_pt: 
+      RVMODEL_BOOT          // [boot code] (BOOT code also has RVMODEL macro defs which are >1 op)
```
jordancarlin:
```suggestion
      RVMODEL_BOOT          // Boot code that is of unknown length
```

tests/env/test_setup.h
jordancarlin: Should this be changed to a jump to `rvtest_entry_pt`?

jordancarlin: The `RVMODEL_IO_INIT` call should be moved to `rvtest_entry_pt` to maintain constant size.

tests/env/test_setup.h
```diff
@@ -173,16 +180,24 @@
...
-    #ifdef rvtest_strap_routine
-      // This is a valid global pte entry w/ all permissions. If at root level, it forms an identity map.
-      rvtest_Sroot_pg_tbl:
-        RVTEST_PTE_IDENT_MAP(0,LVLS,RVTEST_ALLPERMS)
-      #ifdef rvtest_vtrap_routine
-        rvtest_Vroot_pg_tbl:
-        RVTEST_PTE_IDENT_MAP(0,LVLS,RVTEST_ALLPERMS)
-      #endif
+  #ifdef rvtest_strap_routine
+//this is a valid global pte entry w/ all permissions. IF at root level, it forms an identity map.
```
jordancarlin: Fix indentation.

tests/env/test_setup.h
```diff
@@ -173,16 +180,24 @@
...
+      .align 14
+      rvtest_Hroot_pg_tbl:
+      RVTEST_PTE_IDENT_MAP(0,LVLS,RVTEST_ALLPERMS)
+      .align 14
+    #endif
+    #ifdef rvtest_vtrap_routine
+      .align 12
+      rvtest_Vroot_pg_tbl:
+      RVTEST_PTE_IDENT_MAP(0,LVLS,RVTEST_ALLPERMS)
     #endif
```
jordancarlin: Why do we need separate `#ifdef`s for `rvtest_htrap_routine` and `rvtest_vtrap_routine`?

tests/env/arch_test.h
```diff
@@ -122,14 +178,14 @@
...
 // set defaults
 #ifndef   NUM_SPECD_INTCAUSES
-  #define NUM_SPECD_INTCAUSES 16
-  #define INT_CAUSE_MSK ((1<<4)-1)
+  #define NUM_SPECD_INTCAUSES 24
+  #define INT_CAUSE_MSK ((1<<5)-1)
 #endif
 
-// set defaults
+        // set defaults
```
jordancarlin:
```suggestion
// set defaults
```

tests/env/arch_test.h
```diff
@@ -138,45 +194,74 @@
 // This only gets executed if xTVEC is not writable to point to the trap trampoline,
 // and if it isn't writable, the model better have the zifencei extension implemented.
 //==========================================================================================
+// RVMODEL_XX macros that are more than 4bytes (if not defaulted) are defined in the RVMODEL_BOOT macro
```
jordancarlin: This seems more complicated than necessary. Why don't we instantiate all of the RVMODEL_* macros after the end of the code and put a label (maybe just the lowercase version of the macro name) before each one. Then when one of the RVMODEL macros is needed in the test, we just jump to that label. We then have a `ret` after the macro before the label corresponding toe the next label.

I'm worried that many people will mess up with the current approach of saying the macros must be a jump to code in the `RVMODEL_BOOT` macro if they are longer than one instruction. We need to make setting up the macros as easy as possible for users.

tests/env/arch_test.h
```diff
@@ -138,45 +194,74 @@
...
-                        & 0xFFFFFFFF)
-#endif
-#ifndef SET_ABS_TVAL_MSK
-#define SET_ABS_TVAL_MSK ((1<<CAUSE_ILLEGAL_INSTRUCTION) & 0xFFFFFFFF)
-#endif
+        // For each cause, this bit mask defines whether the TVAL contents are 
+        // addresses that may need relocation to match reference model values.
+        // Mask bits should be cleared for implementation options that store zero, or for
+        // exceptions that are not addresses, e.g. opcodes/cause values/implentation defined.
+        // Separately, the SET_ZERO_TVAL_MASK must be set for exceptions that always set TVAL to zero
```
jordancarlin: Is all of this logic still needed now that Sail supports a configurable memory map to match the DUT? @allenjbaum @UmerShahidengr

tests/env/arch_test.h
```diff
@@ -256,70 +411,112 @@
 #define _ARG1(_1ST,...) _1ST
 #define NARG(...) _ARG5(__VA_OPT__(__VA_ARGS__,)4,3,2,1,0)
 
+// this implements the self-checking of trap signatures, using a tempreg and the offset in the entry as parameters
+.macro TRAP_SIGUPD _TMPREG_,    _OFF_
+  #ifdef RVTEST_SELFCHK
```
jordancarlin: The rest of the framework has just been using `#ifdef SELFCHECK`. If you want to switch it to `RVTEST_SELFCHECK` (which might be a good idea) then it needs to be updated everywhere else.

tests/env/arch_test.h
```diff
@@ -256,70 +411,112 @@
 #define _ARG1(_1ST,...) _1ST
 #define NARG(...) _ARG5(__VA_OPT__(__VA_ARGS__,)4,3,2,1,0)
 
+// this implements the self-checking of trap signatures, using a tempreg and the offset in the entry as parameters
+.macro TRAP_SIGUPD _TMPREG_,    _OFF_
```
jordancarlin: Also, how about moving this to `signature.h` with all of the other SIGUPD macros?

tests/env/arch_test.h
```diff
@@ -256,70 +411,112 @@
...
-        srli  x15   , \oldreg\(), XLEN-7
-        or    \dstreg\(), \dstreg\(), x15
+.macro DBLSHIFTR dstreg,     oldreg,    tmpreg, shamt       //this is just a rotate  using xtmp as a tmp
+        slli    \tmpreg\(), \oldreg\(),   XLEN-\shamt
+        srli    \dstreg\(), \oldreg\(),        \shamt
+        or      \dstreg\(), \dstreg\(), \tmpreg\()
 .endm
+
 /* init regs, to ensure you catch any errors */
 .macro RVTEST_INIT_GPRS
```
jordancarlin: Since this now sets CSRs in addition to GPRs, it should probably be renamed `RVTEST_INIT_REGS` (or something along those lines).

tests/env/arch_test.h
```diff
@@ -256,70 +411,112 @@
...
+     DBLSHIFTR x14, x13, x15, 7
+     LI (x15, (0xFAB7FBB6FAB7FBB6 & MASK))
 .endm
 
+/*************************************/
+/**** valid mode configs are:     ****/
+/**** M                           ****/
+/**** M  U                        ****/
+/**** MS U                        ****/
+/**** MH U                        ****/
```
jordancarlin, line 447: Does H really mean HS here? Worth clarifying.

jordancarlin (changes requested):
I'm still working my way through this (there are a lot of changes), but here's an initial round of comments.

tests/env/encoding.h
Zain2050: I regenerated it. It produced about 4,500 lines, most of which we don't require but probably will be helpful in the future.

tests/env/arch_test.h
Zain2050: Right. I changed it. Please make sure that I didn't miss it anywhere.

tests/env/test_setup.h
Zain2050: Well, @allenjbaum is the one to implement this. But I think that he kept rvtest_htrap_routine and rvtest_vtrap_routine different, so that tests that don't enter or interact with guest modes can just define rvtest_htrap_routine. It will reduced the size of code including V trap handler and save area.

tests/env/arch_test.h
Zain2050: Most probably. At many places in arch_test, H is used instead of HS and V instead of VS/VU.

tests/env/test_setup.h
Zain2050: Yes, as different DUTs will have different different IO_INITs, it must be placed at rvtest_entry_pt.

Zain2050: Yes, it should be.

tests/env/arch_test.h
Zain2050: @allenjbaum @UmerShahidengr

tests/env/test_setup.h
UmerShahidengr: rvtest_htrap_routine is a flag showing that H-extension is enabled, and HS-mode is supported (along with the regular U-mode), vtrap_routine is showing that VS/VU-modes are also supported. We have defined an extra save area for VS/VU-mode for handling 2-stage address translation.

tests/env/arch_test.h
UmerShahidengr: RVTEST_SELFCHECK will be a better name. Thank you @Zain2050 for renaming it

tests/env/arch_test.h
UmerShahidengr: Yes, H means HS

tests/env/test_setup.h
jordancarlin: It makes sense to have two separate save areas, but I don't see the need to have separate flags for them. If H mode is supported, then VS/VU modes are always also supported. It seems like it would simplify the interface to have a single `#define` for both of them (probably `rvtest_htrap_routine` to be consistent with the extension name).

tests/env/arch_test.h
jordancarlin: @Zain2050 There is at least one more instance of `SELFCHECK` in `makefile_gen.py`. Can you do a check for any other instances of `SELFCHECK` that need to be updated?

tests/env/arch_test.h
allenjbaum: Ah - this might be the case; relocation isn't required if the memory maps match. Note that this does mean that self-checking tests that trap will always need to run a reference model configured like the DUT to capture expected results before running tests on the DUT (but you knw that...)

tests/env/arch_test.h
allenjbaum: I'm a bit puzzled; the rest of the comments outline exactly what you propose, in a bit more detail (saving and restoring registers, etc). I just noticed that if an RVMODEL macro is expected to work in Umode, it isn't possible to save and restore using the approach I outlined; they will need to have a dedicated register point to a save area (or they and the tests will have to be written to not need save anything. I think that may be a problem...

All RVMODEL macro definitions must be inside RVMODEL_BOOT (there isn't any other place DUT provided code gets loaded) , and that has already been placed after the test.
This first sentence is just an (admittedly unnecessary) optimization to replace the JAL with the actual macro definition if it is a single op, the same size as the JAL.

tests/env/arch_test.h
allenjbaum: The reason there was both SIGUPD and TRAP_SIGUPD was that my original implementation used an ECALL to get to the failure code. With the removal of the requirement to be able to continue after a test failure, this isn't necessary anymore.

tests/env/arch_test.h
allenjbaum: Agreed; it made a lot of the code simpler because it was always a single character.
Note that an implementation might support HS/VS and S - just not at the same time (if misa.H is RW).

Also there is at least one place in the test where V is used to indicate a "variable" segment (segment=code/data/signature/variable) where the V segment can be configured to be inside any of them.
How necessary that is now that we aren't doing relocation is unclear to me, but probably isn't needed any longer.

tests/env/arch_test.h
jordancarlin: @allenjbaum What I'm suggesting is that instead of defining the macros as jumps into RVMODEL_BOOT, the macros are defined to do whatever the DUT needs and the test itself jumps to a label where the macro is instantiated (instead of instantiating it inline). For example:

```c
// main test code
jal ra, rvmodel_clr_msw_int
.....


// At end of test
rvmodel_clr_msw_int:
  RVMODEL_CLR_MSW_INT
  ret
```

That way the user does not need to think about where code is going to be instantiated.

allenjbaum:
I'm still unclear how that is different than what is described in the
following comments.
The actual code for RVMODEL_XX macros will reside inside RVMODEL_BOOT (
provided by the vendor, #included after the actual test) at label
rvmodel_xx: .

For tests to execute it, they jal to that label - just as you describe (so
not a jump, but a jal, but you probably meant that)
The special case  I documented was if the macro was a single instruction,
so it wasn't necessary to JAL to it, it could just be instantiated inline.
This is just an optimization, and likely wouldn't save much. If you want to
remove that comment, feel free.

Ah, the difference is the way the test looks, but not the actual mechanics

// main test code
RVMODEL_CLR_MSW_INT (which is always #defined to be jal ra, rvmodel_clr_msw_int)

.....


// At end of test ***inside RVMODEL_BOOT**
rvmodel_clr_msw_int:
  code provided by vendor
  ret

 It appears to be different in the code, but always does exactly the same thing.

Either way works. A caveat either way is that ra must be reserved in the test.

davidharrishmc:
This trap handler is on the critical path for merging the VM and PMP tests and also for other Phase 1 tests.  @Zain2050 would you be able to push on it daily until it is merged?  @allenjbaum @UmerShahidengr @jordancarlin could you keep an eye on it and review quickly when it needs your attention?

generators/testgen/src/testgen/coverpoints/special/cp_custom_sc.py
```diff
@@ -192,6 +192,7 @@ def make_custom_sc(instr_name: str, instr_type: str, coverpoint: str, test_data:
                 load_int_reg("rs2", params.rs2, params.rs2val, test_data),
                 load_int_reg("temp_reg", params.temp_reg, params.temp_val, test_data),
                 f"LA(x{params.rs1}, scratch) # rs1 = base address",
+                "nop",
```
jordancarlin: Can you add a comment with a TODO explaining why this `nop` is being added and a link to the spike issue once you create it?

tests/env/utils.h
```diff
@@ -46,15 +46,20 @@
 # FLEN specific macros
 #define FREGWIDTH (FLEN>>3)      // in units of #bytes
 
-#if FLEN==32
+#ifdef ZFINX
+  #define FLREG LREG
+  #define FSREG SREG
```
jordancarlin, line 51: This doesn’t seem quite right for Zfinx. What if you have rv32i_zdinx or rv64i_zfinx? In either case, loading an fp value is not the same as LREG.

tests/env/test_setup.h
```diff
@@ -120,24 +132,13 @@
...
-    RVMODEL_BOOT
+  // When the test starts, it jumps here (to the end of the test)
+  // Here we can place model specific macros, so that the code size remains constant
+  rvtest_entry_pt:
+    RVMODEL_BOOT                  // Boot code that is of unknown length
     RVMODEL_IO_INIT(T1, T2, T3)
-    jr x1
-    nop // Padding to ensure valid memory after jr in case it's at the edge of the .text section
+    LA (T1, rvtest_init)
+    jr T1                         // Jump back to the start of the test
```
jordancarlin: After lots of discussion, we decided to give each RVMODEL macro its own label that we jump to when we want to execute that function instead of having all of that code live in RVMODEL_BOOT. Since the other RVMODEL macros will have their own labels/instantiations, the comments and label here can be reverted back to being specific to boot code.

tests/env/test_setup.h
```diff
@@ -182,16 +183,24 @@
   // TODO: Is this still needed?
```
jordancarlin: Seems like this TODO can be dropped? I these page tables are being used since you are adding more to them.

tests/env/test_setup.h
```diff
@@ -182,16 +183,24 @@
...
   #ifndef RVTEST_NO_IDENTY_MAP
-    #ifdef rvtest_strap_routine
-      // This is a valid global pte entry w/ all permissions. If at root level, it forms an identity map.
-      rvtest_Sroot_pg_tbl:
-        RVTEST_PTE_IDENT_MAP(0,LVLS,RVTEST_ALLPERMS)
-      #ifdef rvtest_vtrap_routine
-        rvtest_Vroot_pg_tbl:
-        RVTEST_PTE_IDENT_MAP(0,LVLS,RVTEST_ALLPERMS)
-      #endif
+  #ifdef rvtest_strap_routine
```
jordancarlin: Inconsistent indentation with all of these ifdefs.

tests/env/test_setup.h
jordancarlin: It's still unclear to me why we need separate `#ifdefs` for `rvtest_htrap_routine` and `rvtest_vtrap_routine`. I understand the need for both save areas, but why not always instantiate both of them if we are doing a hypervisor test? The cost to test size is minimal and it seems like it would make debugging easier if the trap handler is always set up to handle both (assuming hypervisor is enabled).

tests/env/arch_test.h
```diff
@@ -17,97 +18,152 @@
...
-        rvtest_[M/V/S]trap_routine
-        GOTO_[M/S/U]MODE, INSTANTIATE_MODE_MACRO (prolog/handler/epilog/savearea)
+        RVMODEL_SET_[M/H/S/V][SW]_INT
+        RVMODEL_CLR_[M/H/S/V][SW/TIMTER/EXT]_INT
+        rvtest_[M/H/S/V]trap_routine
+        GOTO_[M/H/S/U]MODE, INSTANTIATE_MODE_MACRO (prolog/handler/epilog/savearea)
+   The following are custom macros for selfchk mode
+        RVMODEL_SELFCHECK
+        RVMODEL_WR_STR  optional for status writes
+        RVMODEL_WR_BIN  optional for status writes
```
jordancarlin: A lot of these comments are no longer accurate.

tests/env/arch_test.h
```diff
@@ -17,97 +18,152 @@
...
-#define T1      x6
-#define T2      x7
-#define T3      x8
-#define T4      x9
-#define T5      x10
-#define T6      x11
+
+#ifndef T1
+  #define T1      x6
+#endif
```
jordancarlin, line 145: What's the rationale for adding all of the `#ifndef` checks? Are we expecting `T1`, etc. to be defined elsewhere?

tests/env/arch_test.h
jordancarlin: After many more discussions, we decided to go with the approach that each `RVMODEL_*` macro is instantiated at an appropriate `rvtest_*` label (with the name otherwise being the same as the macro). These labels are jumped to instead of calling the RVMODEL macros directory. We are creating RVTEST macro wrappers that do the actual jump. This makes it easier for ACT users to create their macros without thinking about the size.

tests/env/arch_test.h
```diff
@@ -138,45 +194,74 @@
...
-#define SET_REL_TVAL_MSK ((1<<CAUSE_MISALIGNED_FETCH | 1<<CAUSE_FETCH_ACCESS                             | 1<<CAUSE_BREAKPOINT   | \
-                           1<<CAUSE_MISALIGNED_LOAD  | 1<<CAUSE_LOAD_ACCESS  | 1<<CAUSE_MISALIGNED_STORE | 1<<CAUSE_STORE_ACCESS | \
-                           1<<CAUSE_FETCH_PAGE_FAULT | 1<<CAUSE_LOAD_PAGE_FAULT                          | 1<<CAUSE_STORE_PAGE_FAULT) \
-                        & 0xFFFFFFFF)
-#endif
-#ifndef SET_ABS_TVAL_MSK
-#define SET_ABS_TVAL_MSK ((1<<CAUSE_ILLEGAL_INSTRUCTION) & 0xFFFFFFFF)
-#endif
+        // For each cause, this bit mask defines whether the TVAL contents are
+        // addresses that may need relocation to match reference model values.
```
jordancarlin, line 209: Are relocations to match the ref model ever needed now? The ref model memory map should be the same as the DUT memory map, so the addresses should be the same. Can this be removed?

tests/env/arch_test.h
```diff
@@ -138,45 +194,74 @@
...
 #ifndef _ADDR_SZ_
-  #if XLEN==64
-    #define _ADDR_SZ_ 57
-  #else
+  #if XLEN==32
     #define _ADDR_SZ_ 32
+  #else
+    #define _ADDR_SZ_ 57
   #endif
 #endif
```
jordancarlin: If this is the `MAX(phy_addr_size, VADDR_SZ)`, shouldn't XLEN=32 be 34? If not, please clarify the comment.

UmerShahidengr:
@Zain2050 , @jordancarlin what is left in this PR? It seems like this trap handler is good to go, can we merge this one?

tests/env/arch_test.h
```diff
@@ -260,66 +415,95 @@
...
-.macro RVTEST_INIT_GPRS
+.macro RVTEST_INIT_REGS
+    #ifdef rvtest_mtrap_routine
+     LI  (x1, 0)
+     // Initialising CSR registers (mpec, mtval, mstatus, mip)
+     csrw  CSR_MSTATUS,    x1
+     csrw  CSR_MEPC,       x1
+     csrw  CSR_MIP,        x1
+     csrw  CSR_MTVAL,      x1
+     csrw  CSR_MCAUSE,     x1
```
jordancarlin: Could simplify this to use a `csrwi` and drop the `LI`.

tests/env/arch_test.h
```diff
@@ -260,66 +415,95 @@
...
 #ifdef RVTEST_ENAB_INSTRET_CNT
      csrr  x14, CSR_MSCRATCH
      csrr  x15, CSR_MINSTRET
-     SREG  x15, tramp_sz+4*8(x14)               // this replaces initial canary val w/ instret counter val
-
-     DBLSHIFT7 x14, x13
-     LI (x15, (0xFAB7FBB6FAB7FBB6 & MASK))
+     LREG  x14, sig_bgn_off(x14)        // pointer to sig begin
+     SREG  x15, -8(x14)                 // this replaces initial canary val w/ instret counter val
 #endif
```
jordancarlin: Is `RVTEST_ENAB_INSTRET_CNT` ever going to be usable in self-checking mode? If not, let's drop it. I don't think it will work as is.

tests/env/arch_test.h
```diff
@@ -260,66 +415,95 @@
...
-      #ifdef rvtest_vtrap_routine
-        \MACRO_NAME V // actual v-mode prolog/epilog/handler code
-      #endif
+    \MACRO_NAME M         // actual m-mode prolog/epilog/handler code
+  #endif
+  #ifdef rvtest_strap_routine
+    #ifdef rvtest_htrap_routine
+      \MACRO_NAME H     // actual s-mode prolog/epilog/handler code
+    #endif
+    \MACRO_NAME S // actual v-mode prolog/epilog/handler code
```
jordancarlin: Comments seem messed up.

tests/env/arch_test.h
```diff
@@ -341,82 +525,130 @@
...
+//| Int Deleg | midleg    | hideleg   |      sideleg       | separate for, mnged by H
+//| EnvCfg    | menvcfg   | henvcfg   |      senvcfg       | separate for, mnged by H
+//| Cnter-En  | mcounteren| hcounteren|   scounteren       | separate for, mnged by H
+//+-----------+-----------+-----------+---------+----------+
+//| TrapVec   | mtvec     |         stvec       | vstvec   |    HS/S shared
+//| Scratch   | mscratch  |         sscratch    | vsscratch|    HS/S shared
+//| ExcepPC   | mepc      |         sepc        | vsepc    |    HS/S shared
+//| TrapCause | mcause    |         scause      | vscause  |    HS/S shared
+//+-----------+-----------+-----------+---------+----------+
+//| Timecmp   | mtimecmp  |     x     |mtimecmp |vstimecmp |    new for H
```
jordancarlin: Should this be `stimecmp` for S? Worth a comment explaining this one because mtimecmp is MMIO while stimecmp and vstimecmp are CSRs.

tests/env/arch_test.h
```diff
@@ -528,119 +760,158 @@
...
-/**** determines the caller's mode, uses it to select it's CODE_BEGIN, and uses ****/
-/**** to calculate it offset from Mmode's CODE_BEGIN, adjusts MEPC by that amt  ****/
-/**** to convert it to an Mmode address, restores saved regs, and branches to   ****/
-/**** the relocated addr+4, immediately following the ECALL, but now in Mmode   ****/
-/**** **NOTE**: this destroys T2 and clears x2 (AKA sp)                                               ****/
-/**** **NOTE**: this works from any mode but MUST not be used if                ****/
-/****       medeleg[<GOTO_M_OP_cause>]==1 to prevent infinite delegation loops.   ****/
+/**** This must be used before using RVTEST_GOTO_LOWER_MODE and at CODE_END.    ****/
+/**** It sets a1 to 0 to signal that this is not an explict ECALL, and that it  ****/
+/**** return normally. The handler will check that trap cause==ecallx, & divert ****/
```
jordancarlin:
```suggestion
/**** return normally. The handler will check that trap cause==ecall, & divert ****/
```

tests/env/arch_test.h
```diff
@@ -528,119 +760,158 @@
...
+/**** to a spcl rtn_fm_mmode: handler if x3=0. That code translates MEPC from   ****/
```
jordancarlin, line 734: Inconsistent use of abi (`a1`) and architectural (`x3`) register names in this comment.

tests/env/arch_test.h
```diff
@@ -528,119 +760,158 @@
...
 /****  with an op that causes a different exception cause that isn't delegated. ****/
 /***********************************************************************************/
 
+#ifndef GOTO_M_OP
+    #define GOTO_M_OP   ecall  // default; this must be called with x3=0
+#endif
+
+#ifndef CAUSE_SPCL_GO2MMODE_OP // make sure this default can be overwritten (e.g. to illegal fetch addr)
+    #define ALT_GOTO_M_CAUSE CAUSE_ILLEGAL_INSTRUCTION
+    #define ALT_GOTO_M_OP    .word 0
```
jordancarlin: Should this be `.insn` since it represents an instruction instead of `.word`?

tests/env/arch_test.h
```diff
@@ -528,119 +760,158 @@
...
 
     GOTO_M_OP                   /* ECALL: traps always, but returns immediately to */
-                                /* the next op if x2=0, else handles trap normally */
-    mv   x2, t0                 /* FIXME: Hacky way to preserve x2 as stack pointer by trashing t0 instead */
-
- #endif
+                                /* the next op if x3=0, else handles trap normally */
+    mv   x3, t0
+#endif
 .option pop
```
jordancarlin: Indent everything inside the macro.

tests/env/arch_test.h
```diff
@@ -528,119 +760,158 @@
...
-    mv   t0, x2                 /* FIXME: Hacky way to preserve x2 as stack pointer by trashing t0 instead */
-    li   x2, 0                  /* Ecall w/x2=0 is handled specially to rtn here */
-// Note that if illegal op trap is delegated , this may infinite loop
-// The solution is either for test to disable delegation, or to
-// redefine the GOTO_M_OP to be an op that will trap  to mmode
+    mv   t0, x3                 /* FIXME: Hacky way to preserve x3 by trashing t0 instead */
+    li   x3, 0                  /* Ecall w/x3=0 is handled specially to rtn here */
+// Note that if ecalls are delegated , this may infinite loop
+// The solution is either for test to disable delegation, or to redefine GOTO_M_OP
+// to be an op that will trap to mmode (e.g. jump to odd address, or .word 0)
```
jordancarlin: Is this comment still true? Seems like the solution now is to use `RVTEST_GOTO_DELEGATED_MMODE` instead?

tests/env/arch_test.h
```diff
@@ -528,119 +760,158 @@
...
+.option pop
+.endm
 
-/**** This is a helper macro that causes harts to transition from    ****/
-/**** M-mode to a lower priv mode at the instruction that follows    ****/
-/**** the macro invocation. Legal params are VS,HS,VU,HU,S,U.        ****/
-/****  This uses T1,T2&T4. The H,U variations leave V unchanged.     ****/
-/**** NOTE: this MUST be executed in M-mode. Precede with GOTO_MMODE ****/
-/**** FIXME - SATP & VSATP must point to the identity map page table ****/
+/**** This is a helper macro that causes transition harts from M-mode       ****/
```
jordancarlin:
```suggestion
/**** This is a helper macro that causes harts to transition from M-mode       ****/
```

tests/env/arch_test.h
```diff
@@ -528,119 +760,158 @@
...
+/**** to the following instruction, at lower priv mode. Legal params are    ****/
+/**** VSmode,HSmode,VUmode, HUmode,Smode & Umode. The H,U variations leave  ****/
```
jordancarlin: What is HUmode?

tests/env/arch_test.h
```diff
@@ -1334,76 +1756,102 @@ spcl_\__MODE__\()handler:               // case table branch to special handler
...
+        .dword  1                        // int cause  6 no vmode
+        .dword  \__MODE__\()clr_Mtmr_int // int cause  7  Mmode Tmr int
+ //****************************************************************
+        .dword  0                        // int cause  8 is reserved, error
+        .dword  \__MODE__\()clr_Sext_int // int cause  9  Smode Ext int
+        .dword  1                        // int cause  A no vmode
+        .dword  \__MODE__\()clr_Mext_int // int cause  B  Mmode Ext int
+ //****************************************************************
+  #else
+    #if defined(rvtest_strap_routine)   // M/S/U only
```
jordancarlin: Can this be a single case with an or instead of two identical interrupt tables?

tests/env/arch_test.h
```diff
@@ -1334,76 +1756,102 @@ spcl_\__MODE__\()handler:               // case table branch to special handler
...
+        .dword  1                        // int cause  A no vmode
+        .dword  \__MODE__\()clr_Mext_int // int cause  B  Mmode Ext int
 //****************************************************************
+    #endif
+  #endif
 #endif
+
  .rept NUM_SPECD_INTCAUSES-0xC
-        .dword  1                       // int cause c..NUM_SPECD_INTCAUSES is reserved, just return
+        .dword  1                       // int cause c..NUM_SPECD_INTCAUSES is rsvd, just rrtn
```
jordancarlin: Not sure why this comment was changed. Was more clear before. Why is `rrtn` being used as short for return?

tests/env/arch_test.h
```diff
@@ -1475,7 +1925,7 @@ excpt_\__MODE__\()hndlr_tbl:            // handler code should only touch T2..T6
 
 \__MODE__\()clr_Sext_int:               // int 9 default to just return after saving IntID in T3
         RVMODEL_CLR_SEXT_INT
-        SREG    T3, 3*REGWIDTH(T1)      // save 4rd sig value, (intID)
+        TRAP_SIGUPD(T4, 3)              // save 4rd sig value, (intID)
```
jordancarlin: Several other instances of this.
```suggestion
        TRAP_SIGUPD(T4, 3)              // save 4th sig value, (intID)
```

tests/env/arch_test.h
```diff
@@ -1550,52 +2011,70 @@ excpt_\__MODE__\()hndlr_tbl:            // handler code should only touch T2..T6
...
+        LREG    T4, 4+xedeleg_sv_off(T1)        // get saved xedelegh
 #endif
-  .else
-//FIXME: if Umode-int-extension or anything like it is implemented, uncomment the following
-//      csrw    CSR_XEDELEG, T2 //this handles S  mode restore
+.ifnc \__MODE__ , S
+  .ifnc \__MODE__ , V
+        csrw    CSR_XEDELEG,  T2
+// TODO: Commenting this code only because CSR_M/S/HEDELEGH is not defined in reference + spike
+// Writing this CSR is causing infinite traps of illegal access
```
jordancarlin, line 1990: Is there an issue opened with Spike for this yet? If so, it's good to include the issue link here. If not, let's make sure to open one.

tests/env/arch_test.h
```diff
@@ -1550,52 +2011,70 @@ excpt_\__MODE__\()hndlr_tbl:            // handler code should only touch T2..T6
...
+// #if (XLEN==32)
+//         csrw    CSR_XEDELEGH, T4
+// #endif
```
jordancarlin, line 1993: I don't think this will work as is for `hedelgh` because access to it is controlled by `mstateen0`.

tests/env/signature.h
```diff
@@ -51,6 +51,23 @@
...
+// Used to compare/write signatures while handling traps.
+// In Self Check mode, compare reference and DUT signatures and jump to
+// test_failure in case of a mismatch.
+// If not in Self Check mode, just store the signatures to the signature region
+#ifdef RVTEST_SELFCHECK
+  #define TRAP_SIGUPD(_TMPREG, _OFF)                         \
+    LREG       _TMPREG, _OFF*REGWIDTH(T1)                   ;\
+    beq        _TMPREG, T3, 2f                              ;\
+    jal  T2, failedtest_x5_x4                               ;\
+    2:                                                      ;
```
jordancarlin: This will need some further changes for proper error reporting. Fine to defer to a future PR, but I want to make sure it isn't forgotten.

jordancarlin (dismissed):
Finally got through this PR. This should be my last big round of feedback. Once it's addressed we should try to get this in ASAP since several other PRs are depending on it.

tests/env/utils.h
UmerShahidengr: @allenjbaum wrote that part, and I am quite sure this nested-if-else are fine, but I will wait for Allen's comment on this one.

tests/env/arch_test.h
Zain2050: @allenjbaum

tests/env/arch_test.h
Zain2050: @allenjbaum

tests/env/arch_test.h
jordancarlin: Actually, can’t rv64 have physical addresses that are 64 bits long? Why is it 57 right now? The comment about this was coming from the YAML also needs to be updated.

Zain2050: Yes, this doesn't make much sense. I don't know why it was originally added.

tests/env/arch_test.h
Zain2050: Well, in self-checking mode, I think this stored minstret value could be read back and a delta can be calculated to make sure that the correct number of instructions retired.

tests/env/signature.h
Zain2050: Done

tests/env/arch_test.h
Zain2050: Done

tests/env/arch_test.h
jordancarlin: Any RVMODEL macros that are used between when we start counting and end counting would break this. At a minimum, that means we need to be sure we only enable it for unpriv tests since the trap handler uses a number of RVMODEL macros. The current form (where it replaces the initial canary value) also wouldn't work well because we compare that initial canary at the beginning of the test to ensure the signature was loaded correctly.

We also have a dedicated Zicntr test that checks instret before and after a small set of instructions, so I'm not sure that this really adds much to the test suite.

tests/env/arch_test.h
jordancarlin: It looks like everything inside the #ifdef is indented, but not everything inside the actual assembly macro.

tests/env/arch_test.h
UmerShahidengr: This is not getting used anywhere in the latest trap handler, so we can safely remove this variable, probably it used to show the Physical Address size for RV32/Rv64 but it was never referred in any macro.

tests/env/arch_test.h
UmerShahidengr: I think no issue was added in either Spike or sail, I will follow up on this one

tests/env/arch_test.h
UmerShahidengr: I think for this particular line, mstateen0 is not required. hedelegh is just a normal CSR getting used to read/write a certain bit for signature matching, we dont require any additional line here

jordancarlin: Section 4.2 of the priv spec says

> The P1P13 bit in `mstateen0` controls access to the `hedelegh` introduced by Privileged Specification Version 1.13.

That seems to mean `hedelegh` is not readable or writable if `mstateen0.P1P13` is not set and the Smstateen extension is enabled. Are you interpreting that differently?

allenjbaum:
This was only used to initialize satp.mode field and calculate the number
of pages tables levels there are, and is specifically for VAs, not PAs

tests/env/arch_test.h
Zain2050: I just uncommented it to test it out, but it turns out the encoding.h file that I regenerated doesn't define CSR_MEDELEGH, while it has HEDELEGH. Pretty weird. Should I hand-modify the encoding file or just add a comment as a reminder to fix this piece of code later.

Zain2050:
@jordancarlin I was going through the code and I realized that there was a bug in TRAP_SIGUPD macro due to which DUT signature was being compared to itself instead of the reference. The fix was simple. However, I started getting mismatches. It seems that `#include SIGNATURE_FILE` is missing for `mtrap_sigptr` in test_setup.h. Why is the framework is not embedding trap signatures into the elf?

jordancarlin:
> @jordancarlin I was going through the code and I realized that there was a bug in TRAP_SIGUPD macro due to which DUT signature was being compared to itself instead of the reference. The fix was simple. However, I started getting mismatches. It seems that `#include SIGNATURE_FILE` is missing for `mtrap_sigptr` in test_setup.h. Why is the framework is not embedding trap signatures into the elf?

Self-checking had not been implemented for the trap handler yet, so the signature was not included. Good catch and that is definitely something that needs to be addressed for self-checking priv tests. Can you handle that in this PR? Or do you want to disable the checking portion of `TRAP_SIGUPD` and add that in a follow up PR?

Zain2050:
Right. I'll probably implement that in a follow up PR because it might take some time, and we don't want to stall this PR further as a lot of work depends on it.

jordancarlin:
Sounds good. Let's make `TRAP_SIGUPD` a `nop` for now and add a TODO comment.

tests/env/arch_test.h
Zain2050: @UmerShahidengr suggested to keep this one.

tests/env/arch_test.h
Zain2050: This will be removed in a follow up PR by @UmerShahidengr

tests/env/arch_test.h
Zain2050: Changed ADDR_SZ to VA_SZ.

tests/env/arch_test.h
jordancarlin: I'd rather not hand-modify the file because the change will likely be forgotten the next time it is updated. Maybe add a comment for now and also open an issue/PR for riscv-opcodes.

Zain2050: Currently medelegh is not being generated by riscv-opcodes. Will fix it on riscv-opcodes and push the change later on and uncomment the line after making sure it works.

tests/env/utils.h
Zain2050: Resolving it assuming that it's correct. Will push a patch if it breaks anything.

tests/env/test_setup.h
jordancarlin: This seems like the last big outstanding comment. @Zain2050 @UmerShahidengr

UmerShahidengr: rvtest_htrap_routine will define a save area for HS/Umode, whereas rvtest_vtrap_routine defines the save area for VSmode when V=1. When you have VS/VU mode, you always require an extra save area to store items for regular Umode, thats why we require 2 save areas for all VS/VU tests.

UmerShahidengr: I am resolving this one for now, we can come back to this code section if it will start bugging test developers

UmerShahidengr (approved):
LGTG

tests/env/arch_test.h
Zain2050: https://github.com/riscv/riscv-opcodes/pull/407

PR #852, opened 2026-01-15
Phase 0 Parameters in CTP, fix name of coverpoint

davidharrishmc:


ctp/src/a.adoc
```diff
@@ -36,13 +38,36 @@ include::{testplansdir}/Zaamo.adoc[]
...
+
+TODO: On implementations that support the C extension, compressed forms of the
 I instructions permitted inside constrained LR/SC sequences, as
-described in <<sec:lrscseq>>, are also perm
+described in <<sec:lrscseq>>, are also permitted.
+
+*UDB Parameters:* None. The LRSC_MISALIGNED_BEHAVIOR parameter affects the expected results of ExceptionsZalrsc/cp_{load/store}_address_misaligned (<<t-ExceptionsZalrsc-coverpoints>>).
+LRSC_RESERVATION_STRATEGY affects the expected results of sc_w/cp_custom_sc_address.
+LRSC_FAIL_ON_NON_EXACT_LRSC affects the expected results of sc_d/cg_cp_custom_sc_lrsc and sc_w/cp_custom_sc_address.
+LRSC_FAIL_ON_VA_SYNONYM is not tested.
```
jordancarlin: We should make these bulleted lists or tables or something like that. Otherwise, I think asciidoc renders them as a single paragraph with no line breaks.

ctp/src/a.adoc
```diff
@@ -36,13 +38,36 @@ include::{testplansdir}/Zaamo.adoc[]
...
 
-***
-On implementations that support the C extension, compressed forms of the
+
+TODO: On implementations that support the C extension, compressed forms of the
 I instructions permitted inside constrained LR/SC sequences, as
-described in <<sec:lrscseq>>, are also perm
+described in <<sec:lrscseq>>, are also permitted.
+
+*UDB Parameters:* None. The LRSC_MISALIGNED_BEHAVIOR parameter affects the expected results of ExceptionsZalrsc/cp_{load/store}_address_misaligned (<<t-ExceptionsZalrsc-coverpoints>>).
```
jordancarlin: Seems like this isn't none?

README.md
```diff
@@ -209,6 +209,10 @@ The ACT Framework uses a selection of assembly macros to run DUT-specific code t
 - `RVMODEL_DATA_SECTION`
 - `RVMODEL_BOOT`
 
+**Timer Macros**: Can be left blank if timer is not tested.
```
jordancarlin: This is more consistent with how we phrased the other macros. We should always be testing it if it is supported.

```suggestion
**Timer Macros**: Can be left blank if `mtime` is not supported.
```

ctp/src/trickbox.adoc
```diff
@@ -26,6 +26,7 @@ The trick box is implemented with DUT-specific macros in given in <<t-trickbox>>
 |RVMODEL_HALT_FAIL|Terminate test with a fail indication.  When the test is run in simulation, this should end the simulation.
 |RVMODEL_IO_INIT(_R1, _R2, _R3)|Initialization steps needed prior to writing to the console. _R1, _R2, and _R3 are temporary registers that may be trashed by the macro.
 |RVMODEL_IO_WRITE_STR(_R1, _R2, _R3, _STR_PTR)|Write a null-terminated ASCII string to the console, where _R1, _R2, and _R3 are temporary registers that may be trashed by the macro and _STR_PTR is a register containing the address of the first byte of the string
+|RVMODEL_SET_MTIME(_R1, _R2)|Set the machine timer (mtime) to the value in the register _R1. _R2 is a temporary register that may be trashed by the macro.
```
jordancarlin:
```suggestion
|RVMODEL_SET_MTIME(_R1, _R2)|Set the machine timer (`mtime`) to the value in the register _R1. _R2 is a temporary register that may be trashed by the macro.
```

generators/testgen/src/testgen/asm/csr.py
```diff
@@ -28,8 +28,8 @@ def gen_csr_read_sigupd(check_reg: int, csr_name: str, test_data: TestData) -> s
     """
     test_data.sigupd_count += 1
     return (
-        f"# Read {csr_name} into x{check_reg} and check against expected.\n"
-        f"RVTEST_SIGUPD_CSR_READ({csr_name}, x{check_reg}, test_{test_data.test_count}) "
+        f"\t# Read {csr_name} into x{check_reg} and check against expected.\n"
```
jordancarlin, line 31: We need to think about the formatting for this. So far we've been avoiding tabs in all of our code. Eventually we are going to check the generated tests into the repo, and this version would have tabs in those generated tests. We could exclude the generated tests from those checks, or we could have the generator produce spaces instead of tabs.

Doesn't need to be resolved now, but something to think about.

generators/testgen/src/testgen/priv/extensions/sm.py
```diff
@@ -338,6 +368,202 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+            f"{coverpoint}",
+            "Attempt to write read-only CSRs.  Should throw illegal instruction",
+        ),
+    )
+
+    lines.append("\tLI(t0, -1)          # t0 = all 1s\n")
+    for csr in range(0xC00, 0x1000):
+        lines.extend(
+            [
+                f"\n{covergroup}_{coverpoint}_{csr}:\n",
```
jordancarlin: Can this use the `test_data.add_testcase` function instead of hardcoding the label?

generators/testgen/src/testgen/priv/extensions/sm.py
```diff
@@ -338,6 +368,202 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+        ]
+    )
+    for csr in cntrsh:
+        lines.extend(csr_access_test(test_data, csr, covergroup, coverpoint))
+
+    lines.extend(
+        [
+            "#endif",
+        ]
+    )
```
jordancarlin:
```suggestion
    lines.append("#endif")
```

generators/testgen/src/testgen/priv/extensions/sm.py
```diff
@@ -338,6 +368,202 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+    for csr in range(0xC00, 0x1000):
+        lines.extend(
+            [
+                f"\n{covergroup}_{coverpoint}_{csr}:\n",
+                f"\tCSRW(0x{csr:03x}, t0)    # attempt to write read-only CSR {csr:03x}; should get illegal instruction\n",
+            ]
+        )
+
+    # Test access to all M-mode counters
+    coverpoint = "cp_cntr_access"
```
jordancarlin, line 430: Might be worth splitting this into another function here when you start the cntr coverpoints to make the logic easier to follow.

generators/testgen/src/testgen/priv/extensions/sm.py
```diff
@@ -338,6 +368,202 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+    coverpoint = "cp_inhibit_mcycle"
+    lines.append(
+        comment_banner(
+            f"{coverpoint}",
+            "Inhibit mcycle",
+        ),
+    )
+    lines.extend(
+        [
+            f"{covergroup}_{coverpoint}:",
```
jordancarlin: Can this use the test_data.add_testcase function instead of hardcoding the label?

generators/testgen/src/testgen/priv/extensions/sm.py
```diff
@@ -338,6 +368,202 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+    coverpoint = "cp_inhibit_minstret"
+    lines.append(
+        comment_banner(
+            f"{coverpoint}",
+            "Inhibit minstret",
+        ),
+    )
+    lines.extend(
+        [
+            f"{covergroup}_{coverpoint}:",
```
jordancarlin: Can this use the test_data.add_testcase function instead of hardcoding the label?

generators/testgen/src/testgen/priv/extensions/sm.py
```diff
@@ -338,6 +368,202 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+    coverpoint = "cp_mtime_write"
+    lines.append(
+        comment_banner(
+            f"{coverpoint}",
+            "Write mtime and read back time",
+        ),
+    )
+    lines.extend(
+        [
+            f"{covergroup}_{coverpoint}:",
```
jordancarlin: Can this use the test_data.add_testcase function instead of hardcoding the label?

jordancarlin (changes requested):

generators/testgen/src/testgen/asm/csr.py
davidharrishmc: Postpone

jordancarlin (approved):

PR #856, opened 2026-01-19
CV32e20 config for ACT4.0

ateeb04:
This PR provides config files to run ACT4.0 certification tests on CV32E20. The config is aligned with core testbench of the CV32E20, which can be checked out [here](https://github.com/openhwgroup/cv32e20-dv).


config/duts/cve2/cv32e20/cv32e20.yaml
jordancarlin, line 1: Looks like this is still based on the older version of UDB. UDB recently finished a significant refactor, and there were some changes to the config files. The latest version of the act4 branch of riscv-arch-test is now using the newer UDB version, so this will need to be updated. It should give you detailed error messages if you pull the latest version into your branch.

jordancarlin (commented):
Not sure that we should be checking these config files into the riscv-arch-test repo, but fine for purposes of review for now.

jordancarlin:
Superseded by #897. Closing.

PR #875, opened 2026-01-27
[ACT4] Add option to exclude parts of priv testsuite name in extensions list

jordancarlin:


generators/testgen/src/testgen/constants.py
```diff
@@ -51,6 +51,14 @@
     "Misalign": "MISALIGNED_LDST: true",
 }
 
+# Priv test name components to exclude from extension list
+PRIV_TEST_EXCLUDE_COMPONENTS = frozenset(
+    {
+        "Exceptions",
```
davidharrishmc, line 57: Should this add Endian or PMP?

davidharrishmc (commented):
Should this add Endian or PMP?

davidharrishmc (approved):

PR #876, opened 2026-01-27
Added Zalrsc, Zaamo coverpoints.  Adjusted parameter generator to inc…

davidharrishmc:
…lude parameter table from normative rule table to simplify main asciidoc, and to generate a None listing when there is no YAML for an extension parameters


coverpoints/norm/Zalrsc.yaml
```diff
@@ -0,0 +1,125 @@
...
+
+  # TODO: Add description for constrained_lrsc_retry_code
+  - name: constrained_lrsc_retry_code
+    coverpoint: ["Untested"]
+
+  # TODO: Add description for lrsc_eventuality_region
+  - name: lrsc_eventuality_region
+    coverpoint: ["Untested"]
+
+  # TODO: Add description for lrsc_same_address_and_size
```
jordancarlin: Are all of these TODO comments necessary? Doesn't this file just need the norm rule name and the corresponding coverpoint? What description do you want to add?

ctp/src/a.adoc
```diff
@@ -21,14 +13,10 @@ The A extension is subdivided into Zaamo and Zalrsc.
 .Zaamo Instruction Coverpoints
```
jordancarlin, line 13: If the normative rule table header is now in the generated file, the coverpoint table header should be too.

jordancarlin (changes requested):
Apparently I forgot to hit submit on my comments.

ctp/src/a.adoc
davidharrishmc: Will do this in a subsequent PR.

jordancarlin (approved):

PR #878, opened 2026-01-28
Adding rv64gcv reference config for vector

georgiatai:
Added rv64gcv config files/folder for vector testing. Mirrored rv32gcv. Confirmed vector running for both rv32gcv and rv64gcv on local machine. Coverage missing but will be addressed later.


ctp/docs-resources
jordancarlin, line 1: This is downgrading the version of the docs-resources submodule.

external/riscv-unified-db
jordancarlin, line 1: This is downgrading the version of the riscv-unified-db submodule.

jordancarlin (changes requested):
Something seems to have gone wrong with the submodules in this PR.

PR #879, opened 2026-01-28
Adding rv64gcv reference config for vector

georgiatai:
Added rv64gcv config files/folder for vector testing. Mirrored rv32gcv. Confirmed vector running for both rv32gcv and rv64gcv on local machine. Coverage missing but will be addressed later. submodules should be fixed now.


config/ref/sail-rv64gcv/sail-rv64gcv.yaml
jordancarlin, line 1: This is an outdated config file and will fail validation. Make sure to test before opening a PR.

jordancarlin (changes requested):

jordancarlin (approved):

PR #889, opened 2026-01-30
S Normative rule to coverpoint mapping

davidharrishmc:
Completed normative rule to coverpoint mapping for S test suite.

Started yaml files for many other extensions.


coverpoints/norm/S.yaml
```diff
@@ -0,0 +1,350 @@
...
+  # Similarly, for U-mode when FIOM=1, if an atomic instruction that accesses a
+  # region ordered as device I/O has its aq and/or rl bit set, then that instruction
+  # is ordered as though it accesses both device I/O and memory.
+  - name: senvcfg-fiom_op_atomic
+    coverpoint: ["OUT-OF-SCOPE"]
+
+  # If satp.MODE is read-only zero (always Bare), the implementation may make FIOM
+  # read-only zero.
+  - name: senvcfg-fiom_acc
+    coverpoint: ["OUT-OF-SCOPE"]
```
jordancarlin: Don't we test writing to FIOM? Seems like this would be in-scope.

jordancarlin (approved):

PR #891, opened 2026-01-31
[ACT4] Fix for c.nop bug in LA macro

jordancarlin:
GCC produces `c.nop` instructions in the `LA` macro when trying to align even when C is not supported. Work around this by jumping around the align directive. This still achieves the appropriate alignment, but the c.nops will not be executed.


jordancarlin:
Discussed with @allenjbaum

davidharrishmc (approved):
Nasty hack but it solves the problem.

PR #897, opened 2026-02-01
[ACT4] Initial CVE2 configuration

MikeOpenHWGroup:
Initial ACT4 configuration for the CV32E20, a specific configuration of the [CVE2](https://github.com/openhwgroup/cve2).  Based on the [cvw-rv32imc](https://github.com/riscv-non-isa/riscv-arch-test/tree/act4/config/duts/cvw/cvw-rv32imc).

The most significant change is the `model_test.h` which is the TrickBox macros.  These are specific to the "core" testbench hosted in the [cv32e20-dv](https://github.com/openhwgroup/cv32e20-dv) repo.


jordancarlin:
@MikeOpenHWGroup thanks for opening this. It looks like it’s still using an old version of the UDB config. Can you merge in the latest act4 branch and make sure that it still generates ELFs correctly when using this config?

MikeOpenHWGroup:
Update that works with the latest UDB.  One imagines this will be a moving target for a while...

config/cores/cve2/cv32e20/cv32e20.yaml
```diff
@@ -0,0 +1,108 @@
...
+  MXLEN: 32
+  #UXLEN: [32] # Why is this a 'Flow Sequence' when MXLEN is an integer?
+  PHYS_ADDR_WIDTH: 32
+
+  # MISA (!VERIFIED)
+  MUTABLE_MISA_M: false
+  MUTABLE_MISA_C: false
+  #MUTABLE_MISA_U: false
+  # UDB says:Parameter is not defined by this config: 'MUTABLE_MISA_U'.
+  # Needs: ((`Sm=1.11.0`{false} || `Sm=1.12.0`{true} || `Sm=1.13.0`{false}) && (!`Sm=1.11.0`{false} || !`Sm=1.12.0`{true}) && (!`Sm=1.11.0`{false} || !`Sm=1.13.0`{false}) && (!`Sm=1.12.0`{true} || !`Sm=1.13.0`{false}) && `U=1.0.0`{false} && `U>=0`{false} && `Sm>=0`{true})
```
jordancarlin, line 34: It's giving that error because you do not have `U` listed as one of the implemented extensions above. Should it be?

config/cores/cve2/cv32e20/README.md
```diff
@@ -0,0 +1,13 @@
...
+Copyright (c) 2026, OpenHW Foundation
+SPDX-License-Identifier: Apache-2.0 WITH SHL-2.1
+--->
+
+## DUT Configuration for the CV32E20
+To build the UDB configuration, coverage files and ELFs run the following
+command from the top of your working copy of this repo:
+```
+$ make CONFIG_FILES=config/duts/cve2/cv32e20/test_config.yaml \
+       REF_CONFIG_FILES=config/sail/sail-RVI20U32/test_config.yaml \
```
jordancarlin: The ref config files line is not needed. That is only used for collected coverage.

config/cores/cve2/cv32e20/README.md
```diff
@@ -0,0 +1,13 @@
...
+       EXTENSIONS=I,M,C
```
jordancarlin: Are there issues with leaving `EXTENSIONS` blank? Right now this is skipping the Zca, Zicsr, Zifencei, and Sm tests, all of which should be supported by CVE2. If you leave it blank it will only compile the tests relevant to your DUT based on your config.

config/duts/cve2/cv32e20/test_config.yaml
```diff
@@ -0,0 +1,8 @@
+name: cv32e20
+compiler_exe: /opt/riscv/bin/riscv64-unknown-elf-gcc # executable name on $PATH or full path to executable
+objdump_exe: /opt/riscv/bin/riscv64-unknown-elf-objdump # executable name on $PATH or full path to executable, optional
+ref_model_exe: /opt/riscv/bin/sail_riscv_sim # executable name on $PATH or full path to executable
```
jordancarlin: Can we change these to drop the `/opt/riscv/bin/` prefix? That way it will find them on the PATH if someone has the tools installed elsewhere.

jordancarlin (changes requested):
Looking very close. We moved the config files under `config/cores/...` recently instead of `config/duts/...`. Can you update this to match?

Also, looks like this is failing the pre-commit check.

config/cores/cve2/cv32e20/cv32e20.yaml
MikeOpenHWGroup: Yes, that is a bit of TODO on my part. I left the comment in to remind me to:
1. Close an internal discussion within the OpenHW cve2 project as to whether or not the CV32E20 supports U-mode or not.
2. Discuss the error message with the UDB people.  The message itself, while accurate, it difficult to read.  Also `(Sm=1.11.0{false} || Sm=1.12.0{true} || Sm=1.13.0{false})` doesn't look right:
  - Why would one add `{ name: Sm, version: "!= 1.11.0" }` to the arch config yaml?  Doesn't `Sm, version: "= 1.12.0"` imply that?  (Maybe I am not following the syntax properly.
  - What does `U=1.0.0{false} && U>=0{false}` even mean?

jordancarlin: Definitely follow up in the UDB repo regarding the error messages. I've included what I believe they mean below.

> `(Sm=1.11.0{false} || Sm=1.12.0{true} || Sm=1.13.0{false})` doesn't look right:
> 
> * Why would one add `{ name: Sm, version: "!= 1.11.0" }` to the arch config yaml?  Doesn't `Sm, version: "= 1.12.0"` imply that?  (Maybe I am not following the syntax properly.

This condition is saying that it requires Sm version 1.11.0, 1.12.0, or 1.13.0. In this case, Sm=1.11.0 is false, Sm=1.12.0 is true, and Sm=1.13.0 is false, so the condition is satisfied.

> * What does `U=1.0.0{false} && U>=0{false}` even mean?

That condition indicates that U must be version 1.0.0 and that U must be >=0 (which is how UDB indicates that the extension must be implemented without imposing version constrains). In this case, both of those conditions are false. Those seem redundant with each other, so I'm not sure why both checks are necessary.

config/cores/cve2/cv32e20/README.md
MikeOpenHWGroup: Do you mean that the ref config is only used to generate, collect and merge functional coverage?  That's useful info, thanks!

In any case, I'm reviewing the generated fcov code, so I'll continue to use it.

config/cores/cve2/cv32e20/README.md
MikeOpenHWGroup: Ahhhhhh.  I wondered about that.  It seemed odd to specify it twice.   Thanks!

config/cores/cve2/cv32e20/README.md
jordancarlin: Exactly. Makes sense to keep using it locally to review the generated coverage files, but I'd rather leave it out of the README that discusses running the CV32E20.

config/cores/cve2/cv32e20/README.md
jordancarlin: Yup. It's mostly a developer option to limit the tests that are run when developing a new extension.

config/duts/cve2/cv32e20/test_config.yaml
MikeOpenHWGroup: Done.

MikeOpenHWGroup:
> We moved the config files under config/cores/... recently instead of config/duts/.... Can you update this to match?

Done.

> Also, looks like this is failing the pre-commit check.

I don't see it...

jordancarlin:
> > Also, looks like this is failing the pre-commit check.
> 
> I don't see it...

https://github.com/riscv-non-isa/riscv-arch-test/actions/runs/21734480924/job/62697553447?pr=897

You can reproduce and automatically fix most of it by running pre-commit locally. `uv run pre-commit run -a`.

jordancarlin (approved):

PR #898, opened 2026-02-02
feat: Add consistency check for compatible settings across multiple config files

### LINKED ISSUE #893: Add consistency check for compatible settings across multiple config files
# Description

Currently, the act.py framework (located at `framework/src/act/act.py`) allows merging multiple configuration files. However, there is no validation to ensure that these configurations share compatible environment settings (e.g., target compiler, header file paths, ABI settings). Mismatched configurations can lead to silent build failures or undefined behavior during test execution.

# Solution
I propose adding a validation step in `framework/src/act/act.py `(specifically in the ACT class or the config loading logic) that:
- Iterates through all provided configuration dictionaries.
- Checks that critical keys (such as `gcc_opts`, `includes`, and `linker_script`) are consistent or compatible across all configuration sources.
- Raises a `ValidationError` (or prints a clear warning) if conflicting settings are detected.

**Additional context**
Referencing TODO in `framework/src/act/act.py`

`# TODO: Add a check that all configs use the same header files/compiler/etc. Otherwise error out or don't use common tests`

Amit-Matth:
## Description

Moved from a TODO in `framework/src/act/act.py`, this PR introduces a consistency check to ensure that when multiple configuration files are used, they share compatible environment settings.

Specifically, the `check_config_consistency` function now:
1.  **Groups Configs by XLEN and E-Extension:** Separates configurations based on both XLEN (32/64) and Base ISA (I/E) to ensure independent validation (e.g., distinguishing RV32I from RV32E).
2.  **Validates Consistency:**
    *   Checks for identical paths/settings for `compiler_exe`, `objdump_exe`, `ref_model_type`, and `ref_model_exe` within each group.
    *   Checks for identical **contents** (byte-for-byte) for `linker_script` and `{dut_include_dir}/model_test.h`, ignoring file path differences.
    *   Checks for identical **memory maps** (parsed from `{dut_include_dir}/sail.json` via pyjson5, if present) to ensure consistent memory configuration for common tests.

This prevents silent failures or undefined behavior that could arise from mixing conflicting toolchains, reference models, or memory layouts during a test run.

### Related Issues

- Fixes #893
- Addresses TODO in `framework/src/act/act.py`: "Add a check that all configs use the same header files/compiler/etc."

- [x] Ratified
- [ ] Unratified

### List Extensions

NA (Framework infrastructure change only)

### Reference Model Used

- [ ] SAIL
- [ ] Spike
- [x] Other - Verified logic with Python reproduction script using dummy configurations.

### Mandatory Checklist:

  - [ ] All tests are compliant with the test-format spec present in this repo ? (NA - Framework change)
  - [ ] Ran the new tests on RISCOF with SAIL/Spike as reference model successfully ? (NA - Framework change)
  - [ ] Ran the new tests on RISCOF in [coverage mode](https://riscof.readthedocs.io/en/stable/commands.html#coverage) (NA)
  - [ ] Link to Google-Drive folder containing the new coverage reports ([See this](https://github.com/riscv-non-isa/riscv-arch-test/blob/main/CONTRIBUTION.md#uploading-test-stats) for more info): NA

### Optional Checklist:


framework/src/act/act.py
```diff
@@ -99,6 +100,75 @@ def run_act(
     print(f"Run make -C {workdir} compile to build all tests.")
 
 
+def validate_configs(configs: list[ConfigData]) -> None:
```
jordancarlin: Move this function to `framework/src/act/config.py`.

framework/src/act/act.py
```diff
@@ -99,6 +100,75 @@ def run_act(
...
+                raise ValueError(
+                    f"Inconsistent linker_script content for XLEN {xlen} between "
+                    f"{ref_config.name} and {config.name}"
+                )
+
+            # Validate model_test.h content
+            model_header = config.dut_include_dir / "model_test.h"
+            if not ref_model_header.exists():
+                 # Force read to raise FileNotFoundError if missing, as implied by requirement
+                 ref_model_header.read_bytes()
```
jordancarlin: Why is this needed? Won't this happen anyway when the comparison is attempted below?

framework/src/act/act.py
```diff
@@ -99,6 +100,75 @@ def run_act(
...
+        if len(xlen_configs) < 2:
+            continue
+
+        ref_data = xlen_configs[0]
+        ref_config = ref_data["config"]
+        ref_model_header = ref_config.dut_include_dir / "model_test.h"
+        ref_linker_script = ref_config.linker_script
+
+        for config_data in xlen_configs[1:]:
+            config = config_data["config"]
```
jordancarlin: It's fairly inefficient to loop over all of the configs multiple times. I think this can be simplified to a single loop. The first time it sees a config of a particular xlen, it sets that as the reference version. Then for each subsequent config of the same xlen, it compares to that reference version. No need to build lists or dicts.

framework/src/act/act.py
```diff
@@ -99,6 +100,75 @@ def run_act(
     print(f"Run make -C {workdir} compile to build all tests.")
 
 
+def validate_configs(configs: list[ConfigData]) -> None:
+    """Validate that configurations are consistent."""
+    configs_by_xlen: dict[int, list[ConfigData]] = {}
+
+    # Group configs by XLEN
```
jordancarlin: I realized I didn't mention this in the issue, but in addition to separating these based on XLEN, I think they also need to be separated based on if the E extension is enabled.

jordancarlin (changes requested):
Thanks for working on this @Amit-Matth. This is a great start. I left a few comments below, but the overall structure looks good.

framework/src/act/act.py
Amit-Matth: You are absolutely right. The read_bytes() call in the comparison check immediately following this block will raise a FileNotFoundError if the file is missing, so this explicit check is redundant. I have removed it.

framework/src/act/act.py
```diff
@@ -83,7 +83,8 @@ def run_act(
             }
         )
 
-    # TODO: Add a check that all configs use the same header files/compiler/etc. Otherwise error out or don't use common tests
+    # Validate configurations
```
jordancarlin, line 86: Flush out this comment:
```suggestion
    # If there are multiple configurations, validate that they are consistent with each other so that the common tests work correctly.
```

framework/src/act/act.py
```diff
@@ -83,7 +83,8 @@ def run_act(
...
+    validate_configs(configs)
```
jordancarlin: For performance reasons, let's gate this entire check on there being multiple configs. If there is only one config, no need to make sure that it is consistent with itself.

framework/src/act/config.py
```diff
@@ -126,3 +127,63 @@ def load_config(config_file: Path) -> Config:
     config = Config.model_validate(yaml_data, context={"config_file_dir": config_file.parent})
     check_ref_model_version(config)
     return config
+
+
+def validate_configs(configs: list[Any]) -> None:
```
jordancarlin: We know that this is a list of `ConfigData`, so be specific with the types. You might need to move where `ConfigData` is defined to make the imports work.

```suggestion
def validate_configs(configs: list[ConfigData]) -> None:
```

framework/src/act/config.py
```diff
@@ -126,3 +127,63 @@ def load_config(config_file: Path) -> Config:
...
+        ref_config = ref_configs[key]
+        ref_model_header = ref_config.dut_include_dir / "model_test.h"
+        ref_linker_script = ref_config.linker_script
+
+        # Validate compiler_exe
+        if ref_config.compiler_exe != config.compiler_exe:
+            raise ValueError(
+                f"Inconsistent compiler_exe for XLEN {xlen}, E-ext {e_ext}: "
+                f"{ref_config.name} uses {ref_config.compiler_exe}, "
+                f"{config.name} uses {config.compiler_exe}"
```
jordancarlin: Let's tweak the error message to be a little more informative. Please apply similar changes to the other error messages.

```suggestion
                f"Inconsistent compiler_exe for XLEN={xlen}, E-ext={e_ext}: "
                f"{ref_config.name} uses {ref_config.compiler_exe}, "
                f"{config.name} uses {config.compiler_exe}."
                "All configs must have the same compiler_exe for common compilation. Update the configs to match or pass the configs to separate runs of the ACT framework."
```

jordancarlin (changes requested):

framework/src/act/act.py
```diff
@@ -83,7 +83,10 @@ def run_act(
             }
         )
 
-    # TODO: Add a check that all configs use the same header files/compiler/etc. Otherwise error out or don't use common tests
+    # Validate configurations
+    # If there are multiple configurations, validate that they are consistent with each other so that the common tests work correctly.
+    if len(configs) > 1:
+        validate_configs(configs)
```
jordancarlin: Let's rename this function to make it clear that it is only checking that configs are consistent with each other, it's not validating the configs themselves. Maybe something like `check_config_consistency`.

framework/src/act/config.py
```diff
@@ -126,3 +127,77 @@ def load_config(config_file: Path) -> Config:
...
     return config
+
+
+class ConfigData(TypedDict):
+    """Type definition for configuration data dictionary."""
+
+    config: Config
+    xlen: int
+    e_ext: bool
+    selected_tests: dict[str, Any]
```
jordancarlin: Why did this change from `selected_tests: dict[str, TestMetadata]` to `selected_tests: dict[str, Any]` when this class moved? Please keep the more precise types.

framework/src/act/config.py
```diff
@@ -126,3 +127,77 @@ def load_config(config_file: Path) -> Config:
...
+                f"{ref_config.name} uses {ref_config.ref_model_exe}, "
+                f"{config.name} uses {config.ref_model_exe}. "
+                "All configs must have the same ref_model_exe for common compilation. Update the configs to match or pass the configs to separate runs of the ACT framework."
+            )
+
+        # Validate linker_script content
+        if ref_linker_script.read_bytes() != config.linker_script.read_bytes():
+            raise ValueError(
+                f"Inconsistent linker_script content for XLEN={xlen}, E-ext={e_ext} between {ref_config.name} and {config.name}. "
+                "All configs must have the same linker_script content for common compilation. Update the configs to match or pass the configs to separate runs of the ACT framework."
```
jordancarlin:
```suggestion
                "All configs must have the same linker_script contents for common compilation. Update the configs to match or pass the configs to separate runs of the ACT framework."
```

framework/src/act/config.py
```diff
@@ -126,3 +127,77 @@ def load_config(config_file: Path) -> Config:
...
+            )
+
+        # Validate model_test.h content
+        model_header = config.dut_include_dir / "model_test.h"
+
+        if ref_model_header.read_bytes() != model_header.read_bytes():
+            raise ValueError(
+                f"Inconsistent model_test.h content for XLEN={xlen}, E-ext={e_ext} between {ref_config.name} and {config.name}. "
+                "All configs must have the same model_test.h content for common compilation. Update the configs to match or pass the configs to separate runs of the ACT framework."
```
jordancarlin:
```suggestion
                "All configs must have the same model_test.h contents for common compilation. Update the configs to match or pass the configs to separate runs of the ACT framework."
```

jordancarlin (changes requested):

framework/src/act/config.py
```diff
@@ -127,3 +130,148 @@ def load_config(config_file: Path) -> Config:
...
+        sail_path = config.dut_include_dir / "sail.json"
+
+        if ref_sail_path.exists() and sail_path.exists():
+            try:
+                ref_sail_content = ref_sail_path.read_bytes()
+                sail_content = sail_path.read_bytes()
+            except (FileNotFoundError, OSError) as e:
+                raise ValueError(f"Error reading sail.json: {e}") from e
+
+            if ref_sail_content != sail_content:
```
jordancarlin: This is not going to work. The sail config will be (and is supposed to be) different for different configurations. The only piece of it that should be compared (if any of it is compared at all) is the memory map.

jordancarlin (changes requested):

framework/src/act/config.py
Amit-Matth: Thanks for catching that! You're absolutely right, checking the entire sail.json file was too strict since other config parameters naturally vary.

I've updated the logic to:
- Parse sail.json using pyjson5 (consistent with how it's read elsewhere).
- Extract specifically the ["memory"]["regions"] section.
- Compare ONLY that memory map across configurations in the same group.

This correctly flags if configs have incompatible memory layouts for common tests, while allowing other settings to differ. I also added robust error handling for missing files or malformed JSON.

jordancarlin:
@Amit-Matth Thanks for all of your work on this. I ended up making changes to how common tests are generated in #914 in an attempt to solve a different problem (common tests were not rebuilt when different config files were used as input), and I think it had the side effect of resolving this issue. Please take a look.

Amit-Matth:
> @Amit-Matth Thanks for all of your work on this. I ended up making changes to how common tests are generated in #914 in an attempt to solve a different problem (common tests were not rebuilt when different config files were used as input), and I think it had the side effect of resolving this issue. Please take a look.

Checked #914 and it looks great! Using a config hash to separate build directories is a much more robust solution than enforcing strict consistency.

**One small observation:** `compute_config_hash` includes the linker script and `model_test.h`, but I didn't see the `sail.json` memory map explicitly hashed. Since the reference model relies on this map during signature generation, differing sail.json files might still matter even if the linker script is identical (though that's rare). Just something to consider adding to the hash!
Thanks!

jordancarlin:
Good observation. Feel free to add the Sail memory map if you’d like. I’m unlikely to get to that for a while otherwise.

Closing this PR since this is resolved now.

PR #900, opened 2026-02-02
[ACT4] Add support for register pairs and instructions with different types/coverpoints per XLEN

jordancarlin:
Full coverage of Zacas extension. This required support for register pairs in test generation.

The coverpoints for rv32 and rv64 are different for `amocas.d`, so `covergroupgen` and `testgen` needed to be extended to use more than just the instruction name as the key for generation.

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/cmp_regs.py
+++ b/generators/testgen/src/testgen/coverpoints/cmp_regs.py
@@ -24,7 +24,7 @@ def make_cmp_rd_rs1(instr_name: str, instr_type: str, coverpoint: str, test_data
     elif coverpoint.endswith("_nx0"):
         regs = range(1, test_data.int_regs.reg_count)  # Exclude x0
     elif coverpoint.endswith("_c"):
-        regs = range(8, 16)  # x8-x15 for compressed instructions"):
+        regs = range(8, 16)  # x8-x15 for compressed instructions
     elif coverpoint.endswith("_nx0_pair"):
         regs = range(2, test_data.int_regs.reg_count, 2)  # Even registers for pair instructions, exclude x0
         is_pair = True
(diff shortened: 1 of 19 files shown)


generators/testgen/src/testgen/coverpoints/cmp_regs.py
```diff
@@ -18,12 +18,19 @@
...
     """Generate tests where rd = rs1."""
     # Determine which rd registers to test based on coverpoint variant
+    is_pair = False
     if coverpoint == "cmp_rd_rs1":
         regs = range(test_data.int_regs.reg_count)
     elif coverpoint.endswith("_nx0"):
         regs = range(1, test_data.int_regs.reg_count)  # Exclude x0
     elif coverpoint.endswith("_c"):
-        regs = range(8, 16)  # x8-x15 for compressed instructions
+        regs = range(8, 16)  # x8-x15 for compressed instructions"):
```
davidharrishmc: Don't think you wanted the "):

generators/testgen/src/testgen/formatters/types/ap_type.py
```diff
@@ -0,0 +1,47 @@
...
+    assert params.rs2 is not None and params.rs2val is not None
+    assert params.rd is not None and params.temp_reg is not None
+
+    # Ensure rs1 is not x0 (base address)
+    if params.rs1 == 0:
+        test_data.int_regs.return_register(params.rs1)
+        params.rs1 = test_data.int_regs.get_register(exclude_regs=[0])
+
+    setup = [
+        load_int_reg("value in memory", params.temp_reg, params.rs1val, test_data),
```
davidharrishmc, line 33: This is also going to need to load the compare value into rd, or into a pair of rds.  We need the ability to write test that will match and mismatch.  I think most tests should match, so rd should have the same value as what is in memory.  But we also need to create a set of custom tests that match no bytes, match only the LSB, match the bottom halfword, match 32 bits, match a 64-bit register, match a pair of 32-bit registers, and match a pair of 64-bit registers, with each type of amocas passing and failing.

jordancarlin: This was following the model of what we were doing for the non-pair amocas instructions. Seems like the Zacas testplan needs to be enhanced to account for these additional tests.

davidharrishmc (approved):

PR #902, opened 2026-02-03
ExceptionsZc python gen 100% coverage 

juliacygong:

## Description

### Related Issues

- [ ] Ratified
- [ ] Unratified

### List Extensions

### Reference Model Used

- [ ] SAIL
- [ ] Spike

### Mandatory Checklist:

### Optional Checklist:


generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
 ##################################
-# priv/extensions/ExceptionsZc.py
+# priv/extensions/exceptionsZc.py
```
davidharrishmc: This name no longer matches the file name.  Either drop it or restore it?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
 
-def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
-    """Generate illegal compressed instructions."""
-    ######################################
-    covergroup = "ExceptionsZc_cg"
-    coverpoint = "cp_illegal_instruction"
-    ######################################
+def _generate_load_address_misaligned_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_load_address_misaligned"
+    addr_reg, base_reg, check_reg, fp_reg = 11, 12, 13, 8  # x11(a1), x12(a2), x13(a3), f8
```
davidharrishmc: Switch to get_register instead of hardwired registers to match Sm and InterruptsSm per Jordan.

tests/env/utils.h
```diff
@@ -360,3 +353,188 @@
 #define SET_MSB(_R) \
     LI(_R, 0x80000000)
 #endif
+
+// Compressed Instruction Macros
```
davidharrishmc: This is weird stuff.  Add a good comment explaining exactly why all of these statements are needed in the macros.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset} (LSBs: {offset:03b})")
+
+        def add_l_test(op: str, is_sp: bool = False, offset: int = offset) -> list[str]:
+            t_lines = []
+            suffix = "_sp" if is_sp else ""
+            t_lines.append(test_data.add_testcase(coverpoint, f"{op.lower()}{suffix}_off{offset}", covergroup))
+
+            t_lines.append(f"    LA(x{addr_reg}, scratch)")
```
davidharrishmc: Rearrange this line and the next few to load the address of scratch directly into either addr_reg or sp, rather than into addr_reg and then into sp.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+                t_lines.append(f"    mv sp, x{addr_reg}")
+
+            target = "sp" if is_sp else f"x{addr_reg}"
+            if offset > 0:
+                t_lines.append(f"    addi {target}, {target}, {offset}")
+
+            reg_prefix = "f" if "F" in op else "x"
+            reg_num = fp_reg if "F" in op else check_reg
+            if is_sp:
+                t_lines.append(f"    {op}({reg_prefix}{reg_num}, 0)")
```
davidharrishmc: Is this 0 required?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+            else:
+                t_lines.append(f"    {op}({reg_prefix}{reg_num}, 0, x{addr_reg})")
```
davidharrishmc: Isn't this usually written 0(x5) rather than 0, x5?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+
+            t_lines.append(f"    CSRR(x{check_reg}, mcause)")
```
davidharrishmc: What is this for?  If there is a trap, it will be taken and show up in the trap signature including the cause.
If there is no trap, then mcause is not updated and isn't a reliable signature anyway.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+            if is_sp:
+                t_lines.append(f"    mv sp, x{base_reg}")
+            t_lines.append(write_sigupd(check_reg, test_data))
```
davidharrishmc: How about storing the value loaded as the signature, rather than mcause?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+            reg_num = fp_reg if "F" in op else val_reg
+            if is_sp:
+                t_lines.append(f"    {op}({reg_prefix}{reg_num}, 0)")
+            else:
+                t_lines.append(f"    {op}({reg_prefix}{reg_num}, 0, x{addr_reg})")
+
+            t_lines.append(f"    CSRR(x{check_reg}, mcause)")
+            if is_sp:
+                t_lines.append(f"    mv sp, x{base_reg}")
+            t_lines.append(write_sigupd(check_reg, test_data))
```
davidharrishmc: How about reading back the 16 bytes starting at scratch and writing them to signature memory instead as the signature?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+    lines = [comment_banner(coverpoint, "Compressed Misaligned Stores")]
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset}")
+
+        def add_s_test(op: str, is_sp: bool = False, offset: int = offset) -> list[str]:
+            t_lines = []
+            suffix = "_sp" if is_sp else ""
+            t_lines.append(test_data.add_testcase(coverpoint, f"{op.lower()}{suffix}_off{offset}", covergroup))
+            t_lines.append(f"    LA(x{addr_reg}, scratch)\n    LI(x{val_reg}, 0xDEADBEEF)")
```
davidharrishmc: Format this like lines 61-68 of generators/testgen/src/testgen/priv/extensions/sm.py with a lines.extend and list of strings, rather than having many lines.append and \n within the line.  It will make the Python more readable.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_store_address_misaligned"
+    addr_reg, base_reg, val_reg, check_reg, fp_reg = 11, 12, 13, 14, 8  # a1, a2, a3, a4, f8
+    lines = [comment_banner(coverpoint, "Compressed Misaligned Stores")]
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset}")
+
+        def add_s_test(op: str, is_sp: bool = False, offset: int = offset) -> list[str]:
+            t_lines = []
+            suffix = "_sp" if is_sp else ""
```
davidharrishmc: I think you could do a is_sp = op.endswith("sp") and then not need to pass is_sp.  Then add_test could take a list of strings and you could shorten up the code that adds tests with a list rather than individually.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+        t_lines = []
+        suffix = "_sp" if is_sp else ""
+        test_label = f"{op.lower()}{suffix}_fault"
+
+        t_lines.append("    .align 3")  # 8-byte alignment
+        t_lines.append(test_data.add_testcase(coverpoint, test_label, covergroup))
+
+        reg_prefix = "f" if "F" in op else "x"
+        reg_num = fp_reg if "F" in op else check_reg
+        t_lines.append(f"    CSRR(x{temp_reg}, mtvec)")
```
davidharrishmc: Why is this needed changing the trap vectors?  Can you just access to the fault address, take the trap, and be done?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+        # For compressed instructions need to skip 2 bytes
+        # Global handler incremented in 8-bytes
+        t_lines.append(f"    la x{temp_reg}, 1f")
+        t_lines.append(f"    CSRW(mepc, x{temp_reg})")
+        t_lines.append("    mret")
+        t_lines.append("1:")
+        if is_sp:
+            t_lines.append(f"    mv sp, x{base_reg}")
+        t_lines.append(f"    CSRW(mtvec, x{temp_reg})")
+        t_lines.append("    # Expected: mcause = 5 (Load Access Fault)")
```
davidharrishmc: I don't think a signature is needed here.  We will always get the fault signature.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
 
     lines = [
-        comment_banner(
-            "cp_illegal_instruction",
-            "Illegal instructions of all 0s and all 1s",
-        ),
+        comment_banner(coverpoint, "Breakpoint (C.EBREAK)"),
+        "    .align 3",  # 8-byte alignment
+        test_data.add_testcase(coverpoint, "c_ebreak", covergroup),
+        f"    CSRR(x{temp_reg}, mtvec)",
```
davidharrishmc: No need to mess around with trap vector.  Use the regular one.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,377 @@
...
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_illegal_instruction"
+    check_reg = 13
 
+    lines = [
+        comment_banner(coverpoint, "Illegal Instruction"),
+        test_data.add_testcase(coverpoint, "illegal_0x0000", covergroup),
+        "    .half 0x0000",
+        f"    CSRR(x{check_reg}, mcause)",
+        write_sigupd(check_reg, test_data),
+        test_data.add_testcase(coverpoint, "illegal_0xFFFF", covergroup),
```
davidharrishmc: On further thought, FFFF is not a compressed instruction and is not covered by a normative rule in Zc.  FFFFFFFF is already tested in ExceptionsSm.  So let's delete FFFF from ExeptionsZc.  I've updated the spreadsheet https://docs.google.com/spreadsheets/d/1W95I4jPbuQBnXzDdIZtOG4kae8vi7djpKtJXWCXPxMU/edit?gid=538759067#gid=538759067, and you'll need to update the ExceptionsZc_coverage.svh.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+        test_data.add_testcase(coverpoint, "illegal_0x0000", covergroup),
+        "    .half 0x0000",  # use two byte for instruction alignment when trapping
+        "    .half 0x0000",
     ]
 
     return lines
 
 
-@add_priv_test_generator("ExceptionsZc")
+@add_priv_test_generator("ExceptionsZc", extensions=["I", "Zicsr", "Zca", "Zcb", "Zcd", "Zcf", "F", "D", "Sm"])
```
jordancarlin: We don’t want to require F, D, Zcb, Zcf, or Zcd  to be supported to run the ExceptionsZc test.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
 def make_exceptionszc(test_data: TestData) -> list[str]:
-    """Generate tests for ExceptionsZc machine-mode extension."""
-    lines: list[str] = []
+    """Main entry point for Zc exception test generation."""
+    lines = []
+
+    # Enable FP and initialize scratch memory
+    lines.extend(
+        [
+            "# Enable floating-point if supported",
```
jordancarlin: This is already done in the RVTEST_BEGIN macro.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+            "#ifdef RVTEST_FP",
+            "    li t0, 0x4000",
+            "    csrs mstatus, t0",
+            "#endif",
+            "",
+            "# Initialize scratch memory with test data",
+            "    LA(x10, scratch)",
```
jordancarlin: Use the register allocation functions instead of hardcoding registers.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+
+def _generate_breakpoint_tests(test_data: TestData) -> list[str]:
+    """Generate breakpoint exception test."""
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_breakpoint"
+    lines = [
+        comment_banner(coverpoint, "Breakpoint"),
+        test_data.add_testcase(coverpoint, "c_ebreak", covergroup),
+        "    c.ebreak",
+        "    nop",
+        "    c.nop",
```
jordancarlin: Add a comment explaining why the c.nop is there. Also, how do we know this failed if the test does not trap when it was supposed to?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+    ]
+
+    return lines
+
+
 def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
-    """Generate illegal compressed instructions."""
```
jordancarlin, line 17: The old comment seems more descriptive.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
-        test_data.add_testcase(coverpoint, "bin_zeros", covergroup),
-        ".instr 0x0000       # illegal instruction",
-        "nop",  # exception handler skips following instruction
-        test_data.add_testcase(coverpoint, "bin_ones", covergroup),
-        ".instr 0xFFFF       # illegal instruction",
-        ".instr 0xFFFF       # fill upper bits",
-        "nop",  # exception handler skips following instruction
+        comment_banner(coverpoint, "Illegal Instruction"),
+        "    .align 2",  # Add alignment
+        test_data.add_testcase(coverpoint, "illegal_0x0000", covergroup),
```
jordancarlin: This should match the actual bin name in the coverpoint.
```suggestion
        test_data.add_testcase(coverpoint, "illegal0", covergroup),
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+        "    .half 0x0000",  # use two byte for instruction alignment when trapping
```
jordancarlin: Why was this changed from .instr to .half?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
-from testgen.asm.helpers import comment_banner
+from testgen.asm.helpers import comment_banner, write_sigupd
 from testgen.data.state import TestData
 from testgen.priv.registry import add_priv_test_generator
 
 
+def _generate_load_address_misaligned_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_load_address_misaligned"
+    addr_reg, base_reg, check_reg, fp_reg = test_data.int_regs.get_registers(
+        4, exclude_regs=[0, 1, 2, 3, 4, 5, 6, 7, *list(range(16, 32))]
```
jordancarlin: Why are all of these registers excluded?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
 
-from testgen.asm.helpers import comment_banner
+from testgen.asm.helpers import comment_banner, write_sigupd
 from testgen.data.state import TestData
 from testgen.priv.registry import add_priv_test_generator
 
 
+def _generate_load_address_misaligned_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_load_address_misaligned"
+    addr_reg, base_reg, check_reg, fp_reg = test_data.int_regs.get_registers(
```
jordancarlin: fp_reg should come from the fp regs, not int regs.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+        4, exclude_regs=[0, 1, 2, 3, 4, 5, 6, 7, *list(range(16, 32))]
+    )
+    lines = [comment_banner(coverpoint, "Compressed Misaligned Loads")]
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset} (LSBs: {offset:03b})")
+
+        def add_l_test(ops: list[str], offset: int = offset) -> list[str]:
```
jordancarlin: Is it really necessary to define a function inside of a function? This seems overly complex.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+                t_lines = []
+                suffix = "_sp" if is_sp else ""
+                t_lines.append(test_data.add_testcase(coverpoint, f"{op.lower()}{suffix}_off{offset}", covergroup))
+
+                # Load address and apply offset
+                if is_sp:
+                    t_lines.append(f"    mv x{base_reg}, sp")  # Save sp
+                    t_lines.append("    LA(sp, scratch)")
+                    if offset > 0:
+                        t_lines.append(f"    addi sp, sp, {offset}")
```
jordancarlin: No harm in doing an addi with 0 when the offset is 0 and it makes the test more consistent.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+            base_ops.extend(["c.ld", "c.ldsp"])
+        lines.extend(add_l_test(base_ops))
+
+        # Zcb
+        lines.append("#ifdef ZCB_SUPPORTED")
+        lines.extend(add_l_test(["c.lh", "c.lhu", "c.lbu"]))
+        lines.append("#endif")
+
+        # Zcf
+        lines.append("#if defined(ZCF_SUPPORTED) && XLEN == 32")
```
jordancarlin: Zcf will only ever be supported when xlen == 32. No need to check here.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+                reg_str = f"f{fp_reg}" if "f" in op.lower() else f"x{check_reg}"
+                sig_reg = fp_reg if "f" in op.lower() else check_reg
+
+                if is_sp:
+                    t_lines.append(f"    {op} {reg_str}, 0(sp)")
+                    t_lines.append(f"    mv sp, x{base_reg}")  # Restore sp immediately
+                else:
+                    t_lines.append(f"    {op} {reg_str}, 0(x{addr_reg})")
+
+                t_lines.append(write_sigupd(sig_reg, test_data))
```
jordancarlin: This needs to set `sig_type=“float”` when Checking an fp register.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,349 @@
...
+        # Zcd
+        lines.append("#ifdef ZCD_SUPPORTED")
+        lines.extend(add_l_test(["c.fld", "c.fldsp"]))
+        lines.append("#endif")
+
+    test_data.int_regs.return_registers([addr_reg, base_reg, check_reg, fp_reg])
+    return lines
+
+
+def _generate_store_address_misaligned_tests(test_data: TestData) -> list[str]:
```
jordancarlin, line 260: There is a lot of duplicated logic in the load/store test functions. Can any of that be moved into a common function that both functions call?

jordancarlin (changes requested):

tests/env/arch_test.h
```diff
@@ -1229,6 +1229,14 @@ chk_\__MODE__\()tval:
         slli    T3, T3, XLEN-1          // put mcause bit# into MSB
         bge     T3, x0, sv_\__MODE__\()tval     // if MSB=0, no adj, sv to ensure tval was cleared
 
+fault_addr_adj_\__MODE__\()tval:
+        LI(     T3, RVMODEL_ACCESS_FAULT_ADDRESS)         // Load fault address
+        andi    T6, T2, ~0x7F                     // Mask bottom 7 bits of mtval
+        andi    T3, T3, ~0x7F                     // Mask of 7 bits of fault address
+        bne     T6, T3, vmem_adj_\__MODE__\()tval // Continue with normal checks if not in fault range
+        li      T3, 0                             // Set adjustment to 0
```
davidharrishmc: Confirm what this adjustment is for and whether this is the right behavior.

davidharrishmc (commented):
Very nice!

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
juliacygong: wanted to change for clarity

jordancarlin: The linker can do strange things sometimes when using .word or .half for instructions. .instr is supposed to be used for instructions and .word/.half/.double is for data.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
juliacygong: I thought to keep it to x8-x15 since compressed instructions typically use these registers

jordancarlin: In that case, the `reg_range` argument is a better fit. Also probably worth including a comment explaining why you are limiting the register range.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
juliacygong: I switched it back to .instr, but I'm now getting an unknown pseudo-op error

davidharrishmc:
Click Resolve Conversation on each suggested change as you address it, and send a message when you are ready for the next round of review.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+    addr_reg: int,
+    base_reg: int,
+    check_reg: int,
+    fp_reg: int,
+) -> list[str]:
+    all_lines = []
+    for op in ops:
+        is_sp = op.endswith("sp")
+        t_lines = []
+        suffix = "_sp" if is_sp else ""
```
davidharrishmc: Why does this suffix need to be added to the bin name?  It's already part of the op.  Global change.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+        if is_sp:
+            t_lines.append(f"    mv x{base_reg}, sp")  # Save sp
+            t_lines.append("    LA(sp, scratch)")
+            t_lines.append(f"    addi sp, sp, {offset}")
+        else:
+            t_lines.append(f"    LA(x{addr_reg}, scratch)")
+            t_lines.append(f"    addi x{addr_reg}, x{addr_reg}, {offset}")
+
+        # Perform load
+        reg_str = f"f{fp_reg}" if "f" in op.lower() else f"x{check_reg}"
```
davidharrishmc: Would this add an f to any instruction with an f in the name?  Do you want to check if op starts with f?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+        sig_reg = fp_reg if "f" in op.lower() else check_reg
+
+        if is_sp:
+            t_lines.append(f"    {op} {reg_str}, 0(sp)")
+            t_lines.append(f"    mv sp, x{base_reg}")  # Restore sp immediately
+        else:
+            t_lines.append(f"    {op} {reg_str}, 0(x{addr_reg})")
+
+        t_lines.append(write_sigupd(sig_reg, test_data, sig_type="float"))
```
davidharrishmc: This is always doing RVTEST_SIGUPD_F even for integer tests.
Be sure to carefully proofread your .S file produced by this generator.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+    all_lines = []
+    for op in ops:
+        is_sp = op.endswith("sp")
+        t_lines = []
+        suffix = "_sp" if is_sp else ""
+        t_lines.append(test_data.add_testcase(coverpoint, f"{op.lower()}{suffix}_off{offset}", covergroup))
+
+        # Load address and apply offset
+        if is_sp:
+            t_lines.append(f"    mv x{base_reg}, sp")  # Save sp
```
davidharrishmc: Query in slack about Jordan's preference on spaces vs. tabs.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset} (LSBs: {offset:03b})")
+
+        # Zca
+        base_ops = ["c.lw", "c.lwsp"]
+        lines.extend(
+            _add_load_test(base_ops, offset, test_data, coverpoint, covergroup, addr_reg, base_reg, check_reg, fp_reg)
+        )
+        lines.append("#if __riscv_xlen == 64")
```
davidharrishmc, line 234: Put \n before all these ifdefs to make them stand out from previous test case.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+                addr_reg,
+                base_reg,
+                check_reg,
+                fp_reg,
+            )
+        )
+        lines.append("#endif")
+
+        # Zcf
+        lines.append("#if defined(ZCF_SUPPORTED)")
```
davidharrishmc: You use ifdef elsewhere.  Concerned that this code gets skipped and you had no way of knowing.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+
+        if is_sp:
+            t_lines.append(f"    mv x{base_reg}, sp")
+            t_lines.append(f"    mv sp, x{addr_reg}")
+            t_lines.append(f"    {op} {reg_str}, 0(sp)")
+            t_lines.append(f"    mv sp, x{base_reg}")
+        else:
+            t_lines.append(f"    {op} {reg_str}, 0(x{addr_reg})")
+
+        t_lines.append("    nop")
```
davidharrishmc: Put a comment in the generated code about why these two nops are necessary.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+            t_lines.append(f"    addi sp, sp, {offset}")
+        else:
+            t_lines.append(f"    LA(x{addr_reg}, scratch)")
+            t_lines.append(f"    addi x{addr_reg}, x{addr_reg}, {offset}")
+
+        # Perform store
+        reg_str = f"f{fp_reg}" if "f" in op.lower() else f"x{check_reg}"
+
+        if is_sp:
+            t_lines.append(f"    {op} {reg_str}, 0(sp)")
```
davidharrishmc: The store does not affect the signature.  If this is intentional, say so in the test and say you are just checking whether a trap takes place.  If you want the test to affect the signature, see  MisalignZca.c-sw.  I would feel better making a signature, because MisalignZca only tests a subset of compressed loads/stores, while this suite tests them all.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+            t_lines.append(f"    mv sp, x{base_reg}")  # Restore sp immediately
+        else:
+            t_lines.append(f"    {op} {reg_str}, 0(x{addr_reg})")
+
+        t_lines.append(f"    li x{check_reg}, 0x{offset:08x}")  # Use offset for signature
```
davidharrishmc: This doesn't write anything related to the store.

davidharrishmc: No point doing a sigupd unless the test affects the result.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
davidharrishmc: Affects multiple functions.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+
+        if is_sp:
+            t_lines.append(f"    mv x{base_reg}, sp")
+            t_lines.append(f"    mv sp, x{addr_reg}")
+            t_lines.append(f"    {op} {reg_str}, 0(sp)")
+            t_lines.append(f"    mv sp, x{base_reg}")
+        else:
+            t_lines.append(f"    {op} {reg_str}, 0(x{addr_reg})")
+
+        t_lines.append("    nop")
```
davidharrishmc: Put comments in the generated code explaining why the two tops are needed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+) -> list[str]:
+    all_lines = []
+    for op in ops:
+        is_sp = op.endswith("sp")
+        t_lines = []
+        suffix = "_sp" if is_sp else ""
+        test_label = f"{op.lower()}{suffix}_fault"
+
+        # 8-byte alignment
+        t_lines.append("    .align 3")
```
davidharrishmc: Why is 8-byte alignment needed?  Are you talking about alignment of code or data?  Not sure why either is needed.  If needed, put a clear comment in the generated code about why this is necessary.

davidharrishmc: Comments here apply to stores as well.

davidharrishmc: Put a space before the alignment.  Right now it seems to be associated with the previous test because there is no space before but a space after.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_load_access_fault"
+    addr_reg, base_reg, check_reg = test_data.int_regs.get_registers(
+        3, exclude_regs=[*range(0, 8), *range(16, 32)]
+    )  # registers x8-x15 used for compressed instructions
+
+    fp_reg = test_data.float_regs.get_register(
+        exclude_regs=[*range(0, 8), *range(16, 32)]
+    )  # exclude registers outside of f8-f15
+
+    lines = [comment_banner(coverpoint, "Load Access Fault")]
```
davidharrishmc: "Load Access Fault" doesn't add very much information to "cp_load_access_fault".  
A better comment would be "Test every type of compressed load to a faulting
address and check that each one throws a load access fault."  Global comment.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+        is_sp = op.endswith("sp")
+        t_lines = []
+        suffix = "_sp" if is_sp else ""
+        test_label = f"{op.lower()}{suffix}_fault"
+
+        # 8-byte alignment
+        t_lines.append("    .align 3")
+        t_lines.append(test_data.add_testcase(coverpoint, test_label, covergroup))
+
+        t_lines.append(f"    li x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS")
```
davidharrishmc: You could get rid of the li here and put it at line 171.  In line 167, li straight into sp rather than into addr_reg.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
-        "",
-        test_data.add_testcase(coverpoint, "bin_zeros", covergroup),
-        ".instr 0x0000       # illegal instruction",
-        "nop",  # exception handler skips following instruction
-        test_data.add_testcase(coverpoint, "bin_ones", covergroup),
-        ".instr 0xFFFF       # illegal instruction",
-        ".instr 0xFFFF       # fill upper bits",
-        "nop",  # exception handler skips following instruction
+        comment_banner(coverpoint, "Illegal Instruction"),
+        "    .align 2",  # Add alignment
```
davidharrishmc: Why is this needed?  If necessary, explain in generated code.  If not, remove.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+        test_data.add_testcase(coverpoint, "illegal0", covergroup),
+        "    .insn 0x0000",  # use two byte for instruction alignment when trapping
```
davidharrishmc: Some of your tests are followed by two nops, some by one, and this by one.  Find what you need and be consistent.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,476 @@
...
+
+def _generate_breakpoint_tests(test_data: TestData) -> list[str]:
+    """Generate breakpoint exception test."""
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_breakpoint"
+
+    lines = [
+        comment_banner(coverpoint, "Breakpoint"),
+        test_data.add_testcase(coverpoint, "c_ebreak", covergroup),
+        "    c.ebreak",
+        "    nop",
```
davidharrishmc: One nop here.  Two after some of the other tests.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,495 @@
...
+        # Two nops ensure trap handler has space to skip to next test
+        " nop",
+        " nop",
     ]
 
     return lines
 
 
-@add_priv_test_generator("ExceptionsZc")
+@add_priv_test_generator("ExceptionsZc", extensions=["I", "Zicsr", "Zca", "Zcd", "Zcb", "Sm"])
```
jordancarlin: This limits the test to only running if Zca, Zcb, and Zcd are all present on the DUT, but most of the test cases apply if any compressed instructions are supported, so this needs to be run if Zca is present even without the other two extensions.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,495 @@
...
+        reg_str = f"f{fp_reg}" if is_float else f"x{check_reg}"
+        sig_reg = fp_reg if is_float else check_reg
+
+        if is_sp:
+            t_lines.append(f" {op} {reg_str}, 0(sp)")
+            t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
+        else:
+            t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
+
+        t_lines.append(write_sigupd(sig_reg, test_data, sig_type="float" if is_float else "int"))
```
davidharrishmc: What is being written as the signature and why?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,495 @@
...
```
davidharrishmc: I though you were going to read back the 16 bytes and store them as the signature to show the misaligned stores worked?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,495 @@
...
+            t_lines.append(f" mv x{base_reg}, sp")
+            t_lines.append(" li sp, RVMODEL_ACCESS_FAULT_ADDRESS")
+            t_lines.append(f" {op} {reg_str}, 0(sp)")
+            t_lines.append(f" mv sp, x{base_reg}")
+        else:
+            t_lines.append(f" li x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS")
+            t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
+
+        # Two nops ensure trap handler has space to skip to next test
+        t_lines.append(" nop")
```
davidharrishmc: Why are two needed instead of one?  If they really are needed, explain.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
davidharrishmc: Zcd and Zcb shouldn't be required because the test will run without them.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,495 @@
...
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset} (LSBs: {offset:03b})")
+
+        # Zca
+        base_ops = ["c.lw", "c.lwsp"]
+        lines.extend(
+            _add_load_test(base_ops, offset, test_data, coverpoint, covergroup, addr_reg, base_reg, check_reg, fp_reg)
+        )
+        lines.append("#if __riscv_xlen == 64")
```
davidharrishmc, line 234: blank line before

davidharrishmc: (many places)

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
davidharrishmc: I thought you were going to store the 16 bytes of scratch as the signature.  See Misaligned.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,495 @@
...
+    all_lines = []
+
+    for op in ops:
+        is_sp = op.endswith("sp")
+        is_float = op.startswith("c.f")
+        t_lines = []
+        test_label = f"{op.lower()}_fault"
+
+        # Ensure proper alignment for exception handling with 8 byte alignment
+        t_lines.append(" .align 3")
```
davidharrishmc: Why is this needed?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,48 +1,495 @@
...
+        if is_sp:
+            t_lines.append(f" mv x{base_reg}, sp")
+            t_lines.append(" li sp, RVMODEL_ACCESS_FAULT_ADDRESS")
+            t_lines.append(f" {op} {reg_str}, 0(sp)")
+            t_lines.append(f" mv sp, x{base_reg}")
+        else:
+            t_lines.append(f" li x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS")
+            t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
+
+        # Two nops ensure trap handler has space to skip to next test
```
davidharrishmc: "The access might throw a trap and the trap handler skips over the next four bytes.  Thus, insert two nops, which each assemble into c.nop, to get 4 bytes of instructions to skip.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
davidharrishmc: Seems like you would need two nops here too because the access might fault.  Same for load.  See if anything else needs it.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -43,15 +43,32 @@ def _add_load_test(
...
-        sig_reg = fp_reg if is_float else check_reg
 
         if is_sp:
             t_lines.append(f" {op} {reg_str}, 0(sp)")
             t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
         else:
             t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
 
-        t_lines.append(write_sigupd(sig_reg, test_data, sig_type="float" if is_float else "int"))
+        # Write 16 bytes from scratch as signature
```
davidharrishmc: Writing 16 bytes to the signature should just be for misaligned stores that affect part of the scratch region.  Shouldn't be done for loads.  Also, the nops should be right after the misaligned load that might fault.  Otherwise something else might get skipped on return from the trap handler.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -43,15 +43,32 @@ def _add_load_test(
...
         reg_str = f"f{fp_reg}" if is_float else f"x{check_reg}"
-        sig_reg = fp_reg if is_float else check_reg
 
         if is_sp:
             t_lines.append(f" {op} {reg_str}, 0(sp)")
             t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
         else:
             t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
 
-        t_lines.append(write_sigupd(sig_reg, test_data, sig_type="float" if is_float else "int"))
```
davidharrishmc: keep this as your signature for loads.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -440,15 +477,20 @@ def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
...
     ]
 
     return lines
 
 
-@add_priv_test_generator("ExceptionsZc", extensions=["I", "Zicsr", "Zca", "Zcd", "Zcb", "Sm"])
+@add_priv_test_generator(
+    "ExceptionsZc",
+    required_extensions=["Zicsr", "Sm", "Zca"],
+    march_extensions=["Zicsr", "Zca", "Zcb", "Zcd", "F", "D"],
```
davidharrishmc: Should this also list Zcf?

davidharrishmc (approved):
Very close to ready.

tests/env/arch_test.h
```diff
@@ -1604,8 +1604,15 @@ chk_\__MODE__\()tval:
...
-        LREG    T2, vmem_bgn_off(T4)              // fetch sig_begin addr
+fault_addr_adj_\__MODE__\()tval:
+        LI(     T3, RVMODEL_ACCESS_FAULT_ADDRESS)         // Load fault address
+        andi    T6, T2, ~0x7F                     // Mask bottom 7 bits of mtval
+        andi    T3, T3, ~0x7F                     // Mask of 7 bits of fault address
+        bne     T6, T3, vmem_adj_\__MODE__\()tval // Continue with normal checks if not in fault range                          // Set adjustment to 0
+        j       adj_\__MODE__\()tval              // jump to perform mtval
+
+vmem_adj_\__MODE__\()tval:         /* T4 still points to sv area of trapping mode */
+        LREG    T3, vmem_bgn_off(T4)            // fetch sig_begin addr
```
jordancarlin: @Zain2050 can you review these trap handler changes to make sure they don't break anything else that you have been doing to the trap handler?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
 #
-# ExceptionsZc privileged extension test generator.
-# jcarlin@hmc.edu Jan 2026
+# ExceptionsZc extension exception test generator.
+# jgong@hmc.edu Jan 2026
 # SPDX-License-Identifier: Apache-2.0
 ##################################
 
-"""ExceptionsZc privileged extension test generator."""
+"""Zc extension exception test generator."""
```
jordancarlin: There is no "Zc extension". Maybe something like this instead:

```suggestion
"""Compressed exceptions test generator."""
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
+            t_lines.append(f" {op} {reg_str}, 0(sp)")
+            t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
+        else:
+            t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
+
+        t_lines.append(write_sigupd(sig_reg, test_data, sig_type="float" if is_float else "int"))
+
+        # Load access may throw a trap and the trap handler skips over the next 4 bytes
+        # Two nops are used to get 4 bytes of instructions to get since each assembles into a c.nop
+        t_lines.append(" nop")
```
davidharrishmc: Needs to go before sigupd

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -44,9 +481,46 @@ def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
     march_extensions=["Zicsr", "Zca", "Zcb", "Zcd", "F", "D"],
 )
 def make_exceptionszc(test_data: TestData) -> list[str]:
-    """Generate tests for ExceptionsZc machine-mode extension."""
-    lines: list[str] = []
+    """Main entry point for Zc exception test generation."""
+    lines = []
+
+    sig_reg, data_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
```
jordancarlin: I'd rather use a different name than `sig_reg` and `data_reg`. These could easily be interpreted as the sig pointer/data pointer, which they are not.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -44,9 +481,46 @@ def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
...
+
+    lines.extend(
+        [
+            "\n#ifdef RVTEST_FP",
+            " li t0, 0x4000",
+            " csrs mstatus, t0",
+            "#endif",
+            "",
```
jordancarlin: Isn't this handled by the default startup code?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
-        "nop",  # exception handler skips following instruction
-        test_data.add_testcase(coverpoint, "bin_ones", covergroup),
-        ".insn 0xFFFF       # illegal instruction",
-        ".insn 0xFFFF       # fill upper bits",
-        "nop",  # exception handler skips following instruction
+        comment_banner(coverpoint, "Illegal Instruction"),
+        # Align to ensure proper instruction fetch and trap handling"
+        " .align 2",  # Add alignment
+        test_data.add_testcase(coverpoint, "illegal0", covergroup),
+        " .insn 0x0000",  # use two byte for instruction alignment when trapping
```
jordancarlin: This might be easier to follow as 
```
.insn 0x00
c.nop
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
+        # Illegal instruction may throw a trap and the trap handler skips over the next 4 bytes
+        # Two nops are used to get 4 bytes of instructions to get since each assembles into a c.nop
+        " nop",
+        " nop",
```
jordancarlin: If these are turning into `c.nop`, let's just use `c.nop` so that it easier to follow. Global change.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
-        "",
-        test_data.add_testcase(coverpoint, "bin_zeros", covergroup),
-        ".insn 0x0000       # illegal instruction",
-        "nop",  # exception handler skips following instruction
-        test_data.add_testcase(coverpoint, "bin_ones", covergroup),
-        ".insn 0xFFFF       # illegal instruction",
-        ".insn 0xFFFF       # fill upper bits",
-        "nop",  # exception handler skips following instruction
+        comment_banner(coverpoint, "Illegal Instruction"),
+        # Align to ensure proper instruction fetch and trap handling"
```
jordancarlin, line 401: Seems like we'd want to include this comment in the generated test, so it needs to be a string. Applies to lots of other comments too. If a comment is only useful when generating the test, make it a Python comment. If a comment is useful to understanding the test itself, then make it a comment in the generated assembly.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
+
+        all_lines.extend(t_lines)
+    return all_lines
+
+
+def _generate_load_address_misaligned_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsZc_cg", "cp_load_address_misaligned"
+    addr_reg, base_reg, check_reg = test_data.int_regs.get_registers(
+        3, exclude_regs=[*range(0, 8), *range(16, 32)]
+    )  # registers x8-x15 used for compressed instructions
```
jordancarlin: Use `test_data.int_regs.get_registers(reg_range=range(8,16)` instead. Global change.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
+
+    lines = [comment_banner(coverpoint, "Compressed Misaligned Loads")]
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset} (LSBs: {offset:03b})")
+
+        # Zca
+        base_ops = ["c.lw", "c.lwsp"]
+        lines.extend(
+            _add_load_test(base_ops, offset, test_data, coverpoint, covergroup, addr_reg, base_reg, check_reg, fp_reg)
```
jordancarlin: Since you're not preloading the registers with anything, it seems like it would simplify the code to call the `get_register` function inside of `_add_load_test` so that you don't need to pass all of the registers around to all of the functions.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
-"""ExceptionsZc privileged extension test generator."""
+"""Zc extension exception test generator."""
 
-from testgen.asm.helpers import comment_banner
+from testgen.asm.helpers import comment_banner, write_sigupd
 from testgen.data.state import TestData
 from testgen.priv.registry import add_priv_test_generator
 
 
+def _add_load_test(
```
jordancarlin, line 16: I think the interface to this would be cleaner if it took a single instruction and generated a single test and the loop was pulled up a level into the coverpoint functions. So something like this:

```py
for instr in ["c.lw", "c.lwsp"]:
    lines.extend(_add_load_test(instr, offset, test_data, coverpoint, covergroup, addr_reg, base_reg, check_reg, fp_reg))
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
+        t_lines.append(write_sigupd(check_reg, test_data))
+        t_lines.append(f" lw x{check_reg}, 8(x{addr_reg})")
+        t_lines.append(write_sigupd(check_reg, test_data))
+        t_lines.append(f" lw x{check_reg}, 12(x{addr_reg})")
+        t_lines.append(write_sigupd(check_reg, test_data))
+
+        # Store access may throw a trap and the trap handler skips over the next 4 bytes
+        # Two nops are used to get 4 bytes of instructions to get since each assembles into a c.nop
+        t_lines.append(" nop")
+        t_lines.append(" nop")
```
jordancarlin: What is the point of these `nops`? There won't be a trap during the sigupd which is what comes right before. Shouldn't these go before the sigupd right after the actual instruction?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
+        # Perform store
+        reg_str = f"f{fp_reg}" if is_float else f"x{check_reg}"
+
+        if is_sp:
+            t_lines.append(f" {op} {reg_str}, 0(sp)")
+            t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
+        else:
+            t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
+
+        # Write 16 bytes from scratch as signature to verify the misaligned store worked
```
jordancarlin:
```suggestion
        # Read 16 bytes from scratch as signature to verify the misaligned store worked
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,475 @@
...
+
+        t_lines.append(test_data.add_testcase(coverpoint, test_label, covergroup))
+
+        reg_str = f"f{fp_reg}" if is_float else f"x{check_reg}"
+
+        if is_sp:
+            t_lines.append(f" mv x{base_reg}, sp")
+            t_lines.append(" li sp, RVMODEL_ACCESS_FAULT_ADDRESS")
+            t_lines.append(f" {op} {reg_str}, 0(sp)")
+            t_lines.append(f" mv sp, x{base_reg}")
```
jordancarlin: The `mv sp, ...` will get skipped over when there is a trap, so `sp` will not get restored. Need `nops` in between. This issue applies to all 4 load/store functions.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -6,340 +6,285 @@
...
+        t_lines.append("# Trap handler skips the next 4 bytes; two c.nops provide 4 bytes")
+        t_lines.append("c.nop")
+        t_lines.append("c.nop")
+        t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
+    else:
+        t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
+        t_lines.append(
+            "# Load access may throw a trap and the trap handler skips over the next 4 bytes. Two c.nops are used to get 4 bytes of instructions"
+        )
+    t_lines.append("c.nop")
```
davidharrishmc: These seem unnecessary for sp.  Should they move into the else?

davidharrishmc: Affects many functions

davidharrishmc (changes requested):

davidharrishmc (approved):

tests/env/arch_test.h
Zain2050: This is no longer required because we have removed the entire xtval check along with relocation in PR #971.

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,399 @@
...
+    else:
+        t_lines.append(f" LA(x{addr_reg}, scratch)")
+        t_lines.append(f" addi x{addr_reg}, x{addr_reg}, {offset}")
+
+    # Perform store
+    reg_str = f"f{fp_reg}" if is_float else f"x{check_reg}"
+
+    if is_sp:
+        t_lines.append(f" {op} {reg_str}, 0(sp)")
+        t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
```
jordancarlin, line 112: The nops between the instruction that might trap and the mv seem to be gone again.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,402 @@
...
+    if is_sp:
+        t_lines.append(f" {op} {reg_str}, 0(sp)")
+        t_lines.append("# Trap handler skips the next 4 bytes; two c.nops provide 4 bytes")
+        t_lines.append("c.nop")
+        t_lines.append("c.nop")
+        t_lines.append(f" mv sp, x{base_reg}")  # Restore sp immediately
+    else:
+        t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
+
+        t_lines.append("c.nop")
```
jordancarlin, line 116: Remove blank line.
```suggestion
        t_lines.append(f" {op} {reg_str}, 0(x{addr_reg})")
        t_lines.append("c.nop")
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -1,38 +1,402 @@
...
+        lines.append("#endif")
+
+        # Zcb
+        lines.append("\n#ifdef ZCB_SUPPORTED")
+        for instr in ["c.lh", "c.lhu", "c.lbu"]:
+            lines.extend(_add_load_test(instr, offset, test_data, coverpoint, covergroup))
+        lines.append("#endif")
+
+        # Zcf
+        lines.append("\n#ifdef ZCF_SUPPORTED")
```
jordancarlin, line 246: I discovered that all of the rv32 configs are missing `ZCF_SUPPORTED` in `rvtest_config.h` and `rvtest_config.svh`, so this code is never being run. Please add it to all of the rv32 configs and make sure everything still passes (and that these instructions show up in the coverage report).

jordancarlin (approved):

PR #908, opened 2026-02-04
(ACT4) Converted Virtual Memory tests to ACT4 framework

Zain2050:
Better to merge it in after PR #843 gets merged.


UmerShahidengr (approved):
LGTM. Great work @Zain2050

tests/priv/Svade/sv32_Svade_Umode.S
```diff
@@ -0,0 +1,281 @@
...
+// 7. PTE.D set, PTE.A unset at level 0, RWX permissions (read, write, execute page):
+//        Then access the page in U-Mode. Expected: Store, Load & Fetch page fault
+// 8. PTE.D unset, PTE.A unset at level 0, RWX permissions (read, write, execute page):
+//        Then access the page in U-Mode. Expected: Store, Load & Fetch page fault
+//
+// Total Expected Faults :: 14
+// --------------------------------------------------------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Sv32]
```
jordancarlin, line 40: @Zain2050 @UmerShahidengr shouldn't all of these Svade tests require Svade?

Zain2050: Yes, these require Svade. But when I was writing these tests, Svade & Svadu were commented out from the config files due to some UDB bug, therefore I removed it from the list of required extension. The bug has been resolved now but I forgot to add Svade back.
I'll be soon pushing tests for Sv variants. I'll make sure to include this change.

jordancarlin: If it's easy to do so, it might be worth pushing as a separate PR so we can merge it quickly so it isn't incorrectly selected for any DUTs. Otherwise, it sounds good to include in your next PR.

PR #914, opened 2026-02-08
[ACT4] Refactor common test selection and simplify config file selection

jordancarlin:
- Only common tests that are actually needed by one of the input configs are compiled.
- Multiple configs with different compilers/linker scripts/etc. are now supported. Common tests are grouped based on these files so that they do not conflict. This also means `make clean` is no longer needed when switching configs.
- `work-ref` has been removed; all tests are now compiled in `work` for both regular runs and coverage runs.
- `REF_CONFIG_FILES` has been renamed to `COVERAGE_CONFIG_FILES` for clarity.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 
 # Directories and files
 # CONFIG_FILES is used as the input configs when just running `make` and will produce elfs in the `work/<config-name>/elfs` directory.
-# COVERAGE_CONFIG_FILES is used as the input configs when just running `make coverage` and will coverage reports in addition to the elfs.
+# COVERAGE_CONFIG_FILES is used as the input configs when just running `make coverage` and will generate coverage reports in addition to the elfs.
 CONFIG_FILES ?= config/spike/spike-rv32-max/test_config.yaml config/spike/spike-rv64-max/test_config.yaml
 COVERAGE_CONFIG_FILES ?= config/sail/sail-rv64-max/test_config.yaml config/sail/sail-rv32-max/test_config.yaml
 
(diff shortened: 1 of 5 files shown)


framework/src/act/makefile_gen.py
```diff
@@ -26,6 +28,34 @@
...
+    # Hash model_test.h contents
+    model_test_h = config.dut_include_dir / "model_test.h"
+    hasher.update(model_test_h.read_bytes())
+
+    # Hash executable paths (resolved paths to detect different binaries)
+    hasher.update(str(config.compiler_exe.resolve()).encode())
+    hasher.update(str(config.ref_model_exe.resolve()).encode())
+    if config.objdump_exe is not None:
+        hasher.update(str(config.objdump_exe.resolve()).encode())
+
```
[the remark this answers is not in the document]
jordancarlin: `rvtest_config.h` is only used in config-dependent tests, not common tests. Ignoring.

Makefile
```diff
@@ -3,17 +3,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # Directories and files
-# CONFIG_FILES is used as the input configs when just running `make` and will produce els in the `work` directory
-# REF_CONFIG_FILES is used as the input configs when running `make coverage` and will produce elfs and coverage reports in the `work-ref` directory
+# CONFIG_FILES is used as the input configs when just running `make` and will produce elfs in the `work/<config-name>/elfs` directory.
+# COVERAGE_CONFIG_FILES is used as the input configs when just running `make coverage` and will coverage reports in addition to the elfs.
```
davidharrishmc: missing word after will
not sure what "just" is doing here

jordancarlin: Fixed

davidharrishmc (changes requested):
comment clarifications

davidharrishmc (approved):

PR #915, opened 2026-02-09
Added CI support for RVI20 profile on Spike

UmerShahidengr:
Profile RVI20 profile tests are running on Spike in CI now


.github/workflows/regress.yml
```diff
@@ -89,7 +98,7 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: ${{ github.workspace }}/spike
-          key: spike-${{ env.SPIKE_HASH }}-RV${{ matrix.xlen }}
+          key: spike-${{ env.SPIKE_HASH }}-RV${{ matrix.spike.xlen }}
```
jordancarlin: Is there any reason to cache spike differently for different XLEN? The Spike binary is the same.

.github/workflows/regress.yml
```diff
@@ -57,27 +57,36 @@ jobs:
...
   spike:
-    name: Spike Regression (${{ matrix.target }})
+    name: Spike Regression (${{ matrix.spike.target }})
     runs-on: ubuntu-latest
+
     strategy:
       matrix:
-        target: [spike-rv64, spike-rv32]
+        spike:
+          - { xlen: 32, profile: max, target: spike-rv32-max }
```
jordancarlin: The profile key isn't used anywhere and the xlen key can be dropped after the Spike caching change. Can we simplify this to just the list of targets?

config/spike/spike-RVI20U32/spike-RVI20U32.yaml
jordancarlin, line 1: Can we name this `spike-RVI20U32` to match the actual name of the profile and how the Sail configs are named? Same for the rv64 version.

jordancarlin (changes requested):

.github/workflows/regress.yml
UmerShahidengr: Eliminated the XLEN key to make the CI simpler

.github/workflows/regress.yml
UmerShahidengr: Done. Thanks for pointing that out

config/spike/spike-RVI20U32/spike-RVI20U32.yaml
UmerShahidengr: Done.

.github/workflows/regress.yml
jordancarlin: Not introduced in this PR, but can you remove `RV${{ matrix.xlen }}` since it isn't used?

jordancarlin: What about something like the following so that we don't need to add every variant to the Makefile:
```suggestion
        run: |
            DOCKER=1 EXTENSIONS=I,M,F,D,Zca,Zcf,Zcd,Zaamo,Zalrsc,Zifencei CONFIG_FILES=${{ matrix.config_file }} make --jobs $(nproc)
            ./run_tests.py "${{ matrix.run_cmd }}" $(WORKDIR)/${{ config_name }}/elfs
            
```

That would also make it easier to add QEMU or RTL cores to CI.

UmerShahidengr: Sure, good idea.

UmerShahidengr: Should we remove the Spike section in MakeFile completely after committing this change?

UmerShahidengr: Sure, will do it

jordancarlin: Let's keep the existing spike max config in the Makefile so that it is easy to run a Spike config locally but not add any additional configs to avoid complicating the Makefile.

UmerShahidengr: Done

UmerShahidengr: Done

jordancarlin:
@UmerShahidengr looks like a few of the copilot comments need to be addressed and then this looks good to merge.

UmerShahidengr:
@jordancarlin done

config/spike/spike-RVI20U64/rvtest_config.h
```diff
@@ -0,0 +1,4 @@
+#define RVMODEL_ACCESS_FAULT_ADDRESS 0x00000000
+#define ACCESS_FAULT_ADDRESS RVMODEL_ACCESS_FAULT_ADDRESS
+#define RVMODEL_PMP_GRAIN 4
```
jordancarlin, line 2:
```suggestion
#define RVMODEL_ACCESS_FAULT_ADDRESS 0x00000000
#define RVMODEL_PMP_GRAIN 4
```

jordancarlin (approved):

PR #916, opened 2026-02-09
Add Spike support to ACT4 to enable H testgen for CVW Hypervisor Testing

TnnsBeast:
## Description

Adds Spike support to ACT4 framework. Introduces 2 new cvw cores cvw-rv32gch and cvw-rv64gch with hypervisor enabled to allow for testing of CVW hypervisor development with Spike as reference model. PR will be updated with corresponding CVW hypervisor PR.

### Related Issues

NA

- [x] Ratified

### List Extensions

Hypervisor

### Reference Model Used

- [ ] SAIL
- [x] Spike

### Mandatory Checklist:

  - [] Ran the new tests on RISCOF with SAIL/Spike as reference model successfully ?

### Optional Checklist:

  - [x] Have you run these on any hard DUT model ? Please specify name and provide link if possible in the description

Ran generated H tests on cvw-rv32gch and cvw-rv64gch, both with self checking against corresponding Spike config.


tests/env/encoding.h
jordancarlin, line 1: Looks like this was hand-modified. It should be regenerated using `riscv-opcodes` and then any additional definitions that are needed should go in one of the other headers (likely `test_macros.h`).

generators/testgen/src/testgen/io/templates.py
```diff
@@ -149,11 +149,11 @@ def generate_defines_from_extensions(ext_components: list[str]) -> list[str]:
     # TODO: Enable vector extension if needed when vector testgen is integrated
 
     # Enable trap handlers if needed
-    if "H" in ext_components:
-        extra_defines.append("#define rvtest_vtrap_routine")
+    # Note: vtrap is intentionally not enabled by default to avoid toolchain
+    # incompatibilities when H is present but V-mode traps are not exercised.
     if any(ext in ext_components for ext in ["H", "S"]):
-        extra_defines.append("#define rvtest_strap_routine")
+        extra_defines.append("#define rvtest_strap_routine 1")
```
jordancarlin: Why is this change needed? The #define should work without a value.

generators/testgen/src/testgen/io/templates.py
```diff
@@ -149,11 +149,11 @@ def generate_defines_from_extensions(ext_components: list[str]) -> list[str]:
     # TODO: Enable vector extension if needed when vector testgen is integrated
 
     # Enable trap handlers if needed
-    if "H" in ext_components:
-        extra_defines.append("#define rvtest_vtrap_routine")
+    # Note: vtrap is intentionally not enabled by default to avoid toolchain
+    # incompatibilities when H is present but V-mode traps are not exercised.
```
jordancarlin: What do you mean toolchain incompatibilities? Why would we not want to enable the hypervisor trap handler for hypervisor tests?

framework/src/act/makefile_gen.py
```diff
@@ -90,7 +90,7 @@ def gen_compile_targets(
     base_dir: Path,
     xlen: int,
     config: Config,
-    sail_config_path: Path,
+    ref_model_config_path: Path | None,
```
jordancarlin: This should never be None.

framework/src/act/makefile_gen.py
```diff
@@ -326,7 +337,9 @@ def generate_config_makefile(
         final_elf = config_elf_dir / elf_name
         trace_name = test_name.with_suffix(".rvvi")
         trace_path = config_coverage_dir / trace_name
-        sail_config_path = config.dut_include_dir / "sail.json"
+        ref_model_config_path = (
```
jordancarlin: This is only used for Sail, so leave the name as `sail_config_path`.

framework/src/act/makefile_gen.py
```diff
@@ -119,6 +119,14 @@ def gen_compile_targets(
     flen = test_metadata.flen
     test_path = test_metadata.test_path
     ref_model_sig_flags = config.ref_model_type.signature_flags.format(sig_file=sig_file, granularity=int(xlen / 8))
+    ref_model_args_line = f"\t\t{config.ref_model_args} \\\n" if config.ref_model_args else ""
+    trace_all_line = "--trace-all" if (debug and config.ref_model_type == RefModelType.SAIL) else ""
+    trace_output_line = (
+        f"\t\t--trace-output {sig_trace_file} \\\n" if (debug and config.ref_model_type == RefModelType.SAIL) else ""
```
jordancarlin: Let’s combine trace_all_line and trace_output_line into debug_flags and move it to config.py the same way signature_flags is handled. Also add flags for the Spike version to produce a log.

jordancarlin (changes requested):
I haven’t reviewed the hypervisor specific parts yet, but here’s an initial round of comments on the spike integration.

config/cores/cvw/cvw-rv32gch/model_test.h
```diff
@@ -0,0 +1 @@
+../cvw-rv64gc/model_test.h
```
davidharrishmc: model_test.h needs to be updated to rvmodel_macros.h to match latest ACT

tests/env/test_macros.h
```diff
@@ -1,3 +1,197 @@
+// Extra CSR encodings not yet present in generated encoding.h. for H tests
```
jordancarlin: If you regenerate `encoding.h` with the latest version of riscv-opcodes, I think most of these should be present. Please do that and only keep the definitions that are still missing here. All of the CSR addresses at least should be present in encoding.h.

generators/testgen/src/testgen/io/templates.py
```diff
@@ -150,7 +150,7 @@ def generate_defines_from_extensions(ext_components: list[str]) -> list[str]:
 
     # Enable trap handlers if needed
     if "H" in ext_components:
-        extra_defines.append("#define rvtest_vtrap_routine")
+        extra_defines.append("#define rvtest_vtrap_routine 1")
```
TnnsBeast: H builds were failing because rvtest_vtrap_routine was emitted as an empty macro (#define rvtest_vtrap_routine), but arch_test.h uses it in numeric preprocessor expressions (#if (rvtest_vtrap_routine) / #if rvtest_vtrap_routine). With an empty define, those expand to invalid #if forms, causing “missing expression between '(' and ')'” and “#if with no expression.”

Setting it to #define rvtest_vtrap_routine 1 fixes this by making those #if checks valid.

rvtest_strap_routine and rvtest_mtrap_routine do not need 1 because they are checked via #ifdef/defined(...) where empty defines are valid.

tests/env/test_macros.h
```diff
@@ -1,3 +1,197 @@
...
+#endif
+#ifndef CSR_HGEIP
+#define CSR_HGEIP 0xe12
+#endif
+
+#ifndef vsstatus
+#define vsstatus CSR_VSSTATUS
+#endif
+#ifndef vsie
+#define vsie CSR_VSIE
```
jordancarlin: Why are these necessary?

tests/env/test_macros.h
```diff
@@ -1,3 +1,197 @@
...
+#define hgeip CSR_HGEIP
+#endif
+
+// VRET encoding (HRET slot) for assemblers that lack a vret mnemonic.
+#ifndef RVTEST_VRET_DEFINED
+#define RVTEST_VRET_DEFINED
+.macro VRET
+  .insn i 0x73, 0, x0, x0, 0x202
+.endm
+#endif
```
jordancarlin: Looks like gcc calls this `hret`, not `vret`. Does that fix the issue and let you call the instruction directly?

config/cores/cvw/cvw-rv32gch/rvmodel_macros.h
jordancarlin, line 1: Use a symlink to the other cvw `rvmodel_macros.h` files since it is identical so that these new ones don't fall out of date.

tests/env/test_macros.h
jordancarlin: Maybe I'm missing something, but I also don't see `hret` or `vret` in the ISA manual.

config/cores/cvw/cvw-rv64gch/test_config.yaml
```diff
@@ -0,0 +1,9 @@
+name: cvw-rv64gch
+compiler_exe: riscv64-unknown-elf-gcc # executable name on $PATH or full path to executable
+objdump_exe: riscv64-unknown-elf-objdump # executable name on $PATH or full path to executable, optional
+ref_model_exe: spike # executable name on $PATH or full path to executable
+ref_model_type: spike
+ref_model_args: "--isa=rv64gch_zicntr_zihpm_zicbom_zicboz_sstc_svpbmt_svadu"
```
jordancarlin, line 6: Could we generate this from the UDB extensions list instead of needing to specify the list of extensions in yet another place? The framework already produces a file called `extensions.txt` in each config's work directory with the list of extensions. Maybe just combine all of the items in that list and stick XLEN on the beginning of it?

framework/src/act/makefile_gen.py
```diff
@@ -127,6 +127,14 @@ def gen_compile_targets(
     flen = test_metadata.flen
     test_path = test_metadata.test_path
     ref_model_sig_flags = config.ref_model_type.signature_flags.format(sig_file=sig_file, granularity=int(xlen / 8))
+    ref_model_args_line = f"\t\t{config.ref_model_args} \\\n" if config.ref_model_args else ""
+    ref_model_debug_line = ""
+    if debug:
+        ref_model_debug_flags = config.ref_model_type.debug_flags.format(sig_trace_file=sig_trace_file)
+        ref_model_debug_line = f"\t\t{ref_model_debug_flags} \\\n" if ref_model_debug_flags else ""
```
jordancarlin: Can this be simplified to the following?
```suggestion
        ref_model_debug_line = f"\t\t{config.ref_model_type.debug_flags.format(sig_trace_file=sig_trace_file)} \\\n"
```

framework/src/act/makefile_gen.py
```diff
@@ -252,7 +261,9 @@ def generate_common_makefile(
     common_build_dir = common_wkdir / "build"
 
     # Generate maximal Sail config with all extensions enabled and memory map from user's config
-    common_sail_config = generate_sail_config(xlen, e_ext, config.dut_include_dir / "sail.json", common_wkdir)
+    common_sail_config = config.dut_include_dir / "sail.json"
+    if config.ref_model_type == RefModelType.SAIL:
+        common_sail_config = generate_sail_config(xlen, e_ext, common_sail_config, common_wkdir)
```
jordancarlin: Instead of this kind of hacky workaround, let's move `common_sail_config` fully into `gen_compile_targets` and only populate it if the reference model is Sail.

generators/testgen/src/testgen/priv/extensions/h.py
```diff
@@ -0,0 +1,332 @@
...
+def _generate_shvstvecd_tests(test_data: TestData) -> list[str]:
+    """Generate Shvstvecd testplan case: vstvec direct mode base alignment."""
+    covergroup = "Shvstvecd_cg"
+    coverpoint = "cp_shvstvecd"
+
+    save_reg, write_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+    lines = [
+        comment_banner(
+            "cp_shvstvecd",
+            "Testplan add-on: write aligned bases to vstvec with direct mode (MODE=0)",
```
jordancarlin: What do you mean "testplan add-on"?

generators/testgen/src/testgen/priv/extensions/h.py
```diff
@@ -0,0 +1,332 @@
...
+        )
+    ]
+
+    lines.extend(
+        [
+            f"\tCSRR(x{save_reg}, vstvec)      # save vstvec",
+            f"\tla x{write_reg}, shvstvecd_base",
```
jordancarlin: Make sure to use the `LA` and `LI` macros instead of `la` and `li`.

jordancarlin (changes requested):
The Spike side of things is getting close. Here is round 2 of feedback.

I still haven't reviewed the actual hypervisor test generator in detail, but I left a few high level comments on it.

tests/env/test_macros.h
davidharrishmc: I'm uncomfortable that this looks like AI-generated slop, like the non-existent CSRs that AI generated earlier.  I don't want to pollute our repo with garbage.  Writing this stuff by hand isn't too hard, and makes us think about what we are really trying to do.  I'd suggest not using AI for hypervisor code generation after these issues.  It's fine to use it for automation outside of our production code.

TnnsBeast: In templates.py, since rvtest_vtrap_routine is added in extra_defines when H is enabled, the generated H-00.S includes #define rvtest_vtrap_routine, which causes arch_test.h to instantiate V-mode trap handlers (INSTANTIATE_MODE_MACRO around lines 320-321). Then, in the V handler, \__MODE__\()RET (arch_test.h line 1301) expands to VRET which isn't recognized by the assembler. 

Error seen without the macro:
cvw/addins/riscv-arch-test-cvw/tests/env/arch_test.h: Assembler messages:
cvw/addins/riscv-arch-test-cvw/tests/env/arch_test.h:1301: Error: unrecognized opcode `vret'

Since right now all of the written tests are done in M mode, the AI generated macro for vret isn't used at runtime (but it assembles fine), which is why the 'patch' was working. 

I think that this likely requires some fixing to be done in arch_test.h so vret isn't produced at all.

jordancarlin: @Zain2050 are you aware of this issue and is it already addressed in #843?

Zain2050: Yes, we have specifically fixed bugs related to hypervisor. I ran tests with htrap and vtrap_routine defined and it was working fine.

jordancarlin: Great. Then let's drop this change and wait for #843 to be merged.

PR #918, opened 2026-02-10
InterruptsSm tests python generator

SadhviNarayanan:
Added the python generator script for InterruptsSm, along with updating sail_test.h with the macros required for triggering interrupts, and arch_test.h for trap handler configurations. Currently at 77.35% coverage which is what the original looped script was at as well. We don't have a way to trigger external interrupts on Sail, so we are not collecting coverage on this.


tests/env/sail_test.h
```diff
@@ -6,6 +6,22 @@
 #ifndef _COMPLIANCE_MODEL_H
 #define _COMPLIANCE_MODEL_H
 
+#define CLINT_BASE_ADDR 0x02000000
+#define PLIC_BASE_ADDR 0x0C000000
+#define GPIO_BASE_ADDR 0x10060000
```
jordancarlin, line 11: Sail doesn’t have a PLIC or GPIO. Why are these here?

tests/env/sail_test.h
```diff
@@ -85,15 +101,65 @@
 
 #define RVMODEL_CLR_MEXT_INT
 
-#define RVMODEL_SET_MTIMER_INT
-
-#define RVMODEL_CLR_MTIMER_INT
+#define RVMODEL_SET_MTIMER_INT \
```
jordancarlin, line 104: This could generate spurious interrupts on rv32. See https://docs.riscv.org/reference/isa/priv/machine.html#3-2-1-machine-timer-mtime-and-mtimecmp-registers for the recommended code sequence to set a timer interrupt on rv32.

tests/env/sail_test.h
```diff
@@ -85,15 +101,65 @@
...
+#define RVMODEL_CLR_MTIMER_INT \
+  li t0, -1;                    \
+  la t2, MTIMECMP;              \
+  SREG t0, 0(t2);               \
+#ifdef __riscv_xlen \
+  #if __riscv_xlen == 32 \
+      sw t0, 4(t2);             \
+  #endif \
+#else \
+  ERROR: __riscv_xlen not defined; \
```
jordancarlin, line 129: Let’s drop the checks for if __riscv_xlen is not defined everywhere and just have a single global check in `check_defines.h`.

tests/env/sail_test.h
```diff
@@ -85,15 +101,65 @@
...
+      ld t0, 0(t0); \
+      addi t0, t0, 0x100; \
+      sd t0, 0(t4); \
+  #elif __riscv_xlen == 32 \
+      lw t1, 0(t0); \
+      lw t2, 4(t0); \
+      addi t3, t1, 0x100; \
+      bgtu t1, t3, 1f; \
+      j 2f; \
+  1: addi t2, t2, 1; \
```
jordancarlin, line 146: Indentation issue.

tests/env/arch_test.h
```diff
@@ -680,23 +680,23 @@
 #endif
 #ifndef RVMODEL_MCLR_SSW_INT // M-mode interrupt handler for S-mode SW Ints
         //.warning "RVMODEL_CLR_SSW_INT    not defined. Executing this will end test. Define an empty macro to suppress this warning"
-        #define  RVMODEL_MCLR_SSW_INT     RVTEST_DFLT_INT_HNDLR
+        #define  RVMODEL_MCLR_SSW_INT     RVMODEL_CLR_SSW_INT
```
jordancarlin, line 683: Discussed with @allenjbaum already. Let’s just drop `RVMODEL_MCLR_SSW_INT` (and the other mode specific variants of these macros) and use `RVMODEL_CLR_SSW_INT` directly. Same applies to the other interrupt types as well.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+    r1, r2 = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(
+            "cp_trigger_mti",
+            "With mstatus.MIE = {0/1}, and mie = all 1s, use MTIMECMP to cause mip.MTIP",
+        ),
+        "",
+        f"LI x{r1}, -1               # all 1s",
+        f"CSRRW x{r2}, mie, x{r1}    # enable all interrupts",
```
jordancarlin, line 32:
```suggestion
        f"CSRRW(x{r2}, mie, x{r1})    # enable all interrupts",
```

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+
+    r1, r2 = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(
+            "cp_trigger_mti",
+            "With mstatus.MIE = {0/1}, and mie = all 1s, use MTIMECMP to cause mip.MTIP",
+        ),
+        "",
+        f"LI x{r1}, -1               # all 1s",
```
jordancarlin, line 31: Global comment:
```suggestion
        f"LI(x{r1}, -1)               # all 1s",
```

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+
+    # Test 1: mstatus.MIE = 0 should NOT take interrupt
+    lines.extend(
+        [
+            test_data.add_testcase(coverpoint, "mie_0", covergroup),
+            f"    LI x{r1}, 1                # success code",
+            f"    CSRRCI x{r2}, mstatus, 8   # mstatus.MIE = 0",
+            "    RVMODEL_SET_MTIMER_INT     # trigger timer interrupt",
+            "    RVMODEL_CLR_MTIMER_INT     # reset mtimecmp (no interrupt fired)",
+            f"    # x{r1} should still be 1 since no interrupt",
```
jordancarlin, line 44: How do you know that x{r1} won’t still be 1 after taking an interrupt? Doesn’t the trap handler restore most (or all) of the register values?

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+        "",
+    ]
+
+    # Test 1: mstatus.MIE = 0 should NOT take interrupt
+    lines.extend(
+        [
+            test_data.add_testcase(coverpoint, "mie_0", covergroup),
+            f"    LI x{r1}, 1                # success code",
+            f"    CSRRCI x{r2}, mstatus, 8   # mstatus.MIE = 0",
+            "    RVMODEL_SET_MTIMER_INT     # trigger timer interrupt",
```
jordancarlin, line 42: We can’t use the `RVMODEL` macros inline because they will be different lengths. They need to be instantiated at the end of the test. Look at how this is done for `RVMODEL_BOOT` in `test_setup.h`.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+
+    # Test 2: mstatus.MIE = 1 should take interrupt
+    lines.extend(
+        [
+            "",
+            test_data.add_testcase(coverpoint, "mie_1", covergroup),
+            f"    LI x{r1}, 1                # success code",
+            f"    CSRRSI x{r2}, mstatus, 8   # mstatus.MIE = 1",
+            "    RVMODEL_SET_MTIMER_INT     # interrupt fires, handler runs",
+            f"    LI x{r1}, -1               # trap handler skips this",
```
jordancarlin, line 57: The `LI` macro is not clearly one instruction long. How about using an `addi` here instead?

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+            write_sigupd(r1, test_data),
+        ]
+    )
+
+    test_data.int_regs.return_registers([r1, r2])
+    return lines
+
+
+def _generate_trigger_sei_plic_tests(test_data: TestData) -> list[str]:
+    """Generate supervisor external interrupt trigger tests via PLIC."""
```
jordancarlin, line 266: Can’t assume all DUTs have a PLIC.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+            write_sigupd(r1, test_data),
+        ]
+    )
+
+    test_data.int_regs.return_registers([r1, r2])
+    return lines
+
+
+def _generate_trigger_sei_sie_tests(test_data: TestData) -> list[str]:
+    """Generate supervisor external interrupt trigger tests via SIE."""
```
jordancarlin, line 315: What do you mean via SIE? Isn’t that the enable bit?

jordancarlin (changes requested):
Haven’t finished going through the generator yet, but here’s a first round of feedback.

tests/env/sail_test.h
```diff
@@ -85,15 +101,65 @@
...
-
-#define RVMODEL_SET_MSW_INT
-
-#define RVMODEL_CLR_MSW_INT
+la t0, MTIME; \
+la t4, MTIMECMP; \
+#ifdef __riscv_xlen \
+  #if __riscv_xlen == 64 \
+      ld t0, 0(t0); \
+      addi t0, t0, 0x100; \
```
davidharrishmc, line 138: Let's make this 100 a parameter in rvmodel_test.h.  Call it RVMODEL_TIMER_INT_SOON_DELAY

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+    ]
+
+    # Test 1: mstatus.MIE = 0 should NOT take interrupt
+    lines.extend(
+        [
+            test_data.add_testcase(coverpoint, "mie_0", covergroup),
+            f"    LI x{r1}, 1                # success code",
+            f"    CSRRCI x{r2}, mstatus, 8   # mstatus.MIE = 0",
+            "    RVMODEL_SET_MTIMER_INT     # trigger timer interrupt",
+            "    RVMODEL_CLR_MTIMER_INT     # reset mtimecmp (no interrupt fired)",
```
davidharrishmc, line 43: After every SET_*_INT call, we need to give it time to fire, in case the delay is nonzero.  Let's create a macro called RVTEST_IDLE_FOR_INTERRUPT and put it after every call to RVMODEL_SET_*_INT (but not after _SOON).  The macro can contain .rept RVMODEL_INTERRUPT_LATENCY nop.  RVMODEL_INTERRUPT_LATENCY should be defined in rvmodel_macros.h and let's give it a value of 10.

davidharrishmc: A problem with this is that now we can't be guaranteed to run/skip the next instruction after the SET_*_INT depending on whether the interrupt fires, because we don't know the interrupt latency.  I think we should drop codes like line 57 because they won't be reliable.  How about instead of having a success code of r1=1, we have an incrementing success code (maybe a variable sigval that is incremented after each write_sigupd)?

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+        f"CSRRW x{r2}, mie, x{r1}    # enable all interrupts",
+        "",
+    ]
+
+    # Test 1: mstatus.MIE = 0 should NOT take interrupt
+    lines.extend(
+        [
+            test_data.add_testcase(coverpoint, "mie_0", covergroup),
+            f"    LI x{r1}, 1                # success code",
+            f"    CSRRCI x{r2}, mstatus, 8   # mstatus.MIE = 0",
```
davidharrishmc, line 41: Looks like we are throwing away r2.  Let's not even bother reading it if we aren't using it.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
davidharrishmc: Did we want to create RVTEST_SET_MTIME_INT that does a jam to the RVMODEL_SET_MTIME_INT?

jordancarlin: Yes. @SadhviNarayanan were you planning to implement that?

SadhviNarayanan: Yes! I will work on implementing this

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
SadhviNarayanan: How do we want to go about determining whether a DUT has a PLIC? Are there some parameters that need to be implemented?

davidharrishmc: We really are trying to say there are two ways to try generating a supervisor external interrupt: via a SEI pin from a peripheral, or by writing mip.seip.  I think we need to clarify that the RVMODEL_SET_SEXT_INTERRUPT performs the function with a peripheral, and then separately write mip.seip in the other test.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+            "    RVMODEL_SET_MEXT_INT",
```
davidharrishmc, line 142: Needs to implement this RVMODEL_SET_MEXT_INT.  See examples/asm/mei for an implementation for Spike and Wally.  Doesn't work on Sail yet.
Similar for SEXT.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+            f"    LI x{r1}, -1               # trap handler skips this",
+            write_sigupd(r1, test_data),
+        ]
+    )
+
+    test_data.int_regs.return_registers([r1, r2])
+    return lines
+
+
+def _generate_trigger_sti_tests(test_data: TestData) -> list[str]:
```
davidharrishmc, line 165: See the test plan that supervisor interrupts are not part of InterruptsSm anymore.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
davidharrishmc: Global change.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+    r1 = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(
+            "cp_interrupts",
+            "Cross of mstatus.MIE = {0/1}, mtvec.MODE = 00, 3 walking 1s in mip.MTIP/MSIP/MEIP,\n"
+            "3 walking 1s in mie.MTIE/MSIE/MEIE (2 x 3 x 3 bins)",
+        ),
+        "",
+        "CSRRCI t6, mtvec, 3     # mtvec.MODE = 00",
```
davidharrishmc, line 381: Here you are using t6 rather than a get_register.  First, no need to read a signature unless you are saving it.  Second, no guarantees t6 is safe to read.  Third, limit the registers to x0-x15 so tests run on E.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
davidharrishmc, line 32: The CSRW macro is a good one to use for these purposes.  See sm.py for an example.  It writes the register, then reads it back and saves to the signature, and also has a nop in case of a trap.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+            write_sigupd(r1, test_data),
+        ]
+    )
+
+    test_data.int_regs.return_registers([r1, r2])
+    return lines
+
+
+def _generate_interrupt_cross_tests(test_data: TestData) -> list[str]:
+    """Generate interrupt cross-product tests (mstatus.MIE x mtvec.MODE x mip x mie)."""
```
davidharrishmc, line 366: Vectored shouldn't be part of the cross anymore.  Check against the testplan and coverpoints, which have been simplified from last year.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+        "CSRRCI t6, mtvec, 3",
+        "CSRRSI t6, mtvec, 1     # mtvec.MODE = 01",
+        "CSRRSI t6, mstatus, 8   # mstatus.MIE = 1",
+        "CSRRW t6, mie, zero     # clear all enables",
+        "LI s1, 0x888            # MEIE/MTIE/MSIE",
+        "CSRRS t6, mie, s1       # enable all three",
+    ]
+
+    # Raise each interrupt type: s2 = 2 (MEIP), 1 (MTIP), 0 (MSIP)
+    for s2 in [2, 1, 0]:
```
davidharrishmc, line 464: Could this just be for s2 in ["msip", "mtip", "meip"]

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+        "CSRRSI t6, mstatus, 8   # mstatus.MIE = 1",
+    ]
+
+    # Unroll s1 loop: 7 down to 0 (mie values)
+    for s1 in range(7, -1, -1):
+        # Build mie value from s1
+        lines.extend(
+            [
+                "CSRRW t6, mie, zero     # clear all enables",
+                f"LI t0, {s1 & 4}         # check bit 2",
```
davidharrishmc, line 519: Let's be consistent about either always using temp registers or always using a requested register.

generators/testgen/src/testgen/priv/extensions/interruptsSm.py
```diff
@@ -0,0 +1,660 @@
...
+                "slli t0, t0, 9          # mie.MEIE position",
```
davidharrishmc, line 520: There is no need to build mei from s1 in assembly.  Your python knows what she should be, so you could just compute it in python and do a single li.

davidharrishmc (changes requested):

PR #923, opened 2026-02-11
QEMU riscv32 and ricsv64 for ACT4

jgonzalezsalgado:
Configurations for qemu-rv32-max and qemu-rv64-max in progress.


UmerShahidengr:
Will you be sharing the scripts to run the tests on QEMU model?
This PR has the config folder which I think is complete, and following the same template as that of config/sail/* or config/spike/* directory, but it is missing the script to run tests on QEMU

davidharrishmc:
Still in development.

PR #924, opened 2026-02-12
Vector coverage update and Vf running

georgiatai:
Bringing back most of Vx coverage (left with 2 instructions that are not 100%) and Vf running. 

The stuff in RISCV_coverage_common.svh are kept as comment instead of delete because these reference might come in handy when generating config files from UDB.


davidharrishmc (commented):
One minor Copilot suggestion and then looks good.

davidharrishmc (approved):

PR #928, opened 2026-02-12
Improved documentation, especially RVTEST Macros.  Atomic and Counter…

davidharrishmc:
… coverpoint improvements per Allen's review


ctp/src/a.adoc
```diff
@@ -40,7 +40,7 @@ include::{testplansdir}/Zalrsc.adoc[]
 |cp_custom_sc_after_store|4|store followed by sc
 |cp_custom_sc_after_load|6|load followed by sc
 |cp_custom_sc_lrsc|2|lr followed by sc that fails and passes
-|cp_custom_sc_address|16*2|lr followed by sc at address of 0, 8, ..., 120 offset from lr address.  For RV64, exercise both lr.d and lr.w
+|cp_custom_sc_addresses|32*2|lr followed by sc at address of 0, 8, ..., 248 offset from lr address.  For RV64, exercise both lr.d and lr.w
```
jordancarlin: 0 is not included in the coverpoint or test.

```suggestion
|cp_custom_sc_addresses|32*2|lr followed by sc at address of 8, ..., 248 offset from lr address.  For RV64, exercise both lr.d and lr.w
```

generators/testgen/src/testgen/coverpoints/special/cp_cntr.py
```diff
@@ -17,64 +17,86 @@ def make_cntr(instr_name: str, instr_type: str, coverpoint: str, test_data: Test
     """Generate tests for counter coverpoints."""
     test_lines: list[str] = []
 
-    # Allocate some registers for testing.  Restrict them to [8,15] in case the registers are used for compressed instructions
-    r1, r2 = test_data.int_regs.get_registers(2, exclude_regs=[0])
+    # Allocate some registers for testing.
+    r1, r2, r3 = test_data.int_regs.get_registers(3, exclude_regs=[0])
```
jordancarlin, line 21: Might make more sense to move the register allocation call into `gen_cntr_test` so that you don’t need to pass all of the registers as inputs. It also means there will be a wider range of registers tested.

generators/testgen/src/testgen/coverpoints/special/cp_cntr.py
```diff
@@ -17,64 +17,86 @@ def make_cntr(instr_name: str, instr_type: str, coverpoint: str, test_data: Test
...
+    test_data.int_regs.return_registers([r1, r2, r3])
 
     return test_lines
 
 
-def gen_cntr_test(instr_name: str, cntr: str, r1: int, r2: int, test_data: TestData) -> str:
+def gen_cntr_test(instr_name: str, cntr: str, r1: int, r2: int, r3: int, test_data: TestData) -> str:
     """Generate counter test snippet."""
-    mindiff = 1 if cntr in ["instret", "cycle"] else 0  # instret and cycle increment quickly; others may not
+    mindiff = 1 if cntr in ["instret", "cycle", "time"] else 0  # h registers unlikely to increment
```
jordancarlin, line 57: This is misleading. `instret` and the h counters don’t use mindiff, they use an exact check.

generators/testgen/src/testgen/coverpoints/special/cp_cntr.py
```diff
@@ -17,64 +17,86 @@ def make_cntr(instr_name: str, instr_type: str, coverpoint: str, test_data: Test
...
     if cntr != "instret" and not cntr.endswith("h"):
         slt = f"slti x{r1}, x{r1}, {mindiff} # set fail code if difference < {mindiff}"
     else:
         slt = ""  # for minstret, the difference should be exact.  High counters should be exactly zero.
```
jordancarlin, line 61:
```suggestion
        slt = ""  # for instret, the difference should be exact.  High counters should be exactly zero.
```

ctp/src/rvmodel.adoc
```diff
@@ -31,34 +31,25 @@ The RVMODEL DUT-Specific Macros are given in <<t-rvmodel>>. The Mode column indi
...
-|RVMODEL_SET_SSW_INT|Sets the Supervisor Software Interrupt (`mip.SSIP` = 1, if supported) | Machine or Supervisor
-|RVMODEL_CLR_SSW_INT|Clears the Supervisor Software Interrupt (`mip.SSIP` = 0) | Machine or Supervisor
-|RVMODEL_SET_VEXT_INT|Set the Virtual Supervisor External Interrupt | Machine or HS
-|RVMODEL_CLR_VEXT_INT|Clear the Virtual Supervisor External Interrupt | Machine or HS
-|RVMODEL_SET_VTIMER_INT|Set the Virtual Supervisor Timer Interrupt | Machine or HS
-|RVMODEL_CLR_VTIMER_INT|Clear the Virtual Supervisor Timer Interrupt | Machine or HS
-|RVMODEL_SET_VTIMER_INT_SOON|Cause the Virtual Supervisor Timer Interrupt to rise soon, but not immediately. | Machine or HS
-|RVMODEL_SET_VSW_INT|Set the Virtual Supervisor Software Interrupt | Machine or HS
-|RVMODEL_CLR_VSW_INT|Clear the Virtual Supervisor Software Interrupt | Machine or HS
+|RVMODEL_SET_MEXT_INT|Sets the Machine External Interrupt `mip.MEIP` using platform-specific interrupt controller, if supported| Machine or Supervisor
```
jordancarlin, line 34: Triggering a machine interrupt from S mode seems problematic. It is almost certain that S mode won’t have write access to the relevant interrupt controller bits.

ctp/src/rvmodel.adoc
jordancarlin, line 1: The README and check_defines.h need to be updated accordingly.

jordancarlin (changes requested):

ctp/src/rvmodel.adoc
allenjbaum: I would expect that either Mmode would handle this, or it would use Mmode only resources and trap instead of actually setting the interrupt (so, it's either a bug or emulated by Mmode which would both set and service the interrupt?)

ctp/src/rvmodel.adoc
```diff
@@ -31,34 +31,25 @@ The RVMODEL DUT-Specific Macros are given in <<t-rvmodel>>. The Mode column indi
...
+|RVMODEL_MTIME_ADDRESS|The address of the memory-mapped `mtime` register, used for timer accesses and interrupts. | n/a
+|RVMODEL_MTIMECMP_ADDRESS|The address of the memory-mapped `mtimecmp` register, used for timer interrupts. | n/a
 |RVMODEL_DATA_SECTION|Model-specific data that will be placed in a .data section. Used for memory-mapped regions like `tohost`/`fromhost`.  Macro can be empty if not needed. | n/a
 |===
 
-Note that if a macro needs machine mode privileges to perform its function but might be called from any mode, it must change privilege level.  For example, if RVMODEL_IO_WRITE_STR needs machine mode to access a UART, it may call RVTEST_GOTO_M_MODE.  Printing (and halting) is done at the end of a test, so there is no need to return to the previous privilege mode.
+[NOTE]
+RISC-V has standardized generating timer interrupts using memory-mapped MTIME and MTIMECMP, so no RVMODEL macros are needed for DUT-specific timer interrupt generation.  Similarly, no RVMODEL macros are needed to set or read MTIME, or set or clear interrupts using standard CSR accesses.
```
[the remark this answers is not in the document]
allenjbaum: MMode doesn't strictly need a macro for timer interrupts, but it does need the mtime and mtimecompare addresses (which we've been calling RVMODEL macros, even though they merely #define an address (as defined above).  Ditto  for SWI
I would argue that enabling/disabling external interrupts should be using xIE CSRs, but tests may want to use virtual devices that would exercise non-ISA components (e.g. PLIC, APLIC, CLIC) to actually cause interrupts and for that we need non-ISA macros (RVMODEL...)

ctp/src/rvmodel.adoc
davidharrishmc: Unless PMP blocks access to peripherals from supervisor mode, the supervisor will be able to make this request of the interrupt controller.

jordancarlin: Could also be PMA that blocks access, which we don’t have control of.

jordancarlin:
@davidharrishmc looks like there are merge conflicts.

tests/priv/ExceptionsZc/ExceptionsZc-00.S
jordancarlin, line 1: Not supposed to be checked in?

generators/testgen/src/testgen/priv/extensions/sm.py
```diff
@@ -579,15 +579,17 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
         [
             test_data.add_testcase(coverpoint, "", covergroup),
             f"\tLI(x{r1}, 42)        # value to write to mtime",
-            f"\tRVMODEL_SET_MTIME(x{r1}, x{r2})    # write MTIME = 42",
+            f"\tLA(x{r2}, RVMODEL_MTIME_ADDRESS)        # load address of mtime",
```
jordancarlin, line 582: What happens if `RVMODEL_MTIME_ADDRESS` is blank (which you said is allowed in the README)? Seems like this would fail to compile.

config/cores/cve2/cv32e20/rvmodel_macros.h
jordancarlin, line 1: Something went wrong with this file. You dropped all of the header comments and deleted the `RVMODEL_IO_WRITE_STR` macro definition.

jordancarlin (changes requested):

jordancarlin (approved):

PR #932, opened 2026-02-13
UDB Config fixes

davidharrishmc:
Updated cvw_rv32gc Sv version to be consistent with Sm
Enabled svadu and svade now that UDB is patched
Fixed missing vector params in cvw_RVB23S64-optional


config/cores/cvw/cvw-rv64gc/cvw-rv64gc.yaml
```diff
@@ -59,8 +59,8 @@ implemented_extensions:
   - { name: Sv39, version: "= 1.13.0" }
```
jordancarlin: Version 1.12

jordancarlin (approved):

PR #942, opened 2026-02-18
ExceptionsZc extension & misaligned priority fetch

Teemooooooooo:
Changed ExceptionsZc 's extensions to `required_extensions`, added `cp_misaligned_priority_fetch`.


generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -38,7 +38,7 @@ def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
     return lines
 
 
-@add_priv_test_generator("ExceptionsZc")
+@add_priv_test_generator("ExceptionsZc", required_extensions=["I", "Zicsr", "Zca", "Zcb", "Zcd", "Zcf", "F", "D", "Sm"])
```
davidharrishmc: Shouldn't the required_extensions drop Zcb, Zcd, Zcf, F, and D, but all of these extensions should show up as march_extensions?  See Jordan's comment yesterday 10:32 pm.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -38,7 +38,7 @@ def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
     return lines
 
 
-@add_priv_test_generator("ExceptionsZc", required_extensions=["I", "Zicsr", "Zca", "Zcb", "Zcd", "Zcf", "F", "D", "Sm"])
+@add_priv_test_generator("ExceptionsZc", required_extensions=["I", "Zicsr", "Zca", "Sm"], march_extensions=["Zcb", "Zcd", "Zcf", "F", "D"])
```
davidharrishmc: March also needs I,Zicsr,Zca,Sm.  March is what goes to gcc, and required_extensions is what goes to select whether the tests should be run.

davidharrishmc (dismissed):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
jordancarlin: Sm is not accepted by gcc, so don't add that one.

Teemooooooooo: Is there any reason that Zca needs to be in required_extensions but all the other ones for Zc just need to be in march?

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -38,7 +38,11 @@ def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
     return lines
 
 
-@add_priv_test_generator("ExceptionsZc")
+@add_priv_test_generator(
+    "ExceptionsZc",
+    required_extensions=["I", "Zicsr", "Sm", "Zca"],
+    march_extensions=["I", "Zicsr", "Zca", "Zcb", "Zcd", "F", "D"],
```
jordancarlin: Let's remove `I` from the list of extensions since we want this to run on E processors as well.

jordancarlin (approved):

PR #946, opened 2026-02-18
NA4 not consistent with PMP grain > 0

davidharrishmc:
Fix inconsistency preventing cvw-rv32gc from running


jordancarlin (approved):
Oops. Probably got messed up when updating all of the configs.

PR #948, opened 2026-02-18
Converted pmpsm and pmpzca tests for ACT4

hamza-1821:
This PR has both rv32 and rv64 tests for pmpsm and pmpzca


Makefile
```diff
@@ -9,7 +9,8 @@ CONFIG_FILES ?= config/spike/spike-rv32-max/test_config.yaml config/spike/spike-
 COVERAGE_CONFIG_FILES ?= config/sail/sail-rv64-max/test_config.yaml config/sail/sail-rv32-max/test_config.yaml
 
 WORKDIR     ?= work
-EXTENSIONS  ?= I,M,F,D,Zca,Zcf,Zcd,Zaamo,Zalrsc,Zifencei,Zicntr,Sm # Extensions to generate tests for. Leave blank to generate for all tests.
+WORKDIR_REF ?= work-ref
+EXTENSIONS  ?= I,Zca,Zcf,Zcd,Zaamo,Zalrsc,Zifencei,pmp # Extensions to generate tests for. Leave blank to generate for all tests.
```
jordancarlin: Revert changes to default list of extensions. This should match RVI20 for now.

Makefile
```diff
@@ -152,3 +153,41 @@ format:
...
+get_pmp:
+	@echo "Generating pmp_tests.sh from ELFs in $(PMP_ELF_DIR)"
+	@echo "#!/bin/bash" > pmp_tests.sh
+	@echo "mkdir -p logs" >> pmp_tests.sh
+	@for elf in $(PMP_ELF_DIR)/*.elf; do \
+		name=$$(basename $$elf .elf); \
+		echo "wsim rv32gc $$elf --lockstepverbose > logs/$${name}_log.txt" >> pmp_tests.sh; \
+	done
+	@chmod +x pmp_tests.sh
+	@echo "pmp_tests.sh created."
```
jordancarlin: I'd rather we not add all of this very Wally specific code. Can this be a script in the cvw repo instead?

tests/priv/pmp/pmp32/pmpsm_all_entries_check-03.S
```diff
@@ -0,0 +1,324 @@
...
+# CONFIG_DEPENDENT: true
+##### END_TEST_CONFIG #####
+
+// Define the name of this test file for reporting and the size of the signature region
+#define TEST_FILE "pmpm_all_entries_check-03.S"
+#define SIGUPD_COUNT 50
+
+// Enable trap handler
+#define rvtest_mtrap_routine
+#define rvtest_strap_routine
```
jordancarlin: This can't enable `rvtest_strap_routine` unless S-mode is present.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-03.S
```diff
@@ -0,0 +1,324 @@
...
+// Set up test environment and declare main entry point
+RVTEST_BEGIN
+
+#ifndef RVMODEL_PMP_GRAIN
+	#define RVMODEL_PMP_GRAIN 4
+#endif
+
+#ifndef RVMODEL_NUM_PMPS
+	#define RVMODEL_NUM_PMPS 64
+#endif
```
jordancarlin: I'd rather we drop this and require that they be set by the user.

tests/priv/pmp/pmp32/pmpsm_all_entires_check-03.S
```diff
@@ -0,0 +1,324 @@
...
+//               access. Just basic checks since bottom PMP_writable_regs have been 
+//               tested thoroughly. 64 Regions Only, With pmpcfg.L=1, pmpcfg.XWR=101,
+//               attempt to lw and sw to each standard region. Read should succeed and
+//               write should fail, proving all PMP entries are usable. 
+//
+// Dependencies: model_test.h, arch_test.h
+// --------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: ['I','Smpmp']
```
jordancarlin: Instead of requiring the made-up Smpmp extension (which UDB is planning to get rid of soon), can this use the NUM_PMP_ENTRIES param as the condition for test selection?

jordancarlin (changes requested):

Makefile
hamza-1821: oops, i might have pushed my changes accidently.I reverted the makefile to it's original verison from act4 branch.

tests/priv/pmp/pmp32/pmpsm_all_entires_check-03.S
allenjbaum: Um, shouldn't this file be "pmpsm_all_entries_check-03.S" rather than "entires"?

hamza-1821: > Um, shouldn't this file be "pmpsm_all_entries_check-03.S" rather than "entires"?

yes, it's fixed in the latest commit

tests/priv/pmp/pmp32/pmpsm_all_entries_check-03.S
jordancarlin: @hamza-1821 This is still an issue. The tests do not require S but unconditionally enable the S-mode trap handler.

hamza-1821: got it ! ill fix that in the next commit.

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
```diff
@@ -0,0 +1,199 @@
...
+// Test Cases  : Checking that A is writable in each region. For each standard
+//               region, attempt to write pmpcfg.A={OFF, TOR, NA4, NAPOT} with
+//               pmpcfg.L=0, pmpcfg.RWX = 000.  Only able to write NA4 if grain
+//               G = 0. Other pmpcfg bits are writable.
+//
+// Dependencies: model_test.h, arch_test.h
+// --------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS:  ['I','S','Smpmp']
```
jordancarlin: These tests are called PMPSm, but they currently claim to require the S extension. Shouldn't they only require Sm?

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
```diff
@@ -0,0 +1,199 @@
...
+# CONFIG_DEPENDENT: true
+##### END_TEST_CONFIG #####
+
+// Define the name of this test file for reporting and the size of the signature region
+#define TEST_FILE "pmpm_cfg_A_all.S"
+#define SIGUPD_COUNT 50
+
+// Enable trap handler
+#define rvtest_mtrap_routine
+#define rvtest_strap_routine
```
jordancarlin: Why is the smode trap handler needed for PMPSm tests? These should not be doing anything in supervisor mode.

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
```diff
@@ -0,0 +1,199 @@
...
+//                Muhammad Abdullah, Hamza Ali, Muhammad Zain
+//
+// Description : This test verifies the functionality and enforcement of
+//               Physical Memory Protection (PMP) configurations in RISC-V
+//               systems. It specifically tests the Read, Write, and Execute
+//               permissions for a designated memory region, ensuring that
+//               the PMP settings are correctly applied and that the system
+//               behaves as expected when accessing this region.
+//
+// Notes       : - Ensure `model_test.h` is configured to match your PMP grain
```
jordancarlin: `model_test.h` has been renamed to `rvmodel_macros.h`.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-03.S
jordancarlin: This still needs to be resolved.

tests/priv/pmp/pmp32/pmpzca_aligned_napot.S
```diff
@@ -0,0 +1,193 @@
...
+//
+// Test Cases  :  Setting up 3 standard consecutive NAPOT PMP regions with L=1, XWR=111
+//               (XWR = 000 for region 3).Place an uncompressed ret = jalr ra inside
+//               first region and inside second region. Attempt jalr to each ret.
+//
+// Dependencies: model_test.h, arch_test.h
+// ---------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: ['I','C','S','Smpmp']
```
jordancarlin: These should require `Zca`, not `C`. We can run this test even if a DUT only supports Zca but not Zcd (which C would require).

jordancarlin (changes requested):
Beyond the labels that need to be updated, it looks like there are a few outstanding issues with the configuration settings.

Also, have you checked coverage with this version of the tests (by running `make coverage --jobs $(nproc)` and checking the report in `work/sail-rv{32/64}-max/reports`)? It looks like right now it won't run the PMP coverage because the directory name doesn't match the coverage file. The PMPSm tests need to be in a PMPSm directory so they get run against the PMPSm covergroups. Same for PMPZca tests in a PMPZca directory.

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
hamza-1821: The tests from prev version were following a template, that included S extension in m mode tests for some reason in RVTEST_CASE. So It got converted as is. Same for smode trap handler being there in m mode tests. I have started removing it test by test to check if in any rare case scenario something breaks.

jordancarlin: Great. We definitely don't want to use S mode in Sm tests, so glad that we are fixing the tests as part of porting them to ACT4.

@UmerShahidengr, do you recall why S was included in the original ACT3 version?

UmerShahidengr:
@hamza-1821 you are getting a conflict, and CI is still failing

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
UmerShahidengr: S was included only when Virtual memory was enabled and we require Smode save area, or when we are switching to Smode, other than that Mmode tests are not using anything from Smode

hamza-1821:
> @hamza-1821 you are getting a conflict, and CI is still failing

yes, haven't pushed the latest changes yet. That will resolve CI issue. Still working on it.

tests/priv/pmp/pmp32/pmpzca_aligned_napot.S
hamza-1821: noted.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-03.S
hamza-1821: fixed

tests/priv/pmp/pmp32/pmpsm_all_entries_check-03.S
hamza-1821: Done !

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
hamza-1821: I have removed smode trap handlers from m mode tests.

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
hamza-1821: Fixed.

tests/priv/pmp/pmp32/pmpzca_aligned_napot.S
hamza-1821: Done.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+//
+// Coverpoints : cp_pmp64 for PMPM is partially covered in this test file.
+//
+// Test Cases  : Checking all (64-PMP_writable_regs) PMP registers affect load/store
+//               access. Just basic checks since bottom PMP_writable_regs have been
+//               tested thoroughly. 64 Regions Only, With pmpcfg.L=1, pmpcfg.XWR=101,
+//               attempt to lw and sw to each standard region. Read should succeed and
+//               write should fail, proving all PMP entries are usable.
+//
+// Dependencies: model_test.h, arch_test.h
```
jordancarlin: This list of dependencies is outdated. Rather than trying to keep it up to date, we're probably better off just dropping it. All of the tests have the same dependencies.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+// Description : This test verifies the functionality and enforcement of
+//               Physical Memory Protection (PMP) configurations in RISC-V
+//               systems. It specifically tests the Read, Write, and Execute
+//               permissions for a designated memory region, ensuring that
+//               the PMP settings are correctly applied and that the system
+//               behaves as expected when accessing this region.
+//
+// Notes       : - Ensure `rvmodel_macros.h` is configured to match your PMP grain
+//                 and PMP region count before running this test.
+//               - Adjust PMP configurations for specific hardware if required.
```
jordancarlin: What does this note mean and how does it differ from the one above?

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
+// --------------------------------------------------------------------------------
+// Title        : Comprehensive PMP (Physical Memory Protection) Verification
```
jordancarlin: This file (and all of the PMP tests) are missing license headers. Please add an appropriate license. Most of the other tests use Apache 2.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+
+#define SKIP_MEPC
+
+// Include framework macros
+#include "riscv_arch_test.h"
+
+// Set up test environment and declare main entry point
+RVTEST_BEGIN
+
+#define NOP         0x13
```
jordancarlin: If this is needed, should we move it to a more central location? Maybe `utils.h`. Right now it is defined in all of these tests.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+
+.macro VERIFICATION_RWX ADDRESS, TEST_CASE
+
+    LA(a5, \ADDRESS)                                        // Address to be ve
```
jordancarlin: Comment unclear.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+    \TEST_CASE\()_1:
+    sw a4, 0(a5)
+    nop
+    RVTEST_SIGUPD(x2, x5, x4, a4,\TEST_CASE\()_1,test_2_str)
```
jordancarlin: Why does TEST_CASE_1 point at test_2_str?

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+    \TEST_CASE\()_2:
+    lw a4, 0(a5)
+    nop
+    RVTEST_SIGUPD(x2, x5, x4, a4,\TEST_CASE\()_2,test_1_str)
```
jordancarlin: Why does TEST_CASE_2 point at test_1_str?

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+
+#if RVMODEL_PMP_GRAIN != 0
+    #define STANDARD_REGION   (((REGIONSTART >> PMP_SHIFT) & PMP_MASK) | PMP_REGION_SIZE)
+#else
+    #define STANDARD_REGION   (REGIONSTART >> PMP_SHIFT)
+#endif
+
+#ifdef rvtest_strap_routine
+    csrw satp, zero                // Disable address translation.
+#endif
```
jordancarlin: You are never defining `rvtest_strap_routine` in this file, so I think this can safely be dropped.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+
+    LA(x4, PMPADDRESS_TOP_GLOBAL)
+    csrw    pmpaddr63, x4
+
+    //Lock the Background region.
+    LI(x4, (PMP_REGION_GLOBAL))
+    csrw pmpcfg15, x4
```
jordancarlin: What happens on a hart that only has 16 PMP regions? This will trap and we will not end up with a background region.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-02.S
```diff
@@ -0,0 +1,327 @@
...
+// Set up test environment and declare main entry point
+RVTEST_BEGIN
+
+#ifndef RVMODEL_PMP_GRAIN
+    #define RVMODEL_PMP_GRAIN 4
+#endif
+
+#ifndef RVMODEL_NUM_PMPS
+    #define RVMODEL_NUM_PMPS 64
+#endif
```
jordancarlin: Looks like this was removed from some but not all of the tests. Please double check to make sure it is removed from all of the tests since we are going to require the user to always define this.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-02.S
jordancarlin, line 1: What is the point of splitting pmpsm_all_entries_check into 4 parts? Seems like it would be easier to understand what each test does if this was kept together as a single test.

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
```diff
@@ -0,0 +1,188 @@
...
+    csrw pmpcfg3, x4
+
+//                                          Verification Section
+// Test Case 1 : Setting region 0-14 to be NA4, NAPOT, TOR and OFF with L->0, XWR->000
+
+    LI(x4, (NA4 << PMP3_CFG_SHIFT|NA4 << PMP2_CFG_SHIFT|NA4 << PMP1_CFG_SHIFT|NA4 << PMP0_CFG_SHIFT))
+    csrw pmpcfg0, x4
+    csrw pmpcfg1, x4
+    csrw pmpcfg2, x4
+    csrw pmpcfg3, x4
```
jordancarlin: This test seems to be missing any kind of self-checking. We need to read back what we wrote to each CSR to make sure it holds the proper value. Look at the `RVTEST_SIGUPD_CSR_WRITE` macro in `signature.h`

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
```diff
@@ -0,0 +1,301 @@
...
+
+exit:
+
+RVTEST_CODE_END
+RVTEST_DATA_BEGIN
+.align 4
+
+// Test case strings for reporting
+canary_mismatch: .string "Testcase signature canary mismatch!"
+test_1_str: .string "\"test: 1; cp: cp_pmpm64_lw\""
```
jordancarlin: Fine to defer this to a follow-up PR after this one is merged, but the test strings should include the covergroup and bin in addition just the coverpoint. That will probably require the `VERIFICATION_RWX` macro to include the `TEST_CASE` number argument as part of the debug string (similar to what is already done for the labels).

tests/priv/pmp/pmp32/pmpzca_aligned_off.S
```diff
@@ -0,0 +1,158 @@
...
+##### END_TEST_CONFIG #####
+
+// Define the name of this test file for reporting and the size of the signature region
+#define TEST_FILE "pmpzca_aligned_tor.S"
+#define SIGUPD_COUNT 50
+
+// Enable trap handler
+#define rvtest_mtrap_routine
+#define RVTEST_PRIV_TEST
+#define rvtest_strap_routine
```
jordancarlin: Why do the Zca tests still require S and the rvtest_strap_routine?

jordancarlin (changes requested):
Still haven't reviewed most of the tests, but here are more structural comments. Also still need the test directory structure to be recategorized so that coverage works.

tests/priv/pmp/pmp32/pmpzca_aligned_off.S
hamza-1821: might have missed the removal of strap handler from zca tests, and some sm tests. Since these tests only run in M mode, i will remove it.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
hamza-1821: TEST_CASE_1 is label for sw and it points to test_2_str because that string in the data region is for sw. But i will swap them now. test case 1 will point to tetst_1_str (for sw) now. Same for lw.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
hamza-1821: Test plan requirement is to check upper entries (15-62). This should only be a single test testing upper entries. Can we put a test selecting check so that test is only selected if 64 entries are supported?

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
hamza-1821: In the older version, there were some tests requiring pmp grain to be 0. we had checks in regex for pmp grain and pmp count for the test to be selected, and this note gives you a hint that configurations should be matched. We can remove this note in the next commit.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-02.S
hamza-1821: This test was split into 4 parts most probably beacause of the size of the signature region being too large for a single test. We can definitely use a single test that tests the upper pmp entries.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
jordancarlin: Yes. You can do it the same way that `MXLEN` is handled. You can add any other UDB parameters to that list along with the value that they must have for the test to be selected. Reach out on Slack if you need help implementing this.

tests/priv/pmp/pmp32/pmpsm_cfg_A_all.S
jordancarlin: Please double check all of the tests. There are likely other parts that are missing self-checking beyond this one too.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
hamza-1821: > TEST_CASE_1 is label for sw and it points to test_2_str because that string in the data region is for sw. But i will swap them now. test case 1 will point to tetst_1_str (for sw) now. Same for lw.

Done.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
hamza-1821: Got it, i've added the check in this test.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-02.S
hamza-1821: Should be entirely removed now.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -0,0 +1,553 @@
...
+//               attempt to lw and sw to each standard region. Read should succeed and
+//               write should fail, proving all PMP entries are usable.
+//
+// --------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: ['I','Smpmp']
+# params:
+#   MXLEN: 32
+#   RVMODEL_NUM_PMPS: 64
```
jordancarlin: This needs to use the parameter names from UDB. `NUM_PMP_ENTRIES` in this case. See https://riscv.github.io/riscv-unified-db/manual/html/isa/isa_20240411/params/param_list.html for the list of parameters you can use.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -0,0 +1,553 @@
...
+
+#define SKIP_MEPC
+
+// Include framework macros
+#include "riscv_arch_test.h"
+
+// Set up test environment and declare main entry point
+RVTEST_BEGIN
+
+#define NOP         0x13
```
jordancarlin: This should be dropped now that it's in `utils.h`. Global change.

tests/env/utils.h
```diff
@@ -89,6 +89,17 @@
...
+// PMP macros
+#define PMP0_CFG_SHIFT  0
+#define PMP1_CFG_SHIFT  8
+#define PMP2_CFG_SHIFT  16
+#define PMP3_CFG_SHIFT  24
+#define PMP4_CFG_SHIFT  32
+#define PMP5_CFG_SHIFT  40
+#define PMP6_CFG_SHIFT  48
+#define PMP7_CFG_SHIFT  56
+#define NOP32           0x13              //single nop
```
jordancarlin: Should this just be called NOP? Not sure what is meant by NOP32 since instructions are always 32 bits.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -0,0 +1,553 @@
...
+# params:
+#   MXLEN: 32
+#   RVMODEL_NUM_PMPS: 64
+# MARCH: rv32i_zicsr
+# CONFIG_DEPENDENT: true
+##### END_TEST_CONFIG #####
+
+// Define the name of this test file for reporting and the size of the signature region
+#define TEST_FILE "pmpsm_all_entries_check-01.S"
+#define SIGUPD_COUNT 50
```
jordancarlin: This count doesn't seem right. I think there are two calls to SIGUPD per VERIFICATION_RWX macro and 47 calls to VERIFICATION_RWX, so it needs to be at least 94. Please double check this for all tests to make sure the signature is a sufficient size based on the number of SIGUPD macros.

UmerShahidengr:
@hamza-1821 Clang CI was running for almost 2hours, I have cancelled that. CI was stuck at Sm-00.S test.
I cant seem to find anything in this PR which should keep it stuck. Can you check why it is hanging?
CC: @jordancarlin

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_bot.S
```diff
@@ -0,0 +1,229 @@
...
+#define SKIP_MEPC
+
+// Include framework macros
+#include "riscv_arch_test.h"
+
+// Set up test environment and declare main entry point
+RVTEST_BEGIN
+
+#define NOP             0x13
+#define g       (1<<(RVMODEL_PMP_GRAIN+2))
```
jordancarlin, line 54: g will be the same for all tests. Might make sense to define it once in a more central location.

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_check-01.S
```diff
@@ -0,0 +1,206 @@
...
+.macro VERIFICATION_RWX ADDRESS TEST_CASE
+
+    LI(a4, NOP)                                             // Value to write (NOP)
+
+    LA(a5, (\ADDRESS-4))                                    // Address to be verified
+    \TEST_CASE\()_1:
+    sw a4, 0(a5)                                            // word-level store test
+    nop
+    nop
+    RVTEST_SIGUPD(x2, x5, x4, a4, \TEST_CASE\()_1, test_1_str)                               // Signature update
```
jordancarlin, line 62: Why do we need 2 `nop` in all of these? If it traps, doesn't it just skip the next instruction? So only one `nop` should be needed.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk.S
```diff
@@ -0,0 +1,346 @@
...
+    csrw pmpaddri, t0
+    .set pmpaddri, pmpaddri+1
+    .endr
+
+    // Loop to SET ALL pmpcfg REGs to zero
+    .set pmpcfgi, CSR_PMPCFG0
+    .rept RVMODEL_NUM_PMPS/4
+    csrw pmpcfgi , x0
+    .set pmpcfgi, pmpcfgi+1
+    .endr
```
jordancarlin: This is all missing self-checking.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk.S
```diff
@@ -0,0 +1,346 @@
...
+        test_62:
+                RVTEST_SIGUPD_CSR_WRITE(pmpaddr61, t2, test_62, test_62_str)
+        test_63:
+                RVTEST_SIGUPD_CSR_WRITE(pmpaddr62, t2, test_63, test_63_str)
+        test_64:
+                RVTEST_SIGUPD_CSR_WRITE(pmpaddr63, t2, test_64, test_64_str)
+    .endif
+
+    slli t2, t2, 1
+    bnez t2, 9b
```
jordancarlin: I think we decided that we don't want tests to have loops in them. @davidharrishmc thoughts on this one since it is not coming from a Python generator?

tests/priv/pmp/pmp32/PMPZca/pmpzca_aligned_na4.S
```diff
@@ -0,0 +1,147 @@
...
+// Test Cases  :  Setting up 3 standard consecutive NA4 PMP regions with L=1, XWR=111
+//               (XWR = 000 for region 3).Place an uncompressed ret = jalr ra inside
+//               first region and inside second region. Attempt jalr to each ret.
+// ---------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: ['I','Zca','Smpmp']
+# params:
+#   MXLEN: 32
+# MARCH: rv32ic_zicsr
```
jordancarlin: Can we use zca instead of c in the march string here? Applies to all Zca tests.

jordancarlin (dismissed):

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk.S
UmerShahidengr: This loop is applied only to initialise all pmp-regs to zero, so self-checking is not required here.

jordancarlin: It sets all of the pmpaddr CSRs to zeros and then ones, which is listed in the header as one of the things this test is checking.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk.S
davidharrishmc: We've gone to a ton of work to remove loops because it is hard to report failures when there are loops.  If it is practical to remove this loop as well, I'd prefer to do so.

tests/env/utils.h
hamza-1821: I've changed NOP32 to NOP and NOP64 to DOUBLE_NOP(it's simply 2 NOPS)

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_bot.S
hamza-1821: some tests (pmpcfgA=NAPOT) use a different g. 
For example in napot tests, we have a condition for g
#if RVMODEL_PMP_GRAIN != 0
  #define g   (1 << (RVMODEL_PMP_GRAIN + 2))
#else
  #define g   (1 << (RVMODEL_PMP_GRAIN + 3))
#endif
so i think we should keep them inside tests. And g itself is not used by all of the tests.

tests/priv/pmp/pmp32/PMPZca/pmpzca_aligned_na4.S
hamza-1821: This is done in the latest commit.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk.S
UmerShahidengr: Done

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk.S
UmerShahidengr: Loops have now been removed

UmerShahidengr (approved):
LGTM, Good work @hamza-1821

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_check-01.S
hamza-1821: This is future work, we'll get it in the next PR.

tests/priv/pmp/pmp32/pmpsm_all_entries_check-01.S
hamza-1821: Sure. A follow up PR will hanlde it.

PR #950, opened 2026-02-19
[ACT4] Enable `.option norelax` for all tests and add 8k `jal`test

jordancarlin:
- Wrap all tests in `.option norelax` to avoid linker relaxations/optimizations.
  - This fixes a bug that came up in the D-fmul.d tests for QEMU.
  - Disabling relaxation had the side benefit of fixing the Clang issue (all unpriv tests now compile and achieve 100% coverage when compiled with clang) and eliminated the `c.nops` in the `LA` macro, so the hack with a jump was removed.
- Fix issue with `cp_imm_edges_jal` coverpoint testing for values above 12 bits when looking at a 12 bit immediate and add 8k `jal` test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ For more details on uv and alternate installation methods, see the [uv installat
 
 #### 3. RISC-V Compiler
 
-The ACT framework is compatible with GCC or LLVM/Clang. This guide uses GCC, but if you prefer LLVM you just need to set the path for the compiler appropriately when [creating your config file](#act-framework-configuration-file).
+The ACT framework is compatible with GCC or LLVM/Clang. This guide uses GCC, but if you prefer LLVM you just need to set the path for the compiler appropriately when [creating your config file](#act-framework-configuration-file). See [config/sail/sail-rv64-max-clang/test_config.yaml](./config/sail/sail-rv64-max-clang/test_config.yaml) for an example.
 
 > [!NOTE]
 >
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ For more details on uv and alternate installation methods, see the [uv installat
 
 #### 3. RISC-V Compiler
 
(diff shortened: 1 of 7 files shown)


generators/coverage/templates/cp_imm_edges_jal.txt
```diff
@@ -12,8 +10,8 @@
         bins b_512   = {512};
         bins b_1024  = {1024};
         bins b_2048  = {2048};
-        bins b_4096  = {[4064:4096]};
-        bins b_8192  = {8192};
+        bins b_4096  = {4096};
```
davidharrishmc, line 13: How do we hit this bin?  Max positive 12 bit 2s comp is 2047. Multiply by 2 for 4094.

tests/env/test_setup.h
```diff
@@ -17,7 +17,11 @@
   .global rvtest_entry_point
   rvtest_entry_point:
 
-  // Disable assembler/linker optimizations
+  // Globally disable linker relaxation
```
davidharrishmc, line 20: Can you add a longer explanation of why it is needed and what goes wrong if not?

README.md
```diff
@@ -64,7 +64,7 @@ For more details on uv and alternate installation methods, see the [uv installat
 
 #### 3. RISC-V Compiler
 
-The ACT framework is compatible with GCC or LLVM. This guide uses GCC, but if you prefer LLVM you just need to set the path for the compiler appropriately when [creating your config file](#act-framework-configuration-file).
+The ACT framework is compatible with GCC or LLVM/Clang. This guide uses GCC, but if you prefer LLVM you just need to set the path for the compiler appropriately when [creating your config file](#act-framework-configuration-file).
```
davidharrishmc: Should this also link to a config file set up for clang?

generators/coverage/templates/cp_imm_edges_jal.txt
jordancarlin: `jal` has a 20 bit immediate, not 12.

davidharrishmc (approved):

davidharrishmc (approved):

PR #971, opened 2026-02-20
[ACT4] Removed relocation & added feature to report failure on an early exit

Zain2050:
I have removed the relocation logic for XTVAL as it is no longer required. 
Removed RVTEST_PTE_IDENT_MAP as it is not being used in any test. Using .skip to allocate memory for page tables instead.
Now the test reports an error message if it exits prematurely. It jumps to abort_test, which cleans up epilogs and then exits the test with a failure message. Now on an early exit we get an output like this.

```
RVCP-SUMMARY: Test File "Sm-00.S": FAILED
RVCP: Test Info: "The test encountered an early exit!"
RVCP: Instruction: 0xbad0e4b7
RVCP: Address: 0x8002c304
RVCP: Register: 0x0000000a
RVCP: Bad Value: 0xbad0dead
RVCP: Expected Value: 0x00000000

HTIF located at 0x8002e000
begin_signature: 0x8009f250
end_signature: 0x800cbf70
Entry point: 0x80000000
FAILURE: 1
```


tests/env/test_setup.h
```diff
@@ -188,28 +191,25 @@
...
+      .skip(4096)       // 4KB page table
+    #ifdef rvtest_htrap_routine
+      .align 14
+      rvtest_Hroot_pg_tbl:
+        .skip(16276)    // 16KB page table
+    #endif
+    #ifdef rvtest_vtrap_routine
+      .align 12
+      rvtest_Vroot_pg_tbl:
+        .skip(4096)     // 4KB page table
```
jordancarlin: Should we make these sizes configurable so that tests can define the size they need and tests that don't need page tables don't allocate unnecessary space? Could be handled similarly to how SIGUPD_COUNT is currently handled.

tests/env/test_setup.h
```diff
@@ -122,27 +122,30 @@
       RVTEST_TRAP_EPILOG M            // actual m-mode prolog/epilog/handler code
     #endif
 
+    // Check if the test has fully executed
+    LI(     T4, 0xBAD0DEAD)           // T5 = 0xBAD0DEAD when the test is exiting early
+    bne     T4, T5, exit_cleanup      // Not an early exit, exit with a success code
+    jal     T3, failedtest_x8_x7
+    RVTEST_WORD_PTR "early_exit"
+    early_exit: .string "\"The test encountered an early exit!\"";
```
jordancarlin: Can we make this string slightly more descriptive so that it is clear that it failed in the trap handler?

generators/testgen/src/testgen/formatters/types/sc_type.py
```diff
@@ -37,6 +37,7 @@ def format_sc_type(
     setup = [
         load_int_reg("rs2", params.rs2, params.rs2val, test_data),
         f"LA(x{params.rs1}, scratch) # rs1 = base address",
+        "nop",  # Test fails on spike without this nop; nop is a temp fix; TODO: Link to issue after opening it
```
jordancarlin, line 40: Why is the `nop` being moved? I believe these tests are currently passing.

jordancarlin (changes requested):

tests/env/test_setup.h
allenjbaum: I think the issue is that the HW doesn't care how many entries are actually used; they'll blindly go ahead and use a garbage entry. It is safest to fully populate the unused portion with invalid entries (which is what a real OS must do). You can argue that tests won't ever touch those entries because we control the tests, of course, but it will make debugging more difficult, and the cost is actually pretty small (since you can define a single all-invalid page that all levels can use.)

Zain2050: I agree, it's better to populate these with invalid PTEs.

Zain2050: @jordancarlin I don't think we should change the size of a page table as in the tests I try to use random PTEs, instead of just restricting to the top ones. Also, let's say we use only 2KB for the page table, but the test accidentally accesses something beyond that (some other trap handler code), that will also be a problem.

However, one approach could be to encapsulate these in a flag like RVTEST_INSTANTIATE_ROOT_PAGE_TABLES so only tests that use page tables allocate memory for them. But we'll have to do it separately for all 3, as the labels are being used while instantiating RVTEST_TRAP_PROLOG. 
Hroot and Vroot are enabled by htrap and vtrap routines and won't be used much. It will be good for strap_routine ones. Do you think we should do it?

generators/testgen/src/testgen/formatters/types/sc_type.py
Zain2050: After these changes, the spike regression started to fail again, but on a different testcase. I moved the nop to that testcase and it got resolved. Pretty weird. I’ll need to allocate some time to investigate the root cause.

tests/env/test_setup.h
jordancarlin: Sounds good to leave as a static size. Let's leave it as it is for now and we can always add `RVTEST_INSTANTIATE_ROOT_PAGE_TABLES` later if we think it would be helpful.

generators/testgen/src/testgen/formatters/types/sc_type.py
jordancarlin: Yikes. Let's definitely get an issue opened for Spike once we have an idea about what is going wrong.

jordancarlin (approved):

tests/env/test_setup.h
```diff
@@ -122,27 +122,29 @@
       RVTEST_TRAP_EPILOG M            // actual m-mode prolog/epilog/handler code
     #endif
 
+    LI(     T4, 0xBAD0DEAD)           // T5 holds 0xBAD0DEAD if abort_tests was executed
+    bne     T4, T5, exit_cleanup      // Exit with a success message if not being aborted
```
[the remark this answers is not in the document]
Zain2050: I have made sure that RVTEST_TRAP_PROLOG doesn't corrupt T5

tests/env/test_setup.h
```diff
@@ -188,28 +190,25 @@
...
+    #ifdef rvtest_htrap_routine
+      .align 14
+      rvtest_Hroot_pg_tbl:
+        .fill (16384/REGWIDTH), REGWIDTH, 0   // 16KB page table
+    #endif
+    #ifdef rvtest_vtrap_routine
+      .align 12
+      rvtest_Vroot_pg_tbl:
+        .fill (4096/REGWIDTH), REGWIDTH, 0    // 4KB page table
     #endif
```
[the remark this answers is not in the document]
Zain2050: Seems better to just use `.zero`

Zain2050:
I have also pushed updated encoding.h (with medelegh) and un-commented CSR_XEDELEGH from the trap handler for M mode only. We are unable to access hedelegh as sail doesn't support it yet.

tests/env/failure_code.h
```diff
@@ -253,6 +253,8 @@
         .fill 2, 4, 0xfeedf00dbaaaaaad
     ascii_buffer:
         .fill 20, 1, 0          # Buffer for hex string conversion (max "0x" + 16 hex digits + "\n" + null)
+    abort_str:
+        .string "\"The trap handler aborted the test before normal completion!\"";
```
jordancarlin: Move this string after end_failure_scratch. It is a preset string, not part of the scratch space.

allenjbaum:
Has anyone filed a bug against Spike?
 I don't like the idea of rewriting valid sequences as a workaround for a
bug that we've acknowledged isn't officially supported as a reference.
If spike does something that is different than Sail, there is a bug that is
lurking that needs correcting (though I can't say in which from here)

Zain2050:
@allenjbaum This is not a bug. Sail just doesn't have hypervisor implemented yet (therefore we are unable to use hedelegh). I have made the modification just so that we can still use medelegh instead of waiting for Sail to support hedelegh.

jordancarlin:
> @allenjbaum This is not a bug. Sail just doesn't have hypervisor implemented yet (therefore we are unable to use hedelegh). I have made the modification just so that we can still use medelegh instead of waiting for Sail to support hedelegh.

I think @allenjbaum is talking about the nop bug for lrsc in spike, not hedelegh in Sail.

jordancarlin (approved):

UmerShahidengr (approved):

PR #976, opened 2026-02-21
Move CTP and CRD into `docs` and use common `docs-resources` submodule

jordancarlin:


jordancarlin:
@james-ball-qualcomm you should be good to replace the current CRD src with the official version after this is merged.

davidharrishmc (approved):

PR #977, opened 2026-02-21
Added coverpoints in Smstateen.yaml

ayeshaanwaar05:
Smstateen Normative rule to coverpoint mappings


coverpoints/norm/Smstateen.yaml
```diff
@@ -3,79 +3,73 @@
 
 normative_rule_definitions:
   - name: smstateen_mmode_rv64_csrs
-    coverpoint: [""]
+    coverpoint: ["cp_walking_ones"]
```
jordancarlin: See other norm rule mapping files. This should include the covergroup as well.

jordancarlin (changes requested):

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+        bins mstateen1h = {CSR_MSTATEEN1H};
+        bins mstateen2h = {CSR_MSTATEEN2H};
+        bins mstateen3h = {CSR_MSTATEEN3H};
+    `endif
+    }
+
+    // ── Walking ones/zeros coverpoint ─────────────────────────────────────
+    `ifdef XLEN64
+    csr_walk: coverpoint ins.current.rs1_val {
+        //wildcard bins walking1_0  = {64'b???????????????????????????????????????????????????????????????1};
```
davidharrishmc, line 73: Why are these commented out?

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+    }
+    csrr: coverpoint ins.current.insn {
+        wildcard bins csrr = {CSRR};
+    }
+    csrrs: coverpoint ins.current.insn {
+        wildcard bins csrrs = {CSRRS};
+    }
+
+    // ── CSR address coverpoint — all mstateen CSRs ────────────────────────
+    csr: coverpoint ins.current.insn[31:20] {
```
davidharrishmc: It would be clearer to call this something like mstateen_csrs rather than just "csr".

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+    csrrw: coverpoint ins.current.insn {
+        wildcard bins csrrw = {CSRRW};
+    }
+    csrr: coverpoint ins.current.insn {
+        wildcard bins csrr = {CSRR};
+    }
+    csrrs: coverpoint ins.current.insn {
+        wildcard bins csrrs = {CSRRS};
+    }
+
```
davidharrishmc, line 29: Let's add a csrop coverpoint with csrw, csrr, csrs, and csrc, and use it on all of the tests where we have been using CSRRW, to be a bit more comprehensive.

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+`endif
+
+    // ── Sm1p13 + Hypervisor dependent coverpoints (cp_p1p13) ─────────────
+`ifdef SM1P13_AND_H_SUPPORTED
+    p1p13_state: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_CURRENT, "mstateen0", "p1p13") {
+        bins p1p13_disabled = {1'b0};
+        bins p1p13_enabled  = {1'b1};
+    }
+    hedeleg_csr: coverpoint ins.current.insn[31:20] {
+        wildcard bins hedeleg = {CSR_HEDELEG};
```
davidharrishmc: change to HEDELEGH

davidharrishmc: several places

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+    // ── Sctr-dependent coverpoints (cp_ctr) ──────────────────────────────
+`ifdef SCTR_SUPPORTED
+    ctr_state: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_CURRENT, "mstateen0", "ctr") {
+        bins ctr_disabled = {1'b0};
+        bins ctr_enabled  = {1'b1};
+    }
+    ctr_csrs: coverpoint ins.current.insn[31:20] {
+        wildcard bins sctrdepth  = {CSR_SCTRDEPTH};
+        wildcard bins sctrstatus = {CSR_SCTRSTATUS};
+        // sctrl intentionally missing from coverage file
```
davidharrishmc: Please clarify why it was omitted.  If it causes crashes when written, can we at least read these registers?

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+    }
+`endif
+
+
+    cp_csr_illegal_accesses: cross priv_mode_u, csr, csrrw;
+    cp_walking_ones: cross csr, csrrw, csr_walk;
+
+`ifdef ZFINX_SUPPORTED
+    cp_fcsr: cross misa_F, mstateen0_fcsr_bit, csrrw, fscr_csr {
```
davidharrishmc: change csrrw to csrop globally per comment above.

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+        ignore_bins ig1 = binsof(misa_F.F_set)  && binsof(mstateen0_fcsr_bit.fcsr_zero);
```
davidharrishmc, line 377: why do we need to exclude all of the fcsr_zero cases in all the ZFINX tests?  Is there any harm in attempting them and confirming a trap takes place?

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+    }
+`endif
+
+    // Row 8: Always present (envcfg always in mstateen0)
+    cp_envcfg: cross csrrw, senvcfg_csr, envcfg_state;
+
+    // Row 7: Zcmt only
+`ifdef ZCMT_SUPPORTED
+    cp_jvt_access:    cross csrrw, jvt_csr, jvt_state;
+    cp_jvt_lower_mode: cross priv_mode_s_u, csrrw, jvt_csr, jvt_state;
```
davidharrishmc: We don't know for sure if S is supported when running Smstateen.  Could you make your own priv_mode_maybes_u coverpoint that only addresses S_mode if S_SUPPORTED?

davidharrishmc: This affects everywhere priv_mode_s_u is used in this file.

coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -0,0 +1,388 @@
...
+    // Row 14: Ssqosid only
+`ifdef SSQOSID_SUPPORTED
+    cp_srmcfg: cross csrrw, srmcfg_csr, srmcfg_state;
+`endif
+
+    // Row 15: Sctr only
+`ifdef SCTR_SUPPORTED
+    cp_ctr: cross csrrw, ctr_csrs, ctr_state;
+`endif
+
```
davidharrishmc, line 425: Is there a reason cp_csrind is not implemented?

Looks like cp_aia and cp_imsic are not implemented.  Is it possible to add those?

coverpoints/priv/Ssstateen_coverage.svh
```diff
@@ -0,0 +1,350 @@
+`define COVER_SMSTATEEN
+`define SVNAPOT_SUPPORTED
+covergroup Ssstateen_cg with function sample(ins_t ins);
```
davidharrishmc, line 11: many comments from Smstateen carry over here.

davidharrishmc: ssstateen0 only has JVT, FCSR fields.  The test plan still has all the others.  Delete the other fields from the test plan and coverpoints.

mstateen0.se0 must be 1 to access sstateen0.  Add that requirement to all of the crosses in this file.

generators/testgen/src/testgen/priv/extensions/Smstateen.py
```diff
@@ -0,0 +1,853 @@
+##################################
```
davidharrishmc: Have not reviewed this yet.

davidharrishmc (changes requested):

coverpoints/priv/Smstateen_coverage.svh
ayeshaanwaar05: Because all of the bits except envcfg depend on other extensions that haven’t been implemented yet.

coverpoints/priv/Smstateen_coverage.svh
ayeshaanwaar05: I think I was told that AIA and IMSIC states weren't implemented and at that time I had no idea like how to test those two states but I'll look at it again and will try to add cp for those too even if it isn't implemented . Will discuss it with Sir umer

davidharrishmc:
Try

imsic_csrs:
	stopei, vstopei

aia_csrs:
	`ifdef XLEN32
		sieh, siph

`ifdef IMSIC_SUPPORTED
	cp_imsec: cross csrops, imsic_csrs, imsic_state;
`endif

`ifdef AIA_SUPPORTED
	cp_aia: cross csrops, aia_csrs, aia_state;

I can’t seem to find other AIA registers, but look if you get a chance.

coverpoints/priv/Smstateen_coverage.svh
ayeshaanwaar05: No harm in testing a trap but testplan specifies to read and write fcsr only when misa.F = 0 and mstateen0.fcsr = 1. The fcsr_zero cases (where a trap would occur) are outside the scope of what this coverpoint is testing per the testplan, so that's why I excluded those bins.

coverpoints/priv/Smstateen_coverage.svh
davidharrishmc: But they aren't commented out in other csr_walk coverpoints in this file.  We should be consistent.  No harm in attempting all of the values and checking that the DUT matches the reference model.

coverpoints/priv/Smstateen_coverage.svh
davidharrishmc: Let's never go out of our way to test comprehensively, unless the outcome is nondeterministic.

coverpoints/priv/Ssstateen_coverage.svh
```diff
@@ -0,0 +1,108 @@
...
+            bins sstateen1 = {CSR_SSTATEEN1};
+            bins sstateen2 = {CSR_SSTATEEN2};
+            bins sstateen3 = {CSR_SSTATEEN3};
+    }
+
+    `ifdef XLEN64
+    csr_walk: coverpoint ins.current.rs1_val {
+            // bits [3:63] are WRPI
+            wildcard bins walking1_0  = {64'b???????????????????????????????????????????????????????????????1};
+            //wildcard bins walking1_1  = {64'b??????????????????????????????????????????????????????????????1?}; // Uncomment Zfinx is supported
```
davidharrishmc: `ifdef ZFINX_SUPPORTED
wildcardbins
`endif
`ifdef ZCMT_SUPPORTE

davidharrishmc (approved):

jordancarlin:
Removed for the merge queue so that #1504 can get in first. The ifdefs in this PR will need to be changed after that before merging. For example, XLEN64 is no longer defined. Use UDB_MXLEN_64 instead. Similar changes are needed for any other ifdefs.

PR #984, opened 2026-02-22
PMP coverpoints

ChiefAbdu:

## Description

Mapped PMP coverpoints to normative rules and generated CTP for PMP succesfully

### Related Issues

PMPU had no coverpoints 

- [ ] Ratified
- [ ] Unratified

### List Extensions

### Reference Model Used

- [ ] SAIL
- [ ] Spike

### Mandatory Checklist:

### Optional Checklist:

- [ ] If you have modified arch_test.h Please provide a detailed description of the changes in the Description section above.


coverpoints/norm/PMPF.yaml
```diff
@@ -0,0 +1,32 @@
+# Normative rule to coverpoint mappings
+# Generated from riscv-isa-manual normative rule definitions
+
+normative_rule_definitions:
+  # Failed memory operations generate an instruction, load, or store access-fault
+  # exception.
+  - name: pmp_access_fault_exception
+    coverpoint: ["cp_cfg_RW"]
```
jordancarlin: See other norm rule mapping files. This should include the covergroup as well.

UmerShahidengr (approved):

coverpoints/norm/PMPF.yaml
```diff
@@ -0,0 +1,32 @@
+# Normative rule to coverpoint mappings
+# Generated from riscv-isa-manual normative rule definitions
+
+normative_rule_definitions:
+  # Failed memory operations generate an instruction, load, or store access-fault
+  # exception.
+  - name: pmp_access_fault_exception
```
davidharrishmc: I have been writing normative rule mappings to generally have each rule only appear once.  For example, the pmp_access_fault_exception rule is covered by coverpoints in PMPSM_cg, so I haven’t been adding rules in PMPF or elsewhere. I’m not trying to associate every coverpoint with a normative rule.  What do you think?

davidharrishmc: This means PMPF.yaml is probably empty because the spec does not define any rules that apply only to floating-point.  Are you ok with this approach?

docs/ctp/src/pmp.adoc
```diff
@@ -5,62 +5,55 @@
 
 [[t-PMPSm-coverpoints]]
 .PMPSm Coverpoints
-,===
```
davidharrishmc, line 8: Don't delete the links to the testplans.  These define the coverpoints.  Just add the new part with the normative rules.

davidharrishmc (changes requested):

coverpoints/norm/PMPF.yaml
ChiefAbdu: Yeah that is one of the things i noticed while mapping the coverpoints, the spec does not define any rules specific to F type. Im okay with this approach

UmerShahidengr: Done

davidharrishmc (approved):

PR #991, opened 2026-02-23
Updated CI to run RVI20 tests on qemu

UmerShahidengr:


davidharrishmc (commented):
Looks good to me, but @jordancarlin is more qualified to approve this one.

.github/workflows/regress.yml
```diff
@@ -205,14 +205,14 @@ jobs:
     strategy:
       matrix:
         include:
-          - name: rv64-max
-            config_file: config/qemu/qemu-rv64-max/test_config.yaml
-            config_name: qemu-rv64-max
+          - name: rv64-RVI20U64
+            config_file: config/qemu/qemu-RVI20U64/test_config.yaml
+            config_name: qemu-RVI20U64
             run_cmd: qemu-system-riscv64 -nographic -semihosting -icount shift=1 -machine virt -cpu max,pmu-mask=0xfffffff8 -bios
```
jordancarlin: If we're running the rvi20 config, we should update qemu to match instead of using `-cpu max`.

config/qemu/qemu-RVI20U32/sail.json
```diff
@@ -0,0 +1,474 @@
...
+  "$schema": "/opt/riscv/share/sail-riscv/sail_riscv_config_schema.json",
+  "base": {
+    "xlen": 32,
+    "E": false,
+    "writable_misa": false,
+    "writable_fiom": true,
+    "writable_hpm_counters": {
+      "len": 32,
+      "value": "0x0"
+    },
```
jordancarlin, line 11: Is this right? I think QEMU supports counters.

jordancarlin (changes requested):

jordancarlin:
@UmerShahidengr Do you want to update this or should we close it? We might be fine without it since we are running all of the tests on QEMU in CI already with the -max configs.

.github/workflows/regress.yml
UmerShahidengr: Done, changes have been made.

config/qemu/qemu-RVI20U32/sail.json
UmerShahidengr: I think thats right for QEMU. I dont think QEMU supports counters

UmerShahidengr:
@jordancarlin thanks for reminding me this. It totally skipped through my mind. I have made the changes now, it is ready for your review.
I think it is better to have CI running Phase-0 and Phase-1 tests on as many simulators as possible. Other profiles can be ignored.

config/qemu/qemu-RVI20U32/sail.json
jordancarlin: They don't count anything, but the CSRs exist and can be tested. They are controlled by the `pmu-mask` flag.

.github/workflows/regress.yml
```diff
@@ -193,39 +193,44 @@ jobs:
...
           - name: rv64-max
             config_file: config/qemu/qemu-rv64-max/test_config.yaml
             config_name: qemu-rv64-max
+            exclude_extensions: Sm,ExceptionsZc,S,Zalrsc,Zacas,ZacasZabha,InterruptsSm,PMPSm,PMPmisaligned,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS
             run_cmd: qemu-system-riscv64 -nographic -semihosting -icount shift=1 -machine virt -cpu max,pmu-mask=0xfffffff8 -bios
-
           - name: rv32-max
             config_file: config/qemu/qemu-rv32-max/test_config.yaml
             config_name: qemu-rv32-max
+            exclude_extensions: Sm,ExceptionsZc,S,Zalrsc,Zacas,ZacasZabha,InterruptsSm,PMPSm,PMPmisaligned,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS
```
jordancarlin: Seems like the list of exclude_extensions is basically the same for all of the different QEMU cores. Can we stick with the single list? The only different is ExceptionsSm, which can be excluded in test_config.yaml (see my other comment about that).

config/qemu/qemu-RVI20U64/test_config.yaml
```diff
@@ -0,0 +1,7 @@
+name: qemu-RVI20U64
+compiler_exe: riscv64-unknown-elf-gcc # executable name on $PATH or full path to executable
+objdump_exe: riscv64-unknown-elf-objdump # executable name on $PATH or full path to executable, optional
+ref_model_exe: sail_riscv_sim # executable name on $PATH or full path to executable
+udb_config: qemu-RVI20U64.yaml
+linker_script: link.ld
+dut_include_dir: . # Directory containing DUT specific rvmodel_macros.h header
```
jordancarlin: Add `include_priv_tests: False` to this file to exclude all priv tests since RVI20 is unprivileged only.

jordancarlin (changes requested):

.github/workflows/regress.yml
UmerShahidengr: Done

config/qemu/qemu-RVI20U64/test_config.yaml
UmerShahidengr: Done. I did not know such a flag also exists. It has simplified the whole implementation

config/qemu/qemu-RVI20U32/link.ld
jordancarlin, line 1: I missed this in my previous review, but can you symlink the linker script to one of the existing QEMU linker scripts (similar to how rvmodel_macros.h is handled)? It should be the same for all of the QEMU configs and this one is already out of date.

UmerShahidengr: @jordancarlin done

config/qemu/qemu-RVI20U32/sail.json
jordancarlin: @UmerShahidengr I think this one still needs to be resolved.

UmerShahidengr: Done. Updated this field

UmerShahidengr:
@jordancarlin now it feels good to go

jordancarlin (approved):

PR #993, opened 2026-02-24
[ACT4] Update failure logging code

jordancarlin:
- Add the bin name to the debug string
- Update the `add_testcase` function to use a more logical order (bin, coverpoint, optional covergroup)
- Use a label to point to the PC of the failing instruction so the correct PC is consistently reported even with compressed instructions or extra instructions between the testcase and the sigupd
- Add floating point mismatch handling code
  - Prints out fp register value instead of int register
  - Support for reporting mismatches in fflags


davidharrishmc (approved):
Code looks good, but concerned about human error adding test_{test_data.test_count} labels.

PR #1002, opened 2026-02-26
InterruptsSm updated for Phase 1

SadhviNarayanan:
Added the python generator script for InterruptsSm and python helper files/functions, along with many of the header files with the appropriate macros for interrupt setup. Additionally, added separate macros for RVTEST and RVMODEL to abstract the generic DUT independent macro, as well as the DUT dependent macro (which gets configured in the DUT specific header). Macros are now defined in the headers to avoid differing PC lengths for the macros.

Currently at 78.26% coverage which is what we expect and the max coverage we can currently collect. SAIL does not have a way to trigger external interrupts so these bins cannot be hit. Additionally, the wfi timeout time on SAIL is configured to ~2 cycles, so wfi coverpoints can only be hit if timed correctly. There was an issue opened about this to SAIL.


SadhviNarayanan:
TODO: need to check for differing trace patterns in WALLY and SAIL for mstatus.MIE upon an interrupt.

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -155,22 +163,34 @@
 
 ##### Machine Interrupts #####
 
-#define RVMODEL_SET_MEXT_INT
+#define RVMODEL_SET_MEXT_INT(_R1, _R2)
```
davidharrishmc, line 166: QEMU, Spike, and cvw need to implement the RVMODEL macros, not just have placeholders.
The tests should run on each.

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -155,22 +163,34 @@
...
+#define RVMODEL_SET_SSW_INT(_R1, _R2)
+
+#define RVMODEL_CLR_SSW_INT(_R1, _R2)
+
 
-#define RVMODEL_CLR_SEXT_INT
+##### Hypervisor Interrupts #####
 
-#define RVMODEL_SET_SSW_INT
+#define RVMODEL_SET_VSW_INT(_R1, _R2)
```
davidharrishmc: All of the Hypervisor interrupt macros should be deleted because they are not DUT-specific.

coverpoints/priv/InterruptsSm_coverage.svh
```diff
@@ -10,99 +10,123 @@
...
-    cp_priority:         cross priv_mode_m, mstatus_mie_one, mie_meie_mtie_msie, mip_meip_mtip_msip;
-    cp_wfi:              cross priv_mode_m, wfi, mstatus_mie, mstatus_tw, mie_mtie_one, mie_mtip_one;
+   option.per_instance = 0;
+   `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+
+   // building blocks for the main coverpoints
+
+
+   mstatus_mie_one: coverpoint ins.prev.csr[12'h300][3] {
```
davidharrishmc: Add comments about why mstatus_mie_{one} uses .prev and that it may be able to revert to .current when the Sail tracer is fixed.

davidharrishmc: Delete the commented out m_<>_intr coverpoints.

coverpoints/priv/InterruptsSm_coverage.svh
```diff
@@ -10,99 +10,123 @@
...
+   cp_wfi:              cross priv_mode_m, wfi, mstatus_mie, mstatus_tw, mie_mtie_one, mip_mtip_one;
 endgroup
 
+
 function void interruptssm_sample(int hart, int issue, ins_t ins);
-    InterruptsSm_cg.sample(ins);
+   InterruptsSm_cg.sample(ins);
+
+
+   $display("=== InterruptsSm Debug ===");
```
davidharrishmc: Comment out or delete the display functions in the sample function

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -0,0 +1,187 @@
...
+    Args:
+        r_mtime: Register number to hold MTIME address
+        r_mtimecmp: Register number to hold MTIMECMP address
+        r_temp: Temp register number for values
+        r_temp2: Second temp register number for RV32
+    """
+    return [
+        f"la x{r_mtime}, RVMODEL_MTIME_ADDRESS",
+        f"la x{r_mtimecmp}, RVMODEL_MTIMECMP_ADDRESS",
+        f"LREG x{r_temp}, 0(x{r_mtime})",
```
davidharrishmc: the code at line 19-20 is only valid for RV64.  Put it in an #if, and flush out the part in the xlen32 to use the code sequence from page 57 of the priv manual, starting by loading two words from MTIME.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -0,0 +1,187 @@
...
+
+def set_mtimer_int_soon(
+    r_mtime: int, r_mtimecmp: int, r_temp1: int, r_temp2: int, r_temp3: int, r_temp4: int
+) -> list[str]:
+    return [
+        f"    la x{r_mtime}, RVMODEL_MTIME_ADDRESS",
+        f"    la x{r_mtimecmp}, RVMODEL_MTIMECMP_ADDRESS",
+        "#if __riscv_xlen == 64",
+        # Put comparator far away first (avoid stale pending)
+        f"    li x{r_temp1}, -1",
```
davidharrishmc: Why is this needed? What is sale pending and how would it cause trouble?

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -0,0 +1,187 @@
...
+        f"    sltu x{r_temp4}, x{r_temp3}, x{r_temp1}",
+        f"    add x{r_temp2}, x{r_temp2}, x{r_temp4}",
+        f"    sw x{r_temp3}, 0(x{r_mtimecmp})",
+        f"    sw x{r_temp2}, 4(x{r_mtimecmp})",
+        "#endif",
+    ]
+
+
+def set_stimer_int(r_temp: int) -> list[str]:
+    """Set supervisor timer interrupt w Sstc.
```
davidharrishmc: w -> without?

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -0,0 +1,187 @@
...
+        r_temp: Temporary register number
+    """
+    return [
+        f"li x{r_temp}, 0x20",  # mip.STIP = bit 5
+        f"csrrc x{r_temp}, mip, x{r_temp}",  # Clear mip.STIP from M-mode
+    ]
+
+
+def set_stimer_int_sstc(r_mtime: int, r_temp: int, r_temp2: int) -> list[str]:
+    """Set supervisor timer interrupt with Sstc extension.
```
davidharrishmc: Why are there flavors of set/clr_stimer_int with and without sstc?  Why can’t we have one that checks menvcfg.STCE and uses the Sstc code sequence if STCE=1?

tests/env/arch_test.h
```diff
@@ -1730,28 +1710,33 @@ excpt_\__MODE__\()hndlr_tbl:            // handler code should only touch T2..T6
...
+        SREG T5, 0(T2)
+        #if __riscv_xlen == 32
+                sw T5, 4(T2)
+        #endif
         j       resto_\__MODE__\()rtn
 
 \__MODE__\()clr_Mext_int:               // int11 default to just return after saving IntID in T3
-        RVMODEL_CLR_MEXT_INT
+        RVMODEL_CLR_MEXT_INT(T2, T5)
         TRAP_SIGUPD(T4, T3, 3)          // save 4th sig value, (intID)
```
davidharrishmc, line 1733: Why does clr_Mext_int have TRAP_SIGUPD but clr_Msw_int and clr_Mtmr_int don’t?
Or is it just clipped in my view?

tests/env/sail_test.h
```diff
@@ -3,95 +3,194 @@
 # Jordan Carlin jcarlin@hmc.edu October 2025
 # SPDX-License-Identifier: BSD-3-Clause
 
+
 #ifndef _COMPLIANCE_MODEL_H
 #define _COMPLIANCE_MODEL_H
 
+
+#define CLINT_BASE_ADDRESS 0x02000000
+#define THRESHOLD_0     (PLIC_BASE_ADDRESS + 0x200000)
```
davidharrishmc: Has PLIC_BASE_ADDRESS been defined?  Is it needed?  Same with GPIO?

tests/env/sail_test.h
```diff
@@ -3,95 +3,194 @@
...
+
 # Perform boot operations. Can be empty.
 #define RVMODEL_BOOT
 
+
 ##### TERMINATION #####
 
+
 # Terminate test with a pass indication.
 # When the test is run in simulation, this should end the simulation.
```
davidharrishmc, line 28: Add a comment that SAIL uses HTIF to terminate.

tests/env/sail_test.h
```diff
@@ -3,95 +3,194 @@
...
+ sw _R1, 0(_R2)            ; /* Set MEXT interrupt */ \
+
 
-#define RVMODEL_CLR_MEXT_INT
+#define RVMODEL_CLR_MEXT_INT(_R1, _R2)        \
+ li _R2, RVMODEL_MEXT_ADDRESS; \
+ sw zero, 0(_R2)            ; /* Clear MEXT interrupt */ \
+
+
+#define SAIL_MSIP_ADDRESS 0x02000000
```
davidharrishmc: Should this be #define CLINT_MSIP_ADDRESS (CLINT_BASE_ADDRESS + 0)

tests/env/sail_test.h
```diff
@@ -3,95 +3,194 @@
...
+ sw _R1, 0(_R2)            ; /* Set SEXT interrupt */ \
+
+
+#define RVMODEL_CLR_SEXT_INT(_R1, _R2)        \
+ li _R2, SAIL_SEXT_ADDRESS; \
+ sw zero, 0(_R2)            ; /* Clear SEXT interrupt */
+
+
+// TODO: check to see if SAIL support this, and we may want to implement this in WALLY
+#define SAIL_SSIP_ADDRESS 0x0200C000
```
davidharrishmc: CLINT_BASE_ADDRESS + C000?

tests/env/sail_test.h
```diff
@@ -3,95 +3,194 @@
...
+ sw _R1, 0(_R2);
+
+
+#define RVMODEL_CLR_SSW_INT(_R1, _R2)        \
+ li _R2, SAIL_SSIP_ADDRESS;              \
+ sw zero, 0(_R2);
+
+
+
+ ##### Hypervisor Interrupts #####
```
davidharrishmc: Delete all of these.  No longer used.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+def _generate_trigger_mti_tests(test_data: TestData) -> list[str]:
+    """Generate timer interrupt trigger tests."""
+    ######################################
+    covergroup = "InterruptsSm_cg"
+    coverpoint = "cp_trigger_mti"
+    ######################################
+
+    r1, r_mtime, r_mtimecmp, r_temp, r_temp2 = test_data.int_regs.get_registers(5, exclude_regs=[0, 1, 2, 4, 5, 7, 30])
```
davidharrishmc: It would help to comment why those specific registers need to be excluded.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+            "cp_trigger_mti",
+            "With mstatus.MIE = {0/1}, and mie = all 1s, use MTIMECMP to cause mip.MTIP",
+        ),
+        "",
+    ]
+
+    # Setup: Enable all interrupts in mie
+    lines.extend(
+        [
+            test_data.add_testcase("mie_setup", coverpoint, covergroup),
```
davidharrishmc: No need for an add_testcase here because nothing is being tested here and you have a test case just below.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+    # Test 1: mstatus.MIE = 0 should NOT take interrupt
+    lines.extend(
+        [
+            test_data.add_testcase("mie_0", coverpoint, covergroup),
+            f"   test_{test_data.test_count}:",
+            "    CSRRCI zero, mstatus, 8    # mstatus.MIE = 0",
+        ]
+    )
+    lines.extend(["    " + line for line in set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2)])
```
davidharrishmc: Is there a cleaner way to write this?  can it be something like lines += set_mtimer_int(...

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+    )
+
+    lines = [
+        comment_banner(
+            "cp_interrupts",
+            "Cross of mstatus.MIE = {0/1}, mtvec.MODE = 00, 3 walking 1s in mip.MTIP/MSIP/MEIP,\n"
+            "3 walking 1s in mie.MTIE/MSIE/MEIE (2 x 3 x 3 bins)",
+        ),
+        "",
+        f"CSRRCI x{r_csr_tmp}, mtvec, 3     # mtvec.MODE = 00",
```
davidharrishmc: Why are you reading mtvec.MODE?  The result is never used.  csrci would be cleaner.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+        # 3 interrupt enables: s1 = 2 (MEIE), 1 (MTIE), 0 (MSIE)
+        for s1 in [2, 1, 0]:
+            mie_bit = 3 + s1 * 4
+            mie_val = 1 << mie_bit
+            enable_name = ["msie", "mtie", "meie"][s1]
+
+            lines.extend(
+                [
+                    f"CSRRW x{r_csr_tmp}, mie, zero     # Clear all enables first!",  # ← ADD THIS!
```
davidharrishmc: What is the ADD THIS for?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+            # 3 interrupt pending: s2 = 2 (MEIP), 1 (MTIP), 0 (MSIP)
+            for s2 in [2, 1, 0]:
+                int_name = ["msip", "mtip", "meip"][s2]
+                binname = f"mie_{mstatus_mie}_{int_name}_{enable_name}"
+
+                lines.extend(
+                    [
+                        test_data.add_testcase(binname, coverpoint, covergroup),
+                        f"    CSRR x{r_mie_save}, mie",
```
davidharrishmc: Why are you reading mie?  You just set it on 231.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+                        "     RVTEST_IDLE_FOR_INTERRUPT",
```
davidharrishmc: What is this IDLE for?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+        for s1 in [2, 1, 0]:
+            mie_bit = 3 + s1 * 4
+            mie_val = 1 << mie_bit
+            enable_name = ["msie", "mtie", "meie"][s1]
+
+            lines.extend(
+                [
+                    f"CSRRW x{r_csr_tmp}, mie, zero     # Clear all enables first!",  # ← ADD THIS!
+                    f"LI x{r_mie_val}, {mie_val}",
+                    f"CSRRS x{r_csr_tmp}, mie, x{r_mie_val}  # Set only this enable",
```
davidharrishmc: You could replace 229 and 231 with CSRW(mie, x{r_mie_val}).  That writes, reads back, and checks the signature.  No need to clear enables and then set.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+                    lines.append("    RVTEST_SET_MEXT_INT")
+                elif s2 == 1:  # MTIP
+                    lines.extend(["    " + line for line in set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2)])
+                else:  # MSIP
+                    lines.append("    RVTEST_SET_MSW_INT")
+
+                # More settling
+                lines.extend(
+                    [
+                        f"    CSRW mie, x{r_mie_save}",
```
davidharrishmc: I don't see why ie would have changed or needs to be restored.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+    r1, r_mtime, r_mtimecmp, r_temp, r_temp2, r_mie_all, r_mie_save, r_csr_tmp = test_data.int_regs.get_registers(
+        8, exclude_regs=[0, 1, 2, 4, 5, 7, 30]
+    )
+
+    lines = [
+        comment_banner(
+            "cp_vectored",
+            "Cross of mtvec.MODE = 01, mstatus.MIE=1, all 3 of mie.MTIE/MSIE/MEIE,\n"
+            "3 walking 1s in mip.MTIP/MSIP/MEIP (3 bins)",
```
davidharrishmc: Looks like mode is 00/01, and there are 2 * 3 = 6 bins.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+    # Test BOTH modes
+    for mode in [0, 1]:  # 0=direct, 1=vectored
+        mode_name = ["direct", "vectored"][mode]
+
+        # Raise each interrupt type: s2 = 2 (MEIP), 1 (MTIP), 0 (MSIP)
+        for s2 in [2, 1, 0]:
+            lines.extend(
+                [
+                    f"# Set mtvec.MODE = {mode:02b} ({mode_name})",
```
davidharrishmc: Why is mtvec mode being set here rather than around line 301?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+                    f"CSRRCI x{r_csr_tmp}, mtvec, 3",
+                    f"CSRRSI x{r_csr_tmp}, mtvec, {mode}",
+                    f"CSRRSI x{r_csr_tmp}, mstatus, 8",
```
davidharrishmc: Why is MIE being turned on here rather than before the loops?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+                    f"CSRRW x{r_csr_tmp}, mie, zero",
```
davidharrishmc: Why is mie being zeroed, then set with three ones rather than just written with 3 ones?

davidharrishmc: This could also be done outside the loop.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+                    f"CSRRS x{r_csr_tmp}, mie, x{r_mie_all}",
+                    "",
+                ]
+            )
+            int_name = ["msip", "mtip", "meip"][s2]
+
+            lines.extend(
+                [
+                    test_data.add_testcase(f"{mode_name}_{int_name}", coverpoint, covergroup),
+                    f"    CSRR x{r_mie_save}, mie",
```
davidharrishmc, line 297: Seems unnecessary.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+                lines.append("    RVTEST_SET_MEXT_INT")
+            elif s2 == 1:  # MTIP
+                lines.extend(["    " + line for line in set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2)])
+            else:  # MSIP
+                lines.append("    RVTEST_SET_MSW_INT")
+
+            # Settling time
+            lines.extend(
+                [
+                    f"    CSRW mie, x{r_mie_save}",
```
davidharrishmc, line 313: unnecessary?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+    # Unroll s1 loop: 7 down to 0 (mie values)
+    for s1 in range(7, -1, -1):
+        # Build mie value from s1 in python
+        mie_value = ((s1 & 4) << 9) | ((s1 & 2) << 6) | ((s1 & 1) << 3)
+
+        lines.extend(
+            [
+                f"CSRRW x{r_csr_tmp}, mie, zero",
+                f"LI x{r_mie_mask}, {mie_value}",  # Single instruction!
```
davidharrishmc: What is the Single instruction! comment for?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+        8, exclude_regs=[0, 1, 2, 4, 5, 7, 30]
+    )
+
+    lines = [
+        comment_banner(
+            "cp_wfi",
+            "Cross Product of mstatus.MIE = {0/1}, mstatus.TW = {0/1}, mie.MTIE = 1\nWFI instruction",
+        ),
+        "",
+        "# Minimal outer cleanup",
```
davidharrishmc: why is this needed?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+        "RVTEST_CLR_MSW_INT",
+        "RVTEST_CLR_MEXT_INT",
+    ]
+    lines.extend(clr_mtimer_int(r_t0, r_mtimecmp))
```
davidharrishmc: why necessary?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+            binname = f"mie_{mie_val}_tw_{tw_val}"
+
+            lines.extend(
+                [
+                    test_data.add_testcase(binname, coverpoint, covergroup),
+                    "",
+                    "    CSRW mie, zero",
+                ]
+            )
+            lines.extend(["    " + line for line in clr_mtimer_int(r_t0, r_mtimecmp)])
```
davidharrishmc, line 424: why needed?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+            lines.append(f"test_{test_data.test_count}:")
+
+            # Set TW if needed
+            if tw_val:
+                lines.extend(
+                    [
+                        f"    LI x{r_scratch}, 0x200000",
+                        f"    CSRRS x{r_csr_tmp}, mstatus, x{r_scratch}",
```
davidharrishmc: Why are you reading mstatus as well?  If you want to, use a separate CSRR macro to read and log to the signature.

davidharrishmc: Where is TW cleared when it is supposed to be 0?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+            # Enable MTIE
+            lines.extend(
+                [
+                    f"    LI x{r_scratch}, 0x80",
+                    f"    CSRW mie, x{r_scratch}",
+                ]
+            )
+
+            # Set MIE if needed
```
davidharrishmc: MIE should be set or cleared in the outer loop.  This is in the inner loop and only sets.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+@add_priv_test_generator("InterruptsSm", required_extensions=["Sm", "I", "Zicsr"])
+def make_interruptssm(test_data: TestData) -> list[str]:
+    """Generate tests for InterruptsSm machine-mode interrupts."""
+
+    # consumer t2 and t5 for interrupt subroutines, but mark as consumed for whole test since they're used throughout
+    test_data.int_regs.consume_registers([7, 30])
+
+    lines: list[str] = []  # TODO: do we want to clear timer interrupts here?
```
davidharrishmc: drop TODO.  They should be clear in the boot process.

davidharrishmc (changes requested):
Looking good overall.  Lots of detailed feedback, a little of which seems to be preventing the tests from having their intended behavior.

tests/env/arch_test.h
SadhviNarayanan: I didn't write this part of the trap handler, but my guess is that external interrupts come from a PLIC and this saved the ID of the external interrupt to the signature region? Because the other interrupt types are single source the signature update may not be needed.

davidharrishmc: That makes sense.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
SadhviNarayanan: I had this for the cases when the timer interrupt would not be taken, and we would need to clear it ourselves.

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -143,6 +143,14 @@
   j 1b                       ;/* Loop */             \
 3:
 
+#ifndef RVMODEL_INTERRUPT_LATENCY
+#define RVMODEL_INTERRUPT_LATENCY 10
+#endif
```
jordancarlin: No need to have `ifndef` in `rvmodel_macros.h` files.

tests/env/check_defines.h
jordancarlin, line 1: Why is all of the indentation changed from 2 spaces to 1 space in this file? Please revert. 1 space is not enough indentation.

tests/env/sail_test.h
jordancarlin, line 1: Why is all of the indentation changed from 2 spaces to 1 space in this file? Please revert. 1 space is not enough indentation.

tests/env/utils.h
jordancarlin, line 1: Why is all of the indentation changed from 2 spaces to 1 space in this file? Please revert. 1 space is not enough indentation.

jordancarlin (changes requested):
Lots of whitespace changes all over the place that are making this hard to review. Please revert the whitespace changes to use standard 2 or 4 spaces of indentation (match what the rest of the file does).

tests/env/utils.h
```diff
@@ -336,3 +336,81 @@
 #define SET_MSB(_R) \
     LI(_R, 0x80000000)
 #endif
+
+// Interrupt Macros
+// Idle for interrupt latency
+#ifndef RVTEST_IDLE_FOR_INTERRUPT
```
davidharrishmc: What is the ifndef for?  Why is it needed here and not at other #defines?

davidharrishmc (changes requested):
Minor query in a quick review.  When Jordan approves and it passes CI, let's merge.

config/cores/cvw/cvw-rv64gc/rvmodel_macros.h
```diff
@@ -88,22 +88,37 @@
 
 ##### Machine Interrupts #####
 
-#define RVMODEL_SET_MEXT_INT
+#define CVW_CLINT_BASE 0x02000000
+#define CVW_MSIP_ADDRESS (CVW_CLINT_BASE + 0x0)
```
jordancarlin: How about just naming these `MSIP_ADDRESS` (and similar for the other #defines) so that it is easier for someone to copy an `rvmodel_macros.h` file and adapt it for their core.

config/cores/cvw/cvw-rv64gc/rvmodel_macros.h
```diff
@@ -88,22 +88,37 @@
 
 ##### Machine Interrupts #####
 
-#define RVMODEL_SET_MEXT_INT
+#define CVW_CLINT_BASE 0x02000000
+#define CVW_MSIP_ADDRESS (CVW_CLINT_BASE + 0x0)
 
-#define RVMODEL_CLR_MEXT_INT
+#define RVMODEL_SET_MEXT_INT(_R1, _R2)
```
jordancarlin, line 98: If you are changing the macros (by adding register arguments) you need to update the description in the README and in the CTP.

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -143,6 +143,10 @@
   j 1b                       ;/* Loop */             \
 3:
 
+#define RVMODEL_INTERRUPT_LATENCY 10
+
+#define RVMODEL_TIMER_INT_SOON_DELAY 100
+
```
jordancarlin, line 146: These defines are missing from the CVW macro file.

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -143,6 +143,10 @@
   j 1b                       ;/* Loop */             \
 3:
 
+#define RVMODEL_INTERRUPT_LATENCY 10
+
+#define RVMODEL_TIMER_INT_SOON_DELAY 100
```
jordancarlin: The README and CTP need to be updated to include these new macro and clarify what they are supposed to do.

jordancarlin: These don't belong under the IO header. Move them down to the interrupts.

tests/env/utils.h
```diff
@@ -336,3 +336,80 @@
 #define SET_MSB(_R) \
     LI(_R, 0x80000000)
 #endif
+
+// Interrupt Macros
```
jordancarlin, line 340: There are enough interrupt macros (and they are probably going to grow as the hypervisor macros are defined) that it might make sense to break these out into an `interrupts.h` file.

tests/env/utils.h
```diff
@@ -336,3 +336,80 @@
...
+#define RVTEST_CLR_VSW_INT
+#define RVTEST_SET_VEXT_INT
+#define RVTEST_CLR_VEXT_INT
+
+// Timer interrupts (no parameters)
+#define RVTEST_CLR_STIMER_INT
+#define RVTEST_CLR_VTIMER_INT
+
+// RVMODEL macros for SAIL. These implement the actual interrupt setup for the SAIL platform and are invoked by the generic RVTEST macros above.
+#define RVTEST_INTERRUPTS \
```
jordancarlin: This macro has nothing to do with Sail. Update the comment.

tests/env/sail_test.h
```diff
@@ -64,34 +72,89 @@
...
+
+// Interrupt latency configuration
+#define RVMODEL_INTERRUPT_LATENCY 10
+
+#define RVMODEL_TIMER_INT_SOON_DELAY 100
 
 #define RVMODEL_ACCESS_FAULT_ADDRESS 0x00000000
 
+
+// Add to test_setup.h after other macros:
```
jordancarlin: What does this comment mean?

tests/env/sail_test.h
```diff
@@ -64,34 +72,89 @@
...
+#define RVMODEL_MTIMECMP_ADDRESS  0x02004000  /* Address of mtimecmp CSR */
 #define RVMODEL_MTIME_ADDRESS  0x0200BFF8  /* Address of mtime CSR */
+#define RVMODEL_SET_MTIME(_R1, _R2)        \
+  li   _R2, RVMODEL_MTIME_ADDRESS        ; /* MTIME address */ \
+  SREG _R1, 0(_R2)            ; /* Set MTIME low */
+
+
+#define RVMODEL_SET_MTIMEH(_R1, _R2)       \
+  li   _R2, RVMODEL_MTIME_ADDRESS        ; /* MTIME address */ \
+  SREG _R1, 4(_R2)            ; /* Set MTIME high */
```
jordancarlin: This seems to have gotten garbled. The comment reads like it applies to `RVMODEL_MTIMECMP_ADDRESS`, which it doesn't. Also, I thought we dropped the `RVMODEL_SET_MTIME` and `RVMODEL_SET_MTIMEH` macros?

tests/env/arch_test.h
```diff
@@ -1766,18 +1751,18 @@ excpt_\__MODE__\()hndlr_tbl:            // handler code should only touch T2..T6
 \__MODE__\()clr_Stmr_int:               // int 5 default to just return
                                         // S-mode timer interrupts need to be reset differently when raised in M or S mode
         .ifc \__MODE__ , M              // Select the interrupt handler function based on current privilege mode
-            RVMODEL_MCLR_STIMER_INT
+            RVMODEL_CLR_STIMER_INT
```
jordancarlin: Does `RVMODEL_CLR_STIMER_INT` exist anymore? Shouldn't it be `RVTEST`?

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+    Args:
+        r_mtime: Register number to hold MTIME address
+        r_mtimecmp: Register number to hold MTIMECMP address
+        r_temp: Temp register number for values
+        r_temp2: Second temp register number for RV32
+    """
+    return [
+        f"    la x{r_mtime}, RVMODEL_MTIME_ADDRESS",
```
jordancarlin: This needs to use the `LA` macro, not the `la` pseudoinstruction. Global comment.

```suggestion
        f"    LA(x{r_mtime}, RVMODEL_MTIME_ADDRESS)",
```

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -0,0 +1,202 @@
...
+) -> list[str]:
+    return [
+        f"    la x{r_mtime}, RVMODEL_MTIME_ADDRESS",
+        f"    la x{r_mtimecmp}, RVMODEL_MTIMECMP_ADDRESS",
+        "#if __riscv_xlen == 64",
+        # Set mtimecmp to max value first to prevent spurious interrupt.
+        # If mtimecmp < mtime from a previous test,
+        # mip.MTIP would be set immediately when we enable MTIE, causing
+        # an unexpected trap before we're ready.
+        f"    li x{r_temp1}, -1",
```
jordancarlin: Needs to use the `LI` macro, not the li pseudoinstruction. Global change.


```suggestion
        f"    LI(x{r_temp1}, -1)",
```

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -0,0 +1,504 @@
...
+            test_data.add_testcase("mie_0", coverpoint, covergroup),
+            "    CSRRCI zero, mstatus, 8    # mstatus.MIE = 0",
+        ]
+    )
+    lines.extend(indent(set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2)))
+    lines.extend(
+        [
+            "    RVTEST_IDLE_FOR_INTERRUPT",
+        ]
+    )
```
jordancarlin: When only adding one line, use `lines.append`. Otherwise, combine all of the nearby lines into a single `lines.extend`.

jordancarlin (changes requested):

tests/env/utils.h
SadhviNarayanan: Will address this when working on the next round of interrupts!

davidharrishmc (approved):
This is running and has a ton of good stuff.  I'm ready to merge.  @jordancarlin if you have style or clarification points, can you add those after the merge?

jordancarlin (approved):

PR #1007, opened 2026-02-27
ExceptionsSm 100% coverage

juliacygong:


config/sail/sail-rv32-max/rvtest_config.h
```diff
@@ -28,3 +28,5 @@
 #define LMULf8_SUPPORTED
 #define LMULf4_SUPPORTED
 #define LMULf2_SUPPORTED
+
+#define SKIP_MEPC
```
jordancarlin: This can't go in `rvtest_config.h`. It will disable the check for all tests that way. Add it to the `extra_defines` argument of the ExceptionsSm generator instead:

```py
@add_priv_test_generator("ExceptionsSm", required_extensions=[...], march_extensions=[...], extra_defines=["#define SKIP_MEPC"])
```

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+
+    lines.extend(
+        [
+            "# Initialize scratch memory with test data",
+            " LA(x10, scratch)",
+            " LI(x11, 0xDEADBEEF)",
+            " sw x11, 0(x10)",
+            " sw x11, 4(x10)",
+            " sw x11, 8(x10)",
+            " sw x11, 12(x10)",
```
jordancarlin, line 549: I wonder if this is something worth moving  into `test_setup.h` so that scratch memory is always initialized with values. Quite a few tests use scratch memory (including unpriv).

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+    lines = [comment_banner(coverpoint, "Instruction Address Misaligned branch")]
+    lines.append(test_data.add_testcase("taken_branch_pc_6", coverpoint, covergroup))
+
+    branches = ["beq", "bne", "blt", "bge", "bltu", "bgeu"]
+    rs1 = ["x0", "x0", "x0", f"x{temp_reg}", "x0", "x0"]
+    rs2 = ["x0", f"x{temp_reg}", f"x{temp_reg}", "x0", f"x{temp_reg}", "x0"]
+
+    lines.append(f" li x{temp_reg}, 1")
+    lines.append(" .align 2")
+    lines.append(f"test_{test_data.test_count}:")
```
jordancarlin: Instead of repeated calls to `lines.append`, use a single call to `lines.extend`:


```suggestion
    lines.extend(
        [
            f" li x{temp_reg}, 1",
            " .align 2",
            f"test_{test_data.test_count}:",
        ]
    )
```

This applies to lots of places.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+
+    for i in range(6):
+        lines.append(f" {branches[i]} {rs1[i]}, {rs2[i]}, .+6")
+        lines.append(" .word 0x00010013")
```
jordancarlin: Why does this need to be a `.word` instead of a regular `addi` instruction?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+def _generate_instr_adr_misaligned_branch_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_instr_adr_misaligned_branch"
+    temp_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [comment_banner(coverpoint, "Instruction Address Misaligned branch")]
+    lines.append(test_data.add_testcase("taken_branch_pc_6", coverpoint, covergroup))
+
+    branches = ["beq", "bne", "blt", "bge", "bltu", "bgeu"]
+    rs1 = ["x0", "x0", "x0", f"x{temp_reg}", "x0", "x0"]
+    rs2 = ["x0", f"x{temp_reg}", f"x{temp_reg}", "x0", f"x{temp_reg}", "x0"]
```
jordancarlin: It might be easier to follow what is going on here if you just have the list of branches and then inside the loop have an if that determines the registers. Something like this:

```py
if branch == "bge":
    rs1 = f"x{temp_reg}"
else:
    rs1 = "x0"
```

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+def _generate_instr_adr_misaligned_branch_nottaken(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_instr_adr_misaligned_branch_nottaken"
+    check_reg, temp_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    b_op = ["beq", "bne", "blt", "bge", "bltu", "bgeu"]
+    r1 = ["x0", "x0", f"x{temp_reg}", "x0", f"x{temp_reg}", "x0"]
+    r2 = [f"x{temp_reg}", "x0", "x0", f"x{temp_reg}", "x0", f"x{temp_reg}"]
+
+    lines = [
+        comment_banner(coverpoint, "Branch to an unaligned address (PC+6). Should not cause an exception"),
```
jordancarlin: Clarify in the comment that the branch is NOT taken (which is why it should not cause an exception).

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+        " .align 2",
+        f" li x{temp_reg}, 1",
+        test_data.add_testcase("nottaken_branch_pc_6", coverpoint, covergroup),
+        f"test_{test_data.test_count}:",
+        f" li x{check_reg}, 1",
+    ]
+
+    for i in range(6):
+        lines.append(f" {b_op[i]} {r1[i]}, {r2[i]}, .+6")
+    lines.append(write_sigupd(check_reg, test_data))
```
jordancarlin: What is the point of the sigupd is `check_reg` is always going to be 1?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+def _generate_instr_adr_misaligned_jalr_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_instr_adr_misaligned_jalr"
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Address Misaligned JALR"),
+    ]
+
+    rs1_base_offsets = {0: 8, 1: 5, 2: 6, 3: 7}
+    jalr_offsets = {0: 8, 1: 5, 2: 6, 3: 7}
```
jordancarlin, line 108: This probably warrants a comment explaining what is going on.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+
+
+def _generate_instr_access_fault_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_instr_access_fault"
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Access Fault"),
+        test_data.add_testcase("instr_access_fault", coverpoint, covergroup),
+        f" la x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS",
```
jordancarlin: Make sure to use the `LA` macro instead of the `la` pseudoinstruction to ensure the test is a constant length:


```suggestion
        f" LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
```

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+        " li x4, 0xACCE",  # trap handler to use ra as return address
```
jordancarlin: Make sure to use the LI macro instead of the li pseudoinstruction to ensure the test is a constant length:

```suggestion
        "LI(x4, 0xACCE)",  # trap handler to use ra as return address
```

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_breakpoint"
+
+    lines = [
+        comment_banner(coverpoint, "Breakpoint"),
+        test_data.add_testcase("ebreak", coverpoint, covergroup),
+        f"test_{test_data.test_count}:",
+        " ebreak",
+        " nop",
+    ]
+    return lines
```
jordancarlin, line 209: Isn't this already covered in Sm? Do we have a duplicate coverpoint?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,553 @@
...
+
+    lines = [
+        comment_banner(coverpoint, "Ecall Machine Mode"),
+        test_data.add_testcase("ecall_m", coverpoint, covergroup),
+        f"test_{test_data.test_count}:",
+        " ecall",
+        " nop",
+    ]
+
+    return lines
```
jordancarlin, line 387: Isn't this already covered in Sm? Do we have a duplicate coverpoint?

jordancarlin (changes requested):
First round of comments. Generally looking great for a first version!

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
juliacygong: The original ExceptionsSm.S was using .word so I was mainly following what the previous tests were doing. Would you prefer to use addi?

davidharrishmc:
I like addi to emphasize the machine language code.  There ought to be a comment that a branch by 6 when compressed is supported goes into the upper half of the next instruction (which is 0x0001), which is a c.nop.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
juliacygong: Yes it's also covered in Sm

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
juliacygong: Also already covered but similar to ebreak it was in the testplan for ExceptionsSm

davidharrishmc: Yes, covered in Sm and that is ok.  Sm tests instructions are executed and ExceptionsSm tests exceptions occur.  Lets leave this as is.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+    ]
+
+    test_data.int_regs.return_registers([save_reg, mask_reg, arg_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "ExceptionsSm",
+    required_extensions=["I", "Zicsr", "Sm"],
+    march_extensions=["F"],
```
jordancarlin: Why does this require F in the march extensions? ExceptionsSm doesn't use F.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+
+
+def _generate_instr_adr_misaligned_branch_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_instr_adr_misaligned_branch"
+    temp_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [comment_banner(coverpoint, "Instruction Address Misaligned branch")]
+
+    lines.extend(
+        [f" li x{temp_reg}, 1", " .align 2", test_data.add_testcase("taken_branch_pc_6", coverpoint, covergroup)]
```
jordancarlin: This is easier to read if each assembly line is on its own line. Ruff will reformat it like this for you automatically if you add a comma after the last item.
```suggestion
        [
            f" li x{temp_reg}, 1",
            " .align 2",
            test_data.add_testcase("taken_branch_pc_6", coverpoint, covergroup),
        ]
```

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_instr_adr_misaligned_branch_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_instr_adr_misaligned_branch"
+    temp_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [comment_banner(coverpoint, "Instruction Address Misaligned branch")]
+
+    lines.extend(
```
jordancarlin: Merge these two together. No need to create lines and then immediately extend it.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+                " addi x0, x2, 0",
+            ]
+        )
+
+    lines.extend(
+        [
+            f" CSRR(x{check_reg}, mcause)",
+            write_sigupd(check_reg, test_data),
+        ]
+    )
```
jordancarlin: Not sure that this sigupd is needed. If it traps, that will be captured by the trap handler sigupds already.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+        if branch in ("beq", "bge", "bgeu"):
+            rs1 = "x0"
+            rs2 = f"x{temp_reg}"
+        elif branch == "bne":
+            rs1 = "x0"
+            rs2 = "x0"  # 0 == 0, so bne not taken
+        else:  # blt, bltu
+            rs1 = f"x{temp_reg}"
+            rs2 = "x0"
+        lines.append(f" {branch} {rs1}, {rs2}, .+6")
```
jordancarlin: How are we checking that the branch was not taken? Seems like a taken branch would be missed in the signature.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+                    f" auipc x{addr_reg}, 0",
+                    f" addi x{addr_reg}, x{addr_reg}, {base_off}",
+                    test_data.add_testcase(f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup),
+                    f" jalr x1, {jalr_off}(x{addr_reg})",
+                    " nop",
+                ]
+            )
+
+            if rs1_lsb & 1:
+                lines.append(" nop")
```
jordancarlin: Does this nop need to be conditional? What if we always did 2 addi instruction?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+    return lines
+
+
+def _generate_instr_access_fault_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_instr_access_fault"
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Access Fault"),
+        f" LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
```
jordancarlin: Why is there a single space at the beginning of the generated assembly line?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+        "LI(x4, 0xACCE)",  # trap handler to use ra as return address
```
jordancarlin: Not clear what this comment means.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+    lines = [comment_banner(coverpoint, "Illegal Instruction Seed")]
+
+    for instr, bin_name in seed_ops:
+        lines.extend(
+            [
+                test_data.add_testcase(bin_name, coverpoint, covergroup),
+                f" {instr}",
+                " nop",
+            ]
+        )
```
jordancarlin: Not sure that the loop actually makes this shorter or easier to follow. Might be better off just putting each line of assembly into the list directly.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+
+def _generate_load_access_fault_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_load_access_fault"
+    addr_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [comment_banner(coverpoint, "Load Access Fault")]
+
+    load_ops = ["lb", "lbu", "lh", "lhu", "lw"]
+    if test_data.xlen == 64:
+        load_ops.extend(["lwu", "ld"])
```
jordancarlin: Priv tests are only generated once. They are not different per XLEN. These instructions need to always end up in the generated assembly and should have an `#ifdef` around them to check for xlen == 64.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+        f" LA(x{addr_reg}, scratch)",
+        f" addi x{addr_reg}, x{addr_reg}, {offset}",
+        test_data.add_testcase(f"{op}_off{offset}", coverpoint, covergroup),
+        f" {op} x{data_reg}, 0(x{addr_reg})",
+        " nop",
+    ]
+
+    # Read back scratch memory to verify store result
+    t_lines.extend(
+        [
```
jordancarlin: Combine this.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+
+def _generate_store_access_fault_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_store_access_fault"
+    addr_reg, data_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [comment_banner(coverpoint, "Store Access Fault")]
+
+    store_ops = ["sb", "sh", "sw"]
+    if test_data.xlen == 64:
+        store_ops.append("sd")
```
jordancarlin: Priv tests are only generated once. They are not different per XLEN. These instructions need to always end up in the generated assembly and should have an `#ifdef` around them to check for xlen == 64.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+
+    test_values = {"sb": "0xAB", "sh": "0xBEAD", "sw": "0xADDEDCAB", "sd": "0xDEADBEEFDEADBEEF"}
+
+    for op in store_ops:
+        lines.extend(
+            [
+                f" LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+                f" li x{data_reg}, {test_values[op]}",
```
jordancarlin: Make sure to use the LI macro.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_store_access_fault"
+    addr_reg, data_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [comment_banner(coverpoint, "Store Access Fault")]
+
+    store_ops = ["sb", "sh", "sw"]
+    if test_data.xlen == 64:
+        store_ops.append("sd")
+
+    test_values = {"sb": "0xAB", "sh": "0xBEAD", "sw": "0xADDEDCAB", "sd": "0xDEADBEEFDEADBEEF"}
```
jordancarlin: Let's not use deadbeef for the test value. We are initialized a lot of memory to deadbeef already, so it is hard to tell if the test actually does what it is supposed to.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+
+    lines = [
+        comment_banner(coverpoint, "Misaligned Priority Fetch"),
+    ]
+
+    target_label = f"misaligned_fetch_target_{test_data.test_count + 1}"
+    lines.extend(
+        [
+            "\n# misaligned existent",
+            f" la x{addr_reg}, {target_label}",
```
jordancarlin: Use the LA macro. There are a few instances of both the la and li pseudoinstructions being used instead of the macro. Please do a careful review for this.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+            " nop",
+            " .align 4",
+            f"{target_label}:",
+            " nop",
+        ]
+    )
+
+    lines.extend(
+        [
+            "\n# misaligned nonexistent",
```
jordancarlin: Combine. No need for 2 extends.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+
+
+def _generate_mstatus_ie_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsSm_cg", "cp_mstatus_ie"
+    save_reg, mask_reg, arg_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Mstatus Interrupt Enable"),
+        f" csrr x{save_reg}, mstatus",
+        f" li x{mask_reg}, 0x8",
```
jordancarlin: LI macro

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+        "",
+        "# Test ecall with mstatus.MIE = 0",
+        f" csrrc x0, mstatus, x{mask_reg}",
```
jordancarlin: I think we decided all csr accesses should use the `CSRRC`/`CSRRS`/`CSRRW`/etc. macros instead of using the instructions directly. @davidharrishmc, is that still what we want to do?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -0,0 +1,569 @@
...
+        f" li x{arg_reg}, 3",
+        test_data.add_testcase("ecall_mie_0", coverpoint, covergroup),
+        " ecall",
+        " nop",
+        "",
+        "# Test ecall with mstatus.MIE = 1",
+        f" csrrs x0, mstatus, x{mask_reg}",
+        f" li x{arg_reg}, 3",
```
jordancarlin: LI macro

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
davidharrishmc: @jordancarlin feel free to make that change.  Let's not gate this PR on it.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
davidharrishmc: The reason for that was to provide a nop in case the instruction might fault.  Here mstatus access will not fault, so it doesn't really matter.  Sometime in S.py, I have to use csr instructions directly to hit the coverpoints because I can't have a nop in-between.

davidharrishmc (approved):
100% coverage, no known problems.  Merging.  @jordancarlin feel free to make style edits after this.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
jordancarlin: This was not addressed and is a blocking issue to merge. test_data.xlen is never 64 when generating priv tests.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
jordancarlin: Agreed. Was more of a note for myself than anything else.

jordancarlin (approved):

jordancarlin (approved):

PR #1008, opened 2026-02-27
Skip XEPC checks if accessing RVMODEL_ACCESS_FAULT_ADDRESS

Zain2050:


davidharrishmc (approved):
thank you for the fast fix on the trap handler.  I’m going to merge it to unblock Julia.

The CTP defines RVMODEL_ACCESS_FAULT_ADDRESS as the base address of a region of at least 128 bytes that
causes an access fault when any part of the region is read
or written. This is used to test exceptions.

For example, there are Exceptions coverpoints that look at both misaligned and access fault, so those would have offsets of up to 7 bytes.

Would it be easy for you to adjust to accept a 128 byte region?  Would that interfere with any of your other tests?

PR #1013, opened 2026-02-27
S, U, UF, SmF tests at 100% coverage in Sail

davidharrishmc:
Does not match Spike yet because of parameters
No self-checking trap handler yet available


generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -141,20 +140,16 @@ def _generate_priv_inst_tests(test_data: TestData) -> list[str]:
         ),
         "",
         # ecall test
-        f"    li x{check_reg}, 1    # success code",
         test_data.add_testcase("ecall", coverpoint, covergroup),
+        f"test_{test_data.test_count}:",
```
jordancarlin: Merge issue? These labels should not be added back in.

generators/testgen/src/testgen/asm/csr.py
```diff
@@ -116,8 +116,16 @@ def csr_walk_test(test_data: TestData, csr_name: str, covergroup: str, coverpoin
         f"    li x{walk_reg}, 1            # 1 in lsb",
     ]
 
+    # don't test MODE bit of satp
+    if csr_name == "satp":
+        r1 = range(31)
```
jordancarlin: Maybe call these range1 and range2 or something. r1 made me think it was a register.

generators/testgen/src/testgen/asm/csr.py
```diff
@@ -144,34 +152,35 @@ def csr_walk_test(test_data: TestData, csr_name: str, covergroup: str, coverpoin
...
-                "",
-                f"    CSRW({csr_name}, x{temp_reg})    # set all bits",
-                f"    CSRC({csr_name}, x{walk_reg})    # clear walking 1 in column {i}",
-                test_data.add_testcase(f"{csr_name}_clr_bit_{i}", coverpoint, covergroup),
-                gen_csr_read_sigupd(check_reg, csr_name, test_data),
-                f"    slli x{walk_reg}, x{walk_reg}, 1      # walk the 1",
-            ]
-        )
-    lines.append("#endif\n")
+    if csr_name != "satp":  # skip walking 0s for satp
```
jordancarlin: There is enough special casing in here that it might make more sense to just use some custom code for `satp` and not call the function for that CSR.

generators/testgen/src/testgen/priv/extensions/U.py
```diff
@@ -0,0 +1,101 @@
...
+##################################
+
+"""U privileged extension test generator."""
+
+from testgen.asm.helpers import comment_banner
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_priv_inst_tests(test_data: TestData) -> list[str]:
```
jordancarlin, line 16: This function is almost the same between Sm, S, and U. Can we use a common version of the function instead of duplicating it everywhere?

generators/testgen/src/testgen/priv/extensions/U.py
```diff
@@ -0,0 +1,101 @@
...
+            "Attempt to read non-user-mode registers.  Should throw illegal instruction",
+        ),
+    ]
+    for csr in (
+        list(range(0x100, 0x400)) + list(range(0x500, 0x800)) + list(range(0x900, 0xC00)) + list(range(0xD00, 0x1000))
+    ):
+        lines.extend(
+            [
+                test_data.add_testcase(coverpoint, f"{csr}", covergroup),
+                f"\tCSRR(t0, 0x{csr:03x})    # attempt to read CSR {csr:03x}; should get illegal instruction",
```
jordancarlin: Don't use hardcoded registers.

jordancarlin (changes requested):
Still need to review S, but here are comments on the rest.

coverpoints/priv/SmF_coverage.svh
```diff
@@ -64,12 +70,6 @@ covergroup SmF_cg with function sample(ins_t ins);
         wildcard bins csrrc_fcsr   = {32'b000000000011_?????_011_?????_1110011};
         wildcard bins csrrc_frm    = {32'b000000000010_?????_011_?????_1110011};
         wildcard bins csrrc_fflags = {32'b000000000001_?????_011_?????_1110011};
-        `ifdef XLEN32
-            `ifdef D_SUPPORTED
-                wildcard bins fmvh         = {FMVH.X.D};
-                wildcard bins fmvp         = {FMVP.D.X};
-            `endif
-        `endif
```
jordancarlin, line 72: Why are we dropping these?

generators/testgen/src/testgen/priv/extensions/SmF.py
```diff
@@ -0,0 +1,194 @@
+##################################
+# SmF.py
+#
+# SmF floating-point from machine mode privileged extension test generator.
+# jcarlin@hmc.edu Jan 2026
```
jordancarlin: I didn't write this.

generators/testgen/src/testgen/priv/extensions/SmF.py
```diff
@@ -0,0 +1,194 @@
...
+from testgen.asm.helpers import comment_banner, load_float_reg
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _gen_fs_init(fs: int, temp_reg: int) -> str:
+    """Initialize mstatus.FS"""
+    lines = [
+        f"\t# mstatus.FS={fs}",
+        f"\tli x{temp_reg}, {3 << 13}  # 11 in bits 14:13",
```
jordancarlin: LI macro

generators/testgen/src/testgen/priv/extensions/SmF.py
```diff
@@ -0,0 +1,194 @@
...
+    ######################################
+    lines.append(
+        comment_banner(
+            coverpoint,
+            "Read, write all 1s, write all 0s, set all 1s, set all 0s, restore all fp CSRs from machine mode with different mstatus.FS",
+        )
+    )
+
+    ones_reg, check_reg, scratch_reg, temp_reg, save_reg = test_data.int_regs.get_registers(5, exclude_regs=[0])
+    lines.append(f"\tli x{ones_reg}, -1")
```
jordancarlin: Use the LI macro. Global comment.

generators/testgen/src/testgen/priv/extensions/SmF.py
```diff
@@ -0,0 +1,194 @@
...
+            lines.extend(
+                [
+                    "",
+                    _gen_fs_init(fs, temp_reg),
+                    test_data.add_testcase(f"{insn}", coverpoint_full, covergroup),
+                    f"\t{insn} # execute instruction with mstatus.FS={fs}",
+                    gen_csr_read_sigupd(temp_reg, "mstatus", test_data),
+                ]
+            )
+        lines.append("\n  #if __riscv_xlen == 32\n\t#ifdef ZFA_SUPPORTED")
```
jordancarlin: Simplify using `#if ... && ...`.

generators/testgen/src/testgen/priv/extensions/SmF.py
```diff
@@ -0,0 +1,194 @@
...
+            lines.extend(
+                [
+                    "",
+                    _gen_fs_init(fs, temp_reg),
+                    test_data.add_testcase(f"{insn}", coverpoint_full, covergroup),
+                    f"\t{insn} # execute instruction with mstatus.FS={fs}",
+                    gen_csr_read_sigupd(temp_reg, "mstatus", test_data),
+                ]
+            )
+        lines.append("\t#endif\n  #endif\n#endif\n#ifdef ZFA_SUPPORTED")
```
jordancarlin: Worth making this multiple lines in the Python instead of using `\n` because this is very difficult to read right now.

generators/testgen/src/testgen/priv/extensions/UF.py
```diff
@@ -0,0 +1,76 @@
...
+    ######################################
+    lines.append(
+        comment_banner(
+            coverpoint,
+            "Read, write all 1s, write all 0s, set all 1s, set all 0s, restore all fp CSRs from M-mode",
+        ),
+    )
+
+    for csr in csrs:
+        lines.extend(csr_access_test(test_data, csr, covergroup, coverpoint))
```
jordancarlin: Not clear why this coverpoint is needed. Isn't this already tested in SmF?

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+    ]
+
+    for sd in (0, 1):
+        for fs in range(4):
+            for xs in range(4):
+                for vs in range(4):
+                    binname = f"sd_{sd}_fs_{fs:02b}_xs_{xs:02b}_vs_{vs:02b}"
+                    fields = fs << 13 | xs << 15 | vs << 9
+                    test_lines = [
+                        test_data.add_testcase(binname, coverpoint, covergroup),
```
jordancarlin: Move add_testcase to right before gen_csr_write_sigupd so that the PC is closer to accurate. Global comment.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+    ]
+
+    for i in range(24):
+        if i in {14, 17}:  # skip reserved causes
+            continue
+        lines.extend(
+            [
+                test_data.add_testcase(f"b_{i}", coverpoint, covergroup),
+                f"    LI(x{check_reg}, {i})           # exception cause {i}",
+                gen_csr_write_sigupd(check_reg, "scause", test_data),
```
jordancarlin: Move `add_testcase` to right before `gen_csr_write_sigupd` so that the PC is closer to accurate.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+
+    for i in range(14):
+        if i in {0, 4, 8}:  # skip reserved causes
+            continue
+        lines.extend(
+            [
+                test_data.add_testcase(f"b_{i}", coverpoint, covergroup),
+                f"    LI(x{check_reg}, {i})           # interrupt cause {i}",
+                f"    or x{check_reg}, x{check_reg}, x{temp_reg}          # set interrupt bit",
+                gen_csr_write_sigupd(check_reg, "scause", test_data),
```
jordancarlin: Move `add_testcase` to right before `gen_csr_write_sigupd` so that the PC is closer to accurate.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+                    ]
+                    if sd == 1:
+                        test_lines.append(f"    or x{check_reg}, x{check_reg}, x{reg1}      # set SD bit")
+                    test_lines.extend(
+                        [
+                            f"    or x{check_reg}, x{check_reg}, x{reg3}   # value to write to sstatus with SD/FS/XS/VS bits set/clear",
+                            gen_csr_write_sigupd(check_reg, "sstatus", test_data),
+                        ]
+                    )
+                    lines.extend(test_lines)
```
jordancarlin: What is the point of `test_lines`? Can you just directly extend `lines` in the loop?

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+                        [
+                            # Test the write value
+                            test_data.add_testcase(f"{binname}_wval", coverpoint, covergroup),
+                            f"    LI(x{check_reg}, 0x{fields:08x})  # mpp = {mpp:02b} mprv = {mprv} mpie = {mpie} mie = {mie}",
+                            f"    or x{check_reg}, x{check_reg}, x{reg1}          # value to write to mstatus with MPP/MPRV/MPIE/MIE bits set/clear",
+                            f"    LA(x{reg3}, 1f)             # return address after mret",
+                            f"    CSRW(mepc, x{reg3})          # set mepc to return address",
+                            f"    CSRW(mstatus, x{check_reg})       # write mstatus with MPP/MPRV/MPIE/MIE bits set/clear",
+                            "    mret                   # test mret instruction",
+                            f"    li x{check_reg}, -1              # should not be executed",
```
jordancarlin: Needs to use the LI macro.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+        comment_banner(
+            coverpoint,
+            "Read, write all 1s, write all 0s, set all 1s, set all 0s, restore all U-mode CSRs from S-mode",
+        ),
+    )
+
+    lines.extend(["#ifdef F_SUPPORTED"])
+    for csr in ["fflags", "frm", "fcsr"]:
+        lines.extend(csr_access_test(test_data, csr, covergroup, coverpoint))
+    lines.extend(["#endif"])
```
jordancarlin, line 364: We already have UF. Might be worth breaking this out to an SF test.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+
+    ######################################
+    coverpoint = "cp_scsrwalk"
+    ######################################
+    lines.append(
+        comment_banner(
+            coverpoint,
+            "Set and clear each bit individually in all writable S-mode CSRs, excluding satp, which is exercised in Sv tests",
```
jordancarlin: `satp` is exercised right below this. No need to mention Sv tests.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+            "Set and clear each bit individually in satp, excluding satp.mode",
+        ),
+    )
+
+    walk_reg, mask_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines.extend(
+        [
+            "\tcsrw satp, zero      # set satp to 0 to start with",
+            f"\tli x{mask_reg}, -1     # x{mask_reg} = all 1s for walking bit tests",
```
jordancarlin: LI macro.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+def _add_shadow(r1: int, r2: int, wreg: str, rreg: str, coverpoint: str, covergroup: str, test_data: TestData) -> str:
+    """Helper function to generate shadow CSR test lines for writing wreg and reading rreg."""
+    return str.join(
+        "\n",
+        [
+            test_data.add_testcase(f"{wreg}_{rreg}_1s", coverpoint, covergroup),
+            f"\tcsrw {wreg}, x{r1}       # write all 1s to {wreg}",
+            gen_csr_read_sigupd(r2, rreg, test_data),
```
jordancarlin, line 508: Could this be simplified to use `gen_csr_write_sigupd`?

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,536 @@
...
+                            f"    CSRW(mstatus, x{check_reg})       # write mstatus with MPP/MPRV/MPIE/MIE bits set/clear",
+                            test_data.add_testcase(f"{binname}_wval", coverpoint, covergroup),
+                            "    mret                   # test mret instruction",
+                            f"   LI(x{check_reg}, -1)              # should not be executed",
+                            "1:                         # mret should return to here",
+                            "    RVTEST_GOTO_MMODE      # make sure we return to machine mode",
+                            write_sigupd(check_reg, test_data),
+                            # Test the read value
+                            test_data.add_testcase(f"{binname}_rval", coverpoint, covergroup),
+                            gen_csr_read_sigupd(check_reg, "mstatus", test_data),
```
jordancarlin, line 197: Do we really want to add a second testcase for the second sigupd? Seems like it might be more clear for both sigupds to point at the same testcase (the mret). Applies to a few other spots in this test as well.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -0,0 +1,536 @@
...
+                        [
+                            # Test the write value
+                            f"    LI(x{check_reg}, 0x{fields:08x})  # mpp = {mpp:02b} mprv = {mprv} mpie = {mpie} mie = {mie}",
+                            f"    or x{check_reg}, x{check_reg}, x{reg1}          # value to write to mstatus with MPP/MPRV/MPIE/MIE bits set/clear",
+                            f"    LA(x{reg3}, 1f)             # return address after mret",
+                            f"    CSRW(mepc, x{reg3})          # set mepc to return address",
+                            f"    CSRW(mstatus, x{check_reg})       # write mstatus with MPP/MPRV/MPIE/MIE bits set/clear",
+                            test_data.add_testcase(f"{binname}_wval", coverpoint, covergroup),
+                            "    mret                   # test mret instruction",
+                            f"   LI(x{check_reg}, -1)              # should not be executed",
```
jordancarlin: This will expand to more than one instruction, so some of it will be executed. Use an `addi` with a smaller immediate value.

jordancarlin (changes requested):

jordancarlin (approved):

PR #1019, opened 2026-03-02
ExceptionsZalrsc test generator with 92% coverage

Teemooooooooo:
ExceptionsZalrsc python test generator at 92% coverage
- Note that illegal sc.w does not get coverage as SAIL stores content in by bytes instead of giving exceptions
- Sail issue: https://github.com/riscv/sail-riscv/issues/1574

- Note that QEMU also has a bug where sc instructions at illegal address do not trigger access fault exception
- QEMU issue: https://gitlab.com/qemu-project/qemu/-/issues/3323
_generate_store_address_misaligned_tests is not hitting the following coverpoints due to SAIL bug
- Cross cp_store_address_misaligned_illegal_w 
- Cross cp_store_address_misaligned_illegal_d


generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
+##################################
+# ExceptionsZalrsc.py
+#
+# ExceptionsZalrsc privileged extension test generator.
+# ellyu@g.hmc.edu Jan 2026
```
jordancarlin: Not January.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+        lines.append("#if __riscv_xlen == 64")
+        lines.extend(
+            [
+                test_data.add_testcase(f"lr.d_off{offset}", coverpoint, covergroup),
+                f"    lr.d x{check_reg}, (x{addr_reg})",
+                "   nop",
+                "   nop",
+            ]
+        )
+        lines.append("#endif")
```
jordancarlin: Instead of repeated calls to `lines.extend` and `lines.append`, it is easier to follow when you combine the lines if there are no Python loops or anything. Something like the following. This also applies to the other functions in this file.

```suggestion
        lines.extend(
            [
                f"\n# Offset {offset} (LSBs: {offset:03b})",
                f"    LA(x{addr_reg}, scratch)",
                f"    addi x{addr_reg}, x{addr_reg}, {offset}",
                test_data.add_testcase(f"lr.w_off{offset}", coverpoint, covergroup),
                f"    lr.w x{check_reg}, (x{addr_reg})",
                "   nop",
                "   nop",
                "#if __riscv_xlen == 64"),
                test_data.add_testcase(f"lr.d_off{offset}", coverpoint, covergroup),
                f"    lr.d x{check_reg}, (x{addr_reg})",
                "   nop",
                "   nop",
                "#endif",
            ]
        )
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_load_address_misaligned_tests(test_data: TestData) -> list[str]:
+    """Generate load address misaligned exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_address_misaligned"
+
+    addr_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [comment_banner(coverpoint, "Load Address Misaligned")]
```
jordancarlin: If you're going to include a comment beyond the coverpoint name, it should be more descriptive. No need to say "Load Address Misaligned" because that can be inferred easily from the coverpoint name (cp_load_address_misaligned).

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+                f"    addi x{addr_reg}, x{addr_reg}, {offset}",
+            ]
+        )
+
+        lines.extend(
+            [
+                test_data.add_testcase(f"lr.w_off{offset}", coverpoint, covergroup),
+                f"    lr.w x{check_reg}, (x{addr_reg})",
+                "   nop",
+                "   nop",
```
jordancarlin: Why do we need two nops? Isn't one enough?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+        lines.append("#endif")
+
+    test_data.int_regs.return_registers([addr_reg, check_reg])
+    return lines
+
+
+def _generate_store_address_misaligned_tests(test_data: TestData) -> list[str]:
+    """Generate store address misaligned exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_store_address_misaligned"
+    addr_reg, data_reg, rd_reg, temp_reg, base_reg = test_data.int_regs.get_registers(5, exclude_regs=[0, 1])
```
jordancarlin: Why are we excluding x1?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset} (LSBs: {offset:03b})")
+
+        lines.extend(
+            [
+                f"    LA(x{base_reg}, scratch)",
+                f"    andi x{base_reg}, x{base_reg}, -8",  # force LSBs predictable
+                f"    addi x{addr_reg}, x{base_reg}, {offset}",  # addr = aligned base + offset
+                f"    li x{data_reg}, 0xDECAFCAB",
```
jordancarlin: Make sure to use the LI macro.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+def _generate_load_access_fault_tests(test_data: TestData) -> list[str]:
+    """Generate load access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_access_fault"
+    addr_reg, check_regs = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Load Access Fault"),
+        # "    .align 2",
```
jordancarlin: Can this be deleted?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+
+
+def _generate_store_access_fault_tests(test_data: TestData) -> list[str]:
+    """Generate store access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_store_access_fault"
+    addr_reg, data_reg, rd_reg = test_data.int_regs.get_registers(3, exclude_regs=[0, 1])
+
+    lines = [
+        comment_banner(coverpoint, "Store Access Fault"),
+        "    .align 2",
```
jordancarlin: Is this needed?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+        lines.append("#endif")
+
+    test_data.int_regs.return_registers([addr_reg, data_reg, rd_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "ExceptionsZalrsc",
+    required_extensions=["I", "Zicsr", "Zalrsc", "Sm"],
+    march_extensions=["Zicsr", "I", "Zalrsc", "A"],
```
jordancarlin: Unless there are extra extensions only needed for the march and not the required_extensions, let's not set march_extensions. It will default to `required_extensions`.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,293 @@
...
+        )
+        lines.append("#endif")
+
+    test_data.int_regs.return_registers([addr_reg, data_reg, rd_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "ExceptionsZalrsc",
+    required_extensions=["I", "Zicsr", "Zalrsc", "Sm"],
```
jordancarlin: Why is I required? This should work for E as well.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
Teemooooooooo: The original hand written test states that:
# REQUIRED_EXTENSIONS: [I, Zicsr, Zalrsc, Sm]
# MARCH: rv${XLEN}ia_zicsr
in the beginning of the file, I did the required extensions based on that. Is there anywhere else I should be looking to figure out what extensions to include?

jordancarlin: Not sure why the old version had that, but I think Zalrsc and Sm should be sufficient. You can include Zicsr as well if you want, but that is implied by Sm. A includes Zaamo and Zalrsc, which is not needed for this test.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
Teemooooooooo: Should I just remove it? I am not sure what else to put here

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+    for offset in range(8):
+        lines.extend(
+            [
+                f"\n# Offset {offset} (LSBs: {offset:03b})",
+                f"    LA(x{addr_reg}, scratch)",
+                f"    addi x{addr_reg}, x{addr_reg}, {offset}",
+                test_data.add_testcase(f"lr.w_off{offset}", coverpoint, covergroup),
+                f"    lr.w x{check_reg}, (x{addr_reg})",
+                "   nop",
+                "#if __riscv_xlen == 64",
```
davidharrishmc, line 35: Let's add a SIGUPD on check_reg after each lr.w/lr.d.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+    lines = [comment_banner(coverpoint)]
+    # illegal sc.w does not get coverage as SAIL stores content in by bytes instead of giving exceptions
+    # Sail issue: https://github.com/riscv/sail-riscv/issues/1574
+
+    for offset in range(8):
+        lines.extend(
+            [
+                f"\n# Offset {offset} (LSBs: {offset:03b})",
+                f"    LA(x{base_reg}, scratch)",
+                f"    andi x{base_reg}, x{base_reg}, -8",  # force LSBs predictable
```
davidharrishmc: Isn't scratch already aligned to a multiple of 8?  I don't think the anti is needed.  If it were needed, then it would be needed for loads as well.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+                f"    addi x{addr_reg}, x{base_reg}, {offset}",  # addr = aligned base + offset
+                f"    LI(x{data_reg}, 0xDECAFCAB)",
+                f"    LI(x{rd_reg}, 5)",  # rd_gt_one_prev
```
davidharrishmc: Let's pick a crazy value that would never possibly be hit by accident, rather than 5.  Improve comment of why it is used.  Multiple places.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+                f"    lr.w x{temp_reg}, (x{base_reg})",  # establish reservation
+                test_data.add_testcase(f"sc.w_off{offset}", coverpoint, covergroup),
+                f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
+                "#if __riscv_xlen == 64",
```
davidharrishmc, line 84: The 16 bytes of scratch should be stored to the signature to confirm if this happened.  See ExceptionsSm for an example.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+                f"    LI(x{rd_reg}, 5)",  # rd_gt_one_prev
+                f"    lr.d x{temp_reg}, (x{base_reg})",  # establish reservation
+                test_data.add_testcase(f"sc.d_off{offset}", coverpoint, covergroup),
+                f"    sc.d x{rd_reg}, x{data_reg}, (x{addr_reg})",
+                "#endif",
```
davidharrishmc, line 103: Also store to memory.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+    addr_reg, check_regs = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        test_data.add_testcase("lr.w_load_access_fault", coverpoint, covergroup),
+        f"    LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        "   nop",
+        f"    lr.w x{check_regs}, (x{addr_reg})",
+        "   nop",
+        "   nop",
```
davidharrishmc: Why two nops?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+def _generate_load_access_fault_tests(test_data: TestData) -> list[str]:
+    """Generate load access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_access_fault"
+    addr_reg, check_regs = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        test_data.add_testcase("lr.w_load_access_fault", coverpoint, covergroup),
+        f"    LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        "   nop",
```
davidharrishmc: Why nop?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def _generate_load_access_fault_tests(test_data: TestData) -> list[str]:
+    """Generate load access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_access_fault"
+    addr_reg, check_regs = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        test_data.add_testcase("lr.w_load_access_fault", coverpoint, covergroup),
```
davidharrishmc: It would be good to put a value in check_reg before all this, and store it to signature after the nops, to confirm check_reg is not disturbed by the faulting lr.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+            ]
+        )
+    test_data.int_regs.return_registers([addr_reg, data_reg, rd_reg, base_reg, temp_reg])
+    return lines
+
+
+def _generate_load_access_fault_tests(test_data: TestData) -> list[str]:
+    """Generate load access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_access_fault"
+    addr_reg, check_regs = test_data.int_regs.get_registers(2, exclude_regs=[0])
```
davidharrishmc: check_regs -> check_reg

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+            "    sw x11, 0(x10)",
+            "    sw x11, 4(x10)",
+            "    sw x11, 8(x10)",
+            "    sw x11, 12(x10)",
+            "",
+        ]
+    )
+
+    lines.extend(_generate_load_address_misaligned_tests(test_data))
+    lines.extend(_generate_store_address_misaligned_tests(test_data))
```
davidharrishmc: Ordering seems random.  Let's use same order as in spreadsheet, moving store after the three loads.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
davidharrishmc: Also store rd_reg to the signature to show it didn't change.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+    """Generate instruction address misaligned and access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_misaligned_priority"
+    addr_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        f"    LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        f"    addi x{addr_reg}, x{addr_reg}, 1",
+        test_data.add_testcase("lr.w_off1_priority", coverpoint, covergroup),
+        f"    lr.w x{check_reg}, 0(x{addr_reg})",
```
davidharrishmc: lr isn't supposed to have a 0 before the (rs1).  2 places.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+        "     nop",
```
davidharrishmc: For paranoia, initialize check_reg before this and store it to signature after.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_store_access_fault"
+    addr_reg, data_reg, rd_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        f"    LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        test_data.add_testcase("sc.w_fault", coverpoint, covergroup),
+        f"    LI(x{data_reg}, 0xADDEDCAB)",
+        f"    LI(x{rd_reg}, 5)",
+        f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
```
davidharrishmc, line 182: test plan says to put a lr before the sc.  lr will fault.  Don't forget nop after.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+        "#endif",
+        "",
+    ]
+
+    test_data.int_regs.return_registers([addr_reg, data_reg, rd_reg])
+    return lines
+
+
+############NOTE ###########
+# need to set rd to be greater than 1 before the instr
```
davidharrishmc: I don't understand these notes.  Can they be deleted?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+            test_data.add_testcase("sc.w_off1_priority", coverpoint, covergroup),
+            f"    lr.w x{rd_reg}, 0(x{addr_reg})",
+            "     nop",
+            f"    LI(x{rd_reg}, 5)",
+            f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
+            "     nop",
+            "#if __riscv_xlen == 64",
+            f"    LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+            f"    addi x{addr_reg}, x{addr_reg}, 1",
+            f"    LI(x{data_reg}, 0xDECAFCAB)",  # Match original value
```
davidharrishmc: 64-bit value?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,221 @@
...
+            test_data.add_testcase("sc.w_off1_priority", coverpoint, covergroup),
+            f"    lr.w x{rd_reg}, 0(x{addr_reg})",
+            "     nop",
+            f"    LI(x{rd_reg}, 5)",
+            f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
+            "     nop",
+            "#if __riscv_xlen == 64",
+            f"    LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+            f"    addi x{addr_reg}, x{addr_reg}, 1",
```
davidharrishmc: addr_reg already has a value.  Why reload it?

davidharrishmc (changes requested):
Many little things.  Going a good direction!

.github/workflows/regress.yml
```diff
@@ -174,7 +174,7 @@ jobs:
           DOCKER=1 \
           FAST=True \
           EXTENSIONS="" \
-          EXCLUDE_EXTENSIONS=Sm,ExceptionsZc,S \
+          EXCLUDE_EXTENSIONS=Sm,ExceptionsZc,S,ExceptionsZalrsc \
```
jordancarlin: Why does this fail the Spike CI but pass QEMU CI? I thought the expected results coming from Sail were wrong.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,252 @@
...
+    lines = [comment_banner(coverpoint)]
+
+    for offset in range(8):
+        lines.extend(
+            [
+                f"\n# Offset {offset} (LSBs: {offset:03b})",
+                f"    LA(x{addr_reg}, scratch)",
+                f"    addi x{addr_reg}, x{addr_reg}, {offset}",
+                test_data.add_testcase(f"lr.w_off{offset}", coverpoint, covergroup),
+                f"    lr.w x{check_reg}, (x{addr_reg})",
```
jordancarlin, line 32: Initialize `check_reg` to something before the `lr.w` so that we know if it changes.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,252 @@
...
+            [
+                f"\n# Offset {offset} (LSBs: {offset:03b})",
+                f"    LA(x{base_reg}, scratch)",
+                f"    addi x{addr_reg}, x{base_reg}, {offset}",  # addr = aligned base + offset
+                f"    LI(x{data_reg}, 0xDECAFCAB)",
+                f"    lr.w x{temp_reg}, (x{addr_reg})",  # establish reservation
+                "    nop",
+                test_data.add_testcase(f"sc.w_off{offset}", coverpoint, covergroup),
+                f"    LI(x{rd_reg}, 0xBAD)",  # previous rd greater than 1
+                f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
```
jordancarlin, line 72: The `add_testcase` function needs to be right before the instruction we are testing (the `sc.w` in this case).

```suggestion
                f"    LI(x{data_reg}, 0xDECAFCAB)",
                f"    LI(x{rd_reg}, 0xBAD)",  # previous rd greater than 1
                f"    lr.w x{temp_reg}, (x{addr_reg})",  # establish reservation
                "    nop",
                test_data.add_testcase(f"sc.w_off{offset}", coverpoint, covergroup),
                f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
```

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def _generate_load_access_fault_tests(test_data: TestData) -> list[str]:
+    """Generate load access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_access_fault"
+    addr_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        test_data.add_testcase("lr.w_load_access_fault", coverpoint, covergroup),
```
jordancarlin: Move `add_testcase` right before the `lr.w`.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,252 @@
...
+def _generate_load_misaligned_priority_tests(test_data: TestData) -> list[str]:
+    """Generate instruction address misaligned and access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_load_misaligned_priority"
+    addr_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        f"    LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        f"    addi x{addr_reg}, x{addr_reg}, 1",
+        test_data.add_testcase("lr.w_off1_priority", coverpoint, covergroup),
```
jordancarlin: Move `add_testcase` right before the `lr.w`.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,252 @@
...
+
+def _generate_store_access_fault_tests(test_data: TestData) -> list[str]:
+    """Generate store access fault exception tests."""
+    covergroup, coverpoint = "ExceptionsZalrsc_cg", "cp_store_access_fault"
+    addr_reg, data_reg, rd_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint),
+        f"    LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        test_data.add_testcase("sc.w_fault", coverpoint, covergroup),
```
jordancarlin: Move `add_testcase` right before the `sc.w`.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,252 @@
...
+        f"    LI(x{data_reg}, 0xADDEDCAB)",
+        f"    lr.w x{rd_reg}, (x{addr_reg})",
+        "     nop",
+        f"    LI(x{rd_reg}, 0xBAD)",  # previous rd greater than 1
+        f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
```
jordancarlin, line 182: Missing sigupds to check rd_reg.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,252 @@
...
+    lines.extend(
+        [
+            f"    LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+            f"    addi x{addr_reg}, x{addr_reg}, 1",
+            f"    LI(x{data_reg}, 0xDECAFCAB)",  # Match original value
+            f"    LI(x{rd_reg}, 0xDECAFCAB)",
+            test_data.add_testcase("sc.w_off1_priority", coverpoint, covergroup),
+            f"    lr.w x{rd_reg}, (x{addr_reg})",
+            "     nop",
+            write_sigupd(rd_reg, test_data),
```
jordancarlin: If this is a test for stores, shouldn't the sigupd be after the sc.w, not the lr.w?

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
Teemooooooooo: Should I just sigupd for both sc.w and lr.w?

davidharrishmc:
Sure. Put a add_test case just before each.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,273 @@
...
+        f"    LI(x{data_reg}, 0xDEADBEEFDECAFCAB)",
+        f"    LI(x{rd_reg}, 0xBAD)",
+        test_data.add_testcase("lr.d_store_fault", coverpoint, covergroup),
+        f"    lr.d x{rd_reg}, (x{addr_reg})",
+        "     nop",
+        write_sigupd(rd_reg, test_data),
+        f"    LI(x{rd_reg}, 0xBAD)",  # previous rd greater than 1
+        test_data.add_testcase("sc.d_fault", coverpoint, covergroup),
+        f"    sc.d x{rd_reg}, x{data_reg}, (x{addr_reg})",
+        "   nop",
```
davidharrishmc: one too few space

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,273 @@
...
+        test_data.add_testcase("sc.w_fault", coverpoint, covergroup),
+        f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
+        "   nop",
+        write_sigupd(rd_reg, test_data),
+        "#if __riscv_xlen == 64",
+        f"    LI(x{data_reg}, 0xDEADBEEFDECAFCAB)",
+        f"    LI(x{rd_reg}, 0xBAD)",
+        test_data.add_testcase("lr.d_store_fault", coverpoint, covergroup),
+        f"    lr.d x{rd_reg}, (x{addr_reg})",
+        "     nop",
```
davidharrishmc: one too many space.  Look at your generated files and make sure the spaces look good.

davidharrishmc (changes requested):
Good to go after minor spacing fixes and merging into the latest version.

davidharrishmc (approved):

davidharrishmc:
Could you add a comment in the coverage file by the coverpoints that aren't hit indicating why and pointing to the Sail bug, so we don't forget this and have to debug again later?

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
```diff
@@ -0,0 +1,273 @@
...
+                f"    addi x{addr_reg}, x{base_reg}, {offset}",  # addr = aligned base + offset
+                f"    LI(x{data_reg}, 0xDECAFCAB)",
+                f"    LI(x{temp_reg}, 0xBAD)",
+                test_data.add_testcase(f"lr.w_off{offset}", coverpoint, covergroup),
+                f"    lr.w x{temp_reg}, (x{addr_reg})",  # establish reservation
+                "    nop",
+                write_sigupd(temp_reg, test_data),
+                f"    LI(x{rd_reg}, 0xBAD)",  # previous rd greater than 1
+                test_data.add_testcase(f"sc.w_off{offset}", coverpoint, covergroup),
+                f"    sc.w x{rd_reg}, x{data_reg}, (x{addr_reg})",
```
jordancarlin, line 72: There are limits on how many instruction can be between `lr.w` and `sc.w`. To avoid conflating this test with that limit, let's move the `LI(x{rd_reg}, 0xBAD)` before the `lr.w`. So initialize all of the registers, then do the `lr.w` and then do the `sc.w` (with a `nop` in between). Let's also move the `write_sigupd` for the `temp_reg` after the `sc.w` since it should not change in between helps avoid running extra instructions between them. This applies to all `sc.w`/`sc.d` tests.

jordancarlin (changes requested):
One last comment and then this looks good to go.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
Teemooooooooo: One of the reason that I put the `LI(x{rd_reg}, 0xBAD)` before the `sc.w` is because one of the coverpoint is `rd_gt_one_prev`. If I move it, it would just be `nop` before the `sc.w`, should I still move it?

davidharrishmc:
Good point.  We’d best keep that LI before the sc.  We may have to drop the write_sigupd after the lr to prevent too much code between the two.

generators/testgen/src/testgen/priv/extensions/ExceptionsZalrsc.py
jordancarlin: @Teemooooooooo Let's change the `LI` to be a regular `addi` instruction and either move the `write_sigupd` that is checking `temp_reg` after the `sc.w` or drop it.

jordancarlin (approved):

jordancarlin (approved):

PR #1024, opened 2026-03-02
Self-Check support in trap handler

### LINKED ISSUE #968: [ACT4] Reference trap signatures are not being embedded into the ELF
Currently, `mtrap_sigptr` does not preload reference signatures and therefore **lack Self-Check** functionality. This needs to be implemented. 
Also `TRAP_SIGUPD` macro is currently commented out for `RVTEST_SELFCHECK` mode. It needs to re-enabled and tested.
```
    signature_base:
      #ifdef RVTEST_SELFCHECK
        // Preload signature region with correct values for self-checking
        #include SIGNATURE_FILE
      #else
        // Initialize signature region to known value for initial pass
        .fill SIGUPD_COUNT*SIG_STRIDE,4,0xdeadbeef
      #endif

    // Signature region for trap handlers
    #ifdef rvtest_mtrap_routine
      tsig_begin_canary:
        CANARY
      mtrap_sigptr:
        .fill 20000*(XLEN/32),4,0xdeadbeef
      tsig_end_canary:
        CANARY
    #endif
```

Zain2050:
I have implemented most of the mechanism, but there is still work to be done. 

**Approach:**
* As `signature_base` and `mtrap_sigptr` are separate labels, therefore their signatures are supposed to embedded separately. I have modified `sig_modify.py` to remove trap signatures from `.results` file and put them in a separate `.trap.results` file which can be passed in separately. 
* Defined a new function `failed_test_x7_x9`, which was required due to limited availability of trap registers. 
* Defined relevant failure strings.

**TODO**
* We need a way to know the size of the trap signature file, just like `SIGUPD_COUNT`. We can have  `FAULT_COUNT` (because signatures are only written in case of faults/interrupts, not ecall traps), and it could be used to to determine the size of the trap signature region. Currently it is hardcoded to 5000 writes (1 fault writes max 6 signatures) which is enough for current tests.
* We need to a way to check if the correct number of traps occured.


Zain2050:
Currently, CI is failing for `InterruptsSm` due to a trap mismatch, which means that the self-check mechanism is working. 
```
RVCP-SUMMARY: Test File "InterruptsSm-00.S": FAILED
RVCP: Test Info: "Mismatch in trap vector signature! Trap was being handled in M-Mode."
RVCP: Instruction: 0x2c013403
RVCP: Approximate address (failure may be slightly after this): 0x0000000080003bb0
RVCP: Register: x8
RVCP: Bad Value:      0x0000000003100e63
RVCP: Expected Value: 0x0000000003100663
```
If comment this sigupd out, we get 
```
RVCP-SUMMARY: Test File "InterruptsSm-00.S": FAILED
RVCP: Test Info: "Mismatch in mcause value! Trap was being handled in M-Mode."
RVCP: Instruction: 0x00050413
RVCP: Approximate address (failure may be slightly after this): 0x0000000080003c40
RVCP: Register: x8
RVCP: Bad Value:      0x8000000000000003
RVCP: Expected Value: 0x8000000000000007
```
I'll debug this soon, while the rest of the stuff can be reviewed.

tests/env/failure_code.h
```diff
@@ -40,6 +40,19 @@
...
         j failedtest_saveregs
 
+    # Log failure. x7 contains return address of jal from the failure and x9 is a vacant temporary register
+    failedtest_x7_x9:
+        la x9, begin_failure_scratch
+        SREG x7, 104(x9)               # store return address
+        SREG DEFAULT_TEMP_REG, 32(x9)  # save DEFAULT_TEMP_REG
+        SREG DEFAULT_LINK_REG, 40(x9)  # save DEFAULT_LINK_REG
+        SREG x1, 8(x9)                 # save x1 early
+        sw zero, 0(x9)                 # failure_type = 0 (integer)
```
jordancarlin: Should we have a different failure type for the trap handler that prints out different things? For example, seems like it would be useful to always print `xepc` so the user knows which instruction caused the trap that mismatched and maybe even print out the instruction at that address. We could probably drop some of the usual printing as well that isn't relevant to trap handler mismatches.

tests/env/test_setup.h
```diff
@@ -245,34 +246,37 @@
...
-    #ifdef rvtest_mtrap_routine
-      tsig_begin_canary:
-        CANARY
-      mtrap_sigptr:
-        .fill 20000*(XLEN/32),4,0xdeadbeef
-      tsig_end_canary:
+      // Create canary at end of signature region to detect overwrites
+      sig_end_canary:
         CANARY
     #endif
```
jordancarlin: Is there any reason for two back to back canaries at the end of the signature? Seems like we could drop the CANARY at the end of the trap signature and just use the one at the end of the complete signature.

jordancarlin (commented):
Looking pretty good. A few suggestions below. Any idea why the QEMU tests are mismatching?

tests/env/test_setup.h
Zain2050: I was also thinking about removing it.

tests/env/failure_code.h
Zain2050: Yes, this is a good idea. I'll improve the error reporting. 
Currently, I haven't looked at the qemu tests yet, but I'll debug it soon.

Zain2050:
Now mismatches within the trap handler display xepc and the instruction at xepc.

```
RVCP-SUMMARY: Test File "InterruptsSm-00.S": FAILED
RVCP: Test Info: "Mismatch in trap vector signature! Trap was being handled in M-Mode."
RVCP: Instruction: 0x2c013403
RVCP: Approximate address (failure may be slightly after this): 0x0000000080003bb0
RVCP: Register: x8
RVCP: Bad Value:      0x0000000003100e63
RVCP: Expected Value: 0x0000000003100663
RVCP: Address of instruction that trapped (XEPC): 0x0000000080000d08
RVCP: Instruction that trapped: 0x304a9073
```

@jordancarlin The QEMU tests were failing as the trap handler was performing extra sigupds as misa.H is enabled in qemu. Instead of disabling it, I've enclosed all the misa.H checks within `rvtest_htrap_routine`. InterruptsSm is failing on qemu due to some mismatch on WFI. I've disabled it for both qemu and spike.

**ZicsrF** is causing a problem. It has `#define RVTEST_PRIV_TEST` but lacks `#define rvtest_mtrap_routine`. Due to that it doesn't recognize CSR_XEPC. I've put the xpec reporting code within `#ifdef RVTEST_PRIV_TEST`, that's why it fails when `RVTEST_PRIV_TEST` is defined without `rvtest_mtrap_routine`. I can just change the check to #ifdef rvtest_mtrap_routine and it'll work. Should I make this change or should RVTEST_PRIV_TEST be removed from ZicsrF as its unpriv?

jordancarlin:
> Now mismatches within the trap handler display xepc and the instruction at xepc.
> 
> ```
> RVCP-SUMMARY: Test File "InterruptsSm-00.S": FAILED
> RVCP: DEBUG INFORMATION FOLLOWS
> RVCP: Test Info: "Mismatch in trap vector signature! Trap was being handled in M-Mode."
> RVCP: Instruction: 0x2c013403
> RVCP: Approximate address (failure may be slightly after this): 0x0000000080003bb0
> RVCP: Register: x8
> RVCP: Bad Value:      0x0000000003100e63
> RVCP: Expected Value: 0x0000000003100663
> RVCP: Address of instruction that trapped (XEPC): 0x0000000080000d08
> RVCP: Instruction that trapped: 0x304a9073
> RVCP: END OF DEBUG INFORMATION
> ```

Great to include `xepc` and the instruction that trapped. Instead of the generic "mismatch in trap vector signature", and integer register that we are saying mismatched, can we print out which CSR mismatched? Similar to what is done for `fflags` mismatches in floating-point tests where we say `fflags` is the register that mismatched instead of whatever integer register we read it into to compare.

> @jordancarlin The QEMU tests were failing as the trap handler was performing extra sigupds as misa.H is enabled in qemu. Instead of disabling it, I've enclosed all the misa.H checks within `rvtest_htrap_routine`. InterruptsSm is failing on qemu due to some mismatch on WFI. I've disabled it for both qemu and spike.

Great to fix the hypervisor issue. Is InterruptsSm failing in Spike as well as QEMU? If it is only failing for QEMU, let's leave it enabled for Spike. If it is failing for both, then fine to disable both of them for now.

> **ZicsrF** is causing a problem. It has `#define RVTEST_PRIV_TEST` but lacks `#define rvtest_mtrap_routine`. Due to that it doesn't recognize CSR_XEPC. I've put the xpec reporting code within `#ifdef RVTEST_PRIV_TEST`, that's why it fails when `RVTEST_PRIV_TEST` is defined without `rvtest_mtrap_routine`. Was this done intentionally for ZicsrF?

How about putting the xepc reporting code within `#ifdef rvtest_mtrap_routine` instead? Or we can figure out a way to avoid setting `RVTEST_PRIV_TEST` for ZicsrF.

tests/env/failure_code.h
```diff
@@ -408,6 +427,40 @@
...
+        li a1, __riscv_xlen
+        jal failedtest_hex_to_str
+        LA(x9, ascii_buffer)
+        RVMODEL_IO_WRITE_STR(x6, x7, x8, x9)
+    failedtest_report_xepc_instr:
+        # Print instruction at xepc
+        LA(x9, xepcinstrstr)
+        RVMODEL_IO_WRITE_STR(x6, x7, x8, x9)
+        # Check if its a compressed instruction
+        lw a0, 0(a0)        # Load instruction at xepc
```
jordancarlin: This probably needs to do 2 `lh` instead of a single `lw` because the instruction might be misaligned on a processor that doesn't support misaligned loads. See how this is handled above for the regular failing instruction.

Zain2050:
Well the strings do tell us which CSR mismatch except for this trap vector one. This signature is a combination of different values not a single csr. But I guess the comment can be improved.
```

|XLEN-1 31 | 30                              13 | 12 | 11 | 10  6 | 5   2  | 1  0 | 
+----------+------------------------------------+----+----+-------+--------+------+ 
| zeroes   | Xstatus[MPRV,SPVP,MPV,GVA,0,[12:0] |XxIP|XxIE|vector |#entries| mode | 
```

About InterruptsSm, on qemu it's a mismatch on WFI while on spike we're not getting an interrrupt even with an interrupt latency of 2000. Therefore, I've disabled it for both.

jordancarlin:
> Well the strings do tell us which CSR mismatch except for this trap vector one. This signature is a combination of different values not a single csr. But I guess the comment can be improved.
> 
> ```
> 
> |XLEN-1 31 | 30                              13 | 12 | 11 | 10  6 | 5   2  | 1  0 | 
> +----------+------------------------------------+----+----+-------+--------+------+ 
> | zeroes   | Xstatus[MPRV,SPVP,MPV,GVA,0,[12:0] |XxIP|XxIE|vector |#entries| mode | 
> ```

If it is easy to do so, I think it would be helpful to split this into separate checks so that we can indicate exactly which CSR mismatched. If not, then at least include this mapping in the error message so the user can determine which CSR the mismatch corresponds to. I also still think we should drop the integer register name from the error message since it is kind of misleading.

> About InterruptsSm, on qemu it's a mismatch on WFI while on spike we're not getting an interrrupt even with an interrupt latency of 2000. Therefore, I've disabled it for both.

Sounds good to disable both for now.

Zain2050:
> If it is easy to do so, I think it would be helpful to split this into separate checks so that we can indicate exactly which CSR mismatched. If not, then at least include this mapping in the error message so the user can determine which CSR the mismatch corresponds to. I also still think we should drop the integer register name from the error message since it is kind of misleading.

I think its better to leave it this way, because we're interested in specific bits only, and it includes some other trap handler related information as well (vector, #entries, mode). But I'll modify the comment to include this mapping instead of the basic current string.

davidharrishmc:
ZicsrF is intentional to have no trap handler. It will never trap. It
should run on rv20i with no privilege modes so it can’t access privileged
registers.

Zain2050:
Everything is done, except for including the mapping in the error message. I've found some inconsistencies in the signature formation. Me & @UmerShahidengr will discuss and fix that tomorrow.

Now, we are checking if DUT encounters the correct number of traps by comparing the trap sig pointer stored at `Mtrap_sig`. 

Unfortunately, due to this feature some tests have started failing on QEMU.

```
RVCP-SUMMARY: Test File "U-00.S": FAILED
RVCP: Test Info: "Trap signature pointer mismatch! The test likely observed an incorrect number of traps."
RVCP: Instruction: 0x00000013
RVCP: Approximate address (failure may be slightly after this): 0x80006bcc
RVCP: Register: x6
RVCP: Bad Value:      0x8004da70
RVCP: Expected Value: 0x8004da50

```

tests/env/test_setup.h
```diff
@@ -123,10 +123,18 @@
...
-    RVTEST_WORD_PTR "abortstr"
+    bne     T4, T5, check_trap_count
+    jal     T2, failedtest_trap_x7_x9
+    RVTEST_WORD_PTR abort_test
+    RVTEST_WORD_PTR abortstr
+
+    check_trap_count:
+      LA(     T1, Mtrap_sig)
+      LREG    T1, 0(T1)
+      RVTEST_SIGUPD(x2, x5, x4, T1, check_trap_count, trap_count_mismatch)
```
jordancarlin: Instead of recording the final address of the signature pointer, can we record the total number of traps? So subtract the starting address of the signature region from this. This will make debugging more intuitive and should make this check more robust to different memory maps.

Zain2050:
Yes I was going to make this change. I took a look at those failing qemu
tests, and it turns out that the address of signature_base is not the same
for dut and ref due to model specific data. This will fix those.

jordancarlin:
> Yes I was going to make this change. I took a look at those failing qemu tests, and it turns out that the address of signature_base is not the same for dut and ref due to model specific data. This will fix those.

Interesting. I wonder why they are different. We should be matching the QEMU memory map in the `sail.json` config file for the QEMU configs.

allenjbaum:
Trap signatures are variable length of 4,5, or 6 words,  mostly depending
on the hypervisor is enabled for a particular case.
That's why the signature value that concatenates the mode.ie/ip, etc has
the size of the entry embedded in it.
So you can't necessarily look at signature size and divide, you would need
to walk forwards from the base, skipping by the size field, until you reach
the end, and that tells you how many traps there are.
That's a space optimisation, but it a simpler alternative would be to just
have a fixed size and waste some signature space.

jordancarlin:
I was thinking about just reporting the size of the trap region (subtract the initial trap sig pointer from the final trap sig pointer), not the actual number of traps. So the number of entries recorded by a particular trap wouldn't matter. If you think it would be straightforward to report the actual number of traps, that would be helpful for debugging purposes, but is probably not worth it if it would be particularly complex.

Zain2050:
> I was thinking about just reporting the size of the trap region (subtract the initial trap sig pointer from the final trap sig pointer), not the actual number of traps. So the number of entries recorded by a particular trap wouldn't matter. If you think it would be straightforward to report the actual number of traps, that would be helpful for debugging purposes, but is probably not worth it if it would be particularly complex.

With the current approach, its not possible to output the number of traps. As @allenjbaum mentioned, we write different number of signatures depending upon whether hypervisor is enabled or not, or if its an interrupt. Therefore, just calculating a delta between mtrap_sigptr (start) & Mtrap_sig (points to the next free slot in the sig region). Perhaps we can add this functionality in future.

Zain2050:
> Interesting. I wonder why they are different. We should be matching the QEMU memory map in the sail.json config file for the QEMU configs.

The memory maps match, but some stuff gets pushed down due to `RVMODEL_DATA_SECTION` and similary `RVMODEL_IO_WRITE_STR` in exit_cleanup changes address of `INSTANTIATE_MODE_MACRO RVTEST_TRAP_HANDLER`.

jordancarlin:
> > Interesting. I wonder why they are different. We should be matching the QEMU memory map in the sail.json config file for the QEMU configs.
> 
> The memory maps match, but some stuff gets pushed down due to `RVMODEL_DATA_SECTION` and similary `RVMODEL_IO_WRITE_STR` in exit_cleanup changes address of `INSTANTIATE_MODE_MACRO RVTEST_TRAP_HANDLER`.

Makes sense. We might want to think about a way of keeping things more consistent independent of the RVMODEL macros.

tests/env/failure_code.h
```diff
@@ -587,6 +642,70 @@
...
+    trap_sig_offset_mismatch:
+        .string "\"Mismatch in trap signature pointer offset! The test likely observed an incorrect number of traps.\"";
+    sv_Mvect_str:
+        .string "\"Mismatch in trap vector signature! Trap was being handled in M-Mode.\"";
+    sv_Sect_str:
+        .string "\"Mismatch in trap vector signature! Trap was being handled in S-Mode.\"";
+    sv_Hvect_str:
+        .string "\"Mismatch in trap vector signature! Trap was being handled in HS-Mode.\"";
+    sv_Vvect_str:
+        .string "\"Mismatch in trap vector signature! Trap was being handled in VS-Mode.\"";
```
jordancarlin, line 658: Looks like the only thing left is clarifying this error message about what the trap vector signature contains and then I think this is good to merge.

Zain2050: Yes, just waiting on @UmerShahidengr for this.

Zain2050: The code for bits extraction & placement  is actually developed by @allenjbaum. Me & @UmerShahidengr are going to further discuss this with him on Friday before making any change. That will change some code in the trap handler and the comments will be written accordingly. That can be pushed in a separate PR. I am going to open up a issue for it.

UmerShahidengr (approved):
LGTM. Good work @Zain2050

Issue #934, opened 2026-02-14
ACT4 now requires MTIME and MTIMECMP


MikeOpenHWGroup:
As of `c4bd176` (the current head of the act4 branch), the ACTs require MTIME and MTIMECMP, irregardless of whether or not the requested tests exercise them.  The following command:
 
```
$ make CONFIG_FILES=config/cores/cve2/cv32e20/test_config.yaml  EXTENSIONS=I,M
```

Produces the following error:
```
   55 |   #error "RVMODEL_MTIME_ADDRESS not defined. Make sure to define it in rvmodel_macros.h."
      |    ^~~~~
/home/mike/GitHubRepos/MikeOpenHWGroup/riscv-arch-test/act4/tests/env/check_defines.h:59:4: error: #error "RVMODEL_MTIMECMP_ADDRESS not defined. Make sure to define it in rvmodel_macros.h."
   59 |   #error "RVMODEL_MTIMECMP_ADDRESS not defined. Make sure to define it in rvmodel_macros.h."
      |    ^~~~~
```

MTIME and MTIMECMP CSRs are _platform_ resources and should be optional for Sm certification of a core.

jordancarlin:
As long as you’re not running the Sm test, it should be possible to define them as empty macros (like what is done for the interrupt macros). Are you getting that error even when they are defined to be empty?

MikeOpenHWGroup:
My rvmodel_macros.h was out-of-date.  It defined `RVMODEL_MTIME_ADDR` and `RVMODEL_MTIMECMP_ADDR`, rather than `RVMODEL_MTIME_ADDRESS` and `RVMODEL_MTIMECMP_ADDRESS`.

Fixing that allows the test generation to proceed, but errors out on the `Sm` tests:
```
/home/mike/GitHubRepos/MikeOpenHWGroup/riscv-arch-test/act4/tests/priv/Sm/Sm-00.S:49784: Error: illegal operands `la x24,'
/home/mike/GitHubRepos/MikeOpenHWGroup/riscv-arch-test/act4/tests/priv/Sm/Sm-00.S:49795: Error: illegal operands `la x24,'
```

Appending EXCLUDE_EXTENSIONS=Sm to the `make` command run as expected.

jordancarlin:
That makes sense and is a known issue (let’s leave this open to track it). Right now the Sm tests always assume MTIME is available, but it seems like that is going to need to change.

karabambus:
[...]

Issue #15, opened 2018-11-07
fix for ma_fetch


neelgala:
The ma_fetch test has been written in a way that implementations with and without "C" support should generate the same signature. The aim is to avoid 
```
j fail
```
at multiple places. 

However, currently the signatures generated by implementation with and without C support are different. This is because the trap handler sets the return address in mepc to point to 

```
SWSIG(?, TESTNUM)
```
instead of 

```
li TESTNUM, ?
```

[This](https://github.com/riscv/riscv-compliance/pull/14) merge request fixes that and also updates the signature. Has been verified with e-class, riscvOVPsim and spike.

neelgala:
Closed by #14

PR #1031, opened 2026-03-03
Add initial CV32E40P configuration

karabambus:
Configuration for CV32E40P core (Phase 1 CTP research) with all parameters documented and verified against manual.

Baseline configuration: COREV_PULP=0, FPU=0, NUM_MHPMCOUNTERS=1

ISA Extensions:
- Base: I, M, C
- Standard: Zca, Zicsr, Zifencei, Zicntr
- Supervisor: Sm, Smhpm

Parameters: 44 documented with verification status
- 21 VERIFIED (with manual citations)
- 7 FROM MANUAL
- 13 ASSUMED (reasonable defaults)
- 2 QUESTIONABLE (flagged for review)
- 1 UNKNOWN

Research notes and atomic concepts in separate Obsidian vault.


config/cores/cve2/cv32e40p/cv32e40p.yaml
```diff
@@ -0,0 +1,155 @@
...
+  MTVEC_ACCESS: rw                       # VERIFIED — control_status_registers.rst:577 MODE[0] is RW
+  MTVEC_ILLEGAL_WRITE_BEHAVIOR: retain   # ASSUMED — standard trap vector behavior
+  MTVEC_BASE_ALIGNMENT_VECTORED: 256     # VERIFIED — control_status_registers.rst:567 "256-byte aligned"
+  MTVEC_BASE_ALIGNMENT_DIRECT: 256       # VERIFIED — control_status_registers.rst:567 same alignment for both modes
+
+  # Sm params (Section 9 — Sm Parameters)
+  PRECISE_SYNCHRONOUS_EXCEPTIONS: true   # ASSUMED — standard RISC-V Priv Spec § 3.3 (manual doesn't explicitly state)
+  MARCHID_IMPLEMENTED: true              # VERIFIED — control_status_registers.rst: marchid (0xF12) is implemented
+  MIMPID_IMPLEMENTED: true               # VERIFIED — control_status_registers.rst: mimpid (0xF13) is implemented
+  PMA_GRANULARITY: 3                     # QUESTIONABLE — M-mode only, no MMU/TLB; parameter required by UDB but unused
```
davidharrishmc: Do you need a parameter about the number of PMP entries supported (presumably 0)?

config/cores/cve2/cv32e40p/cv32e40p.yaml
```diff
@@ -0,0 +1,155 @@
...
+  # CSR / implementation IDs (Section 3 — Implementation IDs)
+  ARCH_ID_VALUE: 0x00000004             # FROM MANUAL — control_status_registers.rst: marchid=0x4
+  IMP_ID_VALUE: 0x0                     # FROM MANUAL — core_versions.rst: mimpid=0 when COREV_PULP=0, FPU=0
+  VENDOR_ID_BANK: 0xC                   # FROM MANUAL — mvendorid 0x602 decoded: bits[31:7]=0xC (JEDEC bank 13)
+  VENDOR_ID_OFFSET: 0x2                 # FROM MANUAL — mvendorid 0x602 decoded: bits[6:0]=0x2 (manufacturer byte)
+
+  # MTVEC (Section 4 — mtvec WARL)
+  MTVEC_MODES: [0, 1]                    # VERIFIED — control_status_registers.rst:589 "both direct and vectored supported"
+  MTVEC_ACCESS: rw                       # VERIFIED — control_status_registers.rst:577 MODE[0] is RW
+  MTVEC_ILLEGAL_WRITE_BEHAVIOR: retain   # ASSUMED — standard trap vector behavior
```
davidharrishmc: Are there any illegal values in MTVEC?  Worth looking at the RTL on this one.

karabambus: I will try to confirm assumed values that is my next task for completion of this draft.

config/cores/cve2/cv32e40p/cv32e40p.yaml
```diff
@@ -0,0 +1,155 @@
...
+  M_MODE_ENDIANNESS: little              # VERIFIED — intro.rst line 77 (little-endian architecture)
+  #U_MODE_ENDIANNESS: little
+
+  # trap/exception behavior (Section 5 — Trap & Exception Behavior)
+  #TRAP_ON_ECALL_FROM_U: true
+  TRAP_ON_ECALL_FROM_M: true             # VERIFIED — exceptions_interrupts.rst:122 "cannot be disabled"
+  TRAP_ON_EBREAK: true                   # VERIFIED — exceptions_interrupts.rst:117 "always triggers"
+  TRAP_ON_UNIMPLEMENTED_INSTRUCTION: true # VERIFIED — exceptions_interrupts.rst:122–125 "cannot be disabled"
+  TRAP_ON_RESERVED_INSTRUCTION: true     # ASSUMED — implied by illegal instruction handling
+  TRAP_ON_UNIMPLEMENTED_CSR: false       # ASSUMED — standard RISC-V: undefined CSRs return 0, no trap
```
jordancarlin: Whether or not this is true for the cv32e40p, it is definitely not “standard RISC-V”. Most cores and simulators trap.

config/cores/cve2/cv32e40p/cv32e40p.yaml
karabambus: I think I confused this parameter with pmp. I will edit this comment, still need to verify it since i took it from cv32e20 config.

Deleted previous comment since I saw my mistake.

karabambus:
@MikeOpenHWGroup  Should I remove v.1.0.0 with FPU enabeled. In manual mimpid is hardwired to 0, as well mstatus.FS? It seems to me this would fail compliance tests.

davidharrishmc:
I don't see why mimpid matters.  Can't you just set it to the same value (0) in the Sail YAML, and then it will pass?

If mstatus.FS is hardwired to 0, none of the floating-point should run.  This is a major deviation from spec if CV32E40P has a FPU and mstatus but hardwires mstatus.FS to 0.

MikeOpenHWGroup:
> If mstatus.FS is hardwired to 0, none of the floating-point should run. This is a major deviation from spec if CV32E40P has a FPU and mstatus but hardwires mstatus.FS to 0.

The CV32E40P can be configured to support, or not, the 'F' ISA via a top-level instantiation parameter (cleverly called `FPU`).  When FPU is not set, the floating point unit is not compiled/synthesised, the decoder treats 'F' instructions as illegal instructions, misa.F is hardwired to 1'b0 and mstatus.FS is hardwired to 2'b00.  I believe this is compliant to both the ISA and [User Manual](https://docs.openhwgroup.org/projects/cv32e40p-user-manual/en/latest/control_status_registers.html#machine-status-mstatus) and that the CV32E40P RTL faithfully implements the above.  We shall soon know the answer to both!

In the CV32E40P, the configuration where FPU==0 is called the v1.0.0 release.  Setting FPU==1 is informally called the "v2" (even though we have yet to formally release a v2.0.0 version of that core.  In any case, it's the same RTL if you clone from the head of the repo.

MikeOpenHWGroup:
> I don't see why mimpid matters.

In the CV32E40P, the mimpid is always read-only and its value is determined by its top-level instantiation parameters ([link](https://docs.openhwgroup.org/projects/cv32e40p-user-manual/en/latest/control_status_registers.html#machine-implementation-id-mimpid)).

The only normative rule regarding mimpid is [mimpid_always_rd](https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mimpid_always_rd).  Given that, its value does not matter for the certification tests.  As @davidharrishmc suggests, setting it to the same value (0) in the Sail YAML should work.

karabambus:
@MikeOpenHWGroup  I might have got this wrong. I created yaml for cv32p40 v1.0.0 as initial version of this core - mstatus.FS is hardwired to 0 is v1.0.0 implementation with FPU=1, this is different for latest version 1.8.3 that has compliant implementation with FPU=1.

davidharrishmc:
I think you want mstatus.FS hardwired to 0 when FPU=0, not 1.
Otherwise programs won’t be able to use the FPU.

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/sail.json
```diff
@@ -217,7 +216,6 @@
       "supported": true
```
MikeOpenHWGroup, line 218: There are no configurations of the CV32E40P that support double precision floating point (D ISA).

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/sail.json
```diff
@@ -111,7 +111,6 @@
       "supported": false
     },
     "F": {
-      
       "supported": true
     },
     "D": {
```
MikeOpenHWGroup, line 118: Not support - `D` should be `false`.

MikeOpenHWGroup (changes requested):

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/sail.json
karabambus: D is false.

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/link.ld
jordancarlin, line 1: Please update this to match the latest linker scripts for other configs that are already in the repo to avoid the entry point address bug.

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/rvmodel_macros.h
```diff
@@ -0,0 +1,109 @@
...
+        .align 8; .global tohost; tohost: .dword 0;         \
+        .align 8; .global fromhost; fromhost: .dword 0;     \
+        .popsection;
+
+##### STARTUP #####
+
+# Perform boot operations. Can be empty.
+# cv32e40p supports both direct (MODE=0) and vectored (MODE=1) mtvec.
+# Direct mode (MODE=0) requires 4-byte alignment only — use .align 2 before handler.
+#define RVMODEL_BOOT                                                     \
```
jordancarlin: Why are you loading a custom trap handler as part of the boot code? The tests use their own trap handler.

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/rvmodel_macros.h
jordancarlin, line 1: For files that are the same between the different variants of the cv32e40, I’d rather use symlinks so we don’t need to maintain multiple copies.

jordancarlin (changes requested):

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/cv32e40p_v1.0.0_rv32imc.yaml
```diff
@@ -0,0 +1,158 @@
...
+  HPM_COUNTER_EN: # VERIFIED — cv32e40p_cs_registers.sv: NUM_MHPMCOUNTERS=1 default (integration.rst:112); only counter index 3 active; UDB requires indices 0-2 false
+    [
+      false, # index 0 (CY — must be false per UDB schema)
+      false, # index 1 (TM — must be false per UDB schema)
+      false, # index 2 (IR — must be false per UDB schema)
+      true, # index 3 (MHPMCOUNTER3 — only active counter with NUM_MHPMCOUNTERS=1)
+      false,
+      false,
+      false,
+      false, # 4..7 (unimplemented, increment hardwired 0)
```
MikeOpenHWGroup: This `4..7` is a hold-over from an earlier version of this file when I tried to organize the dictionary in byte-wise fashion.  The formatter does not like that, so we are stuck with bit-wise.  Given that, let's update the comments to be `# 7` ... `#15`, etc.

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/cv32e40p_v1.0.0_rv32imc.yaml
```diff
@@ -0,0 +1,158 @@
...
+  COUNTINHIBIT_EN: # VERIFIED — cv32e40p_cs_registers.sv: indices 0,2,3 inhibitable via mcountinhibit_q register; index 1 always false per UDB; unimplemented counters hardwired 0
+    [
+      true, # index 0 (MCYCLE inhibit — mcountinhibit_q[0] gates cycle counter)
+      false, # index 1 (TM — always false, mcountinhibit.TM is read-only-0 per UDB)
+      true, # index 2 (MINSTRET inhibit — mcountinhibit_q[2] gates instret counter)
+      true, # index 3 (MHPMCOUNTER3 inhibit — mcountinhibit_q[3] gates hpmcounter3)
+      false,
+      false,
+      false,
+      false, # 4..7 (unimplemented counters, no inhibit effect)
```
MikeOpenHWGroup: Same comment as before (`#4..7` s.b. `#7`, etc.)

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/cv32e40p_v1.0.0_rv32imc.yaml
```diff
@@ -0,0 +1,158 @@
...
+      false,
+      false,
+      false,
+      false,
+      false,
+      false, # 24..31
+    ]
+  COUNTINHIBIT_EN: # VERIFIED — cv32e40p_cs_registers.sv: indices 0,2,3 inhibitable via mcountinhibit_q register; index 1 always false per UDB; unimplemented counters hardwired 0
+    [
+      true, # index 0 (MCYCLE inhibit — mcountinhibit_q[0] gates cycle counter)
```
MikeOpenHWGroup, line 122: In this context, does `true` mean (or imply) the power-on-reset value of the CSR's bit?  In the [CV32E40P User Manual](https://docs.openhwgroup.org/projects/cv32e40p-user-manual/en/latest/control_status_registers.html#machine-counter-inhibit-register-mcountinhibit), this bit is spec'ed to have a PoR of 1'b0, and its access mode is RW.  This is compatible with the WARL field specification.

Question for @jordancarlin, given that the Privilege spec does not specify a PoR for this CSR, how do we specify it for the ACTs?  Given that the unpriv spec 20190608 (v1.11), the version the CV32E40P is verified to, does not specify a PoR, a WARL legal value of 1'b1 (inhibit) was assigned.

jordancarlin: Per https://github.com/riscv/riscv-unified-db/blob/5f9181a37f8fe0f4be1503674649fab9ffa62832/spec/std/isa/param/COUNTINHIBIT_EN.yaml, `true` indicates that the counter CAN be inhibited. It has nothing to do with the reset value.

Reset values are a separate issue that will need to be figured out, as discussed in #1103.

karabambus:
added requested comment changes @MikeOpenHWGroup , added symlink for redundant files @jordancarlin . Reset values will be added when we get that feature. I didnt put any changes to rvmodel_boot.

jordancarlin:
@karabambus @MikeOpenHWGroup, what's the status of this? Is it ready for a review/merge?

karabambus:
> @karabambus @MikeOpenHWGroup, what's the status of this? Is it ready for a review/merge?

This is ready, though we have #1103 that is blocking - I can work around that by reseting mcountinhibit with RVMODEL_BOOT macro, but I would rather wait for this issue to be fixed.

@jordancarlin I used symlinks. Would you like me to change it to target files in configuration folder of core, or you are ok with this solution where redundant files are in core folder?

MikeOpenHWGroup:
I'd rather not be gated by #1103 as I fear a full resolution of that Issue may take some time.

@davidharrishmc, @karabambus and @jordancarlin, if you are comfortable with "known failures" that are explicitly tied to an open Issue being merged in, this PR is ready to go.

davidharrishmc:
Strongly agree that we don’t gate PRs with open issues about tool chains.

I suport the merge.  However, as I just posted in #1103, I also believe the right solution is for CV32E40P to initialize mcountinhibit=0 in RVMODEL_BOOT, because there is no other place this can be done reliably.

https://github.com/riscv/riscv-arch-test/issues/1103

David

karabambus:
Added requested change to rvmodel_boot macro

davidharrishmc (commented):
Can you do this as

csrwi mcountinhibit, 0

instead of a .word?

karabambus:
> Can you do this as
> 
> csrwi mcountinhibit, 0
> 
> instead of a .word?

ok this is another issue. Tests have their own march, for I there is -march=rv32i. binutils 2.38+ requires explicit _zicsr extension in march. This results in illegal instructions while running basic I tests.

my toolchain versions are following GCC 15.2.0, Binutils 2.46. This should be correct according to readme. 

I will open issue asap.

config/cores/cv32e40p/rvmodel_macros.h
```diff
@@ -0,0 +1,82 @@
...
+        .popsection;
+
+##### STARTUP #####
+
+# Perform boot operations.
+# CV32E40P resets mcountinhibit=0xd (all counters inhibited). Clear it so
+# cycle/instret increment as Zicntr tests expect.
+# Raw encoding: .word 0x32005073 = csrwi mcountinhibit, 0
+#define RVMODEL_BOOT \
+  .word 0x32005073 ;
```
MikeOpenHWGroup: Please use assembly code here.

MikeOpenHWGroup:
1. To compile test-programs for the CV32E40P (v1), `-march` must be set to `rv32imc_zicsr_zicntr_zifencei`.  Fortunately, this is compatible with the Optional Extensions and Dependent Extensions sections of the CRD.  That is a happy coincidence as the v1 release claims to implement the 20190608 (v1.11) release of the unpriv spec.
2. We discussed your GCC toolchain last week, so you _should_ be all set.  If you run into problems, please install and configure your toolchain from source as per the ACT instructions. (It takes a while.)

One last thing: unless it is absolutely necessary to do otherwise, please use real instructions in your code.  Avoid using pseudo-instructions and machine code (.word).

davidharrishmc:
This is a great question about CSR instructions in RVMODEL_BOOT that discovers a flaw in the framework.  The framework will compile I tests without march containing Zicsr because it thinks it is compiling a test program with just I instructions.

@jordancarlin how do you think we should handle this when the macros have to contain instructions that are outside the extensions required to run the tests themselves?  Do we need a RVMODEL_SUPPLEMENTAL_EXTENSIONS field passed to the march string?

karabambus:
> 1. To compile test-programs for the CV32E40P (v1), `-march` must be set to `rv32imc_zicsr_zicntr_zifencei`.  Fortunately, this is compatible with the Optional Extensions and Dependent Extensions sections of the CRD.  That is a happy coincidence as the v1 release claims to implement the 20190608 (v1.11) release of the unpriv spec

Tests have their own march, this is not something I can't just set since it is generated. As David said:  

> This is a great question about CSR instructions in RVMODEL_BOOT that discovers a flaw in the framework. The framework will compile I tests without march containing Zicsr because it thinks it is compiling a test program with just I instructions.

Can we try push this with .word workaround @MikeOpenHWGroup @jordancarlin @davidharrishmc , or should I wait for resolution of #1131 ?

jordancarlin:
@karabambus great discovery regarding the march string. While I work on a more general solution, you can work around this in your RVMODEL_BOOT macro by doing the following:

```
.option push
.option arch, +zicsr
# instructions using Zicsr here
.option pop
```

This will enable the Zicsr extension between the .option directives so you can use `csrwi mcountinhibit, 0` instead of `.word`.

jordancarlin (changes requested):
This is looking very close. Can you add a short README to the main cv32e40p directory with a very brief description of the core and a link to its source GitHub repository? It also might be worth including a very brief description of each variant because the differences are not immediately obvious between v1.0.0 and v1.8.3.

config/cores/cv32e40p/README.md
```diff
@@ -0,0 +1,30 @@
...
+| Config | ISA | Notes |
+|--------|-----|-------|
+| `cv32e40p_v1.0.0_rv32imc` | RV32IMC | Original v1.0.0 release |
+| `cv32e40p_v1.8.3_rv32imc` | RV32IMC | Latest release, integer-only |
+| `cv32e40p_v1.8.3_rv32imcf` | RV32IMCF | Latest release with FPU (F + Zcf) |
+
+v1.0.0 and v1.8.3 (without FPU) should behave identically for integer
+instructions. The v1.8.3 release added the optional FPU and PULP custom
+extensions but did not change the base RV32IMC behavior. Both configs
+exist so that certification can be run against either RTL version.
```
jordancarlin: @MikeOpenHWGroup I'd like your opinion on this since you are more familiar with the core, but if both versions have identical configs, is there a benefit to having separate configs for both versions? Seems like you should be able to use a single config and run the tests on both v1.0.0 and v1.8.3.

Maybe we should just have two configs for cv32e40p, an imc one and an imcf one. If you think it is beneficial to have both, then that is fine with me too.

MikeOpenHWGroup: Yeah, its not obvious what is going on here.  The CV32E40P v1.0.0 release implements RV32IMC.  The v1.8.3 release can be configured to implement RV32IMC configuration.  But they are not the same core (although they are very similar).

Since some organizations have v1.0.0 in production we want to have a specific config for that specific release.  So we'd like to keep all three.  Once this PR is in, I'll add a link to the User Manual that explains the distinction in more detail.

jordancarlin (approved):

config/cores/cv32e40p/README.md
jordancarlin: Great. Sounds good to have both versions as long as we add that link later explaining them.

jordancarlin:
Will merge once @MikeOpenHWGroup  approves.

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/cv32e40p_v1.0.0_rv32imc.yaml
```diff
@@ -0,0 +1,155 @@
+# Copyright (c) 2026, 10xEngineers, Harvey Mudd College and OpenHW Foundation
```
MikeOpenHWGroup: Change copyright to `... and Eclipse Foundation`.

config/cores/cv32e40p/README.md
```diff
@@ -0,0 +1,23 @@
+<!--
+Copyright (c) 2026, OpenHW Foundation
```
MikeOpenHWGroup: Please attribute the copyright to Eclipse Foundation (the OpenHW Foundation is not an independent legal entity and cannot hold a copyright).

config/cores/cv32e40p/README.md
```diff
@@ -0,0 +1,23 @@
...
+--->
+
+## DUT Configuration for the CV32E40P
+
+Three configurations are provided:
+
+| Config                     | ISA      | Notes                             |
+| -------------------------- | -------- | --------------------------------- |
+| `cv32e40p_v1.0.0_rv32imc`  | RV32IMC  | Original v1.0.0 release           |
+| `cv32e40p_v1.8.3_rv32imc`  | RV32IMC  | Latest release, integer-only      |
```
MikeOpenHWGroup: "integer-only" is not strictly correct as that would be RV32I.  How about, "logically equivalent to v1.0.0_rv32imc"?

config/cores/cv32e40p/rvmodel_macros.h
```diff
@@ -0,0 +1,86 @@
...
+# Address to use for load/store fault tests that should cause an access fault on the DUT.
+#define RVMODEL_ACCESS_FAULT_ADDRESS 0x00000000
+
+##### TERMINATION #####
+
+# Terminate test with a pass indication.
+#define RVMODEL_HALT_PASS  \
+  li x1, 123456789                ;\
+  li x2, 0x20000000       ;\
+  write_tohost_pass:      ;\
```
MikeOpenHWGroup: A label with the form `*_tohost_*`  is dangerous as people might confuse this with the [Berkeley Host Target Interface (HTIF)](https://defermelowie.github.io/post/riscv-htif/) and our implementation of the RVMODEL macros has no relationship to the HTIF.

I recommend `write_halt_pass`

config/cores/cv32e40p/rvmodel_macros.h
```diff
@@ -0,0 +1,86 @@
...
+    sw x1, 0(x2)          ;\
+    sw x0, 4(x2)          ;\
+  self_loop_pass:         ;\
+    j self_loop_pass      ;\
+
+# Terminate test with a fail indication.
+#define RVMODEL_HALT_FAIL \
+  li x1, 1                ;\
+  li x2, 0x20000000       ;\
+  write_tohost_fail:      ;\
```
MikeOpenHWGroup: `write_halt_fail`

config/cores/cv32e40p/rvmodel_macros.h
```diff
@@ -0,0 +1,86 @@
...
+2:                           ;                        \
+  li   _R2, 0x10000000       ; /* virtual printer */  \
+  sw   _R1, 0(_R2)           ;                        \
+  addi _STR_PTR, _STR_PTR, 1 ; /* Next char */        \
+  j 1b                       ; /* Loop */             \
+3:
+
+##### Machine Timer #####
+
+#define RVMODEL_MTIME_ADDRESS    0x0200BFF8
```
MikeOpenHWGroup: The CV32E40P does not support either `MTIME` or `MTIMECMP`.

config/cores/cv32e40p/rvtest_config.h
```diff
@@ -0,0 +1,3 @@
+#define ACCESS_FAULT_ADDRESS 0x00000000
```
MikeOpenHWGroup: How is this different than `RVMODEL_ACCESS_FAULT_ADDRESS` defined in rvmodel_macros.h?

jordancarlin: It's not and should be deleted. Looks like a carryover from an earlier version.

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/sail.json
```diff
@@ -0,0 +1,343 @@
+{
```
MikeOpenHWGroup, line 3: Please add the two-line license header:
// Copyright (c) 2026 Eclipse Foundation
// SPDX-License-Identifier: Apache-2.0

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/sail.json
```diff
@@ -0,0 +1,343 @@
...
+    "hartid": 0,
+    "cache_block_size_exp": 6,
+    "reservation_set_size_exp": 3,
+    "clint": {
+      "base": 33554432,
+      "size": 786432
+    },
+    "clock_frequency": 1000000000,
+    "instructions_per_tick": 2,
+    "wfi_is_nop": true
```
MikeOpenHWGroup: Not true, see the [Sleep Unit](https://docs.openhwgroup.org/projects/cv32e40p-user-manual/en/cv32e40p_v1.8.3/sleep.html#wfi) chapter of the CV32E40P user Manual.

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/cv32e40p_v1.8.3_rv32imcf.yaml
```diff
@@ -0,0 +1,165 @@
...
+  COUNTINHIBIT_EN: # VERIFIED — cs_registers.sv:1421-1437; indices 0,2,3 inhibitable via mcountinhibit_q; index 1 always false per UDB; unimplemented counters hardwired 0
+    [
+      true, # index 0 (MCYCLE inhibit — cs_registers.sv:1423-1425: !mcountinhibit_q[0])
+      false, # index 1 (TM — always false, mcountinhibit.TM is read-only-0 per UDB)
+      true, # index 2 (MINSTRET inhibit — cs_registers.sv:1428-1431: !mcountinhibit_q[2])
+      true, # index 3 (MHPMCOUNTER3 inhibit — cs_registers.sv:1434-1437: !mcountinhibit_q[3])
+      false, # 4
+      false,
+      false,
+      false, # 4..7
```
MikeOpenHWGroup: \# 7

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/cv32e40p_v1.8.3_rv32imcf.yaml
```diff
@@ -0,0 +1,165 @@
...
+      false,
+      false, # 16..23
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false, # 4..31 (unimplemented)
```
MikeOpenHWGroup: \# 24..31

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/cv32e40p_v1.8.3_rv32imcf.yaml
```diff
@@ -0,0 +1,165 @@
...
+  HPM_COUNTER_EN: # VERIFIED — cs_registers.sv:41 NUM_MHPMCOUNTERS=1 default; line 1432: only index 3 active when NUM_MHPMCOUNTERS=1; UDB requires first 3 entries false
+    [
+      false, # index 0 (CY — must be false per UDB schema)
+      false, # index 1 (TM — must be false per UDB schema)
+      false, # index 2 (IR — must be false per UDB schema)
+      true, # index 3 (MHPMCOUNTER3 — only active counter, cs_registers.sv:1432)
+      false, # 4
+      false,
+      false,
+      false, # 4..7
```
MikeOpenHWGroup: \# 7

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/cv32e40p_v1.8.3_rv32imcf.yaml
```diff
@@ -0,0 +1,165 @@
...
+      false,
+      false, # 16..23
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false, # 4..31 (unimplemented)
```
MikeOpenHWGroup: \# 24..31

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imc/cv32e40p_v1.8.3_rv32imc.yaml
```diff
@@ -0,0 +1,156 @@
...
+      false,
+      false, # 16..23
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false, # 4..31 (unimplemented)
```
MikeOpenHWGroup: \# 24..31

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/cv32e40p_v1.0.0_rv32imc.yaml
```diff
@@ -0,0 +1,155 @@
...
+      false,
+      false, # 16..23
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false,
+      false, # 4..31 (unimplemented)
```
MikeOpenHWGroup: \# 24..31

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/cv32e40p_v1.0.0_rv32imc.yaml
```diff
@@ -0,0 +1,155 @@
...
```
MikeOpenHWGroup: \# 24..31

MikeOpenHWGroup (changes requested):
OK, I sense we are _almost_ ready to merge this.  Hi @karabambus, I have made a few additional comments, please have a look.  The majority are either license header or suggestions to improve comments.

config/cores/cv32e40p/rvmodel_macros.h
karabambus: This was copied over from cv32e20 rvmodel macros should we do the same for cv32e20?

config/cores/cv32e40p/rvmodel_macros.h
karabambus: This is #934 issue, and if I remove this lines, build will fail on sm tests (make CONFIG_FILES= how it is described in README). What do you think about applying @jordancarlin [solution](https://github.com/riscv/riscv-arch-test/issues/1135#issuecomment-4140522435) on this config. This should skip privileged tests. It should be also explained in README and comments on test_config.yaml. Alternative is that configs wont run without specifying additional parameters (i.e. EXTENSIONS=), and this is not really documented in README .

config/cores/cv32e40p/rvtest_config.h
karabambus: This change should be also made to cv32e20 rvtest_config.h

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/sail.json
karabambus: How does this affect tests? This change had no effect on passing and failing (when comparing signatures with dut). @jordancarlin

config/cores/cv32e40p/rvmodel_macros.h
MikeOpenHWGroup: Good catch @karabambus, I should have caught that.  Let's fix it on a separate PR.

config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imcf/sail.json
MikeOpenHWGroup: WFI is a privileged instruction and is not part of certification testing for RVI20.  (So its a problem for "later".)

jordancarlin: It will impact the interrupts tests (InterruptsSm is already in the repo and available).

karabambus:
I have done flollowing changes
- README: describe v1.8.3_rv32imc as logically equivalent to v1.0.0_rv32imc
- sail.json: set wfi_is_nop=false for v1.8.3_rv32imc and v1.8.3_rv32imcf (WFI is not a NOP on CV32E40P)
- v1.8.3_rv32imcf.yaml: fix HPM_COUNTER_EN comment from 4..31 to 24..31
- rvmodel_macros.h: use write_halt_pass/fail labels, add .option arch +zicsr for csrwi, document MTIME/MTIMECMP absence
- rvtest_config.h: remove stale RVMODEL_ACCESS_FAULT_ADDRESS
- Copyright, license headers, and YAML bit-range comments updated per review

MikeOpenHWGroup (approved):
Thanks @karabambus, looks good.

Hi @jordancarlin, this is ready to merge.

config/cores/cve2/cv32e20/rvmodel_macros.h
```diff
@@ -67,9 +67,9 @@
 
 ##### Machine Timer #####
 
-#define RVMODEL_MTIME_ADDRESS  /* unimplemented */
+#define RVMODEL_MTIME_ADDRESS   0x0200bff8
 
-#define RVMODEL_MTIMECMP_ADDRESS   /* unimplemented */
```
jordancarlin, line 72: @karabambus @MikeOpenHWGroup Why is this changing the MTIME macros for CV32E20?

karabambus: sorry jordan, I made mistake on commit

PR #1034, opened 2026-03-04
Add comparison operators for test param constraints

jordancarlin:
Tests can now specify that a DUT parameter must be >=, <=, >, <, !=, or == to a value for a test to be selected.

This will be needed to get PMP tests working correctly.

### DIFF
--- a/framework/src/act/act.py
+++ b/framework/src/act/act.py
@@ -76,10 +76,13 @@ def run_act(
         selected_tests = select_tests(
             full_test_dict, implemented_extensions, config_params, include_priv_tests=config.include_priv_tests
         )
+        mxlen = config_params["MXLEN"]
+        if not isinstance(mxlen, int):
+            raise TypeError(f"MXLEN must be an integer, got {type(mxlen)}: {mxlen!r}")
         configs.append(
             {
                 "config": config,
-                "xlen": config_params["MXLEN"],
+                "xlen": mxlen,
                 "e_ext": "E" in implemented_extensions,
                 "selected_tests": selected_tests,
             }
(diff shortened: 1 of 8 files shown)


framework/src/act/act.py
```diff
@@ -76,10 +76,13 @@ def run_act(
         selected_tests = select_tests(
             full_test_dict, implemented_extensions, config_params, include_priv_tests=config.include_priv_tests
         )
+        mxlen = config_params["MXLEN"]
+        if not isinstance(mxlen, int):
```
davidharrishmc, line 80: Should this be restricted to 32, 64, 128 instead of integer?

framework/src/act/select_tests.py
```diff
@@ -7,25 +7,62 @@
...
-def check_test_params(test_params: dict[str, Any], config_params: dict[str, Any]) -> bool:
+# Parameter constraint comparison operators
+_COMPARISON_RE = re.compile(r"^(>=|<=|!=|==|>|<)\s*(\d+)$")
+
+
+def _compare_param(test_value: object, config_value: object) -> bool:
+    """Compare a test parameter requirement against a config parameter value.
+
+    Supports comparison operator prefixes on string values:
+    '>=128', '<= 64', '> 0', '<256', '!=0', '==64'.
```
davidharrishmc: e.g.

davidharrishmc (changes requested):

framework/src/act/act.py
jordancarlin: This is coming from UDB, which has already validated that it is a valid config.

framework/src/act/select_tests.py
```diff
@@ -7,25 +7,62 @@
...
+    Falls back to exact equality for all other values.
```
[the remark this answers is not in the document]
jordancarlin: Added support for hex values. Deferring strings until they are actually needed.

davidharrishmc (approved):

PR #1036, opened 2026-03-04
Refactor CRD and CTP Generation

jordancarlin:
- Check CRD in CI regression
- Build and release CRD when it changes (similar to what is done already for the CTP)
- Generate CTP adoc files in build directory instead of src directory
- Share common adoc files between CRD and CTP
- Update CTP Makefile to use updated flow from upstream ISA manual


jordancarlin:
cc @james-ball-qualcomm

davidharrishmc (approved):

PR #1040, opened 2026-03-05
Added coverage file for Ssnpm extension of pointer masking.

Ammarahwakeel:
I am submitting this as a draft PR since the pointer masking test plan is currently under review. Once the test plan is finalized, I may need to update or refine some of the coverpoints accordingly.
There is a minor issue I am currently working through with "cp_mask_priv_mode_only" , specifically related to adding bins for different XLEN patterns of virtual and physical addresses.

In the meantime, I would greatly appreciate an initial review of these files to ensure the overall structure and approach are aligned and to catch any early improvements.


coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,166 @@
...
+        bins misaligned = {[2'b01:2'b11]};
+    }
+
+    misaligned_double: coverpoint (ins.current.rs1_val + ins.current.imm)[2:0]
+    iff (ins.current.insn inside {LD, SD}) {
+        bins aligned    = {3'b000};
+        bins misaligned = {[3'b001:3'b111]};
+    }
+
+    mode_satp_bare: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_CURRENT, "satp", "mode") {
```
jordancarlin: If you want the CSR value when the instruction begins execution (which is what impacts its behavior) use `` `SAMPLE_BEFORE `` or `` `SAMPLE_PREV ``. `` `SAMPLE_AFTER `` or `` `SAMPLE_CURRENT `` gives the value after the instruction runs.

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,166 @@
...
+        `ifdef SV39
+        bins sv39 = {4'b1000};
+        `endif
+        `ifdef SV48
+        bins sv48 = {4'b1001};
+        `endif
+        `ifdef SV57
+        bins sv57 = {4'b1010};
+        `endif
+    }
```
jordancarlin: This naming is confusing. All of this if for rv64 only and calling the one with bare rv64 does not make it clear what the difference between `mode_satp_va` and `mode_satp_rv64` are.

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,166 @@
...
+
+    senvcfg_pmm: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_CURRENT, "senvcfg", "pmm") {
+        bins enabled_10  = {2'b10};
+        bins enabled_11  = {2'b11};
+    }
+
+    amo_op: coverpoint ins.current.insn
+    iff (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "misa", "A") == 1'b1) {
```
jordancarlin: Is the test going to be changing `misa`? If not, this check seems unnecessary. Especially because it is valid to implement `misa` as read-only 0 even if you support the A extension. The Zaamo extension also provides these instructions and does not set `misa.A`.

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,166 @@
...
+        wildcard bins amoadd  = {AMOADD_W};
+        wildcard bins amoswap = {AMOSWAP_W};
+        wildcard bins amoxor  = {AMOXOR_W};
+        wildcard bins amoadd_d  = {AMOADD_D};
+        wildcard bins amoswap_d = {AMOSWAP_D};
+        wildcard bins amoxor_d  = {AMOXOR_D};
+    }
```
jordancarlin: What if a core implements Ssnpm without Zaamo? Then these instructions can't be run. This either needs to be inside of an `` `ifdef ZAAMO_SUPPORTED `` or a separate test suite.

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,166 @@
...
+
+    mxr_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "sstatus", "mxr") {
+        bins enabled  = {1'b1};
+    }
+
+   uxlen: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "sstatus", "uxl")
+   iff (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "misa", "mxl") == 2'b10){
+        bins xlen_32 = {2'b01};
+        bins xlen_64 = {2'b10};
+    }
```
jordancarlin: How does this work for a core that does not support mutable `UXLEN` (most don't)?

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,166 @@
...
+    cp_pm_misaligned_word_write: cross priv_mode_u, senvcfg_pmm,mode_satp_rv64,  misalign_write_instr_word,  misaligned_word ;
+    cp_pm_misaligned_word_read: cross priv_mode_u, senvcfg_pmm,mode_satp_rv64, misalign_read_instr_word,  misaligned_word ;
+    cp_pm_misaligned_double_write: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64, misalign_write_instr_double, misaligned_double ;
+    cp_pm_misaligned_double_read: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64, misalign_read_instr_double, misaligned_double ;
+    cp_pm_effective_xlen_constraint_write: cross priv_mode_u, senvcfg_pmm,  mode_satp_rv64, uxlen,  write_instr;
+    cp_pm_effective_xlen_constraint_read: cross priv_mode_u, senvcfg_pmm,  mode_satp_rv64, uxlen, read_instr;
+
+    endgroup
+
+    function void Ssnpm_sample(int hart, int issue, ins_t ins);
```
jordancarlin: Fix indentation. Both of these lines should be unindented.

jordancarlin: Our convention is to make the sample function name lowercase.
```suggestion
    endgroup

    function void ssnpm_sample(int hart, int issue, ins_t ins);
```

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,166 @@
...
+    cp_mask_priv_mode_only_write: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64,   write_instr;
+    cp_mask_priv_mode_only_read: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64,  read_instr ;
+    cp_pm_mxr_disable_write: cross priv_mode_u, senvcfg_pmm, mxr_bit, mode_satp_va,  write_instr ;
+    cp_pm_mxr_disable_read: cross priv_mode_u, senvcfg_pmm, mxr_bit, mode_satp_va, read_instr ;
+    cp_pm_misaligned_half_write: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64, misalign_write_instr_half, misaligned_half ;
+    cp_pm_misaligned_half_read: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64,  misalign_read_instr_half, misaligned_half ;
+    cp_pm_misaligned_word_write: cross priv_mode_u, senvcfg_pmm,mode_satp_rv64,  misalign_write_instr_word,  misaligned_word ;
+    cp_pm_misaligned_word_read: cross priv_mode_u, senvcfg_pmm,mode_satp_rv64, misalign_read_instr_word,  misaligned_word ;
+    cp_pm_misaligned_double_write: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64, misalign_write_instr_double, misaligned_double ;
+    cp_pm_misaligned_double_read: cross priv_mode_u, senvcfg_pmm, mode_satp_rv64, misalign_read_instr_double, misaligned_double ;
```
jordancarlin: There is probably a more concise way to write these that combines them into `cp_pm_misaligned_read` and `cp_pm_misaligned_write` crosses without needing all of the variants.

jordancarlin (changes requested):
I haven't reviewed the crosses in detail yet (I'll wait until the testplan is done), but here are some initial comments on the building block coverpoints.

coverpoints/priv/Ssnpm_coverage.svh
Ammarahwakeel: Hmm, okay removing it from this file. I will make a separate coverage file for this test, which will only run if mutable UXLEN is supported , sounds good?

coverpoints/priv/Ssnpm_coverage.svh
Ammarahwakeel: Earlier I was using a single mode_satp coverpoint (which I later changed to mode_satp_rv64). With that approach, some cross combinations required ignore_bins because certain cases are only valid for PA while others are only valid for VA. To avoid introducing ignore_bins, I separated the modes for PA and VA into different coverpoints.

coverpoints/priv/Ssnpm_coverage.svh
Ammarahwakeel: I will look into it.

PR #1043, opened 2026-03-06
Release MC100 rev0.1 and RVI20 rev0.6 CRDs

james-ball-qualcomm:
Getting ready for CSC F2F tomorrow and vote on RVI20 rev 0.6 CRD for next CSC requirements WG meeting (a week from tomorrow).

### DIFF
--- a/docs/crd/src/extension_defs.adoc
+++ b/docs/crd/src/extension_defs.adoc
@@ -141,6 +141,24 @@ Microcontroller compressed instructions
 Description::
 Additional compressed instructions for microcontrollers including push/pop and table jump.
 
+[[extension:Smepmp]]
+=== Extension Smepmp
+
+Long Name::
+Enhanced PMP
+
+Description::
+Original PMP with additional features such as guards against additional attack vectors.
+
+[[extension:Smpmp]]
+=== Extension Smpmp
+
+Long Name::
+Original PMP
+
+Description::
+Original PMP as defined in the "Physical Memory Protection" section of the RISC-V Privileged Architecture Machine-mode chapter, without the additional features of Smepmp.
+
 [[extension:Zicntr]
 === Extension Zicntr
 
--- a/docs/crd/src/extension_defs.adoc
+++ b/docs/crd/src/extension_defs.adoc
@@ -59,7 +59,7 @@ All Machine-mode features defined as mandatory by the RISC-V Privileged Architec
 .RV32/64 mandatory CSRs (many can be read-only zero)
 [%autowidth]
 |===
-| CSR name | Normative rule definining CSR as mandatory | Notes
+| CSR name | Normative rule defining CSR as mandatory | Notes
 
 | misa | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#misa_always_rd[misa_always_rd] |
 | mvendorid | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mvendorid_always_rd[mvendorid_always_rd] |
@@ -87,7 +87,7 @@ All Machine-mode features defined as mandatory by the RISC-V Privileged Architec
 .Additional RV32-only mandatory CSRs (many can be read-only zero)
 [%autowidth]
 |===
-| CSR name | Normative rule definining CSR as mandatory | Notes
+| CSR name | Normative rule defining CSR as mandatory | Notes
 
 | mstatush | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mstatush_acc[mstatush_acc] | Mentioned as mandatory in the Priv ISA manual "Preface to Version 20211203"
 | mcycleh | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mcycleh_op[mcycleh_op] |
@@ -101,7 +101,7 @@ All Machine-mode features defined as mandatory by the RISC-V Privileged Architec
 .Memory-mapped registers (not CSRs)
 [%autowidth]
 |===
-| Register name | Normative rule definining register as mandatory | Notes
+| Register name | Normative rule defining register as mandatory | Notes
 
 | mtime | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mtime_acc[mtime_acc] |
 | mtimecmp | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mtimecmp_sz[mtimecmp_sz] |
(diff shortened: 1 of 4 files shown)


james-ball-qualcomm:
@jordancarlin, what do you think?

docs/crd/src/rvi20_crd.adoc
```diff
@@ -164,16 +141,6 @@ include::in_scope_param_intro.adoc[]
 | Misaligned load/store instructions are only IN-SCOPE if this parameter is true
 and if any restrictions on misaligned load/store instructions expressed by other parameters are met.
 
-| <<param:MISALIGNED_ATOMICITY_GRANULE_SIZE,MISALIGNED_ATOMICITY_GRANULE_SIZE>>
```
jordancarlin, line 167: Not sure what the idea with the "OUT-OF-SCOPE Parameters" section is (it doesn't seem maintainable to have a complete list there), but should this parameter be added to that list?

docs/crd/src/extension_defs.adoc
```diff
@@ -2,40 +2,195 @@
...
+[[extension:Sm]]
+=== Extension Sm
+
+Long Name::
+Minimal Machine Mode
+
+Description::
+All Machine-mode features defined as mandatory by the RISC-V Privileged Architectures specification.
+
+.RV32/64 mandatory CSRs (many can be read-only zero)
```
jordancarlin, line 59: Some of these are required by Sm, some by Zicntr, some by Zihpm. If we want to include this list of CSRs, then we need to be clear about which extension each CSR belongs to. That seems like a lot of overhead and might be beyond the scope of the CRD. I'm not sure that this list of CSRs provides much value since they are defined by the extensions in the ISA manual already. We don't include a list of the instructions in each extension, and it seems like we should treat CSRs similarly.

docs/crd/src/extension_defs.adoc
```diff
@@ -2,40 +2,195 @@
...
+| CSR name | Normative rule defining CSR as mandatory | Notes
 
-| [#extension:A]#A#
-| Atomic instructions
-| Comprised of instructions from the Zaamo and Zalrsc extensions
+| misa | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#misa_always_rd[misa_always_rd] |
+| mvendorid | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mvendorid_always_rd[mvendorid_always_rd] |
+| marchid | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#marchid_always_rd[marchid_always_rd] |
+| mimpid | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mimpid_always_rd[mimpid_always_rd] |
+| mhartid | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mhartid_always_rd[mhartid_always_rd] |
```
jordancarlin, line 68: Future enhancement would be to have this automatically pull in the actual text of the normative rule. We do this in the CTP already and you might be able to reuse some of the scripts from the `generators/ctp` directory.

docs/crd/src/extension_defs.adoc
```diff
@@ -2,40 +2,195 @@
...
+| Instruction name | Normative rule defining instruction as mandatory | Notes
 
-| [#extension:Zifencei]#Zifencei#
-| Instruction fence
-|
+| ecall | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#ecall_op[ecall_op] | Implicitlymandatory
+| ebreak | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#ebreak_op[ebreak_op] | Implicitlymandatory
+| c.ebreak | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#c-ebreak_op[c-ebreak_op] | Implicitlymandatory
+| mret | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#mret_presence[mret_presence] |
+| wfi | https://riscv.github.io/riscv-isa-manual/snapshot/norm-rules/norm-rules.html#wfi_op[wfi_op] | Implicitlymandatory
```
jordancarlin, line 119: Why is Sm the only extension with a list of instructions?

docs/crd/src/extension_defs.adoc
```diff
@@ -2,40 +2,195 @@
...
+Original PMP with additional features such as guards against additional attack vectors.
+
+[[extension:Smpmp]]
+=== Extension Smpmp
+
+Long Name::
+Original PMP
+
+Description::
+Original PMP as defined in the "Physical Memory Protection" section of the RISC-V Privileged Architecture Machine-mode chapter, without the additional features of Smepmp.
```
jordancarlin, line 160: Smpmp is not an actual extension and UDB is in the process of dropping these made-up extensions. I'd prefer to avoid reintroducing it in the CRD. PMP is really just part of M-mode (Sm) and is controlled by a few parameters (NUM_PMP_ENTRIES being the primary one).

docs/crd/src/mc100_crd.adoc
jordancarlin, line 67: Since Sm is part of MC100, I think `ecall`, `ebreak`, and `c.ebreak` are all IN-SCOPE now.

docs/crd/src/mc100_crd.adoc
```diff
@@ -64,53 +69,36 @@ execution environments to handle such exceptions.
...
-| <<extension:Zihpm,Zihpm>>
-| ~> 2.0
+| REQ-EXT-A | <<extension:A,A>> | ~> 2.1
+| REQ-EXT-M | <<extension:M,M>> | ~> 2.0
+| REQ-EXT-Smepmp | <<extension:Smepmp,Enhanced PMP (Smepmp)>> | ~> 1.13
+| REQ-EXT-Smpmp | <<extension:Smpmp,Original PMP>> | ~> 1.13
+| REQ-EXT-Zicntr | <<extension:Zicntr,Zicntr>> | ~> 2.0
+| REQ-EXT-Zifencei | <<extension:Zifencei,Zifencei>> | ~> 2.0
+| REQ-EXT-Zihpm | <<extension:Zihpm,Zihpm>> | ~> 2.0
+| REQ-EXT-Zce | <<extension:Zce,Zce>> | ~> 1.0
```
jordancarlin, line 101: If MC100 is RVI20 + M-mode, it should not include Smepmp or Zce. Neither of those are part of RVI20.

jordancarlin (changes requested):

docs/crd/src/extension_defs.adoc
james-ball-qualcomm: Seems like it would make the table too busy/wide and you don't have the nice formatting of norm rules like I do in the native HTML I link to. I think as long as there is a stable link, we should be good.

docs/crd/src/mc100_crd.adoc
james-ball-qualcomm: Yeah, when I say RVI20 + M-mode in slides I'm not being as precise as in the CRD. But, I need to fix the MC100 CRD to be more precise. Noted in #1045.

james-ball-qualcomm:
@jordancarlin, thanks for your feedback. Put in #1045 for next rev of docs.

james-ball-qualcomm:
@jordancarlin, can you please approve this so I can merge it and then start working on #1045

jordancarlin (approved):
Feedback will be addressed in the next revision of the CRD.

PR #1048, opened 2026-03-07
ACT4: add Spike support and cvw-rv32gch/rv64gch configs

TnnsBeast:
Enable Spike support for self checking usage in H tests, and create new cvw core configs with H enabled.

Notes
- As recommended by Prof. Harris, generating ref_model_args from UDB extension list is not included and is saved as a TODO for later
- This PR does not include h test generators. This will be included in a separate PR.


config/cores/cvw/cvw-rv32gch/rvmodel_macros.h
```diff
@@ -0,0 +1,109 @@
...
+##### Machine Timer #####
+
+#define RVMODEL_MTIME_ADDRESS  0x0200BFF8  /* Address of mtime CSR */
+
+#define RVMODEL_MTIMECMP_ADDRESS 0x02004000 /* Address of mtimecmp CSR */
+
+##### Machine Interrupts #####
+
+#define RVMODEL_SET_MEXT_INT(_R1, _R2)
+
```
davidharrishmc: Might as well past in the sail code for these interrupt functions from another Sail file that Sadhvi checked in.

config/cores/cvw/cvw-rv32gch/rvtest_config.h
```diff
@@ -0,0 +1,19 @@
+#define ACCESS_FAULT_ADDRESS 0x00000000
```
davidharrishmc, line 1: Should H_SUPPORTED be defined?

config/cores/cvw/cvw-rv32gch/rvtest_config.svh
```diff
@@ -0,0 +1,49 @@
...
+// Base addresses specific for PMP
+`define RAM_BASE_ADDR       32'h80000000  // PMP Region starts at RAM_BASE_ADDR + LARGEST_PROGRAM
+`define LARGEST_PROGRAM     32'h00001000
+
+// Define relevant addresses
+`define ACCESS_FAULT_ADDRESS 64'h00000000
+`define CLINT_BASE 64'h02000000
+
+//define extra supported extensions to collect full coverage in Privileged files
+`define D_SUPPORTED
```
davidharrishmc, line 32: H_SUPPORED?

davidharrishmc (changes requested):
Minor comments.  Ready to go when @jordancarlin approves.

config/cores/cvw/cvw-rv32gch/rvmodel_macros.h
TnnsBeast: Changed both gch configs to use symlinks so that these don't fall out of date. So now cvw-rv32gc, cvw-rv32gch, cvw-rv64gch all have rvmodel_macros.h symlinks that point to that of cvw-rv64gc.

.github/workflows/regress.yml
```diff
@@ -265,7 +265,7 @@ jobs:
           DOCKER=1 \
           FAST=True \
           EXTENSIONS= \
-          EXCLUDE_EXTENSIONS=Sm,ExceptionsZc,S,Zalrsc,Zacas,ZacasZabha \
+          EXCLUDE_EXTENSIONS=Sm,ExceptionsZc,S,Zalrsc,Zacas,ZacasZabha,H \
```
jordancarlin, line 268: There are no H tests being generated. Is this necessary?

config/cores/cvw/cvw-rv32gch/cvw-rv32gch.yaml
```diff
@@ -0,0 +1,391 @@
...
+  - { name: H, version: "= 1.0.0" }
+  - { name: Ssccptr, version: "= 1.0.0" }
+  - { name: Sscounterenw, version: "= 1.0.0" }
+  - { name: Ssstrict, version: "= 1.0.0" }
+  - { name: Sstc, version: "= 1.0.0" }
+  - { name: Sstvala, version: "= 1.0.0" }
+  - { name: Sstvecd, version: "= 1.0.0" }
+  - { name: Sv32, version: "= 1.13.0" }
+  # - { name: Svade, version: "= 1.0.0" }
+  # - { name: Svadu, version: "= 1.0.0" } # TODO: Re-enable when https://github.com/riscv/riscv-unified-db/issues/1217 is fixed
```
jordancarlin, line 61: This seems to be based on an old UDB config. These extensions both work now and are enabled in the other CVW configs. Please update this to match the other CVW configs and just add H (and any other params/extensions) on top of it.

config/cores/cvw/cvw-rv32gch/link.ld
jordancarlin, line 1: This is an old version of the linker script. Please update to match what is currently used by other CVW configs.

framework/src/act/config.py
```diff
@@ -19,16 +19,25 @@
...
+            RefModelType.SPIKE: "+signature={sig_file} +signature-granularity={granularity}",
+        }
+        return flags_map[self]
+
+    @property
+    def debug_flags(self) -> str:
+        """Get the debug/trace flags for this reference model."""
+        flags_map = {
+            RefModelType.SAIL: "--trace-all --trace-output {sig_trace_file}",
+            RefModelType.SPIKE: "-l --log={sig_trace_file}",
```
jordancarlin, line 40: Might be nice to add the `--log-commits` flag to this. It prints out all register and memory updates in Spike and is useful for debugging.

framework/src/act/makefile_gen.py
```diff
@@ -75,6 +76,10 @@ def compute_config_hash(config: Config, xlen: int, e_ext: bool) -> str:
     if config.objdump_exe is not None:
         hasher.update(str(config.objdump_exe.resolve()).encode())
 
+    # Hash reference model settings that affect signature generation command lines
+    hasher.update(config.ref_model_type.value.encode())
+    hasher.update((config.ref_model_args or "").encode())
```
jordancarlin, line 81:
```suggestion
    if config.ref_model_args is not None:
        hasher.update((config.ref_model_args).encode())
```

framework/src/act/makefile_gen.py
```diff
@@ -112,7 +121,7 @@ def gen_compile_targets(
     base_dir: Path,
     xlen: int,
     config: Config,
-    sail_config_path: Path,
+    common_e_ext: bool | None = None,
```
jordancarlin, line 124: `bool | None` is a strange type and I am having trouble following the logic around this. It seem like right now it is overloaded such that `None` means it is not a common test, `true` means it is a common test with E, and `false` means it is a common test without E. What about instead passing in `config_flag` or something and for sail common tests it can pass the generated sail config, for sail not common tests it can pass the DUT sail config, and for spike tests it can pass the spike ISA string?

framework/src/act/makefile_gen.py
```diff
@@ -143,6 +153,17 @@ def gen_compile_targets(
     flen = test_metadata.flen
     test_path = test_metadata.test_path
     ref_model_sig_flags = config.ref_model_type.signature_flags.format(sig_file=sig_file, granularity=int(xlen / 8))
+    ref_model_args_line = f"\t\t{config.ref_model_args} \\\n" if config.ref_model_args else ""
```
jordancarlin, line 156: Seems like we should combine (and rename) ref_model_args and ref_model_config. Right now one is used to configure Spike and the other is used to configure Sail. No need for them to be separate.

jordancarlin (changes requested):
The common test compilation logic seems like it is broken for Spike. If you have a config using Spike as the reference model with `rv32imc` and another with `rv32imfdc` it would compile all of the tests separately even though it should be possible to reuse the I, M, and C tests for both of them.

jordancarlin:
@TnnsBeast sorry this got so out of date. The entire build system has changed since this PR was created, so it will likely need a fairly extensive rework if we want to proceed with it. I'm going to close this for now, but feel free to open a new version if you update it.

PR #1049, opened 2026-03-07
fix(arch_test.h): guard resto_edeleg csrw with rvtest_strap_routine

### LINKED ISSUE #1050: Sm-00 hangs indefinitely on M+U-only (no S-mode) configurations
**Describe the bug**
Running Sm-00 on a no-S-mode configuration hangs indefinitely. The simulation never produces an HTIF exit. With --trace-exception the same trap repeats without bound:  
```
handling exc#illegal-instruction at priv M with tval 0x30239073
```
0x30239073 is csrw medeleg, t2. A single teardown instruction with o loop around it executed 49,174 times in a 5M-instruction run.

**To Reproduce**
```bash
# Normal make invocation — hangs indefinitely on Sm-00, must be killed manually
make --jobs $(nproc) CONFIG_FILES=config/cores/cvw/cvw-rv32imc/test_config.yaml

# Directly, with an instruction limit to observe the behaviour:
sail_riscv_sim \
  --config config/cores/cvw/cvw-rv32imc/sail.json \
  --inst-limit 5000000 \
  --trace-exception \
  --trace-output /tmp/sm00_5m.log \
  --test-signature /tmp/sm00_5m.sig \
  --signature-granularity 4 \
  work/cvw-rv32imc/build/priv/Sm/Sm-00.sig.elf #  hits 5M limit without HTIF exit; test never completes
```

```
handling exc#illegal-instruction at priv M with tval 0x30239073
```

**Additional context**
fixed in PR #1049

karabambus:
On M+U-only (no S-mode) harts, medeleg is not implemented and any access raises illegal-instruction. The resto_edeleg macro in cleanup epilogs unconditionally wrote CSR_XEDELEG, causing an infinite trap loop: the faulting instruction sits past rvtest_code_end, so the trap handler's mepc bounds check fires abort_test, which jumps back to cleanup_epilogs, repeating forever.

The init_edeleg macro already guards the CSR access with #ifdef rvtest_strap_routine (which is only defined when S-mode is present). This commit restores the same guard on resto_edeleg, matching the pre-PR#843 behaviour.

Verified: Sm-00 now exits with SUCCESS (HTIF exit) on an RV32IMC+U+Sm configuration with S-mode disabled in Sail.

Fixes regression introduced by commit 4e467ae2 ("(ACT4) New Trap Handler").

### DIFF
--- a/tests/env/arch_test.h
+++ b/tests/env/arch_test.h
@@ -1879,12 +1879,14 @@ resto_\__MODE__\()edeleg:
 #endif
 .ifnc \__MODE__ , S
   .ifnc \__MODE__ , V
+#ifdef rvtest_strap_routine
         csrw    CSR_XEDELEG,  T2
     .ifc \_MODE__ , M   // TODO: Remove this .ifc when sail supports hedelegh (if Smstateen is supported, set mstateen0.P1P13)
       #if (XLEN==32)
         csrw    CSR_XEDELEGH, T4
       #endif
     .endif
+#endif
   .endif
 .endif
 
--- a/tests/env/arch_test.h
+++ b/tests/env/arch_test.h
@@ -1872,6 +1872,9 @@ exit_\__MODE__\()cleanup:                       // if you enter here from the ab
       .endif
 
 //----------------------
+// Guard below must match init_edeleg. If init skips the csrw (e.g. no S-mode),
+// restore must skip it too. Mismatch causes infinite trap loop on configs where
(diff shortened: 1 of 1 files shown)


jordancarlin:
@Zain2050 @UmerShahidengr @allenjbaum can you review?

Zain2050:
Yes, this is the correct fix.

jordancarlin (approved):

PR #1051, opened 2026-03-07
Update CV32E20 rvmodel macros

karabambus:
Correctly defined parametre macros and added RVMODEL_MTIME_ADDRESS and  RVMODEL_MTIMECMP_ADDRESS according to sail.json configuration

### DIFF
--- a/config/cores/cve2/cv32e20/rvmodel_macros.h
+++ b/config/cores/cve2/cv32e20/rvmodel_macros.h
@@ -67,9 +67,9 @@
 
 ##### Machine Timer #####
 
-#define RVMODEL_MTIME_ADDRESS     0x0200bff8  
+#define RVMODEL_MTIME_ADDRESS     0x0200bff8  // sail.json configuration
 
-#define RVMODEL_MTIMECMP_ADDRESS  0x02004000  
+#define RVMODEL_MTIMECMP_ADDRESS  0x02004000  // sail.json configuration
 
 ##### Machine Interrupts #####
 
--- a/config/cores/cve2/cv32e20/rvmodel_macros.h
+++ b/config/cores/cve2/cv32e20/rvmodel_macros.h
@@ -67,10 +67,9 @@
 
 ##### Machine Timer #####
 
-#define RVMODEL_MTIME_ADDRESS     0x0200bff8  // sail.json configuration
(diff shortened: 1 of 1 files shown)


config/cores/cve2/cv32e20/rvmodel_macros.h
```diff
@@ -67,28 +67,28 @@
 
 ##### Machine Timer #####
 
-#define RVMODEL_MTIME_ADDRESS /* unimplemented */
+#define RVMODEL_MTIME_ADDRESS     0x0200bff8  // sail.json configuration
```
MikeOpenHWGroup: These should be left unimplemented as the CV32E20 does not implement `MTIME` or `MTIMECMP` (because they are _Platform_, not Core resources).  The testbench for the CV32E20 _could_ implement these memory mapped CSRs, but does not because they are not necessary to verify the core.

(FWIW, the Sail module also should not implement these non-core resources.)

config/cores/cve2/cv32e20/rvmodel_macros.h
```diff
@@ -67,28 +67,28 @@
...
 
-#define RVMODEL_MTIMECMP_ADDRESS /* unimplemented */
+#define RVMODEL_MTIMECMP_ADDRESS  0x02004000  // sail.json configuration
 
 ##### Machine Interrupts #####
 
-#define RVMODEL_SET_MEXT_INT
+#define RVMODEL_SET_MEXT_INT(_R1, _R2)
```
MikeOpenHWGroup, line 75: Are these args needed?

MikeOpenHWGroup (approved):
Thanks Marin, this is good.  I will approve, but do not have write access to this repo.  Hi @jordancarlin, can you have a look?

config/cores/cve2/cv32e20/rvmodel_macros.h
jordancarlin: The macro signature needs to match the expected signature, so yes.

config/cores/cve2/cv32e20/rvmodel_macros.h
jordancarlin: I'll merge this for now with them marked an unimplemented, but we need to think about how to handle `mtime` more generally.

jordancarlin (approved):

PR #1055, opened 2026-03-08
loaded xcause into T5 for non-M modes

juliacygong:


tests/env/arch_test.h
jordancarlin: Would it be simpler to just move this line before the `.ifc \__MODE__ ,  M ` so that it is run for all priv modes? That way we don't need to duplicate it below for the non M-mode case.

jordancarlin: Looks like you put the csrr before `.ifc \__MODE__ ,  M` but also kept the one inside of it. The second one can be deleted now since it immediately follows the first.

jordancarlin (approved):

PR #1058, opened 2026-03-09
Make covergroupgen a first-class package with better error messages

jordancarlin:
- Add covergroupgen package to workspace
- Adds progress bar when generated covergoups
- Fix `make clean` to actually remove generated coverpoint files
- Add option to limit which extensions covergroups are generated for (same as testgen)
- Print out similar templates and the directory that templates need to be added to when there is a missing template
- Make covergroupgen more robust in different installation environments


davidharrishmc:
Is there any benefit to not generating all coverage files?  I think we want to check them all into the repo, and they should be independent of the config.  It's helpful for them to all be there for a reader to understand the scope of the ACTs, even when they aren't run.

jordancarlin:
This came up when doing #1056 and I wanted a way to generate coverpoints for a single extension. We will eventually check the coverpoints into the repo just like the tests, but we still have a way to limit which tests we generate.

davidharrishmc (approved):

PR #1061, opened 2026-03-09
InterruptsU Python Tests

SadhviNarayanan:
This is the python generator script to generate tests for InterruptsU. This PR also includes some changes to the coverage file in order to reflect what the tests need to check for, and follows a similar structure to the other python scripts - particularly InterruptsSm. 
Currently the coverage is at 91.66% but this is expected, because the only bins that are missing are the ones that correspond to external interrupts which do not exist currently on SAIL. Everything else is collecting full coverage.


coverpoints/priv/InterruptsU_coverage.svh
```diff
@@ -51,28 +54,29 @@ covergroup InterruptsU_cg with function sample(ins_t ins);
...
 
-    // main coverpoints
-    cp_user_mti:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_mtie_one, m_timer_intr, mip_mtip;
-    cp_user_msi:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_msie_one, m_soft_intr,  mip_msip;
-    cp_user_mei:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_meie_one, m_ext_intr,   mip_meip;
-    cp_wfi:         cross priv_mode_u, wfi,        mstatus_mie, mstatus_tw, mie_mtie_one, m_timer_intr;
-    cp_wfi_timeout: cross priv_mode_u, wfi,        mstatus_mie, mstatus_tw, mie_mtie_one, timeout;
+    cp_user_mti:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_mtie_one, mip_mtip;
+    cp_user_msi:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_msie_one,  mip_msip;
+    cp_user_mei:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_meie_one,   mip_meip;
```
davidharrishmc, line 69: Add a comment that this isn't hit in Sail because of lack of external interrupt support.  If there is a issue or PR to link to, please add.

davidharrishmc: Same for InterruptsSm.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,481 @@
...
+        for mstatus_mie in [0, 1]:
+            mode_name = ["direct", "vectored"][mtvec_mode]
+            mie_name = f"mie_{mstatus_mie}"
+            binname = f"{mode_name}_{mie_name}"
+
+            lines.extend(
+                [
+                    "",
+                    "# Setup mstatus and mtvec",
+                    "csrci mstatus, 8",
```
davidharrishmc: Always explain what the magic numbers do.
Use the CSR macros instead of the raw commands, to be safer.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,481 @@
...
+            if mtvec_mode:
+                lines.append("csrsi mtvec, 1")
+
+            # Set mstatus.MPIE (not MIE) because mret copies MPIE→MIE when transitioning to U-mode.
+            # Setting MIE directly would be overwritten by mret's automatic MPIE restoration.
+            if mstatus_mie:
+                lines.extend(
+                    [
+                        f"LI(x{r_scratch}, 0x80)",  # MPIE bit mask
+                        f"CSRRS(x{r_scratch}, mstatus, x{r_scratch})",  # Set MPIE
```
davidharrishmc: Why do you bother to read if you ignore the result?  CSRS seems good enough?  Many places.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,481 @@
...
+                ]
+            )
+
+            # Set mtvec.MODE if needed
+            if mtvec_mode:
+                lines.append("csrsi mtvec, 1")
+
+            # Set mstatus.MPIE (not MIE) because mret copies MPIE→MIE when transitioning to U-mode.
+            # Setting MIE directly would be overwritten by mret's automatic MPIE restoration.
+            if mstatus_mie:
```
davidharrishmc: Could you do cmd = ["CSRC", "CSRS"][mstatus_mie] and then avoid replicating the logic in the if?

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,481 @@
...
+                ]
+            )
+
+            # Go to U-mode
+            lines.append("RVTEST_GOTO_LOWER_MODE Umode")
+
+            # Trigger interrupt
+            lines.append(test_data.add_testcase(binname, coverpoint, covergroup))
+
+            lines.extend(indent(set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2)))
```
davidharrishmc: The code in interrupts.py for set_mtimer_int for rv32 says "Per RISC-V Spec".  However, it doesn't match the code below of the spec and I don't see how it would work.  Am I missing some other part of the spec? Please double-check.

# New comparand is in a1:a0.
li t0, -1
la t1, mtimecmp
sw t0, 0(t1) # No smaller than old value.
sw a1, 4(t1) # No smaller than new value.
sw a0, 0(t1) # New value.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,481 @@
...
+                [
+                    f"LI(x{r_scratch}, 0x80)",
+                    f"CSRW(mie, x{r_scratch})",
+                ]
+            )
+
+            lines.append("RVTEST_GOTO_LOWER_MODE Umode")
+
+            # Set timer to fire soon
+            lines.extend(indent(set_mtimer_int_soon(r_mtime, r_mtimecmp, r_temp, r_t1, r_t2, r_temp2)))
```
davidharrishmc: Some issues in set_mtimer_int_soon:

        f"    LI(x{r_temp1}, -1)",
        f"    sd x{r_temp1}, 0(x{r_mtimecmp})",
        # Read current time
        f"    ld x{r_temp1}, 0(x{r_mtime})",
        f"    addi x{r_temp1}, x{r_temp1}, RVMODEL_TIMER_INT_SOON_DELAY",  # RVMODEL_TIMER_INT_SOON_DELAY
        f"    sd x{r_temp1}, 0(x{r_mtimecmp})",


Is there any reason we need to set mtimecmp to all 1s before setting it again to mtime+TIMER_INT_SOON_DELAY?

In general, TIMER_INT_SOON_DELAY could be > 2047, which doesn't work with addi.  It would be more robust to LI the delay into a register and then add to mtime_cmp.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,481 @@
...
+                        f"CSRRS(x{r_scratch}, mie, x{r_scratch})",
+                    ]
+                )
+
+            # Clear timer (ensure no interrupt)
+            lines.extend(clr_mtimer_int(r_temp, r_mtimecmp))
+
+            lines.extend(
+                [
+                    "nop",
```
davidharrishmc: What is the nop here for?

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,481 @@
...
+
+    test_data.int_regs.return_registers([r_temp, r_mtimecmp, r_scratch])
+    return lines
+
+
+@add_priv_test_generator("InterruptsU", required_extensions=["Sm", "I", "Zicsr"])
+def make_interruptsu(test_data: TestData) -> list[str]:
+    """Generate tests for InterruptsU user-mode interrupt behavior."""
+
+    # Consume t2 and t5 for interrupt subroutines throughout the test
```
davidharrishmc: How do we know these are used for interrupt subroutines? What subroutines?  interrupts.py or a trap handler?  Is this necessary? Is this documented in DeveloperGuide.md?

davidharrishmc (changes requested):

coverpoints/priv/InterruptsU_coverage.svh
```diff
@@ -16,19 +16,28 @@ covergroup InterruptsU_cg with function sample(ins_t ins);
     `include "general/RISCV_coverage_standard_coverpoints.svh"
 
     // building blocks for the main coverpoints
-    mstatus_mie: coverpoint ins.current.csr[12'h300][3]  {
+
+    // Uses ins.prev instead of ins.current because Sail tracer updates CSRs after instruction retirement, so .current shows
+    // post-trap state. May revert to .current depending on implementation consensus
```
jordancarlin: This is not a Sail specific issue. Pretty sure we decided this is just how RVVI works.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -56,33 +57,39 @@ def clr_mtimer_int(r_temp: int, r_mtimecmp: int) -> list[str]:
...
-        # If mtimecmp < mtime from a previous test,
-        # mip.MTIP would be set immediately when we enable MTIE, causing
-        # an unexpected trap before we're ready.
-        f"    LI(x{r_temp1}, -1)",
-        f"    sd x{r_temp1}, 0(x{r_mtimecmp})",
-        # Read current time
+        # Read current time and add delay
         f"    ld x{r_temp1}, 0(x{r_mtime})",
-        f"    addi x{r_temp1}, x{r_temp1}, RVMODEL_TIMER_INT_SOON_DELAY",  # RVMODEL_TIMER_INT_SOON_DELAY
+        f"    LI(x{r_temp2}, RVMODEL_TIMER_INT_SOON_DELAY)",  # Load delay into register
```
jordancarlin: You should probably do the `LI` before the `ld` so that the `mtime` you are adding to is more accurate. Right now, you are running several instructions after reading `mtime`, so the delay is shorter than expected. Applies to RV32 too.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -24,13 +23,15 @@ def set_mtimer_int(r_mtime: int, r_mtimecmp: int, r_temp: int, r_temp2: int) ->
...
-        f"    lw x{r_temp2}, 4(x{r_mtime})",  # Re-read high word
-        # Now write to mtimecmp
-        f"    sw x{r_temp}, 0(x{r_mtimecmp})",  # Write low word
-        f"    sw x{r_temp2}, 4(x{r_mtimecmp})",  # Write high word
+        # Write sequence to prevent spurious interrupts
+        # Read mtime (new comparand will be in temp2:temp)
+        f"    lw x{r_temp}, 0(x{r_mtime})",  # mtime[31:0] -> low word
+        f"    lw x{r_temp2}, 4(x{r_mtime})",  # mtime[63:32] -> high word
+        # 3-step write sequence (new comparand is in temp2:temp)
+        f"    li x{r_mtime}, -1",  # Reuse r_mtime for -1
```
jordancarlin: Make sure to use the `LI` macro. Global comment.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -467,7 +467,9 @@ def _generate_wfi_tests(test_data: TestData) -> list[str]:
 def make_interruptssm(test_data: TestData) -> list[str]:
     """Generate tests for InterruptsSm machine-mode interrupts."""
 
-    # consumer t2 and t5 for interrupt subroutines, but mark as consumed for whole test since they're used throughout
+    # Consume t2 and t5 - reserved by trap handler in arch_test.h for interrupt clearing subroutines
+    # The trap handler uses these registers when calling RVMODEL_CLR_*_INT macros
+    # See: tests/env/arch_test.h, trap handler interrupt dispatch logic
     test_data.int_regs.consume_registers([7, 30])
```
jordancarlin: Should we simplify things and just consume these registers in `priv.py` for all priv tests?

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,447 @@
...
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_user_mti_tests(test_data: TestData) -> list[str]:
+    """Generate undelegated MTI interrupt tests from U-mode."""
+    covergroup = "InterruptsU_cg"
+    coverpoint = "cp_user_mti"
+
+    r_mtime, r_mtimecmp, r_temp, r_temp2, r_scratch = test_data.int_regs.get_registers(5, exclude_regs=[0, 2, 7, 30])
```
jordancarlin: 7 and 30 are already marked as used when you consume the at the beginning of the main test block. What is wrong with 2?

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,447 @@
...
+
+            # Set mstatus.MPIE (not MIE) because mret copies MPIE→MIE when transitioning to U-mode.
+            # Setting MIE directly would be overwritten by mret's automatic MPIE restoration.
+            cmd = ["CSRC", "CSRS"][mstatus_mie]  # CSRC if 0, CSRS if 1
+            lines.extend(
+                [
+                    f"LI(x{r_scratch}, 0x80)",  # mstatus.MPIE bit mask (bit 7)
+                    f"{cmd}(mstatus, x{r_scratch})",
+                ]
+            )
```
jordancarlin: This might be easier to follow:
```suggestion
            lines.extend(
                [
                    f"LI(x{r_scratch}, 0x80)",  # mstatus.MPIE bit mask (bit 7)
                    f"{'CSRS' if mstatus_mie else 'CSRC'}(mstatus, x{r_scratch})",
                ]
            )
```

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,447 @@
...
+            # Enable MTIE
+            lines.extend(
+                [
+                    f"LI(x{r_scratch}, 0x80)",
+                    f"CSRW(mie, x{r_scratch})",
+                ]
+            )
+
+            # Go to U-mode
+            lines.append("RVTEST_GOTO_LOWER_MODE Umode")
```
jordancarlin: Instead of repeated calls to `lines.extend` or `lines.append`, combine into a single call to `lines.extend`. Global comment.

```suggestion
            lines.extend(
                [
                    f"LI(x{r_scratch}, 0x80)",
                    f"CSRW(mie, x{r_scratch})",
                    "RVTEST_GOTO_LOWER_MODE Umode",
                ]
            )
```

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,447 @@
...
+                ]
+            )
+
+            lines.extend(
+                [
+                    "RVTEST_GOTO_MMODE",
+                    "nop",
+                    "RVTEST_CLR_MSW_INT",
+                ]
+            )
```
jordancarlin, line 145: Combine into single `lines.extend`. Global comment.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,447 @@
...
+                    "    nop",
+                ]
+            )
+
+            lines.extend(
+                [
+                    "RVTEST_GOTO_MMODE",
+                ]
+            )
+            lines.extend(clr_mtimer_int(r_temp, r_mtimecmp))
```
jordancarlin: Why is this needed? You cleared timer_int above and haven't set it again since.

jordancarlin (changes requested):

coverpoints/priv/InterruptsU_coverage.svh
```diff
@@ -51,28 +60,32 @@ covergroup InterruptsU_cg with function sample(ins_t ins);
...
-    cp_wfi:         cross priv_mode_u, wfi,        mstatus_mie, mstatus_tw, mie_mtie_one, m_timer_intr;
-    cp_wfi_timeout: cross priv_mode_u, wfi,        mstatus_mie, mstatus_tw, mie_mtie_one, timeout;
+
+    cp_user_mti:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_mtie_one, mip_mtip;
+    cp_user_msi:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_msie_one,  mip_msip;
+
+    // Note: External interrupts (MEI) not supported in Sail simulator yet
+    cp_user_mei:    cross priv_mode_u, mtvec_mode, mstatus_mie, mie_meie_one,   mip_meip;
+    cp_wfi:         cross priv_mode_u, wfi,        mstatus_mie, mstatus_tw, mie_mtie_one;
+    cp_wfi_timeout: cross priv_mode_u, wfi,        mstatus_mie, mstatus_tw_one, mie_mtie;
```
jordancarlin, line 71: Why is the `cp_wfi_timeout` coverpoint changing from `mie_mtie_one` to `mie_mtie`? Do the cases where mie/mtie = 0 add interesting coverage?

SadhviNarayanan: It was based on the testplan

davidharrishmc (approved):
LGTM.
cp_wfi and cp_wfi_timeout look like they match the testplan.

coverpoints/priv/InterruptsU_coverage.svh
davidharrishmc: I agree this matches the testplan.  The testplan might be overkill, but it is working.

davidharrishmc:
@jordancarlin this looks like it may be ready for your next review. @SadhviNarayanan CI is failing.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -401,25 +401,30 @@ def _generate_wfi_tests(test_data: TestData) -> list[str]:
...
-                    "CSRW mie, zero",
-                    "CSRW mstatus, zero",
-                    *clr_mtimer_int(r_t0, r_mtimecmp),
-                    "# Set TW if needed",
-                    f"LI(x{r_scratch}, {tw_val << 21})",  # TW bit, MIE=0
-                    f"csrw mstatus, x{r_scratch}",
-                    "# Enable MTIE",
+                    "CSRW(mie, zero)",
+                    f"LI(x{r_scratch}, 0x200008)",
+                    f"CSRC(mstatus, x{r_scratch})",
```
jordancarlin, line 407: Worth a comment. What is 0x200008?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -401,25 +401,30 @@ def _generate_wfi_tests(test_data: TestData) -> list[str]:
...
-            # Enable MIE if needed, this is set here to avoid a premature interrupt before we go into wfi
-            if mie_val:
-                lines.append("csrsi mstatus, 8")
+            if tw_val:
+                lines.extend(
+                    [
+                        f"LI(x{r_scratch}, 0x200000)",
+                        f"CSRS(mstatus, x{r_scratch})",
+                    ]
+                )
```
jordancarlin: This is misleading. The comment is always generated but the code is only generated if tw_val == 1.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,346 @@
...
+    lines: list[str] = []
+
+    r_temp, r_mtimecmp = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    # Initial setup - clear any pending timer
+    lines.extend(
+        [
+            "CSRW(mideleg, zero)",
+        ]
+    )
```
jordancarlin: if just a single line, use `lines.append`.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,346 @@
...
+
+    # Cross: mtvec.MODE x mstatus.MIE (2x2 = 4 bins)
+    for mtvec_mode in [0, 1]:
+        for mstatus_mie in [0, 1]:
+            mode_name = ["direct", "vectored"][mtvec_mode]
+            mie_name = f"mie_{mstatus_mie}"
+            binname = f"{mode_name}_{mie_name}"
+
+            if mtvec_mode:
+                lines.append("csrsi mtvec, 1")
```
jordancarlin, line 51: Doesn't this need to clear the appropriate bits for the direct case?

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,346 @@
...
+                    f"LI(x{r_scratch}, 0x80)",
+                    f"CSRW(mie, x{r_scratch})",
+                    "RVTEST_GOTO_LOWER_MODE Umode",
+                    test_data.add_testcase(binname, coverpoint, covergroup),
+                    *set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2),
+                ]
+            )
+
+            for _ in range(5000):
+                lines.append("    nop")
```
jordancarlin: When I was talking about a loop, I meant an assembly loop in the generated test. Inserting 5000 nops directly into the test seems less than ideal and will be very hard to read.

generators/testgen/src/testgen/asm/interrupts.py
jordancarlin, line 1: Seems like there might have been a merge issue here or something? Lots of strange spacing changes in this file.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,346 @@
...
+
+            lines.extend(
+                [
+                    f"LI(x{r_scratch}, 0x80)",
+                    f"{'CSRS' if mstatus_mie else 'CSRC'}(mstatus, x{r_scratch})",
+                    f"LI(x{r_scratch}, 0x80)",
+                    f"CSRW(mie, x{r_scratch})",
+                    "RVTEST_GOTO_LOWER_MODE Umode",
+                    test_data.add_testcase(binname, coverpoint, covergroup),
+                    *set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2),
```
jordancarlin, line 61: I'm confused how this is supposed to work. Presumably it will be common for mtimecmp to only be writable from M-mode (using PMAs for example). How does this work in that case? Seems like we'd trap instead of triggering the expected interrupt.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
SadhviNarayanan: I'll keep note of this, and can circle back to it once we figure out more of the details with M-mode.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -28,13 +27,15 @@ def set_mtimer_int(r_mtime: int, r_mtimecmp: int, r_temp: int, r_temp2: int) ->
...
-        f"sw x{r_temp2}, 4(x{r_mtimecmp}) # Write high word",
+        # Write sequence to prevent spurious interrupts
+        # Read mtime (new comparand will be in temp2:temp)
+        f"lw x{r_temp}, 0(x{r_mtime})",  # mtime[31:0] -> low word
+        f"lw x{r_temp2}, 4(x{r_mtime})",  # mtime[63:32] -> high word
+        # 3-step write sequence (new comparand is in temp2:temp)
+        f"LI(x{r_mtime}, -1)",  # Reuse r_mtime for -1
+        f"sw x{r_mtime}, 0(x{r_mtimecmp})",  # Step 1: Write -1 to low (no smaller than old)
+        f"sw x{r_temp2}, 4(x{r_mtimecmp})",  # Step 2: Write high word (no smaller than new)
+        f"sw x{r_temp}, 0(x{r_mtimecmp})",  # Step 3: Write low word (new value)
```
jordancarlin: Why are all of the comments being removed from the generated assembly?

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -61,34 +62,42 @@ def clr_mtimer_int(r_temp: int, r_mtimecmp: int) -> list[str]:
...
-        f"lw x{r_temp1}, 0(x{r_mtime})",
-        f"lw x{r_temp2}, 4(x{r_mtime})",
-        f"addi x{r_temp3}, x{r_temp1}, RVMODEL_TIMER_INT_SOON_DELAY",
-        f"sltu x{r_temp4}, x{r_temp3}, x{r_temp1}",
-        f"add x{r_temp2}, x{r_temp2}, x{r_temp4}",
-        f"sw x{r_temp3}, 0(x{r_mtimecmp})",
-        f"sw x{r_temp2}, 4(x{r_mtimecmp})",
+        f"sw x{r_temp1}, 0(x{r_mtimecmp})",  # Write -1 to low word first
+        f"sw x{r_temp2}, 4(x{r_mtimecmp})",  # Write high word
+        f"sw x{r_temp3}, 0(x{r_mtimecmp})",  # Write low word (final value)
```
jordancarlin, line 100: Seems like some of these comments would be useful in the generated assembly.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,360 @@
...
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_user_mti_tests(test_data: TestData) -> list[str]:
+    """Generate undelegated MTI interrupt tests from U-mode."""
+    covergroup = "InterruptsU_cg"
+    coverpoint = "cp_user_mti"
+
+    r_mtime, r_mtimecmp, r_temp, r_temp2, r_scratch = test_data.int_regs.get_registers(5, exclude_regs=[0, 2])
```
jordancarlin: Why is 2 excluded? Worth a comment.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,360 @@
...
+                    f"CSRW(mie, x{r_scratch})",
+                    "RVTEST_GOTO_LOWER_MODE Umode",
+                    test_data.add_testcase(binname, coverpoint, covergroup),
+                    *set_mtimer_int(r_mtime, r_mtimecmp, r_temp, r_temp2),
+                ]
+            )
+
+            lines.extend(
+                [
+                    f"    LI(x{r_scratch}, 2500)",  # 2500 iterations × 2 instructions = 5000 cycles
```
jordancarlin, line 67: Rather than hardcode 2500 iterations, shouldn't this use `RVMODEL_INTERRUPT_LATENCY`?

jordancarlin: Actually, shouldn't we update `RVTEST_IDLE_FOR_INTERRUPT` to have this loop and then drop the `RVTEST_IDLE_FOR_INTERRUPT` that comes right after this? It doesn't make sense to do both back to back.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -0,0 +1,360 @@
...
+            lines.extend([f"LI(x{r_scratch}, 0x80)", f"CSRW(mie, x{r_scratch})", "RVTEST_GOTO_LOWER_MODE Umode"])
+
+            # WFI - label right before
+            lines.extend(
+                [
+                    *set_mtimer_int_soon(r_mtime, r_mtimecmp, r_temp, r_t1, r_t2, r_temp2),  # Set timer to fire soon
+                    test_data.add_testcase(binname, coverpoint, covergroup),
+                    "    wfi",
+                    "    nop",
+                    "    RVTEST_GOTO_MMODE",
```
jordancarlin: You can drop the spaces.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
SadhviNarayanan: I'd have to change a few macro files, so I think I'll maybe address this with InterruptsS because I will be changing the way I use the macros for this.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -48,7 +49,7 @@ def clr_mtimer_int(r_temp: int, r_mtimecmp: int) -> list[str]:
         r_mtimecmp: Register number to hold MTIMECMP address
     """
     return [
-        f"{INDENT}# Clear machine timer interrupt",
+        f"{INDENT} # Clear machine timer interrupt",
```
jordancarlin:
```suggestion
        f"{INDENT}# Clear machine timer interrupt",
```

jordancarlin (approved):

PR #1063, opened 2026-03-10
add xtvec_mode and Zicfiss to sail-rv32i template

ekattsim:
The common-test signature generation for RV32I uses an outdated schema in the template.

sail-rv32i.json is missing two now-required properties:
- base.reserved_behavior.xtvec_mode
- extensions.Zicfiss

### DIFF
--- a/framework/src/act/data/sail-rv32i.json
+++ b/framework/src/act/data/sail-rv32i.json
@@ -49,6 +49,10 @@
       // "Xenvcfg_Fatal"  – raise a Sail exception, stopping execution.
       // "Xenvcfg_ClearPermissions" – convert CBIE with 0b10 to 0b00.
       "xenvcfg_cbie": "Xenvcfg_ClearPermissions",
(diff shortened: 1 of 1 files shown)


jordancarlin:
The `sail.json` files in the repo are compatible with the 0.10 release of the RISC-V sail model: https://github.com/riscv/sail-riscv/releases/tag/0.10. We will update them the next time there is a release of sail-riscv, but don't plan to update them every time master changes because it would frequently break compatibility with the currently released version.

ekattsim:
I see, you're right! I was able to get the tests generated when using the 0.10 release of the sail model.

Thanks a lot!

jordancarlin:
> I see, you're right! I was able to get the tests generated when using the 0.10 release of the sail model.
> 
> Thanks a lot!

Great! Glad to hear you got the tests generating. Let us know if there is anything that you think can be improved in the documentation to make it easier to get started.

PR #1071, opened 2026-03-11
Fix/cv32e20 boot addr jal

karabambus:
I-jal generated test from ACT4 uses .align 14, requiring .text at a 16KB-aligned address. Therefore 0x2000 value in linker is wrong and needs to be moved to 0x4000.

Sail doesn't use a hardcoded reset vector - it loads the ELF and starts executing from the rvtest_entry_point symbol directly, wherever the linker placed it. So Sail was always running from 0x4000 (following the ELF), while the DUT was reset-fetching from 0x2000, a completely different address. That is why Sail produced a correct signature while the DUT hung.

### DIFF
--- a/config/cores/cve2/cv32e20/link.ld
+++ b/config/cores/cve2/cv32e20/link.ld
@@ -3,7 +3,7 @@ ENTRY(rvtest_entry_point)
 
 SECTIONS
 {
-  . = 0x00002000;
+  . = 0x00004000;
   .text : { *(.text.init) *(.text) }
   . = ALIGN(0x4000);
   .bss : { *(.bss) }


jordancarlin (approved):
The constraint on the starting address is a known issue that no one has been able to solve yet. For now, this looks good.

karabambus:
@jordancarlin @davidharrishmc Maybe it should be noted that for JAL tests, boot address on DUT should be set to 0x4000

jordancarlin:
> @jordancarlin @davidharrishmc Maybe it should be noted that for JAL tests, boot address on DUT should be set to 0x4000

This is fixed in #1089. @karabambus can you test CV32E20 with its normal boot address (0x1000 if I recall) and update the linker script if that is working correctly now? Wait until #1089 is merged first.

karabambus:
@jordancarlin Tested it, it is working. Waiting for to resolve following [issue](https://github.com/openhwgroup/cv32e20-dv/issues/30) on cv32e20 repo before we can revert to old configuration.

PR #1074, opened 2026-03-11
Update cv32e20 to use Zca instead of C extension

karabambus:
Removed 'C' extension to prevent test generator from incorrectly generating floating-point compressed instructions (c.flw/c.fsw) on FPU=0 configuration.


davidharrishmc (approved):

davidharrishmc:
You'll need to update your branch for this to merge.

PR #1075, opened 2026-03-11
Running PMPSm on Spike in CI

UmerShahidengr:


davidharrishmc (approved):

tests/priv/pmp/pmp32/PMPmisaligned/pmpsm_misaligned_na4.S
jordancarlin, line 1: Moving these to PMPmisaligned is going to break coverage. Should we just add `DO_NOT_RUN` (or something similar) as one of the required extensions so that the test selector never chooses the test? And then we can remove that once the new config option is added.

tests/priv/pmp/pmp64/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -65,563 +65,485 @@ RVTEST_BEGIN
...
+
+#endif  // RVMODEL_NUM_PMPS == 64
+
+// ===========================================================================
+// TEST CASES — 16-PMP ONLY (slots 14-0, test cases 1-15)
+// slot 15 is the background region; we test slots 14 down to 0.
+// pmpcfg2 covers pmpaddr8-15  (RV64: one 64-bit CSR = 8 entries)
+// pmpcfg0 covers pmpaddr0-7
+// ===========================================================================
+#if RVMODEL_NUM_PMPS == 16
```
jordancarlin, line 482: Rather than duplicating a lot of the testcases, can PMP regions 14-0 be run unconditionally in this file and just have 63-15 be gated on `RVMODEL_NUM_PMPS == 64`?

jordancarlin (changes requested):

tests/priv/pmp/pmp64/PMPSm/pmpsm_all_entries_check.S
UmerShahidengr: Sorry, just saw these comments, We will address them in the next iteration

tests/priv/pmp/pmp32/PMPmisaligned/pmpsm_misaligned_na4.S
UmerShahidengr: Nice option, will do it in the next iteration along with other tests as this problem will keep popping

PR #1082, opened 2026-03-12
100% coverage for PMPSm and PMPZca

hamza-1821:
***UPDATE****
Some larger tests have been split further, to adjust addresses for coverage. 

For rv64, some of the tests have a larger code region than largest porgram defined in config file. These files are:
1) pmpsm_cfg_XWR_all-01, 02, 03, 04
2) napot legal lwxr
so coverage could not be acheived for these files.
There is a small discrepancy in pmpsm_csr_walk.S test, after we removed loops from it and tried to write it without loops. I will get that in a follow up PR. 
Apart from above mentioned stuff, we should now have full coverage.


UmerShahidengr (approved):
LGTM. Thank you @hamza-1821

PR #1084, opened 2026-03-13
Bug fix & Error reporting improvement in Self-checking Trap handler

Zain2050:
Using `CSR_XEPC` in `failure_code.h` was a bug. CSR_XEPC is hardcoded during compilation to the last definition of CSR_XEPC. and therefore it couldn't handle mismatches that occured in different priv modes. Now with this change, it will read the correct xepc value depending upon the handler mode.

However, I have found a major problem. Along with xepc we are also printing out the instruction at that address. However, what if that instruction doesn't have PMA or PMP permissions (or PTE permissions if being handler in mode!=M). Also whenever we trap while satp.mode!=bare, xepc will contain a virtual address and if that trap is being handled in M-mode, it will raise another fault if we tried to access that virtual address from M-mode. 

For fetch access faults, we can check xcause and read instr at xepc only if its not a fetch access fault. Similary, a solution for Sv could be to only read instr at xepc if satp.mode=bare. Another solution could be to set xstatus.xprv allowing us to load using a virtual address. But all of this is going to require some complex code just to print out the instruction at xepc. 

@jordancarlin Do you think printing instruction at xepc is worth it? Because instructions at xepc are mostly gonna be faulty.


allenjbaum:
In bare mode, VA=PA, so the address in SATP is the PA. But, in any case, if
the exception code is instruction-access-fault, (code=1) then the
instruction at that address is not the cause and there is no need to print
it
The instruction at mepc is probably what you want to print, as that is the
one that caused the attempted fetch of the illegal address.

jordancarlin:
> However, I have found a major problem. Along with xepc we are also printing out the instruction at that address. However, what if that instruction doesn't have PMA or PMP permissions (or PTE permissions if being handler in mode!=M). Also whenever we trap while satp.mode!=bare, xepc will contain a virtual address and if that trap is being handled in M-mode, it will raise another fault if we tried to access that virtual address from M-mode.
> 
> For fetch access faults, we can check xcause and read instr at xepc only if its not a fetch access fault. Similary, a solution for Sv could be to only read instr at xepc if satp.mode=bare. Another solution could be to set xstatus.xprv allowing us to load using a virtual address. But all of this is going to require some complex code just to print out the instruction at xepc.
> 
> @jordancarlin Do you think printing instruction at xepc is worth it? Because instructions at xepc are mostly gonna be faulty.

This is an interesting one. I definitely think that in general printing the instruction at XEPC is useful because it tells the user what instruction went wrong. Agreed that is not simple to do for fetch access faults and may also be complicated when virtual memory is enabled. If it is easy to print it out in most cases and skip the cases where we expect it to cause trouble, maybe we start with that? Then users will at least get helpful debug information most of the time.

jordancarlin (approved):
There are still further things to be fixed, but this is an improvement so I'm merging and we can follow up with another PR.

Zain2050:
> In bare mode, VA=PA, so the address in SATP is the PA. But, in any case, if
> the exception code is instruction-access-fault, (code=1) then the
> instruction at that address is not the cause and there is no need to print
> it
> The instruction at mepc is probably what you want to print, as that is the
> one that caused the attempted fetch of the illegal address.

 @allenjbaum Whenever we jump to an illegal address, mepc is going to hold that illegal address instead of the address of the jump.

Zain2050:
> This is an interesting one. I definitely think that in general printing the instruction at XEPC is useful because it tells the user what instruction went wrong. Agreed that is not simple to do for fetch access faults and may also be complicated when virtual memory is enabled. If it is easy to print it out in most cases and skip the cases where we expect it to cause trouble, maybe we start with that? Then users will at least get helpful debug information most of the time.

@jordancarlin Yes sure, we can add some logic to print the instruction when possible. We could check xcause and not print the instruction if its a fetch access fault. Similary, if we trapped while virtual memory was enabled we could skip it etc.

allenjbaum:
Hmm. The spec says:

If mtval is written with a nonzero value when an instruction access-fault,
page-fault, or hardware-error

exception occurs on a hart with variable-length instructions, then mtval
will contain the virtual address of

the portion of the instruction that caused the fault, while mepc will point
to the beginning of the

instruction

So, I am actually unclear how to interpret this.
IF it were a load instruction, xepc would hold the address of the load, and
xtval would hold the target address
Is the instruction that caused the fault the jump instruction, or whatever
is at that branch target?
This sounds like mepc and mtval will hold the same value, unless it crossed
a PMA or PTE boundary, in which case the will be off by 2.

AI says:
In a RISC-V processor, if a branch at address A branches to a target
address B that has no execute permissions (due to PMP - Physical Memory
Protection or MMU - Virtual Memory page tables), a fetch page-fault
exception or instruction access-fault exception occurs.
The following values are placed in the CSRs (Control and Status Registers):

   - mepc CSR (Machine Exception Program Counter): The address of the
   branching instruction itself, address A.
   - mtval CSR (Machine Trap Value): The invalid branch target address, address
   B.
   [image: Google Groups]Google Groups +2

Detailed Summary:

   - Exception Cause: The mcause register will be set to indicate a "Fetch
   Page Fault" (code 12) or "Instruction Access Fault" (code 1).
   - Why mepc = A: In RISC-V, when an instruction fails to execute because
   of a fault (as opposed to being unaligned), the mepc points to the
   address of the instruction that caused the trap.
   - Why mtval = B: The mtval register is updated with the specific,
   invalid address that the processor attempted to fetch from.
   [image: GitHub]GitHub +3

*Note: In rare cases where compressed instructions are used and the branch
targets an address straddling a page boundary, mepc might point to the
start of the 4-byte instruction, while mtval points to the specific faulty
page boundary address.*

Zain2050:
Section 2.5 of the Unpriv Spec states that:

> If an instruction access-fault or instruction page-fault exception occurs on the target of a jump or
> taken branch, the exception is reported on the target instruction, not on the jump or branch instruction.

PR #1085, opened 2026-03-13
Rev 1.0 of RVI20 CRD. Applied relevant changes to MC100 CRD too.

james-ball-qualcomm:
- Set allowed values for MISALIGNED_LDST_HW_SUPPORT to be "Any" to be consistent with other parameters.
- Moved HPM_COUNTER_SZ to IN-SCOPE parameters and add "and will not trap if read." to its parameter definition.
- Bumped up rev from 0.8 to 1.0

### DIFF
--- a/docs/crd/src/rvi20_crd.adoc
+++ b/docs/crd/src/rvi20_crd.adoc
@@ -1,4 +1,4 @@
-= RVI20 Certification Requirements Document (rev 0.8)
+= RVI20 Certification Requirements Document (rev 1.0)
 :description: RVI20 CRD (Certificate Requirements Document)
 include::../../common/adoc_common.adoc[]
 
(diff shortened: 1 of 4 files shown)


docs/crd/src/rvi20_crd.adoc
```diff
@@ -152,12 +157,16 @@ include::in_scope_param_intro.adoc[]
 |===
 | Parameter | Allowed Value(s) | Impact on Certificate Scope
 
+| <<param:HPM_COUNTER_SZ,HPM_COUNTER_SZ>>
+| Any
+| Only counters with a size of 0 are IN-SCOPE since the behavior of HPM counters is implementation-specific.
```
jordancarlin, line 162: I don't think the size influences whether or not the HPM counter will trap. It influences what values will be read from it, but can't we test reading from all of the counters to ensure none of them trap? Or is there a piece of the spec that I missed that says something different?

jordancarlin: Merging since this was approved as v1.0 by the requirements workgroup. We can continue to refine parameters as the Parameters SIG continues its work.

jordancarlin (approved):

PR #1091, opened 2026-03-15
Svadu, SvaduPMP, Svnapot, Svpbmt, SvZicbo ACTs

Zain2050:
* This PR adds ACTs for Svadu, SvaduPMP, Svnapot, Svpbmt & SvZicbo.
* Added non-leaf PTE DAU bits fault tests, but currently disabled using "NORUN" because Sail doesn't support it yet. Next Sail release will have this support.
* Excluded these from being ran on QEMU and Spike in CI because Sail doesn't support these yet.


tests/priv/Sv/sv32_nleaf_pte_DAU_Smode.S
jordancarlin, line 1: Are there new coverpoints for this new testcase that is bringing added? Or were there existing coverpoints that we were not hitting?

tests/priv/Sv/sv32_nleaf_pte_DAU_Smode.S
```diff
@@ -0,0 +1,184 @@
...
+// 2. PTE with RWX permissions set up at level 0, PTE.A is set at level 1 (non-leaf):
+//        Then, in S-Mode, the page is accessed --> required: Load-page-fault, Store-page-fault, Fetch-page-fault
+// 3. PTE with RWX permissions set up at level 0, PTE.U is set at level 1 (non-leaf):
+//        Then, in S-Mode, the page is accessed --> required: Load-page-fault, Store-page-fault, Fetch-page-fault
+//
+// Total Expected Faults :: 9
+// --------------------------------------------------------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Sv32, NORUN]
```
jordancarlin, line 18: Add a comment anywhere that NORUN is used explaining why the test is disabled and what needs to be fixed before it can be reenabled.

tests/priv/SvaduPMP/sv32_Svadu_no_pmp_perm_Smode.S
```diff
@@ -0,0 +1,299 @@
...
+// 5. Setup a PTE at Level 0 with PTE.D unset. RWX permissions given
+//      → Then, in S-Mode, the page is accessed --> required: Store access fault
+// 6. Setup a PTE at Level 0 with PTE.D & PTE.A unset. RWX permissions given
+//      → Then, in S-Mode, the page is accessed --> required: Store access fault, Load access fault, Instruction access fault
+//
+// Total Expected Faults :: 14
+// --------------------------------------------------------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Sv32, Svadu]
```
jordancarlin: For tests that use PMP, please add Sm as one of the required extensions so we can filter them out for certification (which will not test all M-mode features).

tests/priv/Svnapot/sv39_Svnapot_reserved_enc_Smode.S
```diff
@@ -0,0 +1,243 @@
...
+  // Signature Update
+  RVTEST_SIGUPD(x2, x5, x4, a2, \TEST_CASE\()_store, \TEST_CASE\()_store_str)
+  RVTEST_SIGUPD(x2, x5, x4, a3, \TEST_CASE\()_load, \TEST_CASE\()_load_str)
+  RVTEST_SIGUPD(x2, x5, x4, a4, \TEST_CASE\()_exec, \TEST_CASE\()_exec_str)
+
+.endm
+
+
+main:
+  li a2, 0x800              // Test signature initialization
```
jordancarlin: Shouldn’t this use the LI macro?

tests/priv/Svnapot/sv57_Svnapot_reserved_enc_Smode.S
```diff
@@ -0,0 +1,298 @@
...
+// 7. Setup a PTE at level 0 with PTE.N set and PTE.PPN[0][3:0] (NAPOT encoding) set to 4'b0100 (reserved), RWX permissions given:
+//         Then, in S-Mode, the page is accessed --> required: Store, Load & Fetch page fault
+// 8. Setup a PTE at level 0 with PTE.N set and PTE.PPN[0][3:0] (NAPOT encoding) set to 4'b0000 (reserved), RWX permissions given:
+//         Then, in S-Mode, the page is accessed --> required: Store, Load & Fetch page fault
+//
+// Total Expected Faults :: 24
+// --------------------------------------------------------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Sv57, Svnapot]
```
jordancarlin: Should all of the Svnapot tests have a NORUN for now since Sail doesn’t support it yet?

tests/priv/Svpbmt/sv39_Svpbmt_nonleaf_Smode.S
```diff
@@ -0,0 +1,243 @@
...
+// 5. Setup a PTE at level 0 with RWX permissions, with PTE.PBMT=2 of non-leaf PTE (at level 1):
+//     Then, in S-Mode, the page is accessed --> required: Store, load & fetch page fault
+// 6. Setup a PTE at level 0 with RWX permissions, with PTE.PBMT=3 of non-leaf PTE (at level 1):
+//     Then, in S-Mode, the page is accessed --> required: Store, load & fetch page fault
+//
+// Total Expected Faults :: 18
+// --------------------------------------------------------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Sv39, Svpbmt]
```
jordancarlin: Should all of the Svpbmt tests have a NORUN for now since Sail doesn’t support it yet?

.github/workflows/regress.yml
```diff
@@ -174,7 +174,7 @@ jobs:
           DOCKER=1 \
           FAST=True \
           EXTENSIONS="" \
-          EXCLUDE_EXTENSIONS=Sm,ExceptionsZc,S,InterruptsSm,PMPmisaligned,ExceptionsZalrsc \
+          EXCLUDE_EXTENSIONS=Sm,ExceptionsZc,S,InterruptsSm,PMPmisaligned,ExceptionsZalrsc,Svadu,SvaduPMP,Svnapot,Svpbmt,SvZicbo \
```
jordancarlin, line 177: What is wrong with the SvZicbo tests? They should be fully supported by all the simulators.

tests/priv/Svadu/sv32_Svadu_Smode.S
```diff
@@ -0,0 +1,290 @@
...
+// 5. PTE.D unset, PTE.A set for three distinct leaf PTEs at level 0, RWX permissions (read, write, execute page):
+//        Then, in S-Mode, the pages are accessed --> required: No fault
+// 6. Both PTE.A & PTE.D unset for three distinct leaf PTEs at level 0, RWX permissions (read, write, execute page):
+//        Then, in S-Mode, the pages are accessed --> required: No fault
+//
+// Total Expected Faults :: 0
+// --------------------------------------------------------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Sv32, Svadu]
```
jordancarlin: Should all of the Svadu tests have a NORUN for now since Sail doesn’t support it yet?

jordancarlin (changes requested):

tests/priv/Sv/sv32_nleaf_pte_DAU_Smode.S
Zain2050: There are existing coverpoints.

tests/priv/Svnapot/sv39_Svnapot_reserved_enc_Smode.S
Zain2050: The compiler replaces it with lui & addi, therefore storing the correct value. Let's say even if it doesn't, its not much of a problem as it's just supposed to be some value that will be incremented by 16 after every testcase. Won't break anything.

jordancarlin: I agree that it usually won't be a problem, but we tend to prefer the `LI()` macro over the `li` pseudoinstruction to make the tests more compiler agnostic.

tests/priv/SvaduPMP/sv32_Svadu_no_pmp_perm_Smode.S
Zain2050: **Sv32** can't exist without **S** & I guess **S** can't exist without **Sm**. So do I need to explicitly include **Sm** here?

tests/priv/Svnapot/sv57_Svnapot_reserved_enc_Smode.S
Zain2050: These extensions just show different behavior (faults when not expecting a fault and vice versa). They run fine and pass without aborting or getting stuck. Should I disable them?
If yes, then is adding NORUN to every test required, can't we just add these to EXCLUDE_EXTENSIONS in CI?

tests/priv/SvaduPMP/sv32_Svadu_no_pmp_perm_Smode.S
jordancarlin: That's a good question. While S cannot exist without Sm, for certification we are not testing M-mode features (including things like PMP). That means we need a way of identifying tests that rely on M-mode features so that we know which tests to skip. The current thinking is that we will skip any tests that explicitly list Sm as one of the required extensions. Since this tests uses PMP, it should be skipped and therefore needs to include Sm. This only applies to the SvaduPMP and SvPMP tests, not the rest of the virtual memory tests.

.github/workflows/regress.yml
Zain2050: SvZicbo tests for a lot of exception scenerios including setting DAU bits in non-leaf PTEs. This is the only reason for disabling it. It will work with the next release of Sail.

tests/priv/Svnapot/sv57_Svnapot_reserved_enc_Smode.S
jordancarlin: My concern is for any user that has a DUT that implements Svnapot. They will list Svnapot in their UDB config file which will cause these tests to be included, but they will fail because the results from Sail will not match. I think we need to add NORUN so that these tests are not selected for DUTs.

tests/priv/Sv/sv32_nleaf_pte_DAU_Smode.S
Zain2050: Sure I can add the comments. However, I was thinking of including a markdown file in the Sv folder specifying which tests are disabled and for what reason. Otherwise you won't know if tests are disabled within the folder without searching for NORUNs. What do you say?

tests/priv/SvaduPMP/sv32_Svadu_no_pmp_perm_Smode.S
Zain2050: Okay, makes sense.

tests/priv/Sv/sv32_nleaf_pte_DAU_Smode.S
jordancarlin: Including the markdown file seems fine if you want, but I think the comments are probably worthwhile too so that it is easier to see when we should be removing each NORUN.

jordancarlin (approved):

PR #1107, opened 2026-03-21
Use native UDB gem and drop container dependency

jordancarlin:
Uses the native UDB Ruby gem instead of the UDB submodule. This makes the dependencies much easier and increases performance.

The gem requires Ruby to be installed. To make this easy for users and avoid needing to install both uv and Ruby, the recommended approach is to use mise, which will handle all of the other tools automatically.


jordancarlin:
Currently a draft pending the next release of the udb gem. The current version has a bug, but with a local patch to the gem this is working. Once that patch is upstreamed and a new version of the gem is released, this should be good to go.

davidharrishmc (approved):

PR #1117, opened 2026-03-24
ExceptionsZc 100% coverage, added nop to load/store access

juliacygong:


generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -153,7 +153,8 @@ def _add_load_fault(
         t_lines.append("LI(sp, RVMODEL_ACCESS_FAULT_ADDRESS)")
         t_lines.append(test_data.add_testcase(test_label, coverpoint, covergroup))
         t_lines.append(f"{op} {reg_str}, 0(sp)")
-        t_lines.append(f"{INDENT}# Trap handler skips the next 4 bytes; two c.nops provide 4 bytes")
+        t_lines.append(f"{INDENT}# Trap handler skips the next 6 bytes; two c.nops provide 6 bytes")
```
davidharrishmc: File says two nops but has three. Many places.

davidharrishmc (changes requested):

davidharrishmc (approved):

generators/testgen/src/testgen/priv/extensions/ExceptionsZc.py
```diff
@@ -162,10 +165,11 @@ def _add_load_fault(
         t_lines.append(test_data.add_testcase(test_label, coverpoint, covergroup))
         t_lines.append(f"{op} {reg_str}, 0(x{addr_reg})")
         t_lines.append(
-            f"{INDENT}# Load access will throw a trap and the trap handler skips over the next 4 bytes. Two c.nops are used to get 4 bytes of instructions"
+            f"{INDENT}# Load access will throw a trap and the trap handler skips over the next 6 bytes. Three c.nops are used to get 6 bytes of instructions"
```
jordancarlin: Still not clear why this skips 6 bytes.

jordancarlin (approved):

PR #1134, opened 2026-03-26
Fix #1103: more flexibility in allowable values in Sm counter and hpm tests

### LINKED ISSUE #1103: Sail handling of implementation-defined CSR reset values (e.g. mcountinhibit)
# Context
While running ACT4 Zicntr certification tests for CV32E40P, the Zicntr-csrrc-00 and Zicntr-csrrs-00 tests failed under RTL simulation (QuestaSim) (but passed under Verilator).

CV32E40P resets mcountinhibit to 0x1 (counters inhibited by default, for power saving). Sail assumes 0 (counters running). Running Zicntr tests without addressing this causes counter-increment checks to fail on RTL simulation.

# Problem
The Sail reference model resolves all UNDEFINED_LEGAL CSR reset values to 0. For cores where certain CSRs reset to non-zero values (a valid, spec-compliant choice), this creates a silent divergence between DUT and reference model at test start - causing test failures that are not bugs in the core or the tests.

davidharrishmc:


generators/testgen/src/testgen/asm/csr.py
```diff
@@ -181,3 +182,60 @@ def csr_walk_test(test_data: TestData, csr_name: str, covergroup: str, coverpoin
...
+        covergroup: Covergroup name for testcase strings
+        coverpoint: Coverpoint name for testcase strings
+
+    Returns:
+        List of assembly lines for the access test
+    """
+    save_reg, temp_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        f"\n# CSR Access Tests for {csr_name}",
```
jordancarlin: Use an empty line instead of a leading `\n`:
```suggestion
        "",
        f"# CSR Access Tests for {csr_name}",
```

jordancarlin (commented):
One minor style comment.

jordancarlin (approved):

PR #1146, opened 2026-03-27
Only run sfence.vma in setup code if S is supported

jordancarlin:


tests/env/test_setup.h
```diff
@@ -57,7 +57,9 @@
         CSRW(pmpcfg0, 0xF)   // configure PMP0 to TOR RWX
         LI(t0, -1)
         CSRW(pmpaddr0, t0)   // configure PMP0 top of range to 0xFFF...FFF to allow all addresses
-        sfence.vma
+        #ifdef S_SUPPORTED
```
davidharrishmc: Note that fence.vma might trap even if S is supported.  There should be a way to make sure that is ok.  One approach is to issue it no matter what, but the trap handler should be set up to trap if appropriate, and the reference model should be configured to match the DUT.  Not sure that config exists yet.

```
For implementations that make satp.MODE read-only zero (always Bare), attempts to execute an
SFENCE.VMA instruction might raise an illegal-instruction exception.
```

davidharrishmc (changes requested):

tests/env/test_setup.h
jordancarlin: The parameter for that exists. UDB call it `TRAP_ON_SFENCE_VMA_WHEN_SATP_MODE_IS_READ_ONLY` and Sail calls it `sfence_vma_illegal_if_svbare_only`.

I think a cleaner approach is to have it only run that instruction if SV32_SUPPORTED or SV39_SUPPORTED. That way it should never trap and is only run when needed.

davidharrishmc:
Perfect.  That works with PMP as well, which requires fence on any hart that implements page-based virtual memory.

tests/env/test_setup.h
```diff
@@ -57,7 +57,9 @@
         CSRW(pmpcfg0, 0xF)   // configure PMP0 to TOR RWX
         LI(t0, -1)
         CSRW(pmpaddr0, t0)   // configure PMP0 top of range to 0xFFF...FFF to allow all addresses
-        sfence.vma
+        #if defined(SV32_SUPPORTED) || defined(SV39_SUPPORTED)
```
davidharrishmc, line 64: Could you add a comment why this is needed: "sfence.vma is required after PMP entries are changed to sync the PMP with the virtual memory system and any PMP or address translation caches.  sfence.vma should not be performed in a system that does not support virtual memory because it might raise an illegal instruction."

davidharrishmc (approved):

PR #1150, opened 2026-03-29
refactor(cores): restructure cv32e40p into cve4

karabambus:
Move cv32e40p/ contents directly into cve4/, renaming variants to use
dash-separated short names (v1/v2, rv32imc/rv32imcf). Add F_SUPPORTED to
rvtest_config.h for the rv32imcf variant (real file); others use symlinks.


davidharrishmc (approved):

config/cores/cve2/cv32e40pv2/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## DUT Configuration for the CV32E40P v2
+
+Two configurations are provided:
+
+| Config                | ISA      | Notes                                                  |
+| --------------------- | -------- | ------------------------------------------------------ |
+| `cv32e40pv2-rv32imc`  | RV32IMC  | v1.8.3 release, logically equivalent to v1.0.0_rv32imc |
+| `cv32e40pv2-rv32imcf` | RV32IMCF | v1.8.3 release with FPU (F + Zcf)                      |
+
```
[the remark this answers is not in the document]
karabambus: @davidharrishmc it seems to me that copilot is wrong. Can you confirm?

config/cores/e40pv1/README.md
jordancarlin, line 1: This directory structure seems more confusing. All of these cores are related and should probably still be grouped together. Having `e40pv1` at the root level also makes it harder to associate with the OpenHW Foundation because the name no longer starts with “cv”. What is the motivation behind this?

jordancarlin (changes requested):

config/cores/e40pv1/README.md
karabambus: This [table](https://github.com/openhwgroup/core-v-cores/blob/master/OpenHW-CoresTimeline-Feb2024.png) has cores grouped like this.  I would like to have cv32e40p follow same directory structure as rest of cores. Though I noticed there is some discrepancy how this familys are described in table and [core roadmap](https://github.com/openhwgroup/core-v-cores?tab=readme-ov-file).  I will draft this until @MikeOpenHWGroup reviews.

davidharrishmc:
If it renders properly for you, click resolve to show you are overriding
AI. At least 10% of the AI feedback is wrong. But it is always with running
copilot on your PRs and you can configure GitHub to request a copilot
review automatically when you open the PR.

config/cores/e40pv1/README.md
karabambus: So as I understand this image was wrong, now it is removed from openhw page. I still think there should be [core family]/[core] structure. cv32e40p cores should belong to cve4. So this will be my change to PR. I would like Mike to review this after its ready.

MikeOpenHWGroup: Yes, cve4/cv32e40p_v* makes sense.

MikeOpenHWGroup:
Hi @karabambus, is this PR ready for a formal review?  If so, please remove the 'Draft' label.  Thanks.

config/cores/cve4/rvtest_config.h
jordancarlin, line 1: This file depends on the implemented extensions (which is different for the different configs), so should not be shared. Same for `rvtest_config.svh`.

config/cores/cve4/README.md
jordancarlin, line 1: The README needs to be updated to use the new directory structure.

jordancarlin (changes requested):
I left a few comments. Otherwise, deferring to @MikeOpenHWGroup to review to make sure this arrangement makes sense for the OpenHW cores.

config/cores/cve4/rvtest_config.h
karabambus: rv32_imc v1 and v2 share same config, so they are correctly sharing it. For rv32_imcf it has seperate rvtest_config.h with F defined. For rvtest_config.svh it should be same for all 3 configurations, am I missing something?

jordancarlin: `rvtest_config.svh` should include `` `define F_SUPPORTED `` where relevant. For the imc v1 and v2 configs, I'd be in favor of putting the main file in one of them and then symlinking the other to that. Putting it in the root makes it seem like it applies to all of the cores.

karabambus:
This is ready @MikeOpenHWGroup . @jordancarlin  I added symlink changes

MikeOpenHWGroup (approved):
This LGTM @karabambus, thanks for the clean-up.

It need not gate perging this PR, but I'd like to understand the purpose of `rvtest_config.svh`.  AFAIK, this SystemVerilog file is _not_ used by the ACT4 tools to generate test-programs.  Are we requiring this file to be "tick included" into the certification client's testbench and that the client's testbench should use these macros in a specific way?

davidharrishmc:
rvtest_config.svh is included during coverage.  It can be used to adjust what is covered, such as whether
to measure certain optional subfeatures.  The coverage files already use these `defines and the
testbench typically would not be aware of them.

MikeOpenHWGroup:
Ah.  I get it now.

> The coverage files already use these `defines and the testbench typically would not be aware of them.

If the coverage files use these macros _and_ the certification client is expected to produce `rvtest_config.svh`, then we will need to clearly document these.  A quick look indicates that the "right" place for this documentation would be `act4/docs/ctp/src/rvmodel.adoc`.  If that is the case, I can take a run at generating a PR for this. LMK if that would be helpful, or just a hindrance.

jordancarlin:
> If the coverage files use these macros _and_ the certification client is expected to produce `rvtest_config.svh`, then we will need to clearly document these. A quick look indicates that the "right" place for this documentation would be `act4/docs/ctp/src/rvmodel.adoc`. If that is the case, I can take a run at generating a PR for this. LMK if that would be helpful, or just a hindrance.

We're getting close to auto-generating `rvtest_config.svh` (and `rvtest_config.h`) from the UDB yaml config (see https://github.com/riscv/riscv-unified-db/pull/1777), so I'd say it's probably not worth the effort of documenting it at this point.

jordancarlin (approved):

PR #1152, opened 2026-03-29
Tests and Coverage files for Svbare

Zain2050:
* Moved satp.MODE=Bare tests and coverpoints from Sv to Svbare. 
* Added 2 additional test cases which perform accesses in S & U modes while satp.MODE=Bare.


coverpoints/priv/Svbare_coverage.svh
```diff
@@ -0,0 +1,69 @@
...
+            bins not_zero = {[1:$]};
+        }
+    `else
+        satp_asid_ppn: coverpoint ins.current.csr[12'h180][30:0] {
+            bins all_zero = {31'b0};
+            bins not_zero = {[1:$]};
+        }
+    `endif
+
+    satp_bare_mode:       cross satp_bare, satp_asid_ppn;
```
davidharrishmc: Not sure nonzero satp_asid_ppn is testable.  The behavior is UNSPECIFIED.

"Attempting to select MODE=Bare with a nonzero pattern in the remaining fields has an
UNSPECIFIED effect on the value that the remaining fields assume and an UNSPECIFIED effect on
address translation and protection behavior."

tests/priv/Svbare/Svbare_Smode.S
```diff
@@ -0,0 +1,104 @@
...
+  nop
+
+  // Test execute permission
+  LI(  x4, 0xACCE)            // Store a value which is to be checked in trap handler
+  \TEST_CASE\()_exec:
+  jalr ra, a5, 0
+  nop
+
+  // Signature Update
+  RVTEST_SIGUPD(x2, x5, x4, a2, \TEST_CASE\()_store, \TEST_CASE\()_store_str)
```
davidharrishmc, line 53: I think the SIGUPD should always go right after the test (e.g. the nop).

For store tests, checking a2 doesn't show that the test was correct.  We need to read back the value from memory.  Jordan has examples of that in the I/sw tests.

This likely affects many Sv tests.

tests/priv/Svbare/Svbare_Smode.S
```diff
@@ -0,0 +1,104 @@
...
+  nop
+  .word 0xbeefcaf1          // Random word
+  .word 0xbeefcaf2          // Random word
+  nop
+  jr ra
+
+//---------------------------------------------------------------------------------------------------------------------------------
+
+satp_bare_str:     .string "\"Mismatch in satp value!\""
+test1_store_str:   .string "\"Mismatch during sw in Test Case 1!\""
```
davidharrishmc, line 99: Not essential for this PR, but these tests would be better if the message referenced the coverpoint and bin (e.g. satp_bare_load/S_mode). 
If not now, can that go on the wish list?

coverpoints/priv/Svbare_coverage.svh
davidharrishmc: We have gravitated toward cp_ prefixes for the primary coverpoints in other extensions?  What do you think about adding them for Sv as well?

davidharrishmc (changes requested):

coverpoints/priv/Svbare_coverage.svh
Zain2050: I just moved it over from Sv coverage file. I'll remove it.

tests/priv/Svbare/Svbare_Smode.S
Zain2050: The value read back (into a3) is also being stored as a signature. a2 is just incremented on every instantiation of the verification macro. Storing that signature is just a sanity check to make sure we are in the correct test case.

tests/priv/Svbare/Svbare_Smode.S
Zain2050: These can be fixed in a future PR as there are a lot of other Sv coverpoints as well.

davidharrishmc (approved):

PR #1153, opened 2026-03-29
docs: add notes for local non-root setup to prevent permission issues

DymShanks:
Hi @jordan-carlin,

While doing a clean run-through of the act4 setup documentation on a local Ubuntu environment, I found that the current instructions for compiling GCC and extracting the Sail model default to system-wide installations requiring sudo.

While this is standard for CI/CD pipelines, it causes immediate friction for individual contributors setting up local environments. Running the subsequent ACT4 python/uv framework steps without root access results in Permission Denied errors when attempting to access the sudo-installed toolchain binaries.

Changes:
I've added brief notes to the Compiler and Sail installation sections advising users on how to modify the commands for a local, non-root installation (e.g., using $HOME/riscv-tools). This preserves the existing enterprise instructions while smoothing the onboarding process for new local contributors.

Let me know if you'd like any adjustments to the wording!


README.md
```diff
@@ -93,6 +93,9 @@ cd riscv-gnu-toolchain
 ./configure --prefix=</path/to/install> --with-multilib-generator="rv32e-ilp32e--;rv32i-ilp32--;rv32im-ilp32--;rv32iac-ilp32--;rv32imac-ilp32--;rv32imafc-ilp32f--;rv32imafdc-ilp32d--;rv64i-lp64--;rv64ic-lp64--;rv64iac-lp64--;rv64imac-lp64--;rv64imafdc-lp64d--;rv64im-lp64--;"
 sudo make  # sudo may be required depending on the selected `prefix`
 ```
+> **Note for Local/Non-Root Installations:**
```
jordancarlin: Use the GitHub NOTE callout for this instead of the custom formatting. See the above note about installation taking a while for an example.

README.md
```diff
@@ -93,6 +93,9 @@ cd riscv-gnu-toolchain
...
+> If you are setting up the ACT framework on a personal machine and want to avoid using `sudo` or modifying root directories, you can install the tools locally in your home directory. 
+> Simply replace `</path/to/install>` with a local directory (e.g., `$HOME/riscv-tools/riscv-gnu-toolchain`) and run `make` without `sudo`.
```
jordancarlin: Let’s use $HOME/riscv as the example user directory (that’s commonly used by other riscv projects already). No need for the extra riscv-gnu-toolchain level of hierarchy either.

README.md
```diff
@@ -93,6 +93,9 @@ cd riscv-gnu-toolchain
 ./configure --prefix=</path/to/install> --with-multilib-generator="rv32e-ilp32e--;rv32i-ilp32--;rv32im-ilp32--;rv32iac-ilp32--;rv32imac-ilp32--;rv32imafc-ilp32f--;rv32imafdc-ilp32d--;rv64i-lp64--;rv64ic-lp64--;rv64iac-lp64--;rv64imac-lp64--;rv64imafdc-lp64d--;rv64im-lp64--;"
 sudo make  # sudo may be required depending on the selected `prefix`
 ```
+> **Note for Local/Non-Root Installations:**
+> If you are setting up the ACT framework on a personal machine and want to avoid using `sudo` or modifying root directories, you can install the tools locally in your home directory.
```
jordancarlin: Maybe something like this?
```suggestion
> If you don’t have sudo access, you can install the tools locally in your home directory. 
```

README.md
```diff
@@ -120,6 +123,9 @@ curl --location https://github.com/riscv/sail-riscv/releases/download/0.10/sail-
 
 Add `/path/to/install/bin` to your `PATH` if you used a different directory than for the `riscv-gnu-toolchain`.
 
+> **Note:** Similarly, for local installations, you can extract the Sail model into your home directory without `sudo` by specifying a local path:
```
jordancarlin: Same comments as above.

jordancarlin (changes requested):

README.md
```diff
@@ -120,6 +124,10 @@ curl --location https://github.com/riscv/sail-riscv/releases/download/0.10/sail-
 
 Add `/path/to/install/bin` to your `PATH` if you used a different directory than for the `riscv-gnu-toolchain`.
 
+> [!NOTE]
+> If you don't have `sudo` access, you can extract the Sail model into your home directory by specifying a local path:
+> `curl --location <url> | tar xvz --directory=$HOME/riscv/sail-riscv --strip-components=1`
```
jordancarlin: No need for the extra directory. That requires the user to add two directories to their PATH instead of just one.
```suggestion
> `curl --location <url> | tar xvz --directory=$HOME/riscv--strip-components=1`
```

jordancarlin (changes requested):
One last comment. Otherwise, LGTM.

jordancarlin (approved):

jordancarlin:
@DymShanks Looks like pre-commit is finding formatting issues. Running `uv run pre-commit run -a` should fix the issues. I also recommend running `uv run pre-commit install` to automatically fix these issues in the future.

README.md
```diff
@@ -94,6 +94,11 @@ cd riscv-gnu-toolchain
 sudo make  # sudo may be required depending on the selected `prefix`
 ```
 
+> [!NOTE]
+> for Local/Non-Root Installations:\*\*
+> If you don't have sudo access, you can install the tools locally in your home directory.
+> Simply replace `</path/to/install>` with a local directory (e.g., `$HOME/riscv`) and run `make` without `sudo`.
```
jordancarlin, line 99: Didn't notice this before, but this seems to be rendering strangely:

Image

I don't think we want the ** and the For local ... part should probably be on its own line.

jordancarlin (changes requested):

README.md
DymShanks: The markdown formatter must have escaped the asterisks. I've cleaned up the syntax and added a line break so it renders properly.

README.md
```diff
@@ -94,6 +94,12 @@ cd riscv-gnu-toolchain
 sudo make  # sudo may be required depending on the selected `prefix`
 ```
 
+> [!NOTE]
+> for Local/Non-Root Installations:
```
jordancarlin: Inconsistent to have this header for this note and not the Sail one. I'd recommend dropping it, but also fine if you'd rather add it to the Sail one as well.

jordancarlin (approved):

PR #1156, opened 2026-03-30
Changes to sv_area_sz offset for S-mode without H-extension

juliacygong:
Issue came out from when working on cp_medeleg_msu for ExceptionsS, my test is failing when medeleg = 0x0002 using the S-mode trap handler. Looking at the trace, the handler does addi x2, x2, -0x298 and then reads lw x6, -0x80(x2), which is below the expected region and causes bltu x6, x9, -0xc38 to abort the test.
Looking at arch_test.h around line 1216, sv_area_off is set to -1*sv_area_sz for S-mode with a comment saying "up 2 save areas (M←S)". But earlier in the file around lines 816-818 there's a comment saying "1*svarea_sz is Smode when rvtest_htrap_routine is not defined". Line 1223 does addi sp, sp, -1*sv_area_sz to move sp to Mtramptbl_sv, but the additional -sv_area_sz from sv_area_off causes the read to land 0x80 bytes below Mtramptbl_sv.


jordancarlin:
@juliacygong Can you update the PR title/description to explain what this is fixing?

tests/env/arch_test.h
```diff
@@ -1213,7 +1213,11 @@ spcl_\__MODE__\()chk4ecall:
         .set sv_area_off, ( 0*sv_area_sz)       // get trapsig_ptr val  up 1 save areas   (M<-HS)
 .else
    .ifc \__MODE__ , S
-        .set sv_area_off, (-1*sv_area_sz)       // get trapsig_ptr val  up 2 save areas   (M<-S)
+     #ifdef rvtest_htrap_routine
+        .set sv_area_off, (-1*sv_area_sz)       // get trapsig_ptr val  up 2 save areas   (M<-HS<-S)
```
davidharrishmc: The comment doesn't make sense.  There is not both a HS and S.

davidharrishmc (changes requested):

tests/env/arch_test.h
```diff
@@ -1213,7 +1213,11 @@ spcl_\__MODE__\()chk4ecall:
         .set sv_area_off, ( 0*sv_area_sz)       // get trapsig_ptr val  up 1 save areas   (M<-HS)
 .else
    .ifc \__MODE__ , S
+     #ifdef rvtest_htrap_routine
         .set sv_area_off, (-1*sv_area_sz)       // get trapsig_ptr val  up 2 save areas   (M<-S)
+     #else
+        .set sv_area_off, ( 0*sv_area_sz)       // get trapsig_ptr val  up 1 save area    (M<-S)
```
UmerShahidengr, line 1219: I am not sure if this is the correct approach to handle this offset. This offset `(-1*sv_area_sz)` was added to translate the overall offset value because `(2*sv_area_sz)` & `(3*sv_area_sz)` was going beyond the offset range of load/store instructions. Resetting this translation might affect the Umode tests when Hypervisor will be enabled. There are a total of 4 save areas (M, HS, SU, VS/VU). Umode shares the save area with Smode save area whereas HSmode and VS/VU modes have their own save areas to store important register/label values. 
Their offsets `(n*sv_area_sz)` were very carefully managed while running Sv tests (with or without Hmode enabled) under the constraints of 12bit offset range of load/store instructions. What problem are you facing with these offset values? Probably this fix will resolve your problem but will create bugs in other places where this offset is getting used while running Sv tests or tests which are delegating traps in lower modes.

UmerShahidengr (changes requested):

UmerShahidengr:
@Zain2050 can you run Sv tests with this change, check if it breaks the logic of trap return i.e, this [code](https://github.com/riscv/riscv-arch-test/blob/b27b0cd9e00a35b267cebd9a0ce2a6445bc47d64/tests/env/arch_test.h#L1805-L1851) section.

UmerShahidengr:
The most important logic that can get affected with this change is `RVTEST_GOTO_LOWER_MODE` logic ([here](https://github.com/riscv/riscv-arch-test/blob/b27b0cd9e00a35b267cebd9a0ce2a6445bc47d64/tests/env/arch_test.h#L799-L834)). When we have virtualization enabled, and try to switch from Mmode to Smode/Umode, then `pc` value is needed to be translated to virtual address, this translation is done by pulling values from lower mode save area very carefully. 
It has to be tested if this change keep LOWER_MODE macro stable.

Zain2050:
This seems familiar. I think I provided this to @juliacygong about a month ago. @juliacygong can you confirm?

tests/env/arch_test.h
Zain2050: This is being used with Smode handler while Sv tests only utilize Mmode handler (with or without hypervisor enabled). 

When `rvtest_htrap_routine` is not defined `Smode save_area is at 1*sv_area_sz` and when `htrap_routine` is defined `HSmode save_area is at 2*sv_area_sz`. If we include the offset `(-1*sv_area_sz) `it becomes 0\*sv_area_sz and 1\*sv_area_sz.
It was always selecting 1\*sv_area_sz irrespective of rvtest_htrap_routine, while Smode save area is at 0*sv_area_sz when htrap_routine isn't defined.

juliacygong:
@Zain2050 Yes I remember you providing this when I was debugging medeleg_msu for ExceptionsSm. This is the link to what you provided: [https://github.com/Zain2050/riscv-arch-test/commit/b6db6f46d6bea945dc4427a6dc3a251c0121046d](https://github.com/Zain2050/riscv-arch-test/commit/b6db6f46d6bea945dc4427a6dc3a251c0121046d)

Zain2050:
@UmerShahidengr This works. I provided it after testing it.

UmerShahidengr (approved):
If all tests are working with this change, then it looks good to me.

davidharrishmc:
@Zain2050 Is the (M <- S) comments correct?  The number of save areas and the names of the save areas don't seem to agree.

tests/env/arch_test.h
```diff
@@ -1213,7 +1213,11 @@ spcl_\__MODE__\()chk4ecall:
         .set sv_area_off, ( 0*sv_area_sz)       // get trapsig_ptr val  up 1 save areas   (M<-HS)
 .else
    .ifc \__MODE__ , S
+     #ifdef rvtest_htrap_routine
         .set sv_area_off, (-1*sv_area_sz)       // get trapsig_ptr val  up 2 save areas   (M<-S)
```
Zain2050: Please modify the comments to these
```suggestion
        .set sv_area_off, (-1*sv_area_sz)       // get trapsig_ptr val  up 2 save areas   (M<-S<-HS)
```

tests/env/arch_test.h
```diff
@@ -1213,7 +1213,11 @@ spcl_\__MODE__\()chk4ecall:
...
+     #else
+        .set sv_area_off, ( 0*sv_area_sz)       // get trapsig_ptr val  up 1 save area    (M<-S)
+     #endif
    .else
       .ifc \__MODE__ , V
         .set sv_area_off, (-2*sv_area_sz)       // get trapsig ptr val  up 3 save areas,  (M<-HS<-VS))
```
Zain2050:
```suggestion
        .set sv_area_off, (-2*sv_area_sz)       // get trapsig ptr val  up 3 save areas,  (M<-S<-HS<-VS))
```

jordancarlin (approved):

davidharrishmc (approved):

PR #1161, opened 2026-03-30
Container image for generating ELFs

### LINKED ISSUE #1155: Container image for generating ELFs
I have created a Dockerfile for myself that allows building ELF files inside of it. Would you be interested in me opening a PR with it? I can also add CI workflow to automatically build and publish it to GitHub Container Registry on new releases or on demand with manual trigger.

Yes, it currently requires `--privileged` due to Podman inside the container, but at least nothing except Podman or Docker is needed to use it and Docker is already present in GitHub Actions CI, so will be fairly easy to use there.

I would rather not build it in CI myself due to how long GCC build takes, so having something upstream would be really nice and I imagine others might appreciate the convenience too.

nazar-pc:
This introduces Dockerfile containing build environment with everything necessary to build ELF files like described in the readme.

I had to customize the build to make it faster and smaller.

There is also a CI workflow that is triggered when tags are created or can also be triggered manually from GitHub UI. It will publish `ghcr.io/riscv/act4-build` image with tag corresponding to the Git tag or branch name when triggered manually.

This works equally well in this repo and in forks in case contributors want to experiment with it.

You can see an example of it here: https://github.com/nazar-pc/riscv-arch-test/pkgs/container/act4-build ([CI run](https://github.com/nazar-pc/riscv-arch-test/actions/runs/24293190485)).

No privileged access is needed (should work with both Docker and Podman), can be running under any user (GitHub Actions is running under user with ID 1001 for example, while Linux desktops usually use ID 1000 by default and this is what the image defaults to).

I didn't update readme with instructions because there is nothing published yet, but you can find necessary details in comments of the Dockerfile itself.

x86-64 image size is 291 MB (1.21 GB on disk), aarch64 image is 285 MB (1.2 GB on disk).

Resolves https://github.com/riscv/riscv-arch-test/issues/1155

Some potential next steps (don't need to be done in this PR, but maybe for the future):
* Since the image is tied to the source code, maybe ship the repo contents with the image itself?
* If this is expected to be rebuilt often, caching can be enabled to speed-up rebuilds
* Once images are published, add a section to the readme briefly explaining how to use it (trivial, but still)
* Fix toolchain repo commit to make sure it is only updated when desired


Dockerfile
```diff
@@ -0,0 +1,250 @@
...
+                 libc6-dev-amd64-cross ;; \
+        *) echo "Unsupported TARGETARCH=${TARGETARCH}" >&2; exit 1 ;; \
+      esac \
+    fi \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN git clone --depth 1 https://github.com/riscv/riscv-gnu-toolchain /tmp/riscv-gnu-toolchain \
+ && cd /tmp/riscv-gnu-toolchain \
+ && sed -i 's/c,c++/c/g' Makefile.in \
+ && sed -i 's/c,c++,fortran/c/g' Makefile.in \
```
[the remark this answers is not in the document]
nazar-pc: Readme didn't suggest a specific commit, hence checking out the default branch

jordancarlin:
@nazar-pc sorry about the delay in reviewing this PR. Now that #1107 is merged, can you please update the container to work with the new dependencies and drop the `--privileged` flag? Once that is done, I will review this in detail and then we can try to get the container image merged and automatically published in CI going forward.

nazar-pc:
Added one more commit and rebased on `act4` branch. Tested and confirmed that it builds an image and the image can be used to compile things successfully both locally an in CI with GitHub Actions.

I also updated PR description to reflect the current version.

nazar-pc:
@jordancarlin friendly ping here

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+#   - sail-riscv           (RISC-V reference model)
+#
+# Usage:
+#   docker build -t riscv-act4 .
+#   docker run -it --rm -v .:/mnt riscv-act4
+#   Or if your user does not have ID 1000:
+#   docker run -it --rm -u $(id -u):$(id -g) -v .:/mnt riscv-act4
+#
+#   Inside the container the repo root is at /mnt; run:
+#     CONFIG_FILES=config/cores/<vendor>/<dut>/test_config.yaml make
```
jordancarlin, line 19: What about simplifying this even further so that the default entrypoint is to compile the tests? We should also consider including the repo itself in the image so that users can just pull the image and point it at their config files with no other setup needed (cloning the repo, etc.).

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+
+# Global ARGs - declared before any FROM so they are overridable across all stages.
+# Each stage that uses them must redeclare them with a bare ARG (no default) to bring them into scope.
+ARG RISCV_TOOLCHAIN_PREFIX=/opt/riscv
+ARG SAIL_PREFIX=/opt/sail
```
jordancarlin: What's the rationale for splitting these into separate locations? Seems like we could simplify it by installing everything to `/opt/riscv`.

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+ARG SAIL_VERSION=0.10
+
+# Stage 1: build riscv-gnu-toolchain
```
jordancarlin, line 27: If the final image is going to use Ubuntu 24.04, why not use the prebuilt binaries to speed things up? https://github.com/riscv-collab/riscv-gnu-toolchain/releases

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+rv32im-ilp32--;\
+rv32iac-ilp32--;\
+rv32imac-ilp32--;\
+rv32imafc-ilp32f--;\
+rv32imafdc-ilp32d--;\
+rv64i-lp64--;\
+rv64ic-lp64--;\
+rv64iac-lp64--;\
+rv64imafdc-lp64d--;\
+rv64im-lp64--;" \
```
jordancarlin: I know this is copied from the README, but I think we can actually trim this list down significantly (or maybe even drop it entirely). We don't use any of the standard libraries, so I don't think we need all of the multilibs. They take up significant space so would be good to drop to trim down the image.

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+    ca-certificates \
+    curl \
+ && rm -rf /var/lib/apt/lists/*
+
+RUN SAIL_OS="$(uname -s)" \
+ && SAIL_ARCH="$(uname -m)" \
+ && mkdir -p "${SAIL_PREFIX}" \
+ && curl --location --fail \
+      "https://github.com/riscv/sail-riscv/releases/download/${SAIL_VERSION}/sail-riscv-${SAIL_OS}-${SAIL_ARCH}.tar.gz" \
+    | tar xz --directory="${SAIL_PREFIX}" --strip-components=1
```
jordancarlin: Seeing as the only thing this layer does is fetch a binary, is it actually useful to split it into a dedicated layer? Seems like it could just go straight in the final image.

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+# We pre-warm the gem cache here so the first `make` inside the container doesn't need network access for Ruby deps.
+#
+# mise installs everything under MISE_DATA_DIR (/opt/mise) so it can be cleanly COPY'd into the final image.
+FROM ubuntu:24.04 AS mise-fetcher
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# mise needs curl + ca-certificates to download tools.
+# The remaining packages are required by ruby-build to compile Ruby from source (mise uses ruby-build under the hood and
+# does not use the distro Ruby package).
```
jordancarlin: mise should be fetching a prebuilt Ruby binary from https://github.com/jdx/ruby. It didn't use to default to that, but I changed the setting recently. Can you test again with a stripped down list of dependencies?

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+# .mise.toml pins the uv and Ruby versions; the Gemfile(s) declare the riscv-unified-db gem. No full repo clone needed.
+COPY .mise.toml /tmp/act/
+COPY framework/src/act/data/Gemfile framework/src/act/data/Gemfile.lock /tmp/act/
+
+# Install uv and Ruby at the versions pinned in .mise.toml.
+# Tools land in /opt/mise/installs/, shims in /opt/mise/shims/.
+RUN cd /tmp/act && mise install
+
+# Pre-install the riscv-unified-db gem into the mise-managed Ruby's gem dir so `bundle install` is a no-op at runtime
+RUN cd /tmp/act && mise exec -- bundle install --gemfile=Gemfile
```
jordancarlin: It would be good to run `uv sync` with the `pyproject.toml` files as well to pre-download all of the Python dependencies too.

Dockerfile
```diff
@@ -0,0 +1,242 @@
...
+ENV MISE_CACHE_DIR=/opt/mise/cache
+ENV MISE_YES=1
+# Writable by any user for mise and other tools to store data
+ENV HOME=/tmp/home
+
+# Full PATH: RISC-V toolchain → Sail → mise shims (uv, ruby, bundle, …) → mise itself
+ENV PATH="${RISCV_TOOLCHAIN_PREFIX}/bin:${SAIL_PREFIX}/bin:/opt/mise/shims:/opt/mise/bin:${PATH}"
+
+# Runtime-only packages:
+#   - python3: uv-managed venvs need a system Python as fallback
```
jordancarlin: uv will download it's own version of Python by default. I think you can skip the system version and let it manage Python (especially coupled with the previous comment to run `uv sync`).

Dockerfile
jordancarlin: With the Ruby switch, I'm wondering if this extra stage is actually useful anymore. It seems like just about everything that is produced is copied over, so it may be simpler to just fold this whole stage into the runtime image. I doubt it would materially change the size of it.

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,31 @@
+name: Build image
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - "*"
```
jordancarlin: We probably want to build the image whenever the Dockerfile changes so the latest version is always available.

jordancarlin: Also, I think the pre-release versions of the documentation cause tags to be created, and we don't want to regenerate the image for those tags.

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  act4-build-image:
+    uses: docker/github-builder/.github/workflows/build.yml@2497a7d1e7d8683af4949c9d6d62012bc16ed59c # v1.5.0
```
jordancarlin, line 13: Why isn't this using the standard action for building Docker images in GitHub Actions? Usually `docker/setup-buildx-action` and `docker/build-push-action`.

jordancarlin (changes requested):
Thanks for working on this! Overall, it looks great. I've left a first round of comments below.

.github/workflows/build-image.yml
nazar-pc: Check the docs and feel free to update/suggest what you want to do exactly. Right now `workflow_dispatch:` allows you to trigger the build on any branch, so you can trigger it on a custom branch and get an image tagged with a branch name (+ you can always pull an image by its hash too).

Yes, this will cause any tags to build images, including pre-releases. Can be changed to only build on a published release instead: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#release

If you really want the latest version (though given how much this repo has changed recently in a breaking way I'm not sure this is a great idea), it can be configured to build `:latest` (default) image on every push to `act4`. With caching enabled this should not take a very long time since the most expensive steps of toolchain compilation will be reused from cache.

.github/workflows/build-image.yml
nazar-pc: This is essentially a successor to `docker/build-push-action`. The biggest difference is that it spawns child runners to build platform-specific images on more physical machines and it tries to pick the native architecture for each image too.

So previously I was building both x86-64 and aarch64 on the same machine, which both caused them to share the CPU (meaning the build was slower) and it also meant I had to enable cross-compilation (see earlier revision of the PR), which became quite complex with the whole mess around Mise and the fact that it compiles various Ruby-related stuff now.

Now with x86-64 image built on x86-64 machine and aarch64 image on aarch64 machine, the Dockerfile is simpler, the build is faster and that action takes care of re-assembling multiarch image automatically for you.

All this can be done with `docker/build-push-action` and a lot of wiring around it (requiring similar reusable action for parallelism), so it is absolutely in no way worth it comparing to the amazing simplicity of what we've got here.

Here is the PR I learned about this action from (was reviewing some changes before upgrading Pi-hole): https://github.com/pi-hole/docker-pi-hole/pull/2008
It gives you a perspective of what was involved before the action was created.

Dockerfile
nazar-pc: I think the current approach is a decent compromise. In CI I'm calling it like this:
```bash
docker run \
  -v ./res/riscv-arch-test:/mnt \
  -v ./res/abundance:/mnt/config/abundance:ro \
  -e CONFIG_FILES="config/abundance/abundance-rv32i-max/test_config.yaml config/abundance/abundance-rv64i-max/test_config.yaml" \
  ${{ steps.build.outputs.imageid }} \
  make
```

If you want to build everything (which is what I believe default `make` without environment variables does) then:
```bash
docker run \
  -v ./res/riscv-arch-test:/mnt \
  ${{ steps.build.outputs.imageid }} \
  make
```

I wouldn't change entrypoint just to avoid `make` and I wanted to avoid writing custom entrypoint script too.

Including the repo into container image is something I've been thinking about and mentioned in PR description, but that, ideally, requires introducing `.dockerignore` file since not everything needs to be included and Docker doesn't read `.gitignore` when building context. However, this is usually not a concern in CI since fresh checkout is used there by default and [Git context](https://github.com/docker/build-push-action#git-context) even ignored `.dockerignore` apparently.

So if you want, I can change that for CI, but for building images (of reasonable size) locally, `.gitignore` would need to be replicated in `.dockerignore` too (could be a symlink, but note that the syntax is not 100% compatible).

Dockerfile
nazar-pc: Just didn't feel right to overlay multiple tools on top of each other. Both would work I guess, but this makes it a bit nicer

Dockerfile
nazar-pc: Because I didn't know about them? Readme here said to build from source and I had painful experience with Ubuntu packages, so I wrote build instructions in Dockerfile 😅

But I'll have to see what they are building there and the question of what release to use remains open ("latest" seems like a bad target because it can change any time and is bad for reproducibility).

UPD: Looks like with those releases the simplicity of the Dockerfile and build time will be traded for the final image size.

Dockerfile
nazar-pc: I wouldn't know what to trim, so you'll have to be more specific.

Also https://github.com/riscv-collab/riscv-gnu-toolchain/releases have both riscv32 and riscv64 builds, does that mean they don't use multilib or something? I don't really know much about this stuff.

Dockerfile
nazar-pc: Considering that both `curl` and `ca-certificates` are both already there it could, but downloading will not be concurrent (all steps that do not depend on each other run in parallel). Not that it matters much I guess.

Dockerfile
nazar-pc: Yes, latest `act4` branch completes this much faster and doesn't seem to build Ruby itself at very least, much better! I'd say it is not bad to run this concurrently with other things though.

Dockerfile
nazar-pc: I'll try to add that then, though I'm not working with either Python or Ruby, so all this is unknown to me

Dockerfile
nazar-pc: Hm... Okay, I'll see about that

nazar-pc (commented):
I'll wait for responses to comments before applying changes

Dockerfile
jordancarlin: Seems fine to get this in as an initial version of the container and then look at including the repo itself in it in a subsequent version.

Also, why is the `--privileged` flag needed? Is that just a carryover from when it needed a container inside the container? Hopefully it can be dropped now.

Dockerfile
jordancarlin: How much larger is the image when using one of the prebuilt images? If it is not significantly different, it might be worth it.

In terms of the version, I agree that it should not be left floating. Let's pin it to the current latest commit (`f27c68dd632102a1eab85d97a90f3cdc4e90350c`) if building from source or the current latest release (2026.04.05) if using the prebuilt releases. Renovate can handle automatically updating that tag as new versions are released.

Dockerfile
jordancarlin: I think the performance difference of parallelizing the downloads is probably insignificant when they are that small, and it may actually end up being slower just because of the overhead of downloading more things.

Dockerfile
jordancarlin: Glad that the ruby download is working. I'd be curious to see how much of a performance benefit we actually get by splitting the layers. I impinge it isn't much, in which case it I'd probably prefer the simpler approach of using fewer layers.

.github/workflows/build-image.yml
jordancarlin: I think we should have `workflow_dispatch` for on demand builds, builds for all releases, and a `latest` build that is built whenever the Dockerfile changes (not whenever the repo changes for now). No need for builds for all tags.

Dockerfile
jordancarlin: They don't use multilib, but because we don't use the standard libraries, we don't need the libraries from multilib anyway.

I think we can actually completely drop the multilib list. I just tested a build that didn't specify multilib (I just used `./configure --prefix=<path/to/install>`), and it works totally fine to compile both rv32 and rv64 tests.

Dockerfile
nazar-pc: Updating to include the cloned repo and updating docs.

`--privileged` is not necessary indeed, updated the comment above.

Dockerfile
nazar-pc: I just tried precompiled version and it is unfortunately A LOT larger. I'm sure it can be stripped down by removing extra files after extraction, but not sure it will be much simpler.

Before:
```
IMAGE                ID             DISK USAGE   CONTENT SIZE
riscv-act4-v2:latest 8d3ebd9c1b40       1.16GB          279MB
```
After:
```
IMAGE                ID             DISK USAGE   CONTENT SIZE
riscv-act4-v2:latest 11487c0e094c       4.44GB         1.29GB
```

.github/workflows/build-image.yml
nazar-pc: Everything except "whenever Dockerfile changes" seems perfectly reasonable. Not sure it is worth complicating workflow by trying to specifically track Dockerfile changes though.

Simply having caching enabled should reduce most of the steps to quick checks as is. I'll make it always build `latest` on `act4` instead.

I did some research on releases and decided to keep building on tags pushed, but ignore `cert-docs-*`, which is trivial and should solve your concern with documentation tags.

Dockerfile
nazar-pc: I'll keep this specific layer due to it needing development packages for compiling Ruby stuff, which has non-negligible size.

nazar-pc (commented):
Rebased on the latest `act4`, the only change is in the last commit.

Key changes:
* build `:latest` tag on `act4` branch
* skip `cert-docs-*` tags when building images
* toolchain is built from a fixed tag without multilib (a bit faster now), prebuilt version is too large to use as is
* enabled caching now that image will be built on each push to `act4` (to make things faster for rebuilds)
* clone of the repo now ships with the image
* sail fetch step absorbed into the final image
* Python cache is now included in addition to Ruby
* fewer dependencies installed in the final image

Due to repo clone the overall image size is larger, but this should be a win for consistency and possibly even total download size/time overall:
```
> sudo docker images --tree
IMAGE                             ID             DISK USAGE   CONTENT SIZE
ghcr.io/nazar-pc/act4-build:wip   cf3e53f781f2       7.04GB          1.2GB
├─ linux/arm64                    1381ce89d713       3.48GB          591MB        
└─ linux/amd64                    27c06814c96e       3.56GB          609MB
```

Dockerfile
jordancarlin: That is significant! Sounds good to use the custom build in that case.

Dockerfile
jordancarlin: Maybe I'm missing something, but now that it downloads a pre-compiled version of ruby, what development packages does it need?

nazar-pc: Ruby itself is downloaded as is, but when dependencies are installed after that, those seem to require C toolchain and some other libraries, not sure why and don't care enough to debug

.dockerignore
```diff
@@ -0,0 +1 @@
+.git
```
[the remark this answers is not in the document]
nazar-pc: I'm not sure how much of this needs to be replicated in `.dockerignore`. In CI none of it exists, but locally it might :shrug:

Dockerfile
jordancarlin: Let's leave it for now then. I might look into that later.

.dockerignore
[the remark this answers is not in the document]
jordancarlin: Seems find to skip this for now.

Dockerfile
```diff
@@ -0,0 +1,187 @@
...
+#
+# HOME=/home/shared is used so all caches (mise, uv, udb, etc.) land in a single directory that is copied to the final
+# image.
+FROM ubuntu:24.04 AS mise-fetcher
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# mise needs curl + ca-certificates to download tools.
+# The remaining packages are required by ruby-build to compile Ruby from source (mise uses ruby-build under the hood and
+# does not use the distro Ruby package).
```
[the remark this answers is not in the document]
jordancarlin: This comment suggestion is valid and should be updated. Other than that, this seems good to go for an initial version. Thanks for all your work on this!

jordancarlin (approved):

nazar-pc:
@jordancarlin just an idea: you can disabling building on `act4` branch completely and simply trigger it on `act4` manually whenever a change warranting the rebuild is merged.

This is much easier than automation and is probably good enough in case you still want to prevent extra builds on `act4` branch.

jordancarlin:
> @jordancarlin just an idea: you can disabling building on `act4` branch completely and simply trigger it on `act4` manually whenever a change warranting the rebuild is merged.
> 
> This is much easier than automation and is probably good enough in case you still want to prevent extra builds on `act4` branch.

That’s a good idea and would work if we didn’t include the repo itself in the image, but now that we are including the repo itself, I think we need to rebuild on every push. I have some ideas to significantly improve the caching that I will push soon so that hopefully the rebuild is insignificant.

nazar-pc:
I'm just a bit uneasy that it is hundreds of megabytes increase of package storage on every merge, regardless of whether it is cached during build or not.

If Python and Ruby dependencies doesn't write anything into the repo itself. I saw a bunch of files updated across the tree in general, but I suspect it is not done during dependencies installation specifically. Dockerfile can be updated to only copy a few files before calling those steps, which will remove the need to re-install those dependencies every time, even if it takes just a few minutes. And `/act4` will be filled from the repo directly in the final image at the very end instead of being copied from `mise-fetcher` step.

As for the repo contents, you can actually mount any revision with `-v /path/to/repo:/act4`, it is just the default that is included.

PR #1169, opened 2026-03-31
ExceptionsU Python Tests

zssmgosh:
Currently collecting 100% coverage on Sail, Spike, and QEMU


generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+    ]
+
+    test_data.int_regs.return_registers([save_reg, mask_reg, arg_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "ExceptionsU",
+    required_extensions=["I", "Zicsr", "Sm", "U"],
+    march_extensions=["I", "Zicsr"],
```
jordancarlin: This shouldn't be needed. `march` only needs to be set when there are extra extensions that are needed beyond the `required_extensions`.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+    extra_defines=["#define SKIP_MEPC"],
```
jordancarlin: Maybe worth a comment explaining why this is needed?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+)
+def make_exceptionsu(test_data: TestData) -> list[str]:
+    """Main entry point for U exception test generation."""
+    lines = []
+
+    lines.extend(["\tRVTEST_GOTO_LOWER_MODE Umode  # Run tests in user mode\n"])
```
jordancarlin:
```suggestion
    lines = []

    lines.append("\tRVTEST_GOTO_LOWER_MODE Umode  # Run tests in user mode\n")
```

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+
+
+def _generate_instr_adr_misaligned_branch_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_instr_adr_misaligned_branch"
+    temp_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Address Misaligned branch"),
+        f"LI(x{temp_reg}, 1)",
+        ".align 2",
```
jordancarlin: Is `.align 2` really needed? This test doesn't have compressed instructions, so I don't see how we could end up aligned to anything less than 2.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+            rs2 = f"x{temp_reg}"
+        elif branch == "bne":
+            rs1 = "x0"
+            rs2 = "x0"  # 0 == 0, so bne not taken
+        else:  # blt, bltu
+            rs1 = f"x{temp_reg}"
+            rs2 = "x0"
+        lines.append(f"{branch} {rs1}, {rs2}, .+6")
+
+    lines.append(write_sigupd(check_reg, test_data))
```
jordancarlin: What is the sigupd with check_reg accomplishing? You never write to check_reg after initializing it, so it sees like this will match no matter what the DUT does.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+        if branch in ("beq", "bge", "bgeu"):
+            rs1 = "x0"
+            rs2 = f"x{temp_reg}"
+        elif branch == "bne":
+            rs1 = "x0"
+            rs2 = "x0"  # 0 == 0, so bne not taken
+        else:  # blt, bltu
+            rs1 = f"x{temp_reg}"
+            rs2 = "x0"
+        lines.append(f"{branch} {rs1}, {rs2}, .+6")
```
jordancarlin: Might be worth adding nops here in case the branch (incorrectly) causes a trap to make debugging easier.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+def _generate_instr_adr_misaligned_branch_nottaken(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_instr_adr_misaligned_branch_nottaken"
+    temp_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Branch to an unaligned address is NOT taken (PC+6). Should not cause an exception"),
+        ".align 2",
+        f"LI(x{temp_reg}, 1)",
+        f"LI(x{check_reg}, 1)",
+        test_data.add_testcase("nottaken_branch_pc_6", coverpoint, covergroup),
```
jordancarlin: Should there be a test case per branch? I assume they are all different bins.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Address Misaligned JALR"),
+    ]
+
+    # rs1 is set to PC + base_offset for rs1[1:0]
+    # jalr controls the offset[1:0], which covers all 16 combinations of (rs1+offset)[1:0]
+    rs1_base_offsets = {0: 8, 1: 5, 2: 6, 3: 7}
+    jalr_offsets = {0: 8, 1: 5, 2: 6, 3: 7}
```
jordancarlin: These dicts are the same. Any reason not to just have one?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+
+    for rs1_lsb in range(4):
+        for offset_lsb in range(4):
+            base_off = rs1_base_offsets[rs1_lsb]
+            jalr_off = jalr_offsets[offset_lsb]
+
+            lines.extend(
+                [
+                    f"\n# rs1[1:0]={rs1_lsb:02b}, offset[1:0]={offset_lsb:02b}",
```
jordancarlin: Use an actual empty line instead of `\n` so it is easier to see what this will look like.
```suggestion
                   "",
                    f"# rs1[1:0]={rs1_lsb:02b}, offset[1:0]={offset_lsb:02b}",
```

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
jordancarlin: Why can't this just use (0, 1, 2, 3)?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+                    ".align 2",
+                    f"auipc x{addr_reg}, 0",
```
jordancarlin: Should this be `auipc x{addr_reg}, 8` to account for the `jalr` being a few instructions after the `auipc`?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+            # conditional nop ensures addi below is always 8 bytes after jalr so trap handler returns correctly
+            if rs1_lsb & 1:
+                lines.append("nop")
+
+            lines.extend(
+                [
+                    "# branch by 6 lands in upper half of next instruction 0x0001 which is generated into a c.nop",
+                    "addi x0, x2, 0",
+                ]
+            )
```
jordancarlin: Could this be implied to just always include 2 `addi x0, x2, 0`?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+        else:  # bne, blt, bltu
+            rs1 = "x0"
+            rs2 = f"x{temp_reg}"
+
+        lines.extend(
+            [
+                f"{branch} {rs1}, {rs2}, .+6",
+                "# branch by 6 lands in upper half of next instruction 0x0001 which is generated into a c.nop",
+                "addi x0, x2, 0",
+                ".word 0x00010013",
```
jordancarlin: Why is there an `addi x0, x2, 0` followed by a `.word 0x00010013`, which decodes to `addi x0, x2, 0`? Can they both use assembly instead? Why do we need two of them back to back?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+
+
+def _generate_instr_access_fault_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_instr_access_fault"
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Access Fault"),
+        f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        "LI(x4, 0xACCE)",  # trap handler checks x4 value and determines to use x1 (ra) as return address instead of mepc
```
jordancarlin: What prevents x4 from being allocated for something else by the register scheduler? Seems like we need to call `consume_register` on x4 before this.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+    test_data.int_regs.return_registers([addr_reg])
+    return lines
+
+
+def _generate_illegal_instruction_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_illegal_instruction"
+
+    lines = [
+        comment_banner(coverpoint, "Illegal Instruction"),
+        ".align 2",
```
jordancarlin: Why is the `.align` needed?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+        test_data.add_testcase("seed_csrrc", coverpoint, covergroup),
+        f" csrrc x{dest_regs[1]}, seed, x0",
+        " nop",
+        test_data.add_testcase("seed_csrrsi", coverpoint, covergroup),
+        f" csrrsi x{dest_regs[2]}, seed, 0",
+        " nop",
+        test_data.add_testcase("seed_csrrci", coverpoint, covergroup),
+        f" csrrci x{dest_regs[3]}, seed, 0",
+        " nop",
+    ]
```
jordancarlin: We can only run this test case if Zkr is supported. It should either be broken out into a separate ZkrU test or at least wrapped in `#ifdef ZKR_SUPPORTED`. For an implementation that doesn't implement Zkr, it is reserved behavior.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+
+def _generate_illegal_instruction_seed_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_illegal_instruction_seed"
+    dest_regs = test_data.int_regs.get_registers(4, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Illegal Instruction Seed"),
+        test_data.add_testcase("seed_csrrs", coverpoint, covergroup),
+        f" csrrs x{dest_regs[0]}, seed, x0",
+        " nop",
```
jordancarlin: Might be worth doing a sigupd with the destination register to make sure it wasn't updated by the attempted CSR read.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+        test_data.add_testcase("illegal_0xFFFFFFFF", coverpoint, covergroup),
+        ".word 0xFFFFFFFF",
+        "nop",
+    ]
+    return lines
+
+
+def _generate_illegal_instruction_seed_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_illegal_instruction_seed"
+    dest_regs = test_data.int_regs.get_registers(4, exclude_regs=[0])
```
jordancarlin: Why do we need 4 separate destination registers? Can we reuse the same one for all 4 instructions instead?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+        [
+            "#if __riscv_xlen == 64",
+            f" LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+            test_data.add_testcase("lwu_fault", coverpoint, covergroup),
+            f" lwu x{check_reg}, 0(x{addr_reg})",
+            " nop",
+            f" LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+            test_data.add_testcase("ld_fault", coverpoint, covergroup),
+            f" ld x{check_reg}, 0(x{addr_reg})",
+            " nop",
```
jordancarlin: Remove single space at beginning of each line. This will not be consistent with the rest of the indentation.

```suggestion
            f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
            test_data.add_testcase("lwu_fault", coverpoint, covergroup),
            f"lwu x{check_reg}, 0(x{addr_reg})",
            "nop",
            f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
            test_data.add_testcase("ld_fault", coverpoint, covergroup),
            f"ld x{check_reg}, 0(x{addr_reg})",
            "nop",
```

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+
+    load_ops = ["lb", "lbu", "lh", "lhu", "lw"]
+
+    for op in load_ops:
+        lines.extend(
+            [
+                f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+                test_data.add_testcase(f"{op}_fault", coverpoint, covergroup),
+                f"{op} x{check_reg}, 0(x{addr_reg})",
+                "nop",
```
jordancarlin: Add a sigupd to make sure `check_reg` didn't change.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+    op: str,
+    offset: int,
+    test_data: TestData,
+    coverpoint: str,
+    covergroup: str,
+) -> list[str]:
+    addr_reg, data_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    t_lines = [
+        f"LI(x{data_reg}, 0xDEADBEEF)",
```
jordancarlin: Pick a random value other than 0xDEADBEEF so that it is obvious if the store worked or not. Ideally the value would depend on `offset` so we get different values each time we write to scratch.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+    ]
+
+    return lines
+
+
+def _generate_misaligned_priority_load_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_misaligned_priority_load"
+    addr_reg, temp_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0, 1])
+
+    lines = [comment_banner(coverpoint, "Misaligned Priority Load")]
```
jordancarlin: Use a more detailed description for the comment banner. No need to say "Misaligned Priority Load" because that will already be obvious from the cover point name. Give an overview of what this test case is actually doing.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+    load_ops_base = ["lh", "lhu", "lw", "lb", "lbu"]
+    load_ops_64 = ["lwu", "ld"]
+
+    lines.append(f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)")
+
+    for offset in range(8):
+        lines.append(f"\n# Offset {offset} (LSBs: {offset:03b}) - Access fault Misaligned")
+        lines.append(f"addi x{temp_reg}, x{addr_reg}, {offset}")
```
jordancarlin: Combine these into a `lines.extend`.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+
+def _generate_mstatus_ie_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_mstatus_ie"
+    save_reg, mask_reg, arg_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Mstatus Interrupt Enable"),
+        f"csrr x{save_reg}, mstatus",
+        f"LI(x{mask_reg}, 0x8)",
+        f"csrrc x0, mstatus, x{mask_reg}",
```
jordancarlin: How does this work if you don't switch to M-mode first?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,524 @@
...
+        "\tRVTEST_GOTO_LOWER_MODE Umode",
```
jordancarlin: Drop all of the `\t`. Macros should be automatically indented by the generator using spaces.

jordancarlin (changes requested):
Great first version! Here are a bunch of (mostly minor) comments.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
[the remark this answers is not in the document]
davidharrishmc: This is already done in another file, so this comment can be ignored.

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -182,25 +180,33 @@ def generate_illegal_instruction_tests(test_data: TestData, covergroup: str) ->
 
 def generate_illegal_instruction_seed_tests(test_data: TestData, covergroup: str) -> list[str]:
     coverpoint = "cp_illegal_instruction_seed"
-    dest_regs = test_data.int_regs.get_registers(4, exclude_regs=[0])
+    dest_reg = test_data.int_regs.get_register(exclude_regs=[0])
```
davidharrishmc: No need to exclude regs 0 because it is already excluded for priv tests.

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -182,25 +180,33 @@ def generate_illegal_instruction_tests(test_data: TestData, covergroup: str) ->
...
         test_data.add_testcase("seed_csrrci", coverpoint, covergroup),
-        f"csrrci x{dest_regs[3]}, seed, 0",
+        f"csrrci x{dest_reg}, seed, 0",
         "nop",
+        write_sigupd(dest_reg, test_data),
+        "#endif",
     ]
 
-    test_data.int_regs.return_registers(dest_regs)
+    test_data.int_regs.return_registers([dest_reg])
```
davidharrishmc: I believe there is a return_register function that takes a single register instead of a list

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -182,25 +180,33 @@ def generate_illegal_instruction_tests(test_data: TestData, covergroup: str) ->
...
     coverpoint = "cp_illegal_instruction_seed"
-    dest_regs = test_data.int_regs.get_registers(4, exclude_regs=[0])
+    dest_reg = test_data.int_regs.get_register(exclude_regs=[0])
 
     lines = [
-        comment_banner(coverpoint, "Illegal Instruction Seed"),
+        comment_banner(
+            coverpoint, "Illegal instruction on seed CSR which should cause an illegal instruction exception"
+        ),
+        "#ifndef ZKR_SUPPORTED",
```
davidharrishmc: Per the test plan and coverpoints, these should be tested even if ZKR is supported, so take out the ifdef.

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
davidharrishmc, line 76: We are missing a nop after the addi.  We need it here until the trap handler is changed to return immediately after the faulting instruction.

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -88,7 +87,6 @@ def generate_instr_adr_misaligned_jal_tests(test_data: TestData, covergroup: str
         "jal x0, .+6",
         "# branch by 6 lands in upper half of next instruction 0x0001 which is generated into a c.nop",
         "addi x0, x2, 0",
-        "nop",
```
davidharrishmc, line 91: still need a nop after addi

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
davidharrishmc, line 124: need nop after addi

davidharrishmc (commented):
Looks good to me with minor changes.  @jordancarlin should take a pass as well.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,94 @@
...
+    generate_store_address_misaligned_tests,
+)
+from testgen.priv.registry import add_priv_test_generator
+
+_CG = "ExceptionsU_cg"
+
+
+def _generate_mstatus_ie_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsU_cg", "cp_mstatus_ie"
+    save_reg, mask_reg = test_data.int_regs.get_registers(2, exclude_regs=[])
```
jordancarlin: Drop empty exclude_regs list.
```suggestion
    save_reg, mask_reg = test_data.int_regs.get_registers(2)
```

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,94 @@
...
+
+    lines = [
+        comment_banner(coverpoint, "ecall from user mode with MIE=0 vs MIE=1"),
+        "RVTEST_GOTO_MMODE",
+        f"csrr x{save_reg}, mstatus",
+        f"LI(x{mask_reg}, 0x8)",
+        f"csrrc x0, mstatus, x{mask_reg}",
```
jordancarlin: Use `csrc` if you don't care about the result. Global comment.
```suggestion
        f"csrc mstatus, x{mask_reg}",
```

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,94 @@
...
+        "RVTEST_GOTO_LOWER_MODE Umode",
+        test_data.add_testcase("ecall_mie_0", coverpoint, covergroup),
+        "ecall",
```
jordancarlin: Should this also clear `mstatus.MPIE` since it becomes `mie` when you switch to U mode?

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -0,0 +1,94 @@
...
+        f"csrw mstatus, x{save_reg}",
+    ]
+
+    test_data.int_regs.return_registers([save_reg, mask_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "ExceptionsU",
+    required_extensions=["I", "Zicsr", "Sm", "U"],
```
jordancarlin:
```suggestion
    required_extensions=["U"],
```

jordancarlin (commented):
A few small comments, but otherwise LGTM.

jordancarlin (approved):

PR #1171, opened 2026-03-31
ZicntrU/S at 100% coverage

Teemooooooooo:
- priv.py is also changed to exclude x6, x7 and x9 as they are used by the RVTEST_GOTO_LOWER_MODE macro


generators/testgen/src/testgen/generate/priv.py
```diff
@@ -58,7 +58,9 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     tc = test_data.begin_test_chunk()
 
     # Priv tests use x1/ra as the return address for function calls, so reserve it before generating the test
-    test_data.int_regs.consume_registers([1])
+    test_data.int_regs.consume_registers(
+        [1, 6, 7, 9]
+    )  #  x6, x7, x9 are excluded because they are used by the RVTEST_GOTO_LOWER_MODE Smode macro
```
jordancarlin: Move the comment before the line of code to be with the other comment explaining why x1 is excluded.

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_mcounteren_access_s_tests(test_data: TestData) -> list[str]:
+    """Generate mcounteren access s mode tests."""
+    covergroup, coverpoint = "ZicntrS_cg", "cp_mcounteren_access_s"
+
+    read_reg, ori_csr_reg, one_reg, walk_reg, temp_reg = test_data.int_regs.get_registers(5, exclude_regs=[0])
+    # the registers(the ones other than x0) are excluded because they are used by the RVTEST_GOTO_LOWER_MODE Smode macro
```
davidharrishmc: Obsolete

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+        comment_banner(
+            coverpoint,
+            "Write walking 1s and 0s to mcounteren.  Read from corresponding counter and counterh in S-mode",
+        ),
+        "",
+    ]
+    lines.extend(
+        [
+            f"csrr x{ori_csr_reg}, mcounteren",
+            f"LI(x{one_reg}, -1)",
```
davidharrishmc: maybe change to "ones_reg" to emphasize all ones.

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+            coverpoint,
+            "Write walking 1s and 0s to scounteren with mcounteren = all 1s.  Read from corresponding counter and counterh in S-mode",
+        ),
+        "",
+    ]
+    lines.extend(
+        [
+            f"LI(x{one_reg}, -1)",
+            f"csrw mcounteren, x{one_reg}  # enable all counters in M-mode",
+            f"csrr x{ori_csr_reg}, scounteren",
```
davidharrishmc: How about save instead of ori.  ori must be shorthand for original, but it sounds like OR-Immediate.

generators/testgen/src/testgen/generate/priv.py
davidharrishmc: Drop Smode - the macro uses them for any mode?

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+                test_data.add_testcase(f"walking 1 at {i}", coverpoint, covergroup),
+                f"csrrc x{temp_reg}, mcounteren, x{one_reg}  # clear all bits",
+                f"csrrs x{temp_reg}, mcounteren, x{walk_reg}  # set current bit",
+                "RVTEST_GOTO_LOWER_MODE Smode",
+            ]
+        )
+        if i < 3:
+            lines.extend(
+                [
+                    f"csrr x{read_reg}, {reg_list[i]}",
```
davidharrishmc, line 45: There is no "nop" after these instructions that could trap.  Soon it won't be needed because the trap handler will be updated, but still needed for now.

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+        [
+            f"csrr x{ori_csr_reg}, mcounteren",
+            f"LI(x{one_reg}, -1)",
+            f"LI(x{walk_reg}, 1)",
+        ]
+    )
+    for i in range(32):
+        lines.extend(
+            [
+                test_data.add_testcase(f"walking 1 at {i}", coverpoint, covergroup),
```
davidharrishmc: no spaces in bin name.  walking_1_{i}

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+
+    lines.extend(
+        [
+            f"LI(x{walk_reg}, 1)",
+        ]
+    )
+    for i in range(32):
+        lines.extend(
+            [
+                test_data.add_testcase(f"walking 0 at {i}", coverpoint, covergroup),
```
davidharrishmc: walking_0_{i}

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+    lines = [
+        comment_banner(
+            coverpoint,
+            "Write walking 1s and 0s to mcounteren.  Read from corresponding counter and counterh in S-mode",
+        ),
+        "",
+    ]
+    lines.extend(
+        [
+            f"csrr x{ori_csr_reg}, mcounteren",
```
davidharrishmc: Doesn't look like you need to save and restore mcounteren.  You always set it when you want it.  Same with scounteen later.

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+                f"slli x{walk_reg}, x{walk_reg}, 1",
+            ]
+        )
+    lines.append(f"csrrw x{temp_reg}, scounteren, {ori_csr_reg}  # restore original csr value")
+
+    test_data.int_regs.return_registers([read_reg, ori_csr_reg, one_reg, walk_reg, temp_reg])
+    return lines
+
+
+def _generate_scounteren_access_m_tests(test_data: TestData) -> list[str]:
```
davidharrishmc, line 244: the m, s, u versions of cp_scounteren_access are identical except for the priv mode changes.  Pass an input of the priv mode to the function rather than replicating code.

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
```diff
@@ -0,0 +1,587 @@
...
+                f"slli x{walk_reg}, x{walk_reg}, 1",
+            ]
+        )
+    lines.append(f"csrrw x{temp_reg}, scounteren, {ori_csr_reg}  # restore original csr value")
+
+    test_data.int_regs.return_registers([read_reg, ori_csr_reg, one_reg, walk_reg, temp_reg])
+    return lines
+
+
+def _generate_mcounteren_access_u_tests(test_data: TestData) -> list[str]:
```
davidharrishmc: should be called mscounteren_access_u per coverpoint.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
Teemooooooooo: I did not put a "nop" here because based on the spec it should not trap in this specific instruction since the related bits in counteren are enabled. Should I add it in just in case?

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
Teemooooooooo: I did three tests because I thought we want to do it in a way where each test corresponds to its own coverpoint? Or are there some specific cases when we should combine it?

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
davidharrishmc: It shouldn’t trap, but we are testing the system to check that it doesn’t. If the test malfunctions we need the nop.

generators/testgen/src/testgen/priv/extensions/ZicntrS.py
davidharrishmc: I like that you have three tests for three cover points. They are almost identical so you can write one function and call it three times with slightly different segments. This is more maintainable than having may copies.

davidharrishmc:
At you make changes be sure to make resolve conversation on each fix.

davidharrishmc (approved):

Issue #1127, opened 2026-03-25
[CTP] Clarification on Extension Hierarchy and Dependency Listing in Config Files


karabambus:
CTP does not explicitly define the requirements for listing extensions in architectural configuration files when those extensions exist in a superset/subset relationship. Currently, it is unclear whether a configuration file must:

- Only list the highest-level superset.
- Explicitly list all constituent subsets.
- List a specific subset even if the superset is present to trigger specific architectural tests.

I suggest UDB parameters page as candidate, as it is missing implemented extensions parameter and seems to further clarify rules about it.

jordancarlin:
UDB validates the config files and should give fairly descriptive error messages when the appropriate extensions are not all specified. It requires all of the subset extensions to be present in addition to the superset extension. ACT4 relies on the full list of subset extensions since that is the granularity at which tests are selected.

If you think additional explanation in the CTP would make this more clear, please open a PR.

jordancarlin:
@karabambus are you ok closing this now? Or is there still a change that you think should be made?

PR #1173, opened 2026-04-01
ctp: clarify superset/subset extension listing requirement

karabambus:
- added note in strategy section that clarifies superset/subset extension listing requirement

This PR addresses issue #1127


docs/ctp/src/strategy.adoc
```diff
@@ -18,6 +18,11 @@ This file lists:
 * all of the supported extensions
 * all of the parameters required to describe the behavior of the extensions
 
+[NOTE]
+====
+When an extension implies sub-extensions (e.g. `C` implies `Zca`/`Zcf`/`Zcd`; `A` implies `Zaamo`/`Zalrsc`), the `implemented_extensions` list must include *both the superset and all applicable sub-extensions*.  UDB validates this, and the test framework selects test suites by sub-extension name.
```
davidharrishmc, line 23: superset -> primary extension

Are you certain this is true?  If C were omitted, would there be any problems?

karabambus: I relied on Jordan's comments  [here](https://github.com/riscv/riscv-arch-test/issues/1127#issuecomment-4127365643). 

Although it is mentioned that sub extensions ultimately define what tests would be generated.
> ACT4 relies on the full list of subset extensions since that is the granularity at which tests are selected.

Tests showed that removing C would be valid by schema, and produce valid tests, DUT might also pass certification (DUT and Sail sig comparisson).

But that is all with one small problem. We must remove MUTABLE_MISA_C parameter since we don't have C. This would be also case for other extensions, since primary extension defines what parameters are needed in schema. I am not sure how critical is MUTABLE_MISA_C but even in case that it doesn't modify tests in any meaningful way, we can speculate some extensions might have parameters that would be critical and cant be omitted.

Writing this - this might be issue with UDB. That it validates schema that contains sub extensions but not its primary.

davidharrishmc: So it sounds like C is required because this DUT has MUTABLE_MISA_C, not because compressed instructions are supported.  The note introduces ambiguity, possibly implying that C is required anytime Zca, Zcf, Zcd are supported.  

I'd lean toward remaining silent, rather than adding this note.  UDB will complain if the necessary extensions are missing.

karabambus: I agree with omitting the note for now. However, it is worth noting that omitting the primary extension creates a silent bypass of schema validation. Since UDB only expects MUTABLE_MISA_C when C is explicitly defined, removing C prevents the validator from catching missing or mismatched parameters.

My current tests only passed because MUTABLE_MISA_C happened to default to false, matching my current DUTs. If we encounter a configuration where MUTABLE_MISA_C must be true, we would be unable to define it in the schema without the primary extension. In that scenario, we would likely see a certification failure that UDB would fail to flag beforehand.

jordancarlin: Agreed that we should skip this note. UDB might change its approach and this note would get stale. At the end of the day, users need to do whatever UDB requires to get their config validated.

PR #1178, opened 2026-04-01
VF with Custom 100% coverage

jacassidy:


generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
jordancarlin, line 1: Why is this changing InterruptsSm?

generators/testgen/src/testgen/asm/interrupts.py
jordancarlin, line 1: Why is this changing interrupts.py?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_ls_emul_f16.sv
jordancarlin, line 1: We switched all of the other coverpoint templates to be `*.sv` instead of `*.txt`.

generators/coverage/src/covergroupgen/templates/cp_custom_vfncvt_rup_overflow.txt
jordancarlin, line 1: What is the logic for which templates go in the `vector` directory and which don't? Seems like this one should?

generators/coverage/src/covergroupgen/generate.py
```diff
@@ -363,6 +383,12 @@ def write_covergroups(
         lines: list[str] = []
         init_lines: list[str] = []
 
+        # Undef sibling SEW macros so only this arch's macro is active
+        # (prevents conflicts when multiple SEW coverage files are compiled together)
+        if vector:
+            for macro in _get_sibling_sew_macros(arch, all_archs):
+                lines.append(f"`ifdef {macro}\n  `undef {macro}\n`endif\n")
```
jordancarlin: When would this ever happed? We run coverage for each suite individually. Also, it seems like this would just break coverage if running multiple together by disabling other variants?

generators/testgen/scripts/coverpoint_registry.py
jordancarlin, line 1: Is this trying to emulate what regular testgen does? I'm starting to get concerned by the complexity of vector testgen considering it is not a package. This auto discovery trick can fail in strange ways on different systems when the code is not in a package.

docs/ctp/src/strategy.adoc
```diff
@@ -46,12 +46,12 @@ https://docs.google.com/spreadsheets/d/1tFGLbocTp1YNn11aN8JTmvxL_23e7x8SlBNJfOPR
...
-* https://github.com/riscv/riscv-unified-db/blob/main/cfgs/MC100-32.yaml[MC100]
-* https://github.com/riscv/riscv-unified-db/blob/main/cfgs/profile/RVA22S64.yaml[RVA22S64]
-* https://github.com/riscv/riscv-unified-db/blob/main/cfgs/profile/RVB23S64.yaml[RVB23S64]
-* https://github.com/riscv/riscv-unified-db/blob/main/cfgs/profile/RVA23S64.yaml[RVA23S64]
+* https://riscv.github.io/riscv-unified-db/resolved_arch/profile/RVI20U32.yaml[RVI20U32]
+* https://riscv.github.io/riscv-unified-db/resolved_arch/profile/RVI20U64.yaml[RVI20U64]
+* https://github.com/riscv/riscv-unified-db/blob/main/cfgs/mc100-32-full-example.yaml[MC100]
+* https://riscv.github.io/riscv-unified-db/resolved_arch/profile/RVA22S64.yaml[RVA22S64]
+* https://riscv.github.io/riscv-unified-db/resolved_arch/profile/RVB23S64.yaml[RVB23S64]
+* https://riscv.github.io/riscv-unified-db/resolved_arch/profile/RVA23S64.yaml[RVA23S64]
```
jordancarlin: Why are these links being changed?

framework/src/act/build_plan.py
```diff
@@ -249,14 +249,13 @@ def gen_compile_tasks(
...
-    # 4. final.elf – compile with -DRVTEST_SELFCHECK
+    # 4. final.elf – compile without -DRVTEST_SELFCHECK (disabled for coverage)
     final_elf_cmd = [
         *compiler_cmd,
         "-o",
         str(final_elf),
         f"-march={march}",
         f"-mabi={mabi}",
-        "-DRVTEST_SELFCHECK",
         f"-DXLEN={xlen}",
```
jordancarlin: Why are you disabling self check? This doesn't seem right.

framework/src/act/config.py
```diff
@@ -7,6 +7,8 @@
 # Parse test framework configuration files
 ##################################
 
+from __future__ import annotations
```
jordancarlin, line 10: Not needed.

framework/src/act/config.py
```diff
@@ -104,6 +106,11 @@ def resolve_relative_paths(cls, v: str | None, info: ValidationInfo) -> Path | N
         config_file_dir: Path = context["config_file_dir"]
         return config_file_dir.absolute() / path
 
+    @property
+    def compiler_string(self) -> str:
+        """Return the compiler executable as a string for Makefile generation."""
+        return str(self.compiler_exe)
+
```
jordancarlin: Why are you adding this? It isn't used.

framework/src/act/makefile_gen.py
jordancarlin, line 1: This was deleted in the act4 branch.

coverpoints/priv/InterruptsU_coverage.svh
jordancarlin, line 1: Merge issues?

coverpoints/priv/InterruptsSm_coverage.svh
jordancarlin, line 1: Merge issues?

jordancarlin (changes requested):
Lots of merge issues.

framework/src/act/config.py
jacassidy: I was getting lint errors

jordancarlin: That must be caused by some other change you made. This function should not be used anywhere anymore.

coverpoints/priv/InterruptsSm_coverage.svh
jacassidy: No this is intended, I like having it because it allows us to test all the extent of every coverpoint in rv32 because this is a legal implementation

generators/testgen/scripts/coverpoint_registry.py
jacassidy: yes that was my goal, should I just copy the format you were using this was fine for me

framework/src/act/config.py
jacassidy: ruff wont pass without this

coverpoints/priv/InterruptsSm_coverage.svh
jordancarlin: This is a comment on InterruptsSm. Not sure what that has to do with the "extend of every coverpoint in rv32".

framework/src/act/config.py
jordancarlin: Strange. It should work fine since we are using Python 3.12. Fine to leave it though.

framework/src/act/fcov/coverage/RISCV_coverage_common.svh
```diff
@@ -198,6 +218,9 @@
 //   `endif
 // `endif
 
+`define ELEN64
```
jordancarlin: Isn't this DUT specific? Seems like this shouldn't be in the common file.

generators/coverage/src/covergroupgen/generate.py
```diff
@@ -346,6 +346,24 @@ def _gen_instruction_samples(
 ##################################
 
 
+def _get_sibling_sew_macros(arch: str, all_archs: list[str]) -> list[str]:
```
jordancarlin: Unused function?

generators/testgen/scripts/coverpoint_registry.py
jordancarlin: Let's leave it for now. We really need to get vector testgen merged into the main testgen package though to avoid this.

jordancarlin (changes requested):
Haven't reviewed any of the actual vector coverpoints or testgen code yet.

davidharrishmc (approved):

generators/coverage/src/covergroupgen/templates/cp_csr_frm.sv
jordancarlin, line 1: Why is this changing a non-vector coverpoint? And have you confirmed that we still get full coverage for the existing extensions?

generators/coverage/src/covergroupgen/templates/cp_csr_frm_v.sv
jordancarlin, line 1: Why is this changing a non-vector coverpoint? And have you confirmed that we still get full coverage for the existing extensions?

generators/coverage/src/covergroupgen/generate.py
```diff
@@ -363,6 +388,12 @@ def write_covergroups(
         lines: list[str] = []
         init_lines: list[str] = []
 
+        # Undef sibling SEW macros so only this arch's macro is active
+        # (prevents conflicts when multiple SEW coverage files are compiled together)
+        if vector:
+            for macro in _get_sibling_sew_macros(arch, all_archs):
+                lines.append(f"`ifdef {macro}\n  `undef {macro}\n`endif\n")
```
jordancarlin: Looks like there may have been a git issue? I think I already commented on this and I thought I remember you removing it. Anyway, when would this ever happed? We run coverage for each suite individually. Also, it seems like this would just break coverage if running multiple together by disabling other variants?

generators/coverage/src/covergroupgen/templates/vector/cp_vstart_vl_noop.sv
```diff
@@ -0,0 +1,13 @@
+//////////////////////////////////////////////////////////////////////////////////
+    // cp_vstart_vl_noop
+    //////////////////////////////////////////////////////////////////////////////////
```
jordancarlin: Seems like something strange is going on with the header here? Why are the second two lines of the header comment indented but not the first?

jordancarlin (changes requested):
Looking pretty close. A few last comments.

generators/coverage/src/covergroupgen/templates/cp_csr_frm_v.sv
jacassidy: I confirmed this is still 100% coverage on F and D, this happened because of the issue surrounding only fcsr updating so I changed all of them, I can change these back if desired

jacassidy:
@jordancarlin all set to merge

coverpoints/unpriv/Vf16_coverage.svh
```diff
@@ -51,10 +52,97 @@ covergroup Vf16_vfadd_vf_cg with function sample(ins_t ins);
...
+    cp_custom_vfp_NaN_input : cross std_vec, vs2_element0_sqNAN;
+    `endif
+`endif
+
+    //// end cp_custom_vfp_NaN_input////////////////////////////////////////////////
+    //////////////////////////////////////////////////////////////////////////////////
+    // cp_custom_vfp_flags_nv_nx_of
+    // For instructions that can raise NV, NX, and OF but not DZ/UF
+    // with the standard test vectors: vfadd.vv, vfadd.vf.
+    //////////////////////////////////////////////////////////////////////////////////
```
jordancarlin, line 88: The formatting here is kind of confusing. Can this be cleaned up? At least a blank line between the two maybe?

coverpoints/unpriv/Vf16_coverage.svh
```diff
@@ -448,6 +639,74 @@ covergroup Vf16_vfclass_v_cg with function sample(ins_t ins);
...
+        bins bit_0_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0001};
+        bins bit_1_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0010};
+        bins bit_2_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0100};
+        bins bit_3_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_1000};
+        bins bit_4_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0001_0000};
+        bins bit_5_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0010_0000};
+        bins bit_6_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0100_0000};
+        bins bit_7_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_1000_0000};
+        bins bit_8_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0001_0000_0000};
+        bins bit_9_1  = {64'b0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0010_0000_0000};
```
jordancarlin, line 657: Doesn't need to happen in this PR, but take a look at some of the other coverpoints for an example of a much more concise way to write walking ones coverpoints.

coverpoints/unpriv/Zalrsc_coverage.svh
jordancarlin, line 1: Why is this file changing?

framework/src/act/fcov/coverage/RISCV_coverage_common.svh
```diff
@@ -220,6 +158,16 @@
...
 
+`ifdef SEW64_SUPPORTED
+  `define ELEN64
+`elsif SEW32_SUPPORTED
+  `define ELEN32
+`elsif SEW16_SUPPORTED
+  `define ELEN16
+`elsif SEW8_SUPPORTED
+  `define ELEN8
+`endif
```
jordancarlin: Isn't ELEN a parameter? In which case it should just directly come from `riscv_config.svh`?

jacassidy: yes its also defined as the max sew supported but I can move it

jordancarlin (approved):

PR #1179, opened 2026-04-01
Report no summary line as a failure in run_tests

jordancarlin:
Killing the Imperas iss from another shell when it hung was being reported as a success before.


davidharrishmc:
Thank you.  That had been on my list too.

davidharrishmc (approved):

PR #1181, opened 2026-04-01
feat: implement Zcmop unprivileged test suite

DymShanks:
implements the complete unprivileged test suite for the `Zcmop` (Compressed May-Be-Operations) extension, following the `act4` unprivileged test generation strategy.
As defined in the RISC-V Unprivileged Specification, the `c.mop.n` instructions are nullary and do not write to any register. Therefore, this suite strictly validates execution without fabricating operand or register-value coverpoints. 

##Implementation Details:
* **Testplan (`Zcmop.csv`):** Added all 8 compressed instructions (`c.mop.1` through `c.mop.15`) as `CI` type, tracking only `cp_asm_count`.
* **Coverpoints (`Zcmop.yaml`):** Implemented normative coverpoint specifically targeting `cp_asm_count` to verify execution without raising illegal-instruction exceptions.
* **Disassembler (`disassemble.svh`):** Injected the 8 decoding strings for the framework.
* **Documentation (`unprivmisc23.adoc`):** Uncommented the Zcmop testplan inclusion to generate the CTP tables.

*(While verifying the Imperas configs, I noticed that the `rv64-max` config was missing the flags for both `Zimop` and `Zcmop`. I have gone ahead and added both `--override cpu*/Zimop=T` and `--override cpu*/Zcmop=T` to `imperas-rv64-max/imperas.ic` in this PR to ensure the 64-bit simulator catches both MOP extensions.

### DIFF
--- a/config/imperas/imperas-rv64-max/imperas.ic
+++ b/config/imperas/imperas-rv64-max/imperas.ic
@@ -67,7 +67,5 @@
 --override cpu*/PMP_undefined=T # access to unimplemented PMP registers cause illegal instruction exception
 
 # More extensions
---override cpu*/Zimop=T
---override cpu*/Zcmop=T
 #--override cpu*/Smdbltrp=T
 #--override cpu*/Ssdbltrp=T
--- a/config/imperas/imperas-rv64-max/imperas.ic
+++ b/config/imperas/imperas-rv64-max/imperas.ic
@@ -67,5 +67,7 @@
 --override cpu*/PMP_undefined=T # access to unimplemented PMP registers cause illegal instruction exception
 
 # More extensions
+--override cpu*/Zimop=T
+--override cpu*/Zcmop=T
 #--override cpu*/Smdbltrp=T
 #--override cpu*/Ssdbltrp=T
--- a/coverpoints/norm/Zcmop.yaml
+++ b/coverpoints/norm/Zcmop.yaml
@@ -1,0 +1,10 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2024 RISC-V International
+
+Zcmop_instruction_format:
+  description: >
+    Verify that all c.mop.n instructions correctly execute
+    as May-Be-Operations. They must not write to any register,
+    and must not raise an illegal-instruction exception.
+  coverpoints:
+    - cp_asm_count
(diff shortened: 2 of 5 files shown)


config/imperas/imperas-rv64-max/imperas.ic
```diff
@@ -67,5 +67,7 @@
 --override cpu*/PMP_undefined=T # access to unimplemented PMP registers cause illegal instruction exception
 
 # More extensions
+--override cpu*/Zimop=T
+--override cpu*/Zcmop=T
```
jordancarlin: There are automatically enabled since this uses the RVA23S64 profile as its base (which requires these extensions).

coverpoints/norm/Zcmop.yaml
jordancarlin, line 1: This is not the correct format for the coverpoint norm rule yaml files. Please refer to any of the other yamls in this directory.

testplans/Zcmop.csv
```diff
@@ -0,0 +1,9 @@
+Instruction,Type,RV32,RV64,cp_asm_count
+c.mop.1,CI,x,x,x
```
jordancarlin: CI type doesn't seem like the right fit. Based on the instruction formatter functions, CI type produces instructions with the format `ins rs1, imm`. These take neither an rs1 nor an immediate.

jordancarlin (changes requested):

testplans/Zcmop.csv
DymShanks: thanks for the review. You're right that CI is incorrect — c.mop.n takes no operands. I investigated the formatters directory and found there's no zero-operand compressed formatter currently. c.nop in Zca.csv also uses type CN but cn_type.py doesn't exist yet in the formatters.
Should I create a new cn_type.py formatter for zero-operand compressed instructions as part of this PR? Or is there an existing type I should use instead?"

jordancarlin: The `cp_asm_count` generator doesn't use the formatters, and since that is the only coverpoint, no Python formatter is needed for the CN type.

DymShanks: Thanks for the clarification, Jordan. Since cp_asm_count only emits the instruction name and doesn't need the Python formatter, I've updated the testplan to use the CN type label. The latest commit reflects this change.

jordancarlin (approved):

Issue #1148, opened 2026-03-28
Vector Selfcheck special cases and agnostic relaxation


georgiatai:
As mentioned by @Jimx- in PR #1038, vector self-check are currently missing special cases for vcompress and needs minor fixes on agnostic relaxation:

- For vcompress, there can be fewer active elements in vd than original vl. In that case, an element can be treated as a tail element even though its index is < original vl.
- According to [Section 30.1](https://docs.riscv.org/reference/isa/unpriv/v-st-ext.html#sec-agnostic) of the spec, mask tails are always treated as agnostic to reduce complexity, regardless of vta. Agnostic relaxation shouldn't be skipped here if mask-prod flag is set.

davidharrishmc:
@fourth-bit is this resolved by your vector self-check PR?  If so, go ahead and close.

fourth-bit:
Fixed in PR #1549.

PR #1183, opened 2026-04-02
vcompress and mask-producing instr tail elements signature fix

georgiatai:
Fixes on signature for vcompress and mask-producing instruction tail elements as mentioned in issue #1148


georgiatai:
Add indentations so `signature.h` is a little unclear on what is changed, the main changes are line 368-369, 375-379, 409-416, and the corresponding non-selfchecking version. 

This PR was closed and reopened because there are things that I thought I haven't fix but I actually did. Sorry for the confusion.

Jimx-:
It looks like vcompress is not the only instruction that can have a different effective vector length of vd than the original `vl`. `vmv.s.x` and all reduction operations write scalar result to element 0 of vd and other elements (0 < index < VLEN/VSEW) are tail elements. They are like vcompress but with only 1 active element in vd.

davidharrishmc:
Solved with #1549.

jordancarlin:
@georgiatai @fourth-bit can this be closed now that #1549 was merged instead? Or are there additional changes in this PR that are still needed?

jordancarlin:
Closing because #1549 was built on top of this.

PR #1184, opened 2026-04-02
fix Sail hang on ExceptionsSm instr_access_fault

karabambus:
The instr_access_fault coverpoint jumps to address 0x0 to trigger an instruction access fault. With the memory region starting at 0x0, Sail fetched 0x0000 (c.illegal) instead of faulting, causing an infinite illegal-instruction trap loop.

Fix: start the memory region at 0x4000 to match link.ld, leaving 0x0-0x3FFF unmapped so Sail correctly raises an instruction access fault.

### DIFF
--- a/config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/sail.json
+++ b/config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/sail.json
@@ -65,11 +65,11 @@
       {
         "base": {
           "len": 64,
-          "value": "0x0"
+          "value": "0x00000080"
         },
         "size": {
           "len": 64,
-          "value": "0x80000000"
+          "value": "0x7FFFFF80"
         },
         "attributes": {
           "cacheable": true,
--- a/config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imc/sail.json
+++ b/config/cores/cv32e40p/cv32e40p_v1.8.3_rv32imc/sail.json
@@ -65,11 +65,11 @@
       {
         "base": {
           "len": 64,
-          "value": "0x0"
+          "value": "0x00000080"
         },
         "size": {
           "len": 64,
-          "value": "0x80000000"
+          "value": "0x7FFFFF80"
         },
         "attributes": {
           "cacheable": true,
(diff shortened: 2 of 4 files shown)


config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/sail.json
```diff
@@ -65,11 +65,11 @@
       {
         "base": {
           "len": 64,
-          "value": "0x0"
+          "value": "0x00000080"
```
davidharrishmc, line 68: The base address should match the DUT memory map.  Where does memory start in each of these cores?

If DUT memory really starts at 0, Sail should not be changed to 0x80 or 0x4000 because then the DUT will behave differently from Sail.  

An alternative is to change RVMODEL_ACCESS_FAULT_ADDRESS to point to an address that is not implemented and will cause an access fault.

davidharrishmc (changes requested):

config/cores/cv32e40p/cv32e40p_v1.0.0_rv32imc/sail.json
karabambus: [cv32e20](https://github.com/openhwgroup/cv32e20-dv) it is set to 0x4000 (hardcoded), for [cv32e40p-dv](https://github.com/karabambus/cv32e40p-dv-review/pull/1) we override it at runtime with elf entry point (this repo is still under review this might not be final).

So this is set accordingly. Though for consistency sake I think I should set it all on same value. Maybe both to 0x80. Or use RVMODEL_ACCESS_FAULT_ADDRESS with some upper bound. What do you suggest?

davidharrishmc:
You are running on RTL and the RTL has s certain memory map.

Sail should be configured to match the memory map of the DUT, not based on RVMODEL_ACCESS_FAULT_ADDRESS.  RVMODEL_ACCESS_FAULT_ADDRESS should be set somewhere that the DUT does not have memory.

davidharrishmc:
Good catch!

karabambus:
this PR is blocked by #1192

davidharrishmc:
I believe this PR is unnecessary after #1248.

PR #1186, opened 2026-04-02
Enhance Zacas coverage and add additional Zimop coverpoints

amanmurad-10xe:
## Zacas (Compare-and-Swap) Updates

* Updated Zacas.csv with comprehensive testcases focusing on rd vs rs1 value comparisons
* Implemented a custom Python generator dedicated to handling Zacas value comparison logic and intelligent register allocation
* Added new SystemVerilog (SV) coverpoints targeting CAS (Compare-and-Swap) success and failure states
* Resolved trace memory-visibility limitations by evaluating pre-instruction and post-instruction rd write-back values to accurately verify CAS logic and memory atomicity

## Zimop Extension Updates

* Added cmp (comparison) coverpoints to mop.r.n and mop.rr.n instructions for enhanced coverage

## Test Coverage & Validation

* Achieved 100% architectural coverage of all coverpoints on both RV32 and RV64 variants
* All new testcases validated against the RISC-V specification requirements


framework/src/act/fcov/disassemble.svh
jordancarlin, line 1: Looks like just white space changes in this file. Please revert.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_eq.sv
```diff
@@ -0,0 +1,5 @@
+cmp_rd_rs1_val_eq : coverpoint (ins.current.rd_val == ins.prev.rd_val) iff (ins.trap == 0) {
+    // Compare RD current to RD previous value which is basically rs1 current value
+    bins rd_val_equal_rs1  = {1}; // This bin is hit when rd_val is equal to rs1 value (which is the previous rd_val)
+    bins rd_val_not_equal_rs1 = {0}; // This bin is hit when rd_val is not equal to rs1 value (which is the previous rd_val)
```
jordancarlin: These comments don't seem to add much. I'd recommend dropping them. The coverpoint will also automatically create bins for 0/1 if those are the only two possibilities, so no need to explicitly declare the bins unless you think giving them names provides significant value over the default 0/1.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_hw.sv
```diff
@@ -0,0 +1,5 @@
+    cmp_rd_rs1_val_hw : coverpoint (ins.current.rd_val[15:0] == ins.prev.rd_val[15:0]) iff (ins.trap == 0) {
+        // Compare the least significant 16 bits of rd and rs1 register values
```
jordancarlin: The comment does not agree with the coverpoint. The comment says it compares rd and rs1 values but the actual coverpoint compares the previous rd value to the new rd value.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_pair_partial_val.sv
```diff
@@ -0,0 +1,9 @@
+    cmp_rd_rs1_pair_partial_val : coverpoint (
+            (ins.current.rd_val[`XLEN-1:0] == ins.prev.rd_val[`XLEN-1:0]) ^
+            (ins.current.rd_val[2*`XLEN-1:`XLEN] == ins.prev.rd_val[2*`XLEN-1:`XLEN])
```
jordancarlin: How does this work? `rd_val` is defined to be XLEN bits wide (see below). `` rd_val[2*`XLEN-1:`XLEN] `` should not exist.

```sv
bit [(XLEN-1):0]          rd_val;
```

testplans/Zacas.csv
```diff
@@ -1,5 +1,5 @@
-Instruction,Type,RV32,RV64,cp_asm_count,cp_rs1,cp_rs2,cp_rd,cp_rs2_edges,cmp_rs1_rs2,cmp_rd_rs1,cmp_rd_rs2,cmp_rd_rs1_rs2,cp_align
-amocas.w,A,x,x,x,nx0,x,x,x,nx0,nx0,x,nx0,word
-amocas.d,A,,x,x,nx0,x,x,x,nx0,nx0,x,nx0,
-amocas.d,AP,x,,x,nx0,pair,pair,x,nx0_pair,nx0_pair,pair,nx0_pair,
-amocas.q,AP,,x,x,nx0,pair,pair,x,nx0_pair,nx0_pair,pair,nx0_pair,
+Instruction,Type,RV32,RV64,cp_asm_count,cp_rs1,cp_rs2,cp_rd,cp_rs2_edges,cmp_rs1_rs2,cmp_rd_rs1,cmp_rd_rs2,cmp_rd_rs1_rs2,cp_align,cmp_rd_rs1_val_eq,cmp_rd_rs1_val_lsb,cmp_rd_rs1_val_hw,cmp_rd_rs1_val_w,cmp_rd_rs1_val_d,cmp_rd_rs1_pair_partial_val,cmp_rd_rs1_sign_ext
```
jordancarlin: All of these new coverpoints need to be documented in the CTP. Please review how other special coverpoints are handled and add theses to the Zacas section. Also make sure to update `coverpoints/norm/Zacas.yaml` appropriately to link these coverpoints to normative rules.

generators/testgen/src/testgen/formatters/types/ap_type.py
```diff
@@ -22,26 +22,67 @@ def format_ap_type(
...
+    assert params.rd is not None and params.rdval is not None
+    assert params.temp_reg is not None
 
     # Ensure rs1 is not x0 (base address)
     if params.rs1 == 0:
         test_data.int_regs.return_register(params.rs1)
         params.rs1 = test_data.int_regs.get_register(exclude_regs=[0])
 
+    # Check if this instruction uses pair registers, which requires additional setup and checking of upper halves of registers and memory
+    is_pair = getattr(params, "rd_is_pair", False)
```
jordancarlin: Why is this needed? The `AP` type is only used for instructions that use register pairs.

generators/testgen/src/testgen/formatters/types/ap_type.py
```diff
@@ -22,26 +22,67 @@ def format_ap_type(
...
+
+    # Apply mask to ensure values fit within xlen, and calculate offset for upper half of memory for pair registers
+    mask = (1 << test_data.xlen) - 1
```
jordancarlin, line 35: The values that `params` is populated with should already be XLEN bits long. Why was this added? Were you seeing larger numbers?

generators/testgen/src/testgen/formatters/types/ap_type.py
```diff
@@ -22,26 +22,67 @@ def format_ap_type(
...
+        load_int_reg("value in memory", params.temp_reg, params.rs1val & mask, test_data),
+        load_int_reg("rs2", params.rs2, params.rs2val & mask, test_data),
+        load_int_reg("rd compare value", params.rd, params.rdval & mask, test_data),
         f"LA(x{params.rs1}, scratch) # load base address into rs1",
-        f"SREG x{params.temp_reg}, 0(x{params.rs1}) # store value into memory at address in rs1",
+        f"SREG x{params.temp_reg}, 0(x{params.rs1}) # store lower value into memory",
     ]
+
+    # For pair registers, setup upper halves and write upper half to memory
+    if is_pair:
```
jordancarlin: This can be combined with the above section since the type is only used for paired registers.

generators/testgen/src/testgen/formatters/types/ap_type.py
```diff
@@ -22,26 +22,67 @@ def format_ap_type(
...
+        ]
+    )
+
+    # Load and check the upper half of memory for pair registers to ensure the instruction correctly updated both halves
+    if is_pair:
+        check.extend(
+            [
+                f"LA(x{params.rs1}, scratch) # reload base address",
+                f"LREG x{params.rs1}, {offset}(x{params.rs1}) # Load the updated upper memory value",
+                write_sigupd(params.rs1, test_data, "int"),
```
jordancarlin: Can we load the scratch address into one register and the data into another so we don't overwrite the address and need to load it twice?

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
```diff
@@ -0,0 +1,344 @@
...
+    desc: str,
+    bin_name: str,
+    rd_val: int,
+    rs1_val: int,
+    load_rd: bool = True,
+) -> TestChunk:
+    """Generate a generic compare test case for CAS instructions"""
+
+    # Check if this instruction uses register pairs (value is 2 * XLEN)
+    is_pair = (instr_name == "amocas.q" and test_data.xlen == 64) or (instr_name == "amocas.d" and test_data.xlen == 32)
```
jordancarlin: Could simplify this by using the instruction type.

jordancarlin (changes requested):
Thanks for working on this! I haven't finished reviewing it yet, but here is some initial feedback.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_pair_partial_val.sv
amanmurad-10xe: Yes, you have pointed this out correctly. This is working fine because it only checks the lower half of the register values and gives result accordingly. 

I have tested by removing the XOR statement which is actually out of bound comparison, i still got the complete coverage report.

I am trying to write a coverpoint which is basically checking the partial comparison between rd and rs1 register values of both the register pairs in amocas.d rv32 and amocas.q rv64. But i am unable to read the upper register which is logically rd+1 or rs1+1. 

Is there any way that we can read rd+1 in sv coverpoint??

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_hw.sv
amanmurad-10xe: Screenshot from 2026-04-13 11-08-26

Here is a brief overview of how Zacas new coverpoints are implemented:

Architecturally, the success or failure of a Zacas instruction is determined by comparing the initial value in rd against the memory value at (rs1). However, since our SystemVerilog coverpoints rely on RVVI traces, they only have visibility into register states and cannot access memory directly during runtime. To bypass this limitation, we observe a specific architectural behavior of Zacas: it always writes the original memory value back into rd upon completion.

Therefore, these coverpoints evaluate the memory comparison by checking the pre-instruction rd state against the post-instruction rd writeback state. By comparing ins.prev.rd_val (the initial comparison value we set) with ins.current.rd_val (the actual memory value written back by the hardware), we can accurately track success, failure, and partial-match edge cases entirely through register states, without needing direct memory access.

I have fixed the comments in this file.

generators/testgen/src/testgen/formatters/types/ap_type.py
amanmurad-10xe: You are right. I assumed that it might be used for some instructions in future.

I have removed is_pair check since AP type is exclusively used for paired-register instructions.

generators/testgen/src/testgen/formatters/types/ap_type.py
amanmurad-10xe: Yes, exactly! Because we are dealing with paired registers (amocas.d requires 64 bits on RV32, and amocas.q requires 128 bits on RV64), the test generator creates the full double-width value and passes it as a single Python integer into params.rdval, rs1val, and rs2val.

Because Python integers can be arbitrarily large, those variables actually hold the full 2*XLEN value.

generators/testgen/src/testgen/formatters/types/ap_type.py
amanmurad-10xe: Great suggestion. I will fix it so that there is no more redundant address loading and overwriting.

testplans/Zacas.csv
amanmurad-10xe: Updated the `coverpoints/norm/Zacas.yaml`. Okay, I will document these in the CTP too.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_pair_partial_val.sv
davidharrishmc: framework/src/act/fcov/coverage/RISCV_instruction_base.svh defines current to be of type riscvTraceData.  

In RISCV_instruction_base.svh, the add_rd method accesses prev.x_wdata[get_gpr_num(ops[offset].key)].  It seems likely this could be modified to read rd+1 where needed.

coverpoints/norm/Zacas.yaml
```diff
@@ -3,40 +3,55 @@
...
   - name: Zacas_rv64_amocas-q_op
-    coverpoint: [""]
+    coverpoint:
+      [
+        "Zacas_amocas_q_cg/{cp_asm_count, cp_rs1, cp_rs2, cp_rd, cmp_rd_rs1_val_eq, cmp_rd_rs1_val_lsb, cmp_rd_rs1_val_hw, cmp_rd_rs1_val_w, cmp_rd_rs1_val_d, cmp_rd_rs1_pair_partial_val}",
+      ]
 
   - name: Zacas_rv64_amocas-q_frst_pair_entry_reg_even
-    coverpoint: [""]
+    coverpoint: ["Zacas_amocas_q_cg/{cp_rs2, cp_rd}"]
```
davidharrishmc: I think these should be cp_rs2_pair, cp_rd_pair.  Similar for other rules.

coverpoints/norm/Zacas.yaml
```diff
@@ -3,40 +3,55 @@
...
-    coverpoint: [""]
+    coverpoint: ["Zacas_amocas_q_cg/{cp_rs2}"]
 
   - name: Zacas_rv64_amocas-q_rd_frst_reg_x0
-    coverpoint: [""]
+    coverpoint: ["Zacas_amocas_q_cg/{cp_rd}"]
 
   - name: Zacas_amocas_rs1_addr_alignment
-    coverpoint: [""]
+    coverpoint: ["cp_align"]
```
davidharrishmc: The Zacas instructions need to be added to ExceptionsZaamo_coverage.svh and the associated generator to test illegal alignments.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_sign_ext.sv
```diff
@@ -0,0 +1,4 @@
+    cmp_rd_rs1_sign_ext : coverpoint (ins.current.rd_val == $signed(ins.current.rd_val[31:0])) iff (ins.trap == 0) {
```
davidharrishmc: I think this should just be
cmp_rd_sign_ext : coverpoint ins.current.rd_val[31] iff (ins.trap == 0) {
}

This checks both signs of the lower 32-bit value being exercised.  The reference model will check that the correct answer comes out (with sign extension).

There is no rs1 involved, so adjusted the name too. That will require renaming the coverpoint in the testplan.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_d.sv
```diff
@@ -0,0 +1,6 @@
+    cmp_rd_rs1_val_d : coverpoint (ins.current.rd_val[63:0] == ins.prev.rd_val[63:0]) iff (ins.trap == 0) {
```
davidharrishmc: Why do we test this, and only for amocas.q?

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_hw.sv
```diff
@@ -0,0 +1,6 @@
+    cmp_rd_rs1_val_hw : coverpoint (ins.current.rd_val[15:0] == ins.prev.rd_val[15:0]) iff (ins.trap == 0) {
```
davidharrishmc, line 1: What do we learn from this coverpoint?

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_lsb.sv
```diff
@@ -0,0 +1,6 @@
+    cmp_rd_rs1_val_lsb : coverpoint (ins.current.rd_val[7:0] == ins.prev.rd_val[7:0]) iff (ins.trap == 0) {
```
davidharrishmc, line 1: What do we learn from this coverpoint?

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_w.sv
```diff
@@ -0,0 +1,6 @@
+    cmp_rd_rs1_val_w : coverpoint (ins.current.rd_val[31:0] == ins.prev.rd_val[31:0]) iff (ins.trap == 0) {
+        // Compare the lowest 32 bits of current rd value to the
```
davidharrishmc, line 2: What do we learn from this coverpoint?

generators/testgen/src/testgen/formatters/types/ap_type.py
davidharrishmc: I'm not familiar with this part, but how do we know the variables have interesting 128-bit values in amocas.q rather than only being 64 bits?  For example, isn't cp_rs2_edges produce XLEN values, not 2*XLEN?  Have you visually inspected the code and find interesting values?

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
```diff
@@ -0,0 +1,341 @@
...
+
+    # Determine if this is a pair register instruction based on the instruction type
+    is_pair = instr_type == "AP"
+
+    # Split values if dealing with a register pair
+    mask = (1 << test_data.xlen) - 1
+
+    mask_val = (1 << (2 * test_data.xlen)) - 1 if is_pair else mask
+
+    rd_val &= mask_val
```
davidharrishmc, line 40: How do we know rd_val initially has at least 2*XLEN interesting bits?

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
```diff
@@ -0,0 +1,341 @@
+#################################
```
davidharrishmc, line 1: It is hard to follow all of this logic.  Could you attach the four amocas .S files instead?

davidharrishmc (changes requested):

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_sign_ext.sv
amanmurad-10xe: Updated.

coverpoints/norm/Zacas.yaml
amanmurad-10xe: Done! I've updated these to cp_rs2_pair and cp_rd_pair here and applied the same fix to the other relevant rules.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_d.sv
amanmurad-10xe: I was trying to catch a partial match on the lower 64 bits for the quad instruction. But since our framework is 64-bit anyway, doing [63:0] is identical to the coverpoint that compares the full rd and rs1 values. I'll remove this.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_hw.sv
amanmurad-10xe: I added this to explicitly track partial matches. A full `rd == rs1` check only catches complete matches or complete mismatches. This coverpoint ensures we are hitting the specific case where just the halfword matches, which acts as an expected failure condition for wider instructions

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_lsb.sv
amanmurad-10xe: I added this to explicitly track partial matches. A full `rd == rs1` check only catches complete matches or complete mismatches. This coverpoint ensures we are hitting the specific case where just the least significant byte (LSB) matches, which acts as an expected failure condition for wider instructions

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_val_w.sv
amanmurad-10xe: I added this to explicitly track partial matches. A full `rd == rs1` check only catches complete matches or complete mismatches. This coverpoint ensures we are hitting the specific case where just the lower 32 bits (word) match, which acts as an expected failure condition for wider instructions like `.d` or `.q`."

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
amanmurad-10xe: [Zacas_Tests.zip](https://github.com/user-attachments/files/27126539/Zacas_Tests.zip)

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
amanmurad-10xe: rd_val is not assumed to initially have 2*XLEN meaningful bits. It’s just a random value generated by the coverpoint.

For pair instructions, we explicitly construct a 2*XLEN value by combining two XLEN-bit halves (hi and lo). For non-pair cases, it remains XLEN-wide.

In generate_cmp_testcase, we then enforce the correct width using masking 2*XLEN for pair instructions and XLEN otherwise.

So we don’t rely on the initial value having 2*XLEN bits, we either construct it (for pairs) or constrain it via masking.

generators/testgen/src/testgen/formatters/types/ap_type.py
amanmurad-10xe: We don’t assume the values have meaningful 2*XLEN bits. The 2*XLEN comes from how pair instructions are represented, not from a single register’s width.

For AP instructions, we’re dealing with register pairs (e.g., rd and rd+1), each XLEN wide. In the generator, we pack both into one Python integer for convenience, which looks like a 2*XLEN value.

The masking and shifting are just to split this combined value back into two XLEN halves and load them correctly into registers and memory. So it’s about handling paired operands, not about having 2*XLEN-wide values.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_pair_partial_val.sv
amanmurad-10xe: I have added support to read rd+1 value.

coverpoints/norm/Zacas.yaml
amanmurad-10xe: Added successfully!

davidharrishmc:
Will get to this review soonish, but I'm in end-of-semester crunch.

testplans/Zacas.csv
davidharrishmc: The docs go in docs/ctp/src/unprivmisc23.adoc under the Zacas heading.  See a.adoc for how additional Zalrsc coverpoints are defined in a table.

amanmurad-10xe: Added.

coverpoints/priv/ExceptionsZaamo_coverage.svh
```diff
@@ -35,6 +36,8 @@ covergroup ExceptionsZaamo_cg with function sample(ins_t ins);
             wildcard bins amomaxu_d = {AMOMAXU_D};
             wildcard bins amomin_d  = {AMOMIN_D};
             wildcard bins amominu_d = {AMOMINU_D};
+            wildcard bins amocas_d  = {AMOCAS_D};
```
davidharrishmc: Why is q applicable to RV64 but d not applicable to RV32?  Seems like d ought to go in RV32.

docs/ctp/src/unprivmisc23.adoc
```diff
@@ -3,8 +3,26 @@
...
+.Zacas Instruction Coverpoints
 include::{testplansdir}/Zacas.adoc[]
 
+[[t-Zacas-coverpoint-definitions]]
+.Zacas Coverpoint Definitions
+[cols="2, 1, 4" options=header]
+|===
+|Coverpoint|Bins|Definition
+3+^|Zacas Value Comparison Coverpoints
+|cmp_rd_rs1_val_eq|1|rd value equals rs1 value (previous rd value).
```
davidharrishmc: Original rd value equals rs1 value.

docs/ctp/src/unprivmisc23.adoc
```diff
@@ -3,8 +3,26 @@
...
+|cmp_rd_rs1_val_lsb|2|Least significant byte of rd and rs1 are equal/not equal.
```
davidharrishmc: Add "original" before each rd

docs/ctp/src/unprivmisc23.adoc
```diff
@@ -3,8 +3,26 @@
...
+|cmp_rd_rs1_val_hw|2|Lowest 16 bits of rd and rs1 are equal/not equal.
+|cmp_rd_rs1_val_w|2|Lowest 32 bits of rd and rs1 are equal/not equal.
+|cmp_rd_rs1_pair_partial_val|2|For register pairs, rd and rs1 have matching high or low halves, but not both.
+|cmp_rd_sign_ext|2|Sign bit of rd (bit 31) is 0 or 1.
+|===
+
 include::norm/Zacas_norm_rules.adoc[]
 
 TODO: Add .d to RV32 and .q to RV64 in testplans directory when testgen is updated to handle register pairs.
```
davidharrishmc: Delete this if it is working.

generators/testgen/src/testgen/formatters/types/ap_type.py
```diff
@@ -22,26 +22,49 @@ def format_ap_type(
...
-        f"LA(x{params.rs1}, scratch) # load base address into rs1",
-        f"SREG x{params.temp_reg}, 0(x{params.rs1}) # store value into memory at address in rs1",
+        # Lower halves of register values and memory value for pair registers
+        load_int_reg("value in memory", params.temp_reg, params.rs1val & mask, test_data),
+        load_int_reg("rs2", params.rs2, params.rs2val & mask, test_data),
+        load_int_reg("rd compare value", params.rd, params.rdval & mask, test_data),
+        # Base address
+        f"LA(x{params.rs1}, scratch)",  # load base address into rs1
+        f"SREG x{params.temp_reg}, 0(x{params.rs1})",  # store lower value into memory
+        # Upper halves of register values and memory value for pair registers
```
davidharrishmc, line 46: Thank you for generating the assembly code.  This is very helpful for identifying issues.  Please take a close look at it and see if you find any other issues that I haven't found.

rd (and rd+1) are not being initialized, so it is unknown whether these amocas will match or not.  Need to add initialization.

It looks like the upper values are not being randomized.  From Zacas-amocas-q-o00.S, the upper values are all 0s or all 1s.  Please randomize everything.

Also, cp_rs1, rs2, rd use random values for the memory vs. rd, so they are extremely unlikely to ever match, and never test writing anything to rd on a successful compare.  Is it easy to turn these into a different coverpoint variant that generates amocas two tests, one matching and one not, by loading the same value into rs2 that is placed in scratch memory?  That way we exercise both cases for every register.  The same thing could apply to amoswap if easy.


```
# Testcase cp_rs1_nx0 (Test source rs1 = x1)
  RVTEST_TESTDATA_LOAD_INT(x3, x8) # load value in memory: x8 = 0xc0ac1a0539c7cb55
  RVTEST_TESTDATA_LOAD_INT(x3, x6) # load rs2: x6 = 0x1f2dbc6490ef8bc0
  RVTEST_TESTDATA_LOAD_INT(x3, x20) # load rd compare value: x20 = 0xcfa8e1a71dea2168
  LA(x1, scratch)
  SREG x8, 0(x1)
  RVTEST_TESTDATA_LOAD_INT(x3, x7) # load rs2 upper: x7 = 0x0000000000000000
  RVTEST_TESTDATA_LOAD_INT(x3, x21) # load rd compare value upper: x21 = 0xffffffffffffffff
  RVTEST_TESTDATA_LOAD_INT(x3, x8) # load value in memory upper: x8 = 0xffffffffffffffff
  SREG x8, 8(x1)
Zacas_amocas_q_cg_cp_rs1_nx0_b1:
  amocas.q x20, x6, (x1)
  # Check if x20 contains the expected result. x2 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
  RVTEST_SIGUPD(x2, x5, x4, x20, Zacas_amocas_q_cg_cp_rs1_nx0_b1, Zacas_amocas_q_cg_cp_rs1_nx0_b1_str)
  # Check if x21 contains the expected result. x2 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
  RVTEST_SIGUPD(x2, x5, x4, x21, Zacas_amocas_q_cg_cp_rs1_nx0_b1, Zacas_amocas_q_cg_cp_rs1_nx0_b1_str)
  LA(x1, scratch)
  LREG x8, 0(x1)
  # Check if x8 contains the expected result. x2 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
  RVTEST_SIGUPD(x2, x5, x4, x8, Zacas_amocas_q_cg_cp_rs1_nx0_b1, Zacas_amocas_q_cg_cp_rs1_nx0_b1_str)
  LREG x8, 8(x1)
  # Check if x8 contains the expected result. x2 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
  RVTEST_SIGUPD(x2, x5, x4, x8, Zacas_amocas_q_cg_cp_rs1_nx0_b1, Zacas_amocas_q_cg_cp_rs1_nx0_b1_str)
```

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
```diff
@@ -0,0 +1,310 @@
...
+    if load_rd:
+        if is_pair:
+            # Load lower half into rd, upper half into rd+1
+            lines.append(load_int_reg("rd compare lo", rd, rd_lo, test_data))
+            lines.append(load_int_reg("rd compare hi", rd + 1, rd_hi, test_data))
+        else:
+            lines.append(load_int_reg("rd compare value", rd, rd_val, test_data))
+
+    # Initialize memory
+    lines.append(f"\tLA(x{rs1}, scratch)")
```
davidharrishmc: Some weird things are happening in cmp_rd_rs1_val_eq:

scratch is getting the same value stored twice.
There is no need to load x21 with LOAD_INT; it could just be copied over with mv.

cmp_rd_rs1_val_lsb, hw etc. also perform two SREG to the same address of the same value.

cmp_rd_rs1_pair_partial_value also performs excess SREG.

```
# Testcase cmp_rd_rs1_val_eq (rd_val == mem_val)
  RVTEST_TESTDATA_LOAD_INT(x17, x21) # load rd compare value: x21 = 0x29156d208bb185da
	LA(x2, scratch)
  RVTEST_TESTDATA_LOAD_INT(x17, x28) # load memory init: x28 = 0x29156d208bb185da
	SREG x28, 0(x2)
  RVTEST_TESTDATA_LOAD_INT(x17, x19) # load value in memory: x19 = 0x29156d208bb185da
  RVTEST_TESTDATA_LOAD_INT(x17, x20) # load rs2: x20 = 0x3064db9c313bd804
  LA(x2, scratch) # load base address into rs1
  SREG x19, 0(x2) # store value into memory at address in rs1
Zacas_amocas_w_cg_cmp_rd_rs1_val_eq_equal:
  amocas.w x21, x20, (x2) # perform operation
```

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_pair_partial_val.sv
```diff
@@ -0,0 +1,9 @@
+    cmp_rd_rs1_pair_partial_val : coverpoint (
+            (ins.current.rd_val == ins.prev.rd_val) ^
```
davidharrishmc, line 2: I don't think this coverpoint is doing what you want.  It is not comparing the value in memory at rs1 to the value in rd.

generators/testgen/src/testgen/priv/extensions/ExceptionsZaamo.py
```diff
@@ -94,6 +94,59 @@ def _generate_amo_address_misaligned_tests(test_data: TestData) -> list[str]:
             )
         lines.append("#endif")
 
+        # Zacas
```
davidharrishmc, line 97: Please share the ExceptionsZaamo.S file (and inspect that it is doing what you intend).

davidharrishmc (changes requested):
Apologies it has taken so long to review.  Busy time right now.  The .S files are very helpful.  Please review them carefully and that will reveal unintended behaviors.

davidharrishmc:
@amanmurad-10xe it would be great to wrap this up and get it merged.

generators/testgen/src/testgen/priv/extensions/ExceptionsZaamo.py
amanmurad-10xe: [ExceptionsZaamo-00.zip](https://github.com/user-attachments/files/28065995/ExceptionsZaamo-00.zip)
I have inspected it, and it is working fine based on my observations. Please let me know if anything still needs to be changed.

generators/coverage/src/covergroupgen/templates/cmp_rd_rs1_pair_partial_val.sv
amanmurad-10xe: As mentioned earlier, the RVVI traces used by our SystemVerilog coverpoints do not provide direct visibility into memory contents at (rs1). To handle this, we rely on the architectural behavior of Zacas instructions, where the original memory value is always written back into rd after execution. Therefore, the coverpoint compares ins.prev.rd_val (the initial compare value) with ins.current.rd_val (the memory value returned by the instruction) to infer whether the compare succeeded, failed, or partially matched, entirely through register state observation.

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
amanmurad-10xe: Thanks for pointing this out. The extra stores were happening because generate_cmp_testcase was duplicating the memory setup already performed by ap_type formatter. I've removed the redundant SREG lines and register loads.
Fixed!!

generators/testgen/src/testgen/formatters/types/ap_type.py
amanmurad-10xe: I have resolved the initial issues.

For the part about adding new coverpoint variants, I am a bit unclear on that. I made some changes in the same generator file for cp_rs1, cp_rs2, and rd. Please let me know if these need to be covered separately, and also whether we are planning to randomize these and add two test variants: matching and non-matching.

If that is the case, do we also need to do the same for the four additional coverpoints: cmp_rd_rs1, cmp_rd_rs2, cmp_rd_rs1_rs2, and cp_rs2_edges?

This part is a bit unclear to me.

I have attached the updated Zacas-amocas-q-o00.S file for the reference.
[Zacas-amocas.q-00.zip](https://github.com/user-attachments/files/28066636/Zacas-amocas.q-00.zip)

coverpoints/priv/ExceptionsZaamo_coverage.svh
amanmurad-10xe: Fixed.

amanmurad-10xe:
Apologies for the delay. I have pushed the changes.

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
```diff
@@ -0,0 +1,303 @@
...
+            # upper equal, lower mismatch
+            mem_lo = random_range(0, all_ones)
+            while mem_lo == lo:
+                mem_lo = random_range(0, all_ones)
+
+            rd_val = (hi << test_data.xlen) | lo
+            rs1_val = (hi << test_data.xlen) | mem_lo
+
+            desc = f"{coverpoint} (upper match)"
+            bin_name = "partial_hi"
```
davidharrishmc, line 264: I see how the lower match part works, but not the upper match.  In lower match, x16 = 0xade8... = value at scratch.  But in upper match, x11 = 0xbfaddb..., which is different than the value in memory 0xfbaf...

/////////////////////////////////
// cmp_rd_rs1_pair_partial_val
/////////////////////////////////

# Testcase cmp_rd_rs1_pair_partial_val (lower match)
  RVTEST_TESTDATA_LOAD_INT(x1, x10) # load value in memory: x10 = 0xade841f7d3721545
  RVTEST_TESTDATA_LOAD_INT(x1, x20) # load rs2: x20 = 0x8d8aeeb54bfc475e
  RVTEST_TESTDATA_LOAD_INT(x1, x16) # load rd compare value: x16 = 0xade841f7d3721545
  LA(x23, scratch)
  SREG x10, 0(x23)
  RVTEST_TESTDATA_LOAD_INT(x1, x21) # load rs2 upper: x21 = 0xf104807eba0278cf
  RVTEST_TESTDATA_LOAD_INT(x1, x17) # load rd compare value upper: x17 = 0xfded3e9fcf7f8fb2
  RVTEST_TESTDATA_LOAD_INT(x1, x10) # load value in memory upper: x10 = 0xf7ff7e9e7dfbe7ba
  SREG x10, 8(x23)
Zacas_amocas_q_cg_cmp_rd_rs1_pair_partial_val_partial_lo:
  amocas.q x16, x20, (x23)
 ...

# Testcase cmp_rd_rs1_pair_partial_val (upper match)
  RVTEST_TESTDATA_LOAD_INT(x1, x20) # load value in memory: x20 = 0x7ee4b139cfacda6d
  RVTEST_TESTDATA_LOAD_INT(x1, x30) # load rs2: x30 = 0xa37e05bbe3b236ab
  RVTEST_TESTDATA_LOAD_INT(x1, x10) # load rd compare value: x10 = 0xa843e57f6b13c03b
  LA(x29, scratch)
  SREG x20, 0(x29)
  RVTEST_TESTDATA_LOAD_INT(x1, x31) # load rs2 upper: x31 = 0x59f49501d682d661
  RVTEST_TESTDATA_LOAD_INT(x1, x11) # load rd compare value upper: x11 = 0xbfaddb7ebffdd6ff
  RVTEST_TESTDATA_LOAD_INT(x1, x20) # load value in memory upper: x20 = 0xfbaf7bfffefd7f73
  SREG x20, 8(x29)
Zacas_amocas_q_cg_cmp_rd_rs1_pair_partial_val_partial_hi:
  amocas.q x10, x30, (x29)

generators/testgen/src/testgen/formatters/types/ap_type.py
davidharrishmc: This looks good enough that the cp_rs1/rs2/rd are tested both matching and non-matching.

generators/testgen/src/testgen/priv/extensions/ExceptionsZaamo.py
davidharrishmc: 05b rather than 03b will print nicer

davidharrishmc:
This looks very good.  A few minor comments.

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
```diff
@@ -0,0 +1,303 @@
...
+@add_coverpoint_generator("cmp_rd_rs1_val_lsb")
+def make_cmp_rd_rs1_val_lsb(instr_name: str, instr_type: str, coverpoint: str, test_data: TestData) -> list[TestChunk]:
+    """Generate CAS tests where the least significant byte of rd value is equal to or not equal to the least significant byte of the memory value (rs1)."""
+
+    test_chunks = []
+    is_pair = instr_type == "AP"
+    all_ones = (1 << (2 * test_data.xlen)) - 1 if is_pair else (1 << test_data.xlen) - 1
+    mask = 0xFF
+
+    for equal_case in [True, False]:
```
davidharrishmc, line 147: Equal_case seems it should just apply to CAS.  I don't know whether this logic is also applied to other AMOs, but amocas seems sufficient.

generators/testgen/src/testgen/coverpoints/special/cmp_values.py
amanmurad-10xe: Great catch. That mismatch was caused by a masking bug in the upper match logic.

The individual register pairs were incorrectly being masked according to $2 \times \text{XLEN}$ instead of just $\text{XLEN}$. I've fixed the issue by updating the logic to mask according to $\text{XLEN}$, which resolves the value discrepancy you seeing in memory.

Updated testcase:

# Testcase cmp_rd_rs1_pair_partial_val (lower match)
  RVTEST_TESTDATA_LOAD_INT(x1, x26) # load value in memory: x26 = 0xade841f7d3721545
  RVTEST_TESTDATA_LOAD_INT(x1, x16) # load rs2: x16 = 0x7324628c9dcc3c9f
  RVTEST_TESTDATA_LOAD_INT(x1, x22) # load rd compare value: x22 = 0xade841f7d3721545
  LA(x2, scratch)
  SREG x26, 0(x2)
  RVTEST_TESTDATA_LOAD_INT(x1, x17) # load rs2 upper: x17 = 0x26077f646bab0527
  RVTEST_TESTDATA_LOAD_INT(x1, x23) # load rd compare value upper: x23 = 0x7a8c582ff5ad3e9e
  RVTEST_TESTDATA_LOAD_INT(x1, x26) # load value in memory upper: x26 = 0x3dc9e6aa3fa369e4
  SREG x26, 8(x2)

# Testcase cmp_rd_rs1_pair_partial_val (upper match)
  RVTEST_TESTDATA_LOAD_INT(x1, x26) # load value in memory: x26 = 0xd682d661a37e05bb
  RVTEST_TESTDATA_LOAD_INT(x1, x24) # load rs2: x24 = 0x62e506172789db74
  RVTEST_TESTDATA_LOAD_INT(x1, x8) # load rd compare value: x8 = 0x63f1d969e3558991
  LA(x22, scratch)
  SREG x26, 0(x22)
  RVTEST_TESTDATA_LOAD_INT(x1, x25) # load rs2 upper: x25 = 0x99ac537eb6bd5473
  RVTEST_TESTDATA_LOAD_INT(x1, x9) # load rd compare value upper: x9 = 0xa843e57f6b13c03b
  RVTEST_TESTDATA_LOAD_INT(x1, x26) # load value in memory upper: x26 = 0xa843e57f6b13c03b
  SREG x26, 8(x22)

generators/testgen/src/testgen/priv/extensions/ExceptionsZaamo.py
amanmurad-10xe: Done!

davidharrishmc:
Looks like amocas.b is failing on Whisper.
https://github.com/riscv/riscv-arch-test/actions/runs/26282255285/job/77363195377?pr=1186

You can click on the gear icon at that page to Download logs, but the Github download is hanging for me after 40 MB, so I can't see the details of the failure.

Have you been able to run these tests on Whisper locally?

amanmurad-10xe:
One issue that I have currently figured out is that after adding the equal/not-equal matching cases in `cp_rd`, `cp_rs1`, and `cp_rs2`, it has disturbed the flow of other extension tests. For example, `E_add` is failing because of this change, and similarly, other extensions that involve these coverpoints may also face issues.

I am currently working on it and will update you soon regarding this. I am thinking of either adding a label so that these equal/not-equal cases are generated only when that label is present; otherwise, it should work as it did previously, so the flow of other extension tests does not get affected.

Let me also check the Whisper issue alongside this, and I will update you on that soon as well.

davidharrishmc:
I think we only want the equal/not equal to apply to Zacas instructions.

amanmurad-10xe:
I have fixed the `check generated tests` failure check. It is now passing successfully.
I am currently working on the Whisper test failure issue and will update you once it is resolved. It is showing some strange behavior.

davidharrishmc:
These tests have uncovered a bunch of bugs in whisper.  If you find the apparent root cause is a Whisper bug, please open an issue on the whisper repo. You can edit config/whisper/ci.yaml to exclude the failing Whisper tests, and paste in a comment with a link to the issue. That way we will be able to pass CI and merge before Whisper fixes the root cause.

See
https://docs.google.com/spreadsheets/d/1EuG4LV1TQDx4wTzYDWEAvq0fwaWjNuxenwDARZmrN-E/edit?gid=1586739087#gid=1586739087

for a bunch of other open issues.

David

davidharrishmc:
@amanmurad-10xe we are trying to release as much of RVA23 as possible before the RISC-V summit June 8.  Is everything else good except Whisper?  If so, can you add Zacas to the exclusions list in whisper ci.yaml and we will merge the part that is working, and then you can do a followup to enable Whisper when the root cause is identified and fixed in Whisper or ACT?

amanmurad-10xe:
I have just added a helper function to handle Zacas randomization masking for the .b, .h, .w, .d, and .q instructions. Surprisingly, after pushing the changes, Whisper is now passing on ci.

davidharrishmc (approved):

davidharrishmc:
@amanmurad-10xe We wanted to get this PR merged before the summit, so we merged as is.  It also unearthed real bugs in QEMU (#1624), which is terrific!

I am creating a new issue #1626 about the unresolved Whisper bugs.  Would you be able to roll back to an earlier commit that reveals those bugs, see if they still exist, and create an issue against Whisper if they are real?  If they are catching things the randomized tests don't, we should understand why and supplement the Zacas test suite with everything relevant.

amanmurad-10xe:
Thanks. I'll investigate this and let you know what I find.

PR #1189, opened 2026-04-02
Updated trap handler, not to avoid one instruction after trap

UmerShahidengr:
Removed `2*WDBYTSZ` offset in `mepc` to avoid skipping one instruction after trap

### DIFF
--- a/config/cores/cvw/ci.yaml
+++ b/config/cores/cvw/ci.yaml
@@ -4,7 +4,7 @@
 # Requires Verilator and the CVW repository
 
 ci_enabled: true
-exclude_extensions: "Sm,S,InterruptsSm,InterruptsS,InterruptsU,InterruptsSstc,PMPSm,ExceptionsZalrsc,ExceptionsZaamo,ExceptionsSvZalrsc,ExceptionsSvZaamo,Svnapot,Sv,SvPMP,SvPMPZicbo,SvZicbo,Svade,SvaduPMP,Svpbmt,Svinval,PMPS,PMPU,PMPF"
+exclude_extensions: "Sm,S,InterruptsSm,InterruptsS,InterruptsU,InterruptsSstc,PMPSm,ExceptionsZalrsc,ExceptionsZaamo,ExceptionsSvZalrsc,ExceptionsSvZaamo,Svnapot,Sv,SvPMP,SvPMPZicbo,SvZicbo,Svade,SvaduPMP,Svpbmt,Svinval,PMPS,PMPU"
 apt_packages: "autoconf flex bison help2man libfl-dev ccache"
 install_script: ".github/scripts/install-cvw.sh"
 setup_script: ".github/scripts/setup-cvw.sh"
--- a/config/sail/ci.yaml
+++ b/config/sail/ci.yaml
@@ -3,4 +3,4 @@
 # CI configuration for Sail reference model
 
 ci_enabled: true
-exclude_extensions: "PMPS,PMPU,PMPF,Sv,SvaduPMP,SvPMP,SvPMPZicbo,Svade,Svinval,SvZicbo,Svnapot,Svpbmt"
+exclude_extensions: "PMPS,PMPU,Sv,SvaduPMP,SvPMP,SvPMPZicbo,Svade,Svinval,SvZicbo,Svnapot,Svpbmt"
--- a/config/spike/ci.yaml
+++ b/config/spike/ci.yaml
@@ -7,6 +7,6 @@ ci_enabled: true
 # TODO: Priv tests mismatch due to config issues. Atomics failure still needs to be diagnosed.
 # TODO: remove za64rs from excluded_extensions when Sail supports UDB_LRSC_FAIL_ON_NON_EXACT_LRSC to match DUT
 # TODO: Remove Zama16b when spike starts supporting it.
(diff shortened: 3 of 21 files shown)


tests/env/arch_test.h
```diff
@@ -1557,7 +1557,7 @@ sv_\__MODE__\()epc:
 
 adj_\__MODE__\()epc_rtn:                // adj mepc so there is at least 4B of padding after op
         andi    T3, T3, ~WDBYTMSK       // adjust mepc to prev 4B alignment (if 2B aligned)
-        addi    T3, T3,  2*WDBYTSZ      // adjust mepc so it skips past op, has padding & 4B aligned
+        addi    T3, T3,  WDBYTSZ        // adjust mepc so it jumps to next instruction, has no padding & 4B aligned
```
davidharrishmc: Is this supposed to be 4B aligned?  If the instruction is compressed or on 2-byte alignment, are any NOPs needed.

allenjbaum:
For reasons lost in the midst of time, i wanted to ensure that returning
from a trap was always to an aligned address - possibly to help debugging?
IF we're not skipping an entire instruction, then skipping half an
instruction defeats the purpose , because I would need either an explicit
noop or an explicit .align,
unless the SIGUPD macro inserted it automatically.

davidharrishmc:
If I understand this correctly, this updated code is now adding either 2 or 4 to EPC to get to the next multiple of 4.  Like Allen says, this doesn't seem to be the correct behavior.  The original code also doesn't seem to be consistent because it adds 6 or 8 to get to the next multiple of 4, which also might not be aligned to the next instruction.

I believe that what we really want to do is add 2 or 4 based on the size of the next instruction.  That is complicated because we have to fetch the next instruction, which might be a virtual or physical address.  In M-mode, we need to set MPRV to fetch with respect to the addressing scheme of the code that trapped.  Below S-mode, we can just fetch, but need SUM and MXR to ensure we can read even when virtual memory is on with XU permissions.

This is the code below from ACT tests/env/WALLY-init-lib.h that I've used to add 2 or 4 assuming the trap handler is in M-mode. It doesn't account for SUM and MXR  So I think we need to bifurcate adj_\__MODE__\()etc_rtn such that if MODE is M, we do something along the following lines with MPRV, and if MODE is lower than M, we leave out the MPRV part.  If MODE is M, we would also need to save and turn on mstatus.SUM/MXR.  If mode is less than M, we would need to save and turn on sstatus.SUM/MXR.  In either case, we need to restore them to their original values before returning.

I have the instruction access fault handling in here and presumably you are doing it somewhere else.

I don't have a deep understanding of the ACT trap handler, so apologies if I'm off base.

```
exception_return:             # add 2 or 4 to mepc for exceptions except Instruction Access Fault
    # First, check if the exception was an Instruction Access Fault
    csrr  t1, mcause              # t1 = exception cause
    addi  t1, t1, -1              # Exception cause code 1 means Instruction Access Fault
    bne   t1, x0, mepc_ret_addr   # If not an IAF, skip ra load
    mv t0, ra
    # If its an IAF then it was called with a JALR (rd = PC +4). So subtract 4 from ra to get faulting address
    addi t0, t0, -4
    j post_ret_mepc               #Use ra instead of mepc if IAF
mepc_ret_addr:
    csrr t0, mepc       # get address of instruction that caused exception if not an IAF
post_ret_mepc:
    li t1, 0x20000
    csrs mstatus, t1    # set mprv bit to fetch instruction with permission of code that trapped
    lh t0, 0(t0)        # get instruction that caused exception
    csrc mstatus, t1    # clear mprv bit to restore normal operation
    li t1, 3
    and t0, t0, t1      # mask off upper bits
    beq t0, t1, instr32 # if lower 2 bits are 11, instruction is uncompresssed
    li t0, 2            # increment PC by 2 for compressed instruction
    j updateepc
instr32:
    li t0, 4
updateepc:
    # First, check if the exception was an Instruction Access Fault
    csrr   t1, mcause            # t1 = exception cause
    addi   t1, t1, -1            # Exception cause code 1 means Instruction Access Fault
    bne    t1, x0, mepc_up_addr  # If not an IAF, skip ra load
    mv t0, ra                    # Jal already increments by 4 so its safe to overwrite t0 for an IAF
    j post_up_mepc               # Use ra instead of mepc (skips next instruction)
mepc_up_addr:
    csrr t1, mepc
post_up_mepc:
    add t1, t1, t0               # add 2 or 4 (from t0) to MEPC to determine return Address
    csrw mepc, t1
trap_return:                     # don't need to update mepc for interrupts
    LREG t0, 0(tp)         # Restore t0, t1, and ra
    LREG t1, -8(tp)
    LREG ra, -16(tp)
    csrrw tp, mscratch, tp   # restore tp
    mret                     # return from trap
```

tests/env/arch_test.h
davidharrishmc, line 1486: I am having trouble following this part.  If there is a fetch page fault or access fault, it seems we would never want to return to that address because we would just get another one.  So why is all of this gated behind T6=ACCE and a SKIP_MEPC?  Also, all of the logic is based on XEPC, but the condition is called SKIP_MEPC.  Should it be SKIP_XEPC if it is necessary?

allenjbaum:
The idea behind the ACT3 handlers is to start re-executing at the next
doubleword boundary that guaranteed at least some padding.
The reason for that is that we had tests that were trapping because they
didn't execute halfword ops, and it kept returning to an illegal boundary -
so we forced the xret address to have a guaranteed legal offset. It forced
EPC to be word aligned (by masking out bit1), and added 8 (to guarantee
skipping past the illegal op with padding. That might not work when 6 or 8
bytes are implemented, and there should be a comment to that effect. That
worked when we had tests that guaranteed those branches would occur at the
end of a page, so we could guarantee that the fall-through code was legal.,
but also meant that tests might be very large depending on offset sizes
being tested.

UmerShahidengr:

PR #1193, opened 2026-04-03
Exclude x0 for all priv tests

jordancarlin:


generators/testgen/src/testgen/generate/priv.py
```diff
@@ -57,8 +57,10 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     # so they can be split for long priv tests (e.g. Ssstrict)
     tc = test_data.begin_test_chunk()
 
+    # Reserve x0 for priv tests so that desired values are actually loaded into registers
     # Priv tests use x1/ra as the return address for function calls, so reserve it before generating the test
-    test_data.int_regs.consume_registers([1])
+    priv_exclude_regs = [0, 1]
```
davidharrishmc, line 62: I thought there were more excluded registers used by the trap handler.

davidharrishmc: For example, see InterruptsSm.  There may be other registers used by the handler.

jordancarlin: I believe there is an open PR from someone else that adds those. I wanted to avoid conflicts, so I’m just adding zero here and will let everyone else add the other registers they need.

davidharrishmc (approved):

PR #1194, opened 2026-04-03
ExceptionsS 100% coverage

juliacygong:


generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+# SPDX-License-Identifier: Apache-2.0
+##################################
+
+"""Exceptions S-mode test generator."""
+
+from testgen.asm.helpers import comment_banner, write_sigupd
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
```
davidharrishmc, line 34: Let's make this test as identical to ExceptionsSm as possible, including the ordering of functions (in ExceptionsSm, _generate_instr_adr_misaligned_branch_tests comes first).

davidharrishmc: Most of this code is identical across ExceptionsSm, ExceptionsS, ExceptionsU.  How about we factor it out into an ExceptionsCommon.py and import it? Each function would need to take covergroup as an argument.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+    test_data.int_regs.return_registers([addr_reg, data_reg, check_reg])
+    return t_lines
+
+
+def _generate_instr_adr_misaligned_branch_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsS_cg", "cp_instr_adr_misaligned_branch"
+    temp_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Address Misaligned branch (taken)"),
```
davidharrishmc: The word "taken" is in this file but not in ExceptionsSm.  Let's keep the two identical by adding taken to ExceptionsSm.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+        f"LI(x{temp_reg}, 1)",
+        ".align 2",
+        test_data.add_testcase("taken_branch_pc_6", coverpoint, covergroup),
+    ]
+
+    branches = ["beq", "bne", "blt", "bge", "bltu", "bgeu"]
+    for branch in branches:
+        if branch in ("bge", "bgeu"):
+            rs1, rs2 = f"x{temp_reg}", "x0"
```
davidharrishmc: This code is changed a bit from ExceptionsSm.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+    covergroup, coverpoint = "ExceptionsS_cg", "cp_instr_adr_misaligned_jalr"
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Address Misaligned JALR"),
+    ]
+
+    # Every combination of rs1[1:0] and offset[1:0]
+    rs1_base_offsets = {0: 8, 1: 5, 2: 6, 3: 7}
+    jalr_offsets = {0: 8, 1: 5, 2: 6, 3: 7}
```
davidharrishmc: These two lists seem to be identical.  Why do we need both?

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+    for rs1_lsb in range(4):
+        for offset_lsb in range(4):
+            base_off = rs1_base_offsets[rs1_lsb]
+            jalr_off = jalr_offsets[offset_lsb]
+
+            lines.extend(
+                [
+                    f"\n# rs1[1:0]={rs1_lsb:02b}, offset[1:0]={offset_lsb:02b}",
+                    ".align 2",
+                    f"auipc x{addr_reg}, 0",
```
davidharrishmc: This logic is hard to follow.  I think Jordan reviewed ExceptionsSm and I apologize for not looking at it closely back then and commenting.  If you want to keep it this way, there should be detailed comments explaining the principles of operation.  I had anticipated a jump forward to a shared bit of code that then returns, which might have been simpler, but this can work so long as it is explained well.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+
+
+def _generate_instr_access_fault_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsS_cg", "cp_instr_access_fault"
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Access Fault"),
+        f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+        "LI(x4, 0xACCE)",
```
davidharrishmc: Is x4 in the list of always-excluded registers for priv?  I don't love hardwiring this.  I had made a suggestion in the open trap handler PR to return using ra instead of mepc whenever there is an instruction access fault, rather than checking x4.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+    test_values = {"sb": "0xAB", "sh": "0xBEAD", "sw": "0xADDEDCAB", "sd": "0xADDEDCABADDEDCAB"}
+
+    for op in store_ops:
+        lines.extend(
+            [
+                f"LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+                f"LI(x{data_reg}, {test_values[op]})",
+                test_data.add_testcase(f"{op}_fault", coverpoint, covergroup),
+                f"{op} x{data_reg}, 0(x{addr_reg})",
+                " nop",
```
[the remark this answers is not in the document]
davidharrishmc: I think LI is correct because it is a #define, not a label.
LA should be changed to LI where this is used elsewhere in ExceptionsS (and ExceptionsSm, which should be automatic when these codes are shared).

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+    return lines
+
+
+def _generate_illegal_instruction_csr_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsS_cg", "cp_illegal_instruction_csr"
+    dest_regs = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Illegal Instruction"),
+        f"LI(x{dest_regs[0]}, 1)",
```
davidharrishmc: Let's load a 0xBOBACAFE value into dest_regs[0] before each csr access and do a sigupd after each access to ensure it doesn't change.  Let's also do this for load access fault and load_address_misaligned to check that trapping instructions don't change the destination register.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+        )
+
+        if priv_mode == 1:
+            lines.append("RVTEST_GOTO_LOWER_MODE Smode")
+        elif priv_mode == 0:
+            lines.append("RVTEST_GOTO_LOWER_MODE Umode")
+        else:
+            lines.append("RVTEST_GOTO_MMODE")
+
+        # Instruction misaligned
```
davidharrishmc, line 129: Instead of repeating all of these tests here inline, please use a call to a shared test generator.  Best not to have repeated code.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+        lines.extend(
+            [
+                test_data.add_testcase(f"instrmisaligned_{tag}", coverpoint, covergroup),
```
davidharrishmc, line 132: This code looks different than the cp_instr_adr_misaligned_jalr.  Please use a library call to the shared code instead of inventing something new.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+                lines.extend(["RVTEST_GOTO_SMODE", "RVTEST_GOTO_MMODE"])
+            else:
+                lines.append("RVTEST_GOTO_MMODE")
+
+    # Clear medeleg
+    lines.extend(
+        [
+            "RVTEST_GOTO_MMODE",
+            "LI(x5, 0)",
+            "csrw medeleg, x5",
```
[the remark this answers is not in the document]
davidharrishmc: We shouldn't be using hardwired registers such as x5 unless they are already part of the trap handler, in which case they should be reserved by the privileged register allocator and there should be a comment about it in the code here.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,926 @@
...
+        "# Delegate instr access fault (medeleg bit 1) to S-mode",
+        f"LI(x{addr_reg}, (1 << 1))",
+        f"csrw medeleg, x{addr_reg}",
+    ]
+
+    if priv_mode == 1:
+        lines.append("RVTEST_GOTO_LOWER_MODE Smode")
+    elif priv_mode == 0:
+        lines.append("RVTEST_GOTO_LOWER_MODE Umode")
+    else:
```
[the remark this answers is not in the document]
davidharrishmc: Hopefully this is part of the priv reg allocator to reserve x6/7/9?

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -0,0 +1,557 @@
...
+            ]
+        )
+
+    test_data.int_regs.return_registers([temp_reg])
+    return lines
+
+
+def generate_instr_adr_misaligned_branch_nottaken(test_data: TestData, covergroup: str) -> list[str]:
+    coverpoint = "cp_instr_adr_misaligned_branch_nottaken"
+    temp_reg, check_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
```
davidharrishmc: exclude_regs 0 no longer needed for priv tests

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -0,0 +1,557 @@
...
+    for branch in branches:
+        if branch in ("beq", "bge", "bgeu"):
+            rs1, rs2 = "x0", f"x{temp_reg}"
+        elif branch == "bne":
+            rs1, rs2 = "x0", "x0"  # 0 == 0, so bne not taken
+        else:  # blt, bltu
+            rs1, rs2 = f"x{temp_reg}", "x0"
+        lines.append(f"{branch} {rs1}, {rs2}, .+6")
+
+    lines.append(write_sigupd(check_reg, test_data))
```
davidharrishmc, line 77: I'm not seeing how check_reg might change and what this sigupd checks

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -0,0 +1,557 @@
...
+def generate_instr_adr_misaligned_jalr_tests(test_data: TestData, covergroup: str) -> list[str]:
+    coverpoint = "cp_instr_adr_misaligned_jalr"
+    addr_reg = test_data.int_regs.get_register(exclude_regs=[0])
+
+    lines = [
+        comment_banner(coverpoint, "Instruction Address Misaligned JALR"),
+    ]
+
+    # rs1 is set to PC + base_offset so that rs1[1:0] == rs1_lsb.
+    # jalr_off controls the offset[1:0], covering all 16 combinations of (rs1+offset)[1:0].
```
davidharrishmc, line 104: This is hard to follow.  Could you clarify exactly where we are trying to land relative to the jalr and what is there?

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -0,0 +1,557 @@
...
+                    ".align 2",
+                    f"auipc x{addr_reg}, 0",
+                    f"addi x{addr_reg}, x{addr_reg}, {base_off}",
+                    test_data.add_testcase(f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup),
+                    f"jalr x1, {jalr_off}(x{addr_reg})",
+                    "nop",
+                ]
+            )
+
+            if rs1_lsb & 1:
```
davidharrishmc: Please check that you want the nop at 120 and possible nop here.  Ellen found that she didn't want them and was getting illegal instruction exceptions even with compressed because she wasn't landing in the middle of the 0x0001 from the addi.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -0,0 +1,470 @@
...
+    generate_misaligned_priority_store_tests,
+    generate_store_access_fault_tests,
+    generate_store_address_misaligned_tests,
+)
+from testgen.priv.registry import add_priv_test_generator
+
+_CG = "ExceptionsS_cg"
+
+
+def _generate_ecall_s_tests(test_data: TestData) -> list[str]:
```
davidharrishmc: Can you eliminate this function and just call line 37 from the main body?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -1,515 +1,45 @@
...
-            "LI(x4, 0xACCE)",  # trap handler checks x4 value to use x1 (ra) as return address instead of mepc
-            test_data.add_testcase("misaligned_nonexistent", coverpoint, covergroup),
-            f"jalr x1, 0(x{addr_reg})",
-            "nop",
-        ]
-    )
-
-    test_data.int_regs.return_registers([addr_reg])
-    return lines
+    return generate_ecall_tests(test_data, _CG, "cp_ecall_m", "ecall_m", "Ecall Machine Mode")
```
davidharrishmc: Why does this belong in a one-line function instead of in make_exceptionssm?

davidharrishmc (changes requested):

davidharrishmc (approved):

PR #1197, opened 2026-04-05
MISA Dependency Tests

davidharrishmc:
Check that misa.MXL is immutable
Check that certain dependent extensions turn off when the controlling extension is turned off
Check that misa.C cannot be set to 0 when PC is an odd multiple of 2.


Makefile
```diff
@@ -106,24 +106,31 @@ qemu-rv64: elfs
 
 
 ##### Whisper targets #####
-.PHONY: whisper whisper-rv64
+.PHONY: whisper whisper-rv64 whisper-rv32
 
 WHISPER_64_CMD := whisper --config config/whisper/whisper-rv64-max/whisper.json
+WHISPER_32_CMD := whisper --config config/whisper/whisper-rv32-max/whisper.json
```
jordancarlin, line 112: The actual whisper rv32 config files seem to be missing.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -15,6 +15,28 @@
 from testgen.priv.registry import add_priv_test_generator
 
 
+def gen_misa_dependencies(
+    misa: str, mask: str, cpbin: str, comment: str, coverpoint: str, covergroup: str, test_data: TestData
+) -> str:
+    """Generate tests for misa dependencies."""
+    r1, rmask, rfail = test_data.int_regs.get_registers(3, exclude_regs=[0])
```
jordancarlin: Drop the `exclude_regs=[0]`.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -428,6 +450,201 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+            coverpoint,
+            "Try to clear misa.C.  Should not change if PC is at 2-byte aligned address",
+        ),
+    )
+
+    r1, r2, rc = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines.extend(
+        [
+            "# Load 1s into msb and msb-1 corresponding to misa.MXL bitfields",
```
jordancarlin: Comment seems wrong.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -428,6 +450,201 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+            "",
+            f"csrs misa, x{rc}     # set misa.C if possible",
+            f"csrr x{r1}, misa          # read misa to check if misa.C was set",
+            f"and x{r1}, x{r1}, x{rc} # mask off all but C bit",
+            ".align 2 # 4-byte alignment",
+            test_data.add_testcase("pc_1_0", coverpoint, covergroup),
+            f"csrc misa, x{rc}      # attempt to clear misa.C with misa.C = 1 and PC 4-byte aligned",
+            f"csrr x{r2}, misa          # read misa to check misa.C changed if writable",
+            f"and x{r2}, x{r2}, x{rc} # mask off all but C bit",
+            f"xor x{r2}, x{r2}, x{r1} # check if misa.C differed before and after clear attempt; might be 4 if misa.C is mutable because it is allowed to differ when PC is 4-byte aligned",
```
jordancarlin, line 624: Not clear why the xor is useful here. Why not just store the result after masking?

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: I don't trust having adequate control over the misa writability in the reference model, so I had coded these to work regardless of writability.  Later I put back in the walking 1s test in misa, which will depend on being able to predict results.  But I'm inclined to leave this because it is more general and can serve if walking 1s is untenable in the near term.

jordancarlin: Walking ones for misa was not added back to the coverpoints.

config/whisper/whisper-rv32-max/link.ld
jordancarlin, line 1: Symlink to the rv64 version to avoid duplication.

config/whisper/whisper-rv32-max/rvmodel_macros.h
jordancarlin, line 1: Symlink to the rv64 version to avoid duplication.

config/whisper/whisper-rv32-max/whisper.json
```diff
@@ -0,0 +1,80 @@
+{
+  "clint": "0x2000000",
+
+  "address_translation_pmms": ["off", "pm57"],
```
jordancarlin: This doesn't seem right. Pointer masking is not supported on rv32.

config/whisper/whisper-rv32-max/whisper.json
```diff
@@ -0,0 +1,80 @@
...
+
+  "isa": "rv32imafdcvhus_zicbom_zicbop_zicboz_zicntr_zicond_zicsr_zihintpause_zihpm_zimop_zawrs_zfa_zfbfmin_zfh_zkr_zcb_zcmop_zba_zbb_zbs_zvbb_zvbc_zvfbfmin_zvfbfwma_zvfh_zvkg_zvkned_zvknhb_zvqdot_sdtrig_smaia_smmpm_smnpm_smrnmi_smstateen_ssaia_sscofpmf_ssnpm_ssqosid_sstc_svadu_svinval_svnapot_svpbmt_svvptc_zifencei_zabha_zbc_zkne_zknd_zknh_zbkb_zbkc_zbkx_zksed_zksh",
```
jordancarlin: Several rv64 only extensions in this list. At least pointer masking and a few of the virtual memory extensions.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: Interesting...they were at one point.  Wonders of git.  Back now.

jordancarlin (approved):

PR #1199, opened 2026-04-06
General cleanup

jordancarlin:


generators/testgen/src/testgen/generate/priv.py
```diff
@@ -57,10 +57,11 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     # so they can be split for long priv tests (e.g. Ssstrict)
     tc = test_data.begin_test_chunk()
 
-    # Reserve x0 for priv tests so that desired values are actually loaded into registers
-    # Priv tests use x1/ra as the return address for function calls, so reserve it before generating the test
+    # Reserve registers for priv tests:
+    #   - x0: avoid so desired values are actually loaded into registers
+    #   - x1/ra: used as the return address for function calls
+    #   - x6, x7, x9: used by the RVTEST_GOTO_LOWER_MODE macro
```
davidharrishmc, line 63: Is this a good time to exclude 16-31 to avoid E problems?

jordancarlin: Added

davidharrishmc (approved):

PR #1203, opened 2026-04-06
Overhaul CI and Makefile

jordancarlin:
- Better documentation in Makefile explaining the different options and targets
- Automatic discovery of configs for Makefile targets and CI using new `run_cmd.txt` and `ci.yaml` files.
  - New configs no longer need to modify the Makefile or CI regress.yml GitHub Action script.


Makefile
```diff
@@ -193,50 +121,118 @@ $(STAMP_DIR)/vector-testgen-unpriv.stamp: generators/testgen/scripts/vector-test
...
+.PHONY: $(ALL_CONFIGS)
+
+define config-target
+$(1): EXTENSIONS =
+$(1): tests
+	$$(eval _DIR := $$(call config-dir,$(1)))
+	@EXTENSIONS="" \
+	CONFIG_FILES="$$(_DIR)/test_config.yaml" \
+	$$(MAKE) elfs
+	./run_tests.py "$$$$(cat $$(_DIR)/run_cmd.txt)" $$(WORKDIR)/$(1)/elfs
```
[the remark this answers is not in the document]
jordancarlin: Incorrect.

Makefile
```diff
@@ -1,38 +1,53 @@
...
 #  - InterruptsSm,PMPSm,PMPZca,PMPmisaligned: Additional testing needed on a wider range of configs. Some missing config options to match ref model.
-DEBUG       ?= # Set to True to generate debug output (signature objdump and trace files). Leave blank for no debug output.
-FAST        ?= # Set to True to disable objdump generation for faster builds. Leave blank for normal builds. Conflicts with DEBUG.
+EXTENSIONS  ?=
+EXCLUDE_EXTENSIONS ?= Sm,S,InterruptsSm,ExceptionsZalrsc,ExceptionsZaamo,PMPSm,PMPZca,PMPmisaligned,Sv,Svade,Svadu,SvaduPMP,SvPMP,SvZicbo
+
+# DEBUG and FAST are runtime options for controlling build output. They are mutually exclusive. Set to True to enable either option.
+# DEBUG enables debug output (signature objdump and trace files). This will slow down ELF generation significantly.
+# FAST disables objdump generation for faster builds. This speeds up ELF generation significantly, but makes debugging mismatches harder.
+DEBUG       ?=
```
davidharrishmc, line 31: When DEBUG=1 is there a way to make a DUT-specific trace?
For example, for Imperas, we need to add --trace --tracemode --tracechange --traceshowicount to the run command.

Makefile
```diff
@@ -193,50 +121,118 @@ $(STAMP_DIR)/vector-testgen-unpriv.stamp: generators/testgen/scripts/vector-test
...
-	$(MAKE) spike || exit_code=1; \
-	$(MAKE) qemu || exit_code=1; \
-	$(MAKE) imperas || exit_code=1; \
+	$(foreach sim,$(SIMULATORS),$(MAKE) $(sim) || exit_code=1; ) \
 	exit $$exit_code
 
-##### Dev targets #####
+
+
+########### Simulators ###########
```
davidharrishmc, line 162: I have mixed feelings here.  This code is only slightly shorter than the original Makefile simulations, and now there is a new level of indirection that makes the code harder to follow.  Is it worth it?

jordancarlin: This also cuts out 50+ lines of code per simulator from the CI script. The other benefit is that it makes it easy to add new configs/simulators in the future. I have a local change that adds cvw so that we can do `make cvw-rv64gc` and it only requires adding a one-line file to the config directory instead of adding all of the code to the Makefile.

Makefile
jordancarlin: We don't have that in the current version. It's something I can look at adding in the future, but it would be a bit tricky because the run_elfs script does not get these variables.

davidharrishmc (approved):

Makefile
jordancarlin: I added support for this. Imperas, Spike, Sail, and Whisper now all include traces in the log file when DEBUG is set.

davidharrishmc (approved):

PR #1226, opened 2026-04-08
Zihintntl, ZihintntlZca 100% all sims

davidharrishmc:
Completely works except that Clang wants C enabled (not just Zc) to compile Zihintntl compressed instructions.


coverpoints/norm/ZihintntlZca.yaml
```diff
@@ -1,11 +1,89 @@
...
+
+  - name: NTL_TRAP_BEHAVIOR
+    coverpoint: ["TODO"]
+
+  # Since the NTL instructions are encoded as ADDs, they can be used within LR/SC
+  # loops without voiding the forward-progress guarantee. But, since using other
+  # loads and stores within an LR/SC loop does void the forward-progress guarantee,
+  # the only reason to use an NTL within such a loop is to modify the LR or the SC.
+  - name: NTL-LR_SC_exception
+    coverpoint: ["TODO"]
```
jordancarlin: These seem like better fits for Zihintntl, not ZihintntlZca.

docs/ctp/src/zi.adoc
```diff
@@ -163,20 +163,19 @@ include::norm/Zihintpause_norm_rules.adoc[]
...
-//,===
+These instructions are hints that behave as nops.  They are simply executed once.
 
+include::{testplansdir}/Zihintntl.adoc[]
 
+include::norm/Zihintntl_norm_rules.adoc[]
 
 ==== ZihintntlZca Compressed Non-temporal Locality Hint
 
 Compressed Zihintntl instructions `c.ntl.p1`, `c.ntl.pall`, `c.ntl.s1`, and `c.ntl.all` are a subset of the `c.add` hints. They are tested by the Zca_c_hint_cg/cp_chint_add coverpoint, so require no additional coverpoints.
```
jordancarlin: This comment is no longer true. They have a dedicated testplan and coverpoints now.

framework/src/act/fcov/disassemble.svh
```diff
@@ -96,15 +96,18 @@ function string disassemble (logic [31:0] instrRaw);
...
-    // NTL_S1:  $sformat(decoded, "NTL.S1");
-    // C_NTL_ALL: $sformat(decoded, "C.NTL.ALL");
-    // C_NTL_PALL: $sformat(decoded, "C.NTL.PALL");
-    // C_NTL_P1: $sformat(decoded, "C.NTL.P1");
-    // C_NTL_S1: $sformat(decoded, "C.NTL.S1");
+    `ifdef ZIHINTNTL_COVERAGE
+    NTL_ALL: $sformat(decoded, "ntl.all");
+    NTL_PALL:$sformat(decoded, "ntl.pall");
+    NTL_P1:  $sformat(decoded, "ntl.p1");
+    NTL_S1:  $sformat(decoded, "ntl.s1");
```
jordancarlin: Indent inside the `` `ifdef `` to be consistent with the `ZICBOP_COVERAGE` section above.

jordancarlin (changes requested):
Is it worth testing these instructions immediately preceding various memory accesses? That is when they actually do something interesting. Even though there is nothing observable to us, it means we are at least exercising the instruction in its intended use case and we can check that the load/store/amo/etc. still operates normally.

jordancarlin (approved):

PR #1237, opened 2026-04-09
fix: derive FLEN from DUT extensions, not test march

### LINKED ISSUE #1223: Wrong FLEN when test march includes D but DUT only implements F
**Describe the bug**
Running `make CONFIG_FILES=config/cores/cve4/cv32e40p-v2-rv32mcf/test_config.yaml` hangs indefinitely on `work/cv32e40p-v2-rv32mcf/build/priv/ZicsrF/ZicsrF-00.sig` (Sail reference model run). Root cause: the build system compiles `ZicsrF-00.sig.elf` with `-DFLEN=64` (because the test's march includes `d`), causing `fsd` (double-precision store) to be emitted. But the DUT's `sail.json` has `D: supported: false`, so Sail traps on `fsd` as an
illegal instruction.

**To Reproduce**
make -j1 CONFIG_FILES=config/cores/cve4/cv32e40p-v2-rv32mcf/test_config.yaml  EXTENSIONS=ZicsrF EXCLUDE_EXTENSIONS= DEBUG=1 JOBS=1
hangs — kill after ~60s
tail work/cv32e40p-v2-rv32mcf/build/priv/ZicsrF/ZicsrF-00.sig.trace

Following changes have to be made to existing config for full run:
- add F_SUPPORTED to rvtest_config.h #1150 adds this
- mtimeaddress and mtimecmp_address added ( #934 workaround)
- move memory region from 0x0 in sail.json or ACCESS_FAULT_ADDRESS outside memory #1187 

```
[3234] [M]: 0x000043F8 fsub.s f7, f13, f13   # executes fine, sets mstatus.FS=Dirty
...
[3244] [M]: 0x00004420 illegal 0x72b027       # fsd — illegal with D=false
```
**Additional context**

### 1. ZicsrF test march includes D

`ZicsrF.py` declares:

```python
@add_priv_test_generator(
    "ZicsrF",
    required_extensions=["Zicsr", "F"],
    march_extensions=["Zicsr", "F", "D", "Zfh"],
)
```

The generator writes `MARCH: rv${XLEN}ifd_zicsr_zfh` into the test's config header. D is intentionally included so that `#ifdef D_SUPPORTED` optional coverpoints can be conditionally compiled in for DUTs that have the D extension.

### 2. FLEN computed from test march, not DUT extensions

`parse_test_constraints.py` — `TestMetadata.flen`:

```python
@property
def flen(self) -> str:
    base = self.march.split("_")[0].lower()
    if "q" in base: return "128"
    if "d" in base or "g" in base: return "64"   # ← "d" is in rv32ifd → FLEN=64
    if "f" in base: return "32"
    return "32"
```

`build_plan.py` — `gen_compile_tasks`:

```python
march = test_metadata.march.replace("${XLEN}", str(xlen))
flen = test_metadata.flen          # → "64"
...
sig_elf_cmd = [..., f"-DFLEN={flen}", ...]   # → -DFLEN=64
```

### 3. FLEN=64 causes fsd to be emitted for F-only code paths

`tests/env/utils.h`:

```c
#if FLEN==32
  #define FSREG fsw
#elif FLEN==64
  #define FSREG fsd     // ← selected because DFLEN=64
#endif
```

`tests/env/signature.h` — `RVTEST_SIGUPD_F` uses `FSREG` to save the FP result to the signature area. With FLEN=64, this emits `fsd` even though the result is a single-precision float (f7 from `fsub.s`).

Confirmed by objdump of `ZicsrF-00.sig.elf`:

```
000043f8 <ZicsrF_cg_cp_fflags_set_m_NV>:
    43f8:  08d6f3d3   fsub.s  f7,f13,f13    # single-precision — F only
    ...
    4420:  0072b027   fsd     f7,0(x5)      # double store — D required ← WRONG
```

The `rvtest_config.h` correctly omits `#define D_SUPPORTED`, so `#ifdef D_SUPPORTED` blocks in the test are compiled out. But FLEN=64 affects FSREG globally, including F-only signature update paths.

karabambus:
FLEN was computed solely from the test's march string, which can diverge from the DUT in both directions:

- Test march wider than DUT (e.g. ZicsrF march includes D for optional coverpoints, but DUT only has F): `fsd` emitted on an F-only DUT, Sail hangs.
- DUT wider than test march (e.g. D-capable DUT compiling an F-only test with `rv32imaf` march): `fsd` emitted in a march context that doesn't include D, assembler error.

Fix: compute FLEN as `min(dut_flen, test_march_flen)`.

- `compute_flen(implemented_extensions, xlen)` derives FLEN from the DUT's extensions (Q skipped on RV32 since toolchains don't support FLEN=128 with XLEN=32).
- `gen_compile_tasks` caps it to the test's march FLEN, preventing wider FP instructions than the test arch allows.

Fixes: #1223

### DIFF
--- a/framework/src/act/act.py
+++ b/framework/src/act/act.py
@@ -21,7 +21,7 @@ from act.build_plan import generate_build_plan
 from act.config import CoverageSimulator, load_config
 from act.coverreport import print_coverage_summary
 from act.parse_test_constraints import generate_test_dict
-from act.parse_udb_config import generate_udb_files, get_config_params, get_implemented_extensions
+from act.parse_udb_config import compute_flen, generate_udb_files, get_config_params, get_implemented_extensions
 from act.select_tests import select_tests
 
 # CLI interface setup
@@ -120,6 +120,7 @@ def run_act(
             generate_build_plan(
                 config,
                 mxlen,
+                compute_flen(implemented_extensions, mxlen),
                 selected_tests,
                 test_dir,
                 coverpoint_dir,
--- a/framework/src/act/build_plan.py
+++ b/framework/src/act/build_plan.py
@@ -57,6 +57,7 @@ def gen_compile_tasks(
     test_metadata: TestMetadata,
     base_dir: Path,
     xlen: int,
+    dut_flen: int,
     config: Config,
     compiler_cmd: list[str],
     debug: bool = False,
@@ -93,9 +94,12 @@ def gen_compile_tasks(
 
     # Metadata — substitute ${XLEN} placeholder used by priv tests
     march = test_metadata.march.replace("${XLEN}", str(xlen))
-    flen = test_metadata.flen
     test_path = test_metadata.test_path
     mabi = f"{'i' if xlen == 32 else ''}lp{xlen}{'e' if test_metadata.e_ext else ''}"
+
+    # Cap FLEN to what the test march actually supports — prevents emitting D/Q instructions
+    # in tests whose march string only covers F, even when the DUT supports wider FP.
+    flen = min(dut_flen, test_metadata.flen)
 
     # 1. sig.elf – compile with -DSIGNATURE
     sig_elf_cmd = [
@@ -387,6 +391,7 @@ def gen_coverage_tasks(
 def generate_build_plan(
     config: Config,
     xlen: int,
+    dut_flen: int,
     selected_tests: dict[str, TestMetadata],
     tests_dir: Path,
     coverpoint_dir: Path,
@@ -416,6 +421,7 @@ def generate_build_plan(
                 test_metadata,
                 config_wkdir,
                 xlen,
+                dut_flen,
                 config,
                 compiler_cmd,
                 debug,
--- a/framework/src/act/parse_test_constraints.py
+++ b/framework/src/act/parse_test_constraints.py
@@ -34,20 +34,20 @@ class TestMetadata(BaseModel):
         return value if isinstance(value, int) else None
 
     @property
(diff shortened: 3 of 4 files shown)


davidharrishmc:
I don't think I fully understand the root problem.  If the DUT supports F but not D and the DUT config files have F but not D, why is fsd being emitted in ZicsrF?

karabambus:
> I don't think I fully understand the root problem. If the DUT supports F but not D and the DUT config files have F but not D, why is fsd being emitted in ZicsrF?

1. Generator for ZicsrF test includes D in its march - so it can cover D extensions if its present. 
2. FLEN is later computed from this march during build phase and it gets length 64 (parse_test_constraints.py). 
3. RVTEST_SIGUPD_F uses FSREG to save the FP result to the signature area. With FLEN=64, this emits fsd. 

To be honest I am not entirely comfortable with this changes. It is not pretty though it works, I am open to suggestions on how can this be done better.

#1223 has it explained in more detail.

jordancarlin:
Thanks for reporting #1223. Unfortunately, I don't think this solution works. If we decrease FLEN, then it will change SIG_STRIDE compared to what the test generation expected, which means we will load the wrong values and no longer test what we want to be testing. I am working on a different solution that splits this into separate `TEST_FLEN` and `CONFIG_FLEN` variables so that we can use one for managing the data in the test and the other for controlling which instruction is used in the SIGUPD. Expect that PR soon. I am going to close this in favor of that approach since this would break coverage.

jordancarlin:
See #1243 for the new solution.

PR #1240, opened 2026-04-10
Preload scratch region

### LINKED ISSUE #1167: [ACT4] Initialize to 0 of the .bss section
**Description**

When  running ACT4 I noticed that the .bss sections is not initial to 0, this is causing some issues when we running the test on the .bss section is not 0  after reset.  becasue the scratch section sit in the .bss
https://github.com/riscv/riscv-arch-test/blob/act4/tests/env/test_setup.h#L217-L220

CC @jordancarlin @allenjbaum @davidharrishmc

jordancarlin:
Replace the `.bss` scratch region with a `.data` section that is preloaded with values for consistent test results when loading from the scratch region. This also means that tests do not need to set the scratch region at the start (which was done inconsistently).

Fixes #1167


tests/env/rvtest_setup.h
```diff
@@ -256,13 +256,29 @@
...
 .macro RVTEST_DATA_BEGIN
   // Scratch region of memory for tests (ie for loads/stores that are not part of signature)
-  .section .bss
+  // Initialized with distinct random values so tests can detect unintended zeroing or aliasing.
+  // 264 bytes = 66 words (needed for atomic reservation tests with offsets up to 256 bytes)
+  .data
   .align 4
   scratch:
-    .space 264 // Reserve enough scratch space (needed for atomic reservation tests with offsets up to 256 bytes)
+    .word 0xEA1FAD7A, 0x3B8596BC, 0x969E25D7, 0x79DE9C3D
```
davidharrishmc: Why these particular values?  Would DEADBEEF be clearer to indicate uninitialized data?

davidharrishmc (changes requested):

tests/env/rvtest_setup.h
jordancarlin: Several tests load from the scratch region and this gives us unique data so we know if it loaded from the wrong address. This is just a set of data that was randomly generated.

davidharrishmc (approved):

PR #1243, opened 2026-04-10
Split FLEN into TEST_FLEN and CONFIG_FLEN to make ZicsrF work on DUTs with only F

### LINKED ISSUE #1223: Wrong FLEN when test march includes D but DUT only implements F

jordancarlin:
Fixes #1223.

### DIFF
--- a/tests/env/failure_code.h
+++ b/tests/env/failure_code.h
@@ -245,7 +245,7 @@
         la x6, scratch
         LREG x7, 0(x6)
         SREG x7, 272(DEFAULT_TEMP_REG)    # failing_value (lower/only)
-    #if FLEN > XLEN
+    #if CONFIG_FLEN > XLEN
         LREG x7, REGWIDTH(x6)
         la x8, failing_value_upper
         SREG x7, 0(x8)                    # failing_value upper half
@@ -265,7 +265,7 @@
         # Load full expected FP value from signature
         LREG x7, 0(x6)
         SREG x7, 280(DEFAULT_TEMP_REG)    # expected_value (lower/only)
-    #if FLEN > XLEN
+    #if CONFIG_FLEN > XLEN
         LREG x7, SIG_STRIDE(x6)
         la x8, expected_value_upper
         SREG x7, 0(x8)                    # expected_value upper half
@@ -395,13 +395,13 @@
         lw a0, failure_type
         li a1, 1
         bne a0, a1, failedtest_report_badval_not_fp
-    #if defined(RVTEST_FP) && FLEN > XLEN
-        # FP with FLEN > XLEN: combined hex "0xUPPER_LOWER"
+    #if defined(RVTEST_FP) && CONFIG_FLEN > XLEN
+        # FP with CONFIG_FLEN > XLEN: combined hex "0xUPPER_LOWER"
         LREG a0, failing_value_upper
         LREG a1, failing_value
         jal failedtest_combined_hex_to_str
     #else
-        # FP with FLEN <= XLEN (or FLEN not defined): standard hex
+        # FP with CONFIG_FLEN <= XLEN (or CONFIG_FLEN not defined): standard hex
         LREG a0, failing_value
         li a1, __riscv_xlen
         jal failedtest_hex_to_str
@@ -422,13 +422,13 @@
         lw a0, failure_type
         li a1, 1
         bne a0, a1, failedtest_report_expval_not_fp
-    #if defined(RVTEST_FP) && FLEN > XLEN
-        # FP with FLEN > XLEN: combined hex "0xUPPER_LOWER"
+    #if defined(RVTEST_FP) && CONFIG_FLEN > XLEN
+        # FP with CONFIG_FLEN > XLEN: combined hex "0xUPPER_LOWER"
         LREG a0, expected_value_upper
         LREG a1, expected_value
         jal failedtest_combined_hex_to_str
     #else
-        # FP with FLEN <= XLEN (or FLEN not defined): standard hex
+        # FP with CONFIG_FLEN <= XLEN (or CONFIG_FLEN not defined): standard hex
         LREG a0, expected_value
         li a1, __riscv_xlen
         jal failedtest_hex_to_str
@@ -556,7 +556,7 @@
         ret
 
 
-#if defined(RVTEST_FP) && FLEN > XLEN
+#if defined(RVTEST_FP) && CONFIG_FLEN > XLEN
     # Convert two XLEN-wide values to combined hex string: "0xUPPER_LOWER\n\0"
     # a0: upper XLEN-bit value
     # a1: lower XLEN-bit value
@@ -628,7 +628,7 @@
         .fill 2, 4, 0xfeedf00dbaaaaaad
     failure_string_ptr:
         .fill 2, 4, 0xfeedf00dbaaaaaad
-#if defined(RVTEST_FP) && FLEN > XLEN
+#if defined(RVTEST_FP) && CONFIG_FLEN > XLEN
     failing_value_upper:
         .fill 2, 4, 0xfeedf00dbaaaaaad
     expected_value_upper:
--- a/tests/env/failure_code.h
+++ b/tests/env/failure_code.h
@@ -242,8 +242,10 @@
         sw x6, 260(DEFAULT_TEMP_REG)      # record failing_reg
 
         # Load bad FP value from scratch memory (written by FSREG in the sigupd macro)
+        # Use FP_LREG so we read exactly the CONFIG_FLEN bits FSREG stored,
+        # zero-extending on RV64+F-only where fsw wrote fewer bytes than LREG reads.
         la x6, scratch
-        LREG x7, 0(x6)
+        FP_LREG x7, 0(x6)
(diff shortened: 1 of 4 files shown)


jordancarlin:
Confirmed this works for DUTs that only have F and still gets full coverage both for DUTs with only F and with both F and D. We need to get a config added to CI that only has F without D to avoid issues like this in the future.

tests/env/signature.h
```diff
@@ -111,24 +111,32 @@
...
 // Floating point values are stored to memory and then loaded back into integer registers
 // for comparison, to avoid issues with NaN that arise from using feq. There is no way to
 // directly transfer a floating point value to an integer register without Zfa when FLEN > XLEN.
-#if FLEN == 128 && XLEN == 32
+#if CONFIG_FLEN == 128 && XLEN == 32
   #error "Q on RV32 is not supported yet."
 #endif
-#if FLEN > XLEN
+#if CONFIG_FLEN > XLEN
   #ifdef RVTEST_SELFCHECK
```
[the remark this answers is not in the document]
jordancarlin: Good catch Copilot! Working on a fix for this.

tests/env/failure_code.h
```diff
@@ -242,10 +242,12 @@
         sw x6, 260(DEFAULT_TEMP_REG)      # record failing_reg
 
         # Load bad FP value from scratch memory (written by FSREG in the sigupd macro)
+        # Use FP_LREG so we read exactly the CONFIG_FLEN bits FSREG stored,
```
davidharrishmc, line 245: "// See tests/env/utils.h for an explanation of CONFIG_FLEN and TEST_FLEN"

tests/env/signature.h
```diff
@@ -111,24 +111,32 @@
 // RVTEST_SIGUPD_F(sigptr, linkreg, tempreg, ftempreg, sigreg, instptr, strptr)
 // Checks both a floating point result register and fflags against the signature.
 // Compares the float value in sigreg with the value in memory at 0(sigptr),
-// then uses RVTEST_SIGUPD_FFLAGS to check fflags. When FLEN > XLEN, the float
-// value requires 2 signature entries (low and high words), plus 1 for fflags
-// (total 3*SIG_STRIDE). When FLEN == XLEN, uses 1 entry for the float value
-// plus 1 for fflags (total 2*SIG_STRIDE).
+// then uses RVTEST_SIGUPD_FFLAGS to check fflags. When CONFIG_FLEN > XLEN, the
```
davidharrishmc, line 114: "// See tests/env/utils.h for an explanation of CONFIG_FLEN and TEST_FLEN"

tests/env/utils.h
```diff
@@ -43,34 +43,101 @@
     #define LREG lq
 #endif
 
-# FLEN specific macros
-#define FREGWIDTH (FLEN>>3)      // in units of #bytes
+// FLEN specific macros
+// ============================================================================
+// Two distinct FLEN values
+// ----------------------------------------------------------------------------
+// TEST_FLEN  — the FLEN this test file was *generated* for. Supplied by the
```
davidharrishmc: Start with explaining the problem that is being solved.

tests/env/utils.h
```diff
@@ -43,34 +43,101 @@
...
+//               needs to be sliced into two integer loads (the
+//               "CONFIG_FLEN > XLEN" path in signature.h).
+//
+//               CONFIG_FLEN must not exceed TEST_FLEN because the assembler
+//               only knows instructions up to that width (e.g. an F-only test
+//               with march=rv64if cannot assemble fsd). It must not exceed
+//               the DUT's capability either (e.g. a priv test generated with
+//               D in its march but run on an F-only DUT must not emit fsd).
+//               See issue #1223.
+// ============================================================================
```
davidharrishmc, line 88: Give examples of TEST_FLEN and CONFIG_FLEN for RV32/64 D/F and show how they solve the problem described above.

tests/env/utils.h
```diff
@@ -43,34 +43,101 @@
...
+#endif
+
+#define FREGWIDTH (TEST_FLEN>>3)     // data/signature slot width, in bytes
+
+// Derive the DUT's raw FP capability from its rvtest_config.h defines.
+#if defined(Q_SUPPORTED)
+  #define _DUT_FLEN 128
+#elif defined(D_SUPPORTED)
+  #define _DUT_FLEN 64
+#elif defined(F_SUPPORTED) || defined(ZFH_SUPPORTED) || defined(ZFHMIN_SUPPORTED) || defined(ZFBFMIN_SUPPORTED)
```
davidharrishmc: Don't all these other Zf* depend on F_SUPPORTED so you only need to check F_SUPPORTED and could leave it as #else with no check?

tests/env/utils.h
davidharrishmc: "was generated for" is unclear because most readers won't fully understand the generation process.  Explain how to know what it was generated for.

tests/env/utils.h
```diff
@@ -278,9 +346,9 @@
   .option pop
 #endif
 
-# Alignment size for LA macro. Must be larger than the longest instruction
-# sequence that the la pseudo-instruction can expand into (to account for the jump hack).
-# On some rv64 targets, this may need to be increased to 6.
+// Alignment size for LA macro. Must be larger than the longest instruction
+// sequence that the la pseudo-instruction can expand into (to account for the jump hack).
+// On some rv64 targets, this may need to be increased to 6.
```
davidharrishmc, line 368: How will we know which rv64 targets need 6 and when to change it?

jordancarlin: This comment has been here for a while. I think @allenjbaum is best suited to answer this.

davidharrishmc (approved):

PR #1258, opened 2026-04-14
InterruptsS_S python tests

SadhviNarayanan:
These are the tests for the supervisor interrupts.
This module generates tests for supervisor-mode interrupts (STIP, SSIP, SEIP), covering both non-delegated (fires in M-mode) and delegated (fires in S-mode) interrupt handling across a wide range of mideleg, mie, and mstatus configurations.
Currently at 76.85% coverage because external interrupts are not implemented.

### DIFF
--- a/config/sail/sail-rv64-max/rvmodel_macros.h
+++ b/config/sail/sail-rv64-max/rvmodel_macros.h
@@ -132,23 +132,18 @@
 
 
 // TODO: check to see if SAIL support this, and we may want to implement this in WALLY
-#define CLINT_SSIP_ADDRESS (CLINT_BASE_ADDRESS + 0xC000)
-// #define RVMODEL_SET_SSW_INT(_R1, _R2)        \
-//   li _R1, 1;                 \
-//   li _R2, CLINT_SSIP_ADDRESS;              \
-//   sw _R1, 0(_R2);
-
-
-// #define RVMODEL_CLR_SSW_INT(_R1, _R2)        \
-//   li _R2, CLINT_SSIP_ADDRESS;              \
-//   sw zero, 0(_R2);
-
+// #define CLINT_SSIP_ADDRESS (CLINT_BASE_ADDRESS + 0xC000)
+#define CLINT_SSIP_ADDRESS (0x80000000)
 #define RVMODEL_SET_SSW_INT(_R1, _R2)        \
-  li _R1, 2;                 \
-  csrs mip, _R1;             /* Set mip.SSIP (M-mode) */
+  li _R1, 1;                 \
+  li _R2, CLINT_SSIP_ADDRESS;              \
+  sw _R1, 0(_R2);
+
 
 #define RVMODEL_CLR_SSW_INT(_R1, _R2)        \
-  li _R1, 2;                 \
-  csrc sip, _R1;             /* Clear sip.SSIP (S-mode handler: only used when mideleg.SSIP=1) */
+  li _R2, CLINT_SSIP_ADDRESS;              \
+  sw zero, 0(_R2);
+
+
 
 #endif // _RVMODEL_MACROS_H
--- a/config/sail/sail-rv64-max/rvmodel_macros.h
+++ b/config/sail/sail-rv64-max/rvmodel_macros.h
@@ -133,14 +133,22 @@
 
 // TODO: check to see if SAIL support this, and we may want to implement this in WALLY
 #define CLINT_SSIP_ADDRESS (CLINT_BASE_ADDRESS + 0xC000)
-#define RVMODEL_SET_SSW_INT(_R1, _R2)        \
-  li _R1, 1;                 \
-  li _R2, CLINT_SSIP_ADDRESS;              \
-  sw _R1, 0(_R2);
+// #define RVMODEL_SET_SSW_INT(_R1, _R2)        \
+//   li _R1, 1;                 \
+//   li _R2, CLINT_SSIP_ADDRESS;              \
+//   sw _R1, 0(_R2);
 
+
+// #define RVMODEL_CLR_SSW_INT(_R1, _R2)        \
+//   li _R2, CLINT_SSIP_ADDRESS;              \
+//   sw zero, 0(_R2);
+
+#define RVMODEL_SET_SSW_INT(_R1, _R2)        \
+  li _R1, 2;                 \
+  csrs mip, _R1;             /* Set mip.SSIP (M-mode) */
 
 #define RVMODEL_CLR_SSW_INT(_R1, _R2)        \
-  li _R2, CLINT_SSIP_ADDRESS;              \
-  sw zero, 0(_R2);
+  li _R1, 2;                 \
+  csrc sip, _R1;             /* Clear sip.SSIP (S-mode handler: only used when mideleg.SSIP=1) */
 
 #endif // _RVMODEL_MACROS_H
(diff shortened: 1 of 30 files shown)


tests/env/rvtest_trap_handler.h
```diff
@@ -1735,10 +1729,11 @@ excpt_\__MODE__\()hndlr_tbl:            // handler code should only touch T2..T6
...
                                         // S-mode software interrupts need to be reset differently when raised in M or S mode
-        .ifc \__MODE__ , M              // Select the interrupt handler function based on current privilege mode
-            RVMODEL_CLR_SSW_INT(T2, T5)
+        .ifc \__MODE__ , M              // M-mode: mideleg.SSIP=0 so sip.SSIP is read-only; must clear via mip directly
+            li T2, 2
+            csrc mip, T2
         .else
-                .ifc \__MODE__ , S
+                .ifc \__MODE__ , S      // S-mode: mideleg.SSIP=1 so sip.SSIP mirrors mip and is writable
                         RVMODEL_CLR_SSW_INT(T2, T5)
```
davidharrishmc, line 1749: sip.SSIP might be set in two ways: via writing to sip, and via the external interrupt controller.  To clear it, you need to do both because you don't know which one set it.  Do a csrc on sip.SSIP, and also do a RVMODEL_CLR_SSW_INT, which should turn off the external interrupt controller pin.  Similar for cleaning from M-mode with mip.SSIP and RVMODEL_CLR_SSW_INT.

generators/testgen/src/testgen/priv/extensions/InterruptsS.py
```diff
@@ -0,0 +1,4743 @@
...
+            else:
+                # Not delegated: set MIE
+                if int_enable_val:
+                    lines.append("csrsi mstatus, 8")
+
+            # 7. Set SSIP
+            lines.append(test_data.add_testcase(binname, coverpoint, covergroup))
+            lines.extend(
+                [
+                    "RVTEST_SET_SSW_INT",
```
davidharrishmc: The test plan says to trigger SSIP with mip.  This is using the macro to trigger SSIP, and the macro is supposed to invoke the external interrupt controller if it exists.  So this should use a csrs instead of the macro.

generators/testgen/src/testgen/priv/extensions/InterruptsS.py
```diff
@@ -0,0 +1,4743 @@
...
+            # 9. Return and cleanup
+            lines.extend(
+                [
+                    "RVTEST_GOTO_MMODE",
+                    "# Complete state cleanup",
+                    "csrci mstatus, 8",  # Clear MIE
+                    "csrci mstatus, 2",  # Clear SIE
+                    "CSRW(mideleg, zero)",  # Clear delegation
+                    "CSRW(mie, zero)",  # Clear all interrupt enables
+                    "RVTEST_CLR_SSW_INT",
```
davidharrishmc: turn off with circa

davidharrishmc (changes requested):

config/sail/sail-rv64-max/rvmodel_macros.h
```diff
@@ -132,7 +132,8 @@
 
 
 // TODO: check to see if SAIL support this, and we may want to implement this in WALLY
-#define CLINT_SSIP_ADDRESS (CLINT_BASE_ADDRESS + 0xC000)
```
davidharrishmc, line 135: Change comment to "Sail does not yet support memory-mapped I/O to cause a supervisor software interrupt.  Change CLINT_SSIP_ADDRSS to appropriate location when implemented in Sail."

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -24,7 +24,7 @@ def _generate_trigger_mti_tests(test_data: TestData) -> list[str]:
 
     # Exclude: x2 (sp), x5 (t0-used by macros),
     # x7 (t2-consumed by interrupt macros), x30 (t5-consumed by interrupt macros)
-    r1, r_mtime, r_mtimecmp, r_temp, r_temp2 = test_data.int_regs.get_registers(5, exclude_regs=[2, 7, 30])
+    r1, r_mtime, r_mtimecmp, r_temp, r_temp2 = test_data.int_regs.get_registers(5, exclude_regs=[0, 2, 7, 30])
```
davidharrishmc: You shouldn't need to exclude 0, 7, 30.  They are already excluded in priv.py at line 64.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -71,7 +71,9 @@ def _generate_trigger_msi_tests(test_data: TestData) -> list[str]:
     coverpoint = "cp_trigger_msi"
     ######################################
 
-    r1, r_mtime, r_mtimecmp, r_temp, r_temp2, r_cleanup = test_data.int_regs.get_registers(6, exclude_regs=[2, 7, 30])
+    r1, r_mtime, r_mtimecmp, r_temp, r_temp2, r_cleanup = test_data.int_regs.get_registers(
+        6, exclude_regs=[0, 2, 7, 30]
```
davidharrishmc: same exclusion.

davidharrishmc: several more places.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -347,6 +351,7 @@ def _generate_priority_tests(test_data: TestData) -> list[str]:
         for mip_bits in range(8):
             lines.extend(
                 [
+                    "CSRW(mie, zero)",  # disable interrupts
```
davidharrishmc: why is this done twice in a row?

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -382,7 +387,7 @@ def _generate_wfi_tests(test_data: TestData) -> list[str]:
     covergroup = "InterruptsSm_cg"
     coverpoint = "cp_wfi"
 
-    r_mtime, r_mtimecmp, r_t0, r_t1, r_t2, r_t3, r_scratch = test_data.int_regs.get_registers(7)
+    r_mtime, r_mtimecmp, r_t0, r_t1, r_t2, r_t3, r_scratch = test_data.int_regs.get_registers(7, exclude_regs=[0])
```
davidharrishmc: no need to exclude 0.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -20,7 +20,7 @@ def _generate_user_mti_tests(test_data: TestData) -> list[str]:
     covergroup = "InterruptsU_cg"
     coverpoint = "cp_user_mti"
 
-    r_mtime, r_mtimecmp, r_temp, r_temp2, r_scratch = test_data.int_regs.get_registers(5)
+    r_mtime, r_mtimecmp, r_temp, r_temp2, r_scratch = test_data.int_regs.get_registers(5, exclude_regs=[0, 2])
```
davidharrishmc: No need to exclude 0

davidharrishmc (changes requested):

davidharrishmc (approved):

PR #1262, opened 2026-04-15
Add IBM coverpoint generator framework

jordancarlin:


generators/testgen/src/testgen/coverpoints/cp_ibm.py
```diff
@@ -0,0 +1,116 @@
+##################################
+# cp_ibm.py
+#
+# IBM-defined floating point coverpoint generator. Each instruction has multiple
```
davidharrishmc: The language is a bit weird here because these are really tests, not coverpoints.  Or the coverpoints are expressed as test cases.

generators/testgen/src/testgen/coverpoints/cp_ibm.py
```diff
@@ -0,0 +1,116 @@
...
+
+# IBM testcase data files live alongside this generator inside the testgen package.
+IBM_DATA_DIR = Path(__file__).resolve().parent / "ibm"
+
+# Valid rounding mode names accepted in the frm column of IBM CSV files.
+VALID_FRM_NAMES = frozenset({"rne", "rtz", "rdn", "rup", "rmm"})
+
+# Input operand value columns that may appear in an IBM CSV. The subset actually
+# required for a given instruction is determined from its formatter config.
+INPUT_VALUE_KEYS = frozenset({"fs1val", "fs2val", "fs3val", "rs1val", "rs2val", "rs3val"})
```
davidharrishmc, line 33: Is there a graceful way for this to play nicely with Zfinx automatically?

generators/testgen/src/testgen/coverpoints/cp_ibm.py
```diff
@@ -0,0 +1,116 @@
...
+
+# Expected-result columns that the CSV may include for DUT verification. These are
+# accepted in the header but currently unused by the test generator.
```
davidharrishmc, line 36: it would be great to print expected results in the test comments.

davidharrishmc (approved):
Approved, but if there are easy ways to address my queries, it would be great to include them.

generators/testgen/src/testgen/coverpoints/cp_ibm.py
jordancarlin: We will need to address Zfinx more holistically later throughout the whole framework. I'd rather defer this until then.

generators/testgen/src/testgen/coverpoints/cp_ibm.py
jordancarlin: @fourth-bit is going to work on continuing to augment this generator. This is an initial implementation. Eventually we will add support for actually using the expected results as the correct result in the test (and can add it as a comment too).

PR #1264, opened 2026-04-16
Ensure consistent addresses between signature and self-checking ELFs and document memory map

### LINKED ISSUE #1151: ExceptionsSm: mtval value mismatch due to different scratch addresses
This look like a general issue similar to https://github.com/riscv/riscv-arch-test/issues/262

I see it in ExceptionsSm_cg_cp_load_address_misaligned_lh_off1

The load address (based on scratch) is different between .elf and .sig.elf.
mtval in signature is therefore different and fails the test.
I don't see an address adjustment for mtval, it is a direct compare.

### LINKED ISSUE #1205: Document memory map
It would be helpful to document the memory map and all of its subtle behaviors.

jordancarlin:
Updates the linker script and `rvtest_setup.h` to place all RVMODEL macros in a separate text section that is located after the `.data` section in memory so addresses stay consistent. Also adds documentation of the test memory layout.

Fixes #1151 and fixes #1205.


docs/ctp/src/rvmodel.adoc
```diff
@@ -86,4 +86,6 @@ If you have a complex DUT, you presumably have a team of sophisticated verificat
 [[linker-script]]
 === Linker Script
 
-Each DUT needs a linker script tailored to its memory map. The linker script in config/<core>/<version>/link.ld defines where to place the code and data regions in the DUT's memory map.  It must contain `.text.init`, `.text`, and `.data` sections, and an `ENTRY` point named `rvtest_entry_point`.  Users with special needs can customize the linker script to start at their own boot code (leaving the `RVMODEL_BOOT` macro blank) and then jump to `rvtest_entry_point`.
+Each DUT needs a linker script tailored to its memory map. The linker script in `config/<core>/<version>/link.ld` defines where to place the code and data regions in the DUT's memory map.  It must contain `.text.init`, `.text.rvtest`, `.data`, and `.text.rvmodel` sections (in that order), and an `ENTRY` point named `rvtest_entry_point`.  Users with special needs can customize the linker script to start at their own boot code (leaving the `RVMODEL_BOOT` macro blank) and then jump to `rvtest_entry_point`.
```
davidharrishmc: Explain the . = 0x...  Explain that this is the only part of a linker file that a simple DUT needs to change, to point to the reset vector.

docs/memory_map.md
```diff
@@ -0,0 +1,70 @@
+# Test Memory Map
+
+The test memory layout is divided into four linker sections, each aligned and ordered so that the DUT ELF and the reference-model ELF produce identical addresses for all test-visible symbols.
+
+## Section Layout
+
+The linker places the following output sections in order:
```
davidharrishmc, line 19: This section is key and could be clearer, especially for somebody who doesn't know linker syntax.  

Clarify that rvtest_entry_point is normally the reset vector and the start of the program.  Most cores just need to change the . = 0x... to where the processor resets.  If you know what you are doing, do whatever you need.

docs/memory_map.md
```diff
@@ -0,0 +1,70 @@
...
+
+## `.text.init` Section Layout
+
+| Symbol / Region      | Purpose                                                                                                  |
+| -------------------- | -------------------------------------------------------------------------------------------------------- |
+| `rvtest_entry_point` | Entry point. Calls `rvmodel_boot` (in the `.text.rvmodel` section) then falls through to `.text.rvtest`. |
+
+`.text.init` is intentionally kept in a separate output section so that `.align` directives in test code do not increase `.text.init`'s alignment and shift `rvtest_entry_point` to an unexpected address.
```
davidharrishmc: For example, cva2... has . = 0x400 and ..., causing rvtest_entry_point to shift to ... if .text.init were combined with .text.rvtest (explain this, or nobody else will understand this subtle one).

docs/memory_map.md
```diff
@@ -0,0 +1,70 @@
...
+| `rvtest_code_begin`           | Start of the test body (signature pointer initialized here).                                                   |
+| _(test code)_                 | The actual test, generated between `RVTEST_CODE_BEGIN` and `RVTEST_CODE_END`.                                  |
+| `rvtest_code_end`             | End of the test body. Switches back to M-mode.                                                                 |
+| `cleanup_epilogs`             | Trap epilogs (restore xTVEC, trampoline, and saved registers per mode).                                        |
+| `exit_cleanup` / `abort_test` | Test termination paths (calls `rvmodel_halt_pass` or `rvmodel_halt_fail`).                                     |
+| Trap handlers                 | One handler per privilege mode (`RVTEST_TRAP_HANDLER`).                                                        |
+| Failure code                  | Failure detection and diagnostic reporting (`RVTEST_FAILURE_CODE`).                                            |
+
+## `.text.rvmodel` Section Layout
```
davidharrishmc: List this section after .data to match order of linker file.

docs/memory_map.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Section Layout
+
+The linker places the following output sections in order:
+
+| Section                             | Permissions | Contents                                                                                                                                   |
+| ----------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
+| `.text.init`                        | R X         | Entry point (`rvtest_entry_point`), boot call, and jump to `.text.rvtest`.                                                                 |
+| `.text.rvtest`                      | R X         | All test code: initialization, trap handlers, the test body, and failure-reporting code.                                                   |
+| `.data` (aligned to 0x4000)         | R W         | All test data and the signature region (see below).                                                                                        |
```
davidharrishmc: Shouldn't this be RWX because some test such as fence.i write an instruction to scratch and then attempt to JAL and run it.

docs/memory_map.md
```diff
@@ -0,0 +1,70 @@
+# Test Memory Map
```
davidharrishmc, line 1: There are subtle issues about how certain addresses have to be exactly the same in the DUT and ref that aren't explicitly explained here.  How about something along the lines:

```
The test memory map has several requirements:

* Place the beginning of the test code at the right place (usually the reset vector, unless the DUT has a separate boot loader)
* Ensure that the addresses of the test code body are the same when running it in Sail to produce an expected signature and on a DUT to check against the expected signature.
* * Ensure the addresses of test data are the same when running in Sail and on the DUT.

The second requirement is subtle.  The program is compiled twice, once to run on Sail to produce an expected signature, and once to run on a DUT to check against the expected signature. The `RVMODEL` macros may differ between Sail and DUT because the two targets might use different methods of printing to the console, terminating a simulation, etc.

The address of the test code body (the part that tests a feature and checks expected results) must be the same because some instructions such as `auipc` produce an expected result that depends on the address.  Moreover, instructions that trap record the trap address in `xepc`, which should match on Sail and the DUT.  Therefore, the `RVMODEL` macros cannot be called directly in the test code body because they might expand to different sizes in Sail vs. the DUT.  Instead, they go in a different section placed after  the test code body and are called from the body via `jal`.

File: docs/memory_map.md [hunk 7]
```diff
@@ -0,0 +1,70 @@
...
+The test memory layout is divided into four linker sections, each aligned and ordered so that the DUT ELF and the reference-model ELF produce identical addresses for all test-visible symbols.
+
+## Section Layout
+
+The linker places the following output sections in order:
+
+| Section                             | Permissions | Contents                                                                                                                                   |
+| ----------------------------------- | ----------- | ------------------------------------------------------------------------------------------------------------------------------------------ |
+| `.text.init`                        | R X         | Entry point (`rvtest_entry_point`), boot call, and jump to `.text.rvtest`.                                                                 |
+| `.text.rvtest`                      | R X         | All test code: initialization, trap handlers, the test body, and failure-reporting code.                                                   |
```
davidharrishmc, line 24: How do we know trap handler and failure-reporting code are exactly the same size?  Does failure reporting instantiate character I/O macro directly?

davidharrishmc: I think I see below that you are changing to call rvmodel_io_write_str.  Worth explaining.

tests/env/rvtest_setup.h
```diff
@@ -197,10 +204,21 @@
...
 
-  // rvtest macros are used to invoke the rvmodel specific interrupt macros and those rvmodels macros need to be at the end of the program because their length is variable and we don't want their lengths to affect the relative addresses of program
+  rvmodel_io_write_str:
+    // a0 = string pointer; T1-T3 (x6-x8) are scratch. Clobbers ra.
+    RVMODEL_IO_WRITE_STR(T1, T2, T3, a0)
+    ret
+
+  rvmodel_halt_pass:
+    RVMODEL_HALT_PASS
+    // does not return
```
davidharrishmc: should it have an infinite loop here in case RVMODEL_HALT_PASS is buggy and doesn't terminate?  Same with fail.

davidharrishmc (changes requested):
This is complicated stuff and the docs don't yet fully explain the subtle points.  They need more emphasis on the why and not just the what.  I've made a few suggestions, but it's worth another close path to make sure everything is explained to somebody unfamiliar with the framework.

davidharrishmc (approved):

Issue #1172, opened 2026-04-01
ExceptionsZc: mcause mismatch in ExceptionsZc-00.S when running ACT on CV32A65x


arshiarifat-10x:
We are integrating ACT 4.0 tests for the CV32A65x configuration in CVA6 and running the riscv-arch-test suite through the ACT generation flow.

While running the Zc exception tests we observe a failure in:

priv/ExceptionsZc/ExceptionsZc-00.S

The failure is due to a mismatch in the expected mcause value.

Environment:
Core: CV32A65x (CVA6)
Simulator: Verilator

Observed behavior:

Instruction that trapped: 0x9002  (c.ebreak)
Address: 0x80002bbc

Expected mcause: 0x5
Observed mcause: 0x3

Log excerpt:

RVCP-SUMMARY: TEST FAILED - Test File "ExceptionsZc-00.S"

RVCP: Test Info: "Mismatch in mcause value! Trap was being handled in M-Mode."
RVCP: Instruction: 0x00050413
RVCP: Register: x8
RVCP: Bad Value:      0x00000003
RVCP: Expected Value: 0x00000005
RVCP: Address of instruction that trapped (XEPC): 0x80002bbc
RVCP: Instruction that trapped: 0x9002

Workaround:

Currently we run ACT generation with:

include_priv_tests: False

This allows tests to generate and execute successfully.

Could you please confirm whether this behavior is expected or if additional fixes are required for the ExceptionsZc tests?

@jordancarlin 
@MikeOpenHWGroup

davidharrishmc:
The observed mcause of 3 is “Breakpoint” while the expected mcause of 5 is “Load access fault.”  
For some reason, Sail seems to have encountered a load access fault when generating the expected signature,
while the DUT is correctly causing a breakpoint exception when executing c.ebreak.

This usually means something is not configured properly in the sail.json file in the cv32a64x config directory.

It might be visible by inspection

You can set DEBUG = True in the Makefile and look at the Sail trace in the work/cv32a64x/build directory around PC 80002bbc to loook
for clues about why Sail was predicting a load access fault.

Otherwise, Jordan is a wizard at this stuff and is likely to have ideas.

include_priv_tests = False hides the mismatch by not running privilege tests, so the DUT is not being fully tested.

MikeOpenHWGroup:
[...]

PR #1271, opened 2026-04-16
Replace LI with LA to address #1172

### LINKED ISSUE #1259: check_access_fault_address_defined()` uses linker symbol `scratch` in `LI()` macro - assembler error
PR #1248  introduced `check_access_fault_address_defined()` in `helpers.py` to make `RVMODEL_ACCESS_FAULT_ADDRESS` optional for DUTs that do not generate access faults. When the macro is not defined by the platform, the helper emits:

```c
#ifndef RVMODEL_ACCESS_FAULT_ADDRESS
  #define RVMODEL_ACCESS_FAULT_ADDRESS scratch
#endif
```
The intent was as I understand to substitute the test framework's `scratch` memory region - a safe area that won't cause an unexpected fault - so the test compiles and runs on DUTs without access fault support.

**Describe the bug**
scratch` is a linker symbol. Its address is assigned by the linker at link time, after the assembler has already run. However, `LI()` (tests/env/utils.h) is not  simple instruction  it expands to assembler `.set` and `.if` directives that evaluate their argument as a **compile-time constant in assembly pass**:

```asm
.set immx,   (RVMODEL_ACCESS_FAULT_ADDRESS & MASK)   #  assembler arithmetic
.if  ((absimm >> IMMSGN) == 0)                        #  assembler conditional
```

When `RVMODEL_ACCESS_FAULT_ADDRESS` expands to `scratch`, the assembler cannot resolve `scratch & MASK` — the value is not yet known. This produces:

```
Error: non-constant expression in ".if" statement
Error: illegal operands `lui x8,(((immx>>12)+cry)&...)'
```

Note that `LA(x, scratch)` works fine because `LA` uses `%pcrel_hi`/`%pcrel_lo`  that are resolved at link time.

Affected generators:
- ExceptionsZc
- ExceptionsZaamo
- ExceptionsZalrsc
- ExceptionsZicboS
- ExceptionsZicboU

**To Reproduce**
During testing cv32e20 configuration.

**Additional context**
This also introduce problems with ExceptionsSm where it failed with linker error since macro access fault was undefined.

Changing RVMODEL_ACCESS_FAULT to safe address works as temporary workaround (but still fails for ExceptionsSm). 

ExceptionsSm still doesnt skip cover points with Access fault. Was this intentional in this PR?

davidharrishmc:
Fixes #1259

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
@@ -391,7 +391,7 @@ def generate_store_access_fault_tests(test_data: TestData, covergroup: str) -> l
         lines.extend(
             [
                 f"\n# Testcase: {op} access fault",
-                f"LI(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+                f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
                 f"LI(x{data_reg}, {test_values[op]})",
                 test_data.add_testcase(f"{op}_fault", coverpoint, covergroup),
                 f"{op} x{data_reg}, 0(x{addr_reg})",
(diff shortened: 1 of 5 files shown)


davidharrishmc:
This did not work.  See #1172.  Also, testing instruction access faults jump to a weird instruction at scratch if they do run.  Next idea would be to gate all access fault tests with #ifdef RVMODEL_ACCESS_FAULT_ADDRESS.  

Closing this PR.

Issue #1141, opened 2026-03-27
Generate supplemental config files from UDB config


nazar-pc:
Right now configs like `sail.json` need to be written manually. The readme suggests they will be auto-generated (I guess from `<dut-name>.yaml`? Would be really nice!), but I do not see a tracking issue for it, hence creating this issue.

jordancarlin:
This is definitely something we want to support. Most of the infrastructure to generate the other config files will likely come from https://github.com/riscv/riscv-unified-db.

The files that we plan to generate from the UDB config (`<dut_name>.yaml`) are:

- [x] `rvtest_config.h`
- [x] `rvtest_config.svh`
- [ ] `sail.json`

davidharrishmc:
rvtest_config.{h/svh} are now generated.

sail.json is still pending.

PR #1272, opened 2026-04-16
Remove duplicate RVMODEL_ACCESS_FAULT_ADDRESS definitions

jordancarlin:
`RVMODEL_ACCESS_FAULT_ADDRESS` was defined in both `rvtest_config.h` and `rvmodel_macros.h` for many configs.


davidharrishmc (approved):

karabambus:
It is also defined in rvtest_config.svh . Although it doesnt have same effect for tests, this redundancy is error prone. May I also ask why do we use macros in svh instead of packages?

jordancarlin:
> It is also defined in rvtest_config.svh . Although it doesnt have same effect for tests, this redundancy is error prone. May I also ask why do we use macros in svh instead of packages?

`rvtest_config.svh` will be handled as part of #1141.

PR #1281, opened 2026-04-17
Add GitHub Actions workflow for autofixing tests and coverpoints

jordancarlin:
PRs that change the test or coverpoint generators will automatically have the PR updated to contain the changes to the generated files. It is easy to forget to commit the generated files, so this workflow helps ensure they are always up to date.


davidharrishmc (approved):
LGTM

PR #1303, opened 2026-04-17
Fix InterruptsSm mstatus.MIE bit

jordancarlin:


jordancarlin:
Also fixed coverage in InterruptsU. Both are now at 100% coverage when using the upcoming Sail release with the new external interrupt generator.

davidharrishmc (approved):

PR #1304, opened 2026-04-18
VLS 100% coverage -- Vector Unpriv 100% coverage

jacassidy:


config/sail/sail-rv32-max/rvtest_config.svh
```diff
@@ -66,3 +66,6 @@
 `define LMULf8_SUPPORTED
 `define LMULf4_SUPPORTED
 `define LMULf2_SUPPORTED
+
+// Maximum supported index EEW for indexed load/store
+`define MAXINDEXEEW32
```
jordancarlin, line 72: Shouldn't the rv32 configs still support EEW of 64?

coverpoints/coverage/RISCV_instruction_sample.svh
```diff
@@ -2615,223 +2615,271 @@
...
+`ifdef MAXINDEXEEW_GE16
         "vloxei16.v"     : begin
             ins.add_vd(0);
             ins.add_rs1(1);
             ins.add_vs2(2);
             ins.add_vm(3);
             ins.add_v0();
         end
+`endif
+`ifdef MAXINDEXEEW_GE32
```
jordancarlin: Why are all of these ifdefs needed? Shouldn't it always be same to sample the instruction?

framework/src/act/build_plan.py
```diff
@@ -231,20 +231,25 @@ def gen_rvvi_tasks(
 
     # Paths
     coverage_dir = base_dir / "coverage"
-    elf_dir = base_dir / "elfs"
-    elf = elf_dir / test_name.with_suffix(".elf")
+    build_dir = base_dir / "build"
+    # Use sig.elf (non-selfcheck) for coverage traces: it runs all tests to
+    # completion without halting on selfcheck mismatches, producing a complete
+    # trace for every instruction.
+    elf = build_dir / test_name.with_suffix(".sig.elf")
```
jordancarlin: Why is this change being made? We definitely want to be collecting coverage using the self-checking version in case the self-checking version of the macros changes alignment or something like that. This can hide subtle bugs.

framework/src/act/build_plan.py
```diff
@@ -231,20 +231,25 @@ def gen_rvvi_tasks(
...
     if not fast and config.objdump_exe is not None:
-        objdump_orig_file = Path(f"{elf}.objdump")
+        elf_dir = base_dir / "elfs"
+        final_elf = elf_dir / test_name.with_suffix(".elf")
+        objdump_orig_file = Path(f"{final_elf}.objdump")
         tasks.append(
             BuildTask(
                 outputs=(objdump_link,),
-                deps=(elf,),
+                deps=(final_elf,),
```
jordancarlin: We already were symlinking the final elf objdump. Why is this change needed?

generators/coverage/src/covergroupgen/templates/vector/cp_custom_ffLS.sv
```diff
@@ -12,7 +12,7 @@
         bins unmasked = {1'b0};
     }
 
-    rs1_at_fault_addr: coverpoint (unsigned'(ins.current.rs1_val) == `ACCESS_FAULT_ADDRESS) {
+    rs1_at_fault_addr: coverpoint (unsigned'(ins.current.rs1_val) == `RVMODEL_ACCESS_FAULT_ADDRESS) {
```
jordancarlin: Anywhere that uses RVMODEL_ACCESS_FAULT_ADDRESS needs to be in an `` `ifdef RVMODEL_ACCESS_FAULT_ADDRESS `` because not all DUTs have addresses that cause access faults. Same applies to the tests themselves.

generators/coverage/src/covergroupgen/templates/covergroup_sample_header_vector.sv
```diff
@@ -1,4 +1,5 @@
 function void ARCH_sample(int hart, int issue, ins_t ins);
 
-    if (get_csr_val(hart, issue, `SAMPLE_BEFORE, "vtype", "vsew") == EFFVSEW) begin
+    if (get_csr_val(hart, issue, `SAMPLE_BEFORE, "vtype", "vsew") == EFFVSEW ||
+        get_csr_val(hart, issue, `SAMPLE_BEFORE, "vtype", "vill") == 1) begin
```
jordancarlin, line 6: It is worth at least a comment explaining why you want to sample when `vill` is set. Seems like we'd want to skip that for most of the coverpoints?

generators/coverage/src/covergroupgen/generate.py
jordancarlin, line 1: Not clear to me why any of the changes in this file are needed. It seems like this should work fine without the `` `ifdef ``, we would just never see that instruction so it wouldn't be sampled.

generators/testgen/scripts/vector-testgen-unpriv.py
```diff
@@ -99,12 +100,12 @@
...
-  "vlseg4e32.v",
-  "vsseg3e64.v",
-  "vsseg3e32.v",
-  "vwredusum.vs"
+  # "vlseg3e32ff.v",  # resolved 2026-04-08: Spike PASS on all SEW/RV variants
+  # "vlseg3e32.v",    # resolved 2026-04-08: Spike PASS on all SEW/RV variants
+  # "vlseg4e32.v",    # resolved 2026-04-08: Spike PASS on all SEW/RV variants
+  # "vsseg3e64.v",    # Sail/Spike disagree on masking (2026-04-08) — Sail runs fine, only Spike comparison fails. NOT a coverage blocker.
+  # "vsseg3e32.v",    # Sail/Spike disagree on masking (2026-04-08) — Sail runs fine, only Spike comparison fails. NOT a coverage blocker.
+  # "vwredsum.vs",    # resolved 2026-04-08: Spike PASS on all SEW/RV variants (was typo: vwredusum.vs)
```
jordancarlin: Can we just delete this list if everything in it is commented out?

generators/testgen/scripts/vector_testgen_common.py
```diff
@@ -1767,11 +1795,19 @@ def writeVecTest(instruction, cp, vd, sew, testline, *scalar_registers_used, tes
 
     writeLine(testline)
 
+    # Restore valid vtype after force_vill test so signature save ops don't trap
+    if force_vill:
+      lmulflag = getLmulFlag(lmul)
+      writeLine(f"vsetivli x0, 1, e{sew}, m{lmulflag}, tu, mu", "# Restore valid vtype after vill test")
```
jordancarlin, line 1913: Is this something that we should be checking for in the signature code? @georgiatai

Makefile
```diff
@@ -120,7 +120,8 @@ elfs: tests
 		$(if $(FAST),--fast) \
 		$(if $(VERBOSE),--verbose) \
 		$(if $(COVERAGE),--coverage) \
-		$(if $(COVERAGE),--coverage-simulator $(COVERAGE_SIMULATOR))
+		$(if $(COVERAGE),--coverage-simulator $(COVERAGE_SIMULATOR)) \
```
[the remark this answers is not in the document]
jordancarlin: Please fix this. Legitimate find by Copilot.

jordancarlin (changes requested):

config/sail/sail-rv32-max/rvtest_config.svh
jacassidy: Sail currently doesnt

generators/coverage/src/covergroupgen/templates/covergroup_sample_header_vector.sv
jacassidy: for most we could skip it but vill being set in unpriv is pretty uncommon, its only when we want to test instructions explicitly still function while vill is set, and then for priv we will want to sample even when vill is set. The only issue I can see this causing is slowing down debugging if a ton of illegal vtype tests are generated which is super unlikely at this point

generators/coverage/src/covergroupgen/templates/vector/cp_custom_ffLS.sv
```diff
@@ -2,20 +2,54 @@
...
+    //
+    // Strategy: mask OFF element 0 (v0 bit 0 = 0) so element 0 does NOT access
+    // memory. Element 1+ are active and access the fault address region, triggering
+    // the vl-trimming behavior without trapping.
+    //
+    // NOTE for implementers: RVMODEL_ACCESS_FAULT_ADDRESS must point to a region
+    // where at least ceil(VLEN*LMUL/8) bytes are faulting (i.e. the entire page
+    // containing that address should be unmapped). With LMUL=2, the test accesses
+    // up to 2*VLEN/8 bytes starting from the fault address base. For typical
+    // implementations where address 0 is unmapped, the entire first page suffices.
```
jordancarlin: This note should not be in the coverpoint template. No-one will see it here. It should go in the CTP/README where we describe RVMODEL_ACCESS_FAULT_ADDRESS.

generators/coverage/src/covergroupgen/generate.py
```diff
@@ -513,5 +577,5 @@ def generate_covergroups(testplan_dir: Path, output_dir: Path, extensions: str =
 
     templates = read_covergroup_templates()
     write_covergroups(test_plans, templates, output_dir)
-    write_coverage_headers(all_test_plans, output_dir, templates)
+    write_coverage_headers(test_plans, output_dir, templates)
```
jordancarlin: Why is this being reverted?

jordancarlin (changes requested):

generators/coverage/src/covergroupgen/generate.py
jacassidy: thats probably what has been causing my weird diffs thats annoying

jordancarlin (approved):

PR #1305, opened 2026-04-18
Added RVA23 coverage and test generators

Ammarahwakeel:


generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+    base_reg, tmp_reg, val_reg, envcfg_reg = test_data.int_regs.get_registers(4, exclude_regs=[0])
+
+    lines = [
+        comment_banner(
+            coverpoint,
+            "Zic64b + Zicboz (RVA23U64): 64-byte cache blocks.\n"
+            "Test cbo.zero (U-mode) at offsets 0..60 & 64 on 128B all-1s buffer; verify via signature.",
+        ),
+        "",
+        "#if __riscv_xlen == 64",
```
jordancarlin: What is the point of restricting this test to rv64 only? Seems like we could run it just as easily on an rv32 core.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+        ]
+    )
+
+    test_data.int_regs.return_registers([base_reg, tmp_reg, val_reg, envcfg_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "Zic64bzicboz",
+    required_extensions=["I", "Zicsr", "Zicboz", "Zic64b", "U"],
```
jordancarlin: Why does this test need U mode? Seems like it could work just as well it any other mode.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+_SAMPLE_OFFSETS: list[int] = [0, 60, 64, 124]
+
+
+def _generate_zic64b_tests(test_data: TestData) -> list[str]:
+    """Generate cp_zic64bzicboz tests for Zic64bZicboz (U-mode only)."""
+
+    covergroup = "Zic64bzicboz_cg"
+    coverpoint = "cp_zic64bzicboz"
+
+    base_reg, tmp_reg, val_reg, envcfg_reg = test_data.int_regs.get_registers(4, exclude_regs=[0])
```
jordancarlin: No need for `exclude_regs=[0]`. It is automatically excluded for priv tests.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+# priv/extensions/Za64rs.py
+#
+# Za64rs privileged extension test generator.
+# ammarahwakeel9@gmail.com UET (April 2026)
+# SPDX-License-Identifier: Apache-2.0
+##################################
+
+"""Za64rs extension test generator.
+Reservation sets are contiguous, naturally aligned, and at most 64 bytes.
+RVA23U64 profile requires LR/SC to execute in U-mode.
```
jordancarlin: If this is just trying to test the Za64rs extension, it should not require U mode. If it is trying to test Za64rs specifically in U mode, the test should be renamed. I think it should just test Za64rs on its own and we are planning to make the default priv mode for tests be U-mode, so it will get that part for free.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    test_data.int_regs.return_registers([save_reg, val_reg, check_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "Ssu64xl",
+    required_extensions=["S", "Zicsr", "Ssu64xl"],
+    march_extensions=["S", "Zicsr"],
```
jordancarlin: Do we need S in the march string? I thought the compiler usually rejects S in the march string.

jordancarlin (changes requested):
I haven't reviewed it in detail yet, but here are a few initial comments. Also, if we are adding Ss1p13 tests, it might make sense to add Sm1p13 tests at the same time.

Ammarahwakeel:
> I haven't reviewed it in detail yet, but here are a few initial comments. Also, if we are adding Ss1p13 tests, it might make sense to add Sm1p13 tests at the same time.

Yes, I have worked on Sm1p13  too, I just have  not pushed it here yet.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
Ammarahwakeel: As these extensions fall under RVA23U64 profile, so i just restricted these tests to be defined for 64-bit user mode to stay aligned with profile.
But of course the test logic itself is not limited to these constraints.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
Ammarahwakeel: Again , I found these Zic64b + Zicboz extension under RVA23U64 profile.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
Ammarahwakeel: That makes sense. The intent of this test is to validate the Za64rs extension itself, not specifically its behavior in U-mode.

I had initially restricted it to U-mode to align with the RVA23U64 profile, but I agree that this is not a requirement for testing Za64rs functionality.

Should I Update all these tests to not be restricted for 64 bit user mode environment ?

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
Ammarahwakeel: Hmm, that's interesting. I did not see any errors. So I think it might be getting accepted or possibly ignored.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
jordancarlin: I think so. Someone could implement the Za64rs extension on an M-mode only processor and we should support testing that.

coverpoints/priv/S_coverage.svh
```diff
@@ -104,6 +114,24 @@ covergroup S_sstatus_cg with function sample(ins_t ins);
     // main coverpoints
     cp_sstatus_sd_write: cross priv_mode_s, csrrw, sstatus, cp_sstatus_sd, cp_sstatus_fs, cp_sstatus_vs, cp_sstatus_xs;
 
+    `ifdef SS1P13_SUPPORTED
```
davidharrishmc: Let's group these coverpoints with cp_ss1p13_scause in one `ifdef.  Indent within the ifdef for readability.

coverpoints/priv/S_coverage.svh
```diff
@@ -80,6 +80,16 @@ covergroup S_scause_cg with function sample(ins_t ins);
     // main coverpoints
     cp_scause_write_exception: cross priv_mode_s, csrrw, scause, scause_exception_values, scause_exception; // CSR write of scause in S mode with interesting values
     cp_scause_write_interrupt: cross priv_mode_s, csrrw, scause, scause_interrupt_values, scause_interrupt; // CSR write of scause in S mode with interesting values
+
+    `ifdef SS1P13_SUPPORTED
+
+    scause_ss1p13_values: coverpoint ins.current.rs1_val[XLEN-2:0] {
+        bins b_18_software_check = {18};
```
davidharrishmc: This is already in cp_scause_write_exception.  We can delete this `ifdef section and just put a comment in the testplan that it is covered by cp_scause_write_exception.

coverpoints/priv/S_coverage.svh
```diff
@@ -104,6 +114,24 @@ covergroup S_sstatus_cg with function sample(ins_t ins);
     // main coverpoints
     cp_sstatus_sd_write: cross priv_mode_s, csrrw, sstatus, cp_sstatus_sd, cp_sstatus_fs, cp_sstatus_vs, cp_sstatus_xs;
 
+    `ifdef SS1P13_SUPPORTED
+
+    sstatus_uxl_after: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "sstatus", "uxl") {
```
davidharrishmc: This coverpoint depends on uxl actually getting the value of 01.  There's no reason to be sure that the DUT will allow this value; for many DUTs, it will be hardwired to 10.  We just need to attempt to write sstatus.UXL to 01 and 10 in the coverpoint, and read back sststaus in the test.

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+        bins offset_24 = {7'd24};
+        bins offset_28 = {7'd28};
+        bins offset_32 = {7'd32};
+        bins offset_36 = {7'd36};
+        bins offset_40 = {7'd40};
+        bins offset_44 = {7'd44};
+        bins offset_48 = {7'd48};
+        bins offset_52 = {7'd52};
+        bins offset_56 = {7'd56};
+        bins offset_60 = {7'd60};
```
davidharrishmc: Add offset of 64

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+
+
+`define COVER_ZA64RS
+
+covergroup Za64rs_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+    `ifdef XLEN64
```
davidharrishmc: Shouldn't be limited to xlen64.  This extension could apply to any glen

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+        bins offset_64 = {7'd64};
+    }
+
+    sc_fail: coverpoint ins.current.rd_val {
+        bins fail = {[1:$]};
+    }
+
+    cp_za64rs_lr:      cross priv_mode_u, lr_w_instr, lr_w_base_aligned;
+    cp_za64rs_success: cross priv_mode_u, sc_w_instr, sc_offset_success;
+    cp_za64rs_fail:    cross priv_mode_u, sc_w_instr, sc_offset_fail, sc_fail;
```
davidharrishmc: No need for this coverpoint.

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+
+    sc_offset_fail: coverpoint ins.current.rs1_val[6:0] {
+        bins offset_64 = {7'd64};
+    }
+
+    sc_fail: coverpoint ins.current.rd_val {
+        bins fail = {[1:$]};
+    }
+
+    cp_za64rs_lr:      cross priv_mode_u, lr_w_instr, lr_w_base_aligned;
```
davidharrishmc: Delete this coverpoint.

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+`define COVER_ZA64RS
+
+covergroup Za64rs_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+    `ifdef XLEN64
+
+    lr_w_instr: coverpoint ins.current.insn {
```
davidharrishmc: change to ins.prev

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+        bins offset_36 = {7'd36};
+        bins offset_40 = {7'd40};
+        bins offset_44 = {7'd44};
+        bins offset_48 = {7'd48};
+        bins offset_52 = {7'd52};
+        bins offset_56 = {7'd56};
+        bins offset_60 = {7'd60};
+    }
+
+    sc_offset_fail: coverpoint ins.current.rs1_val[6:0] {
```
davidharrishmc: delete

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+        bins offset_64 = {7'd64};
+    }
+
+    sc_fail: coverpoint ins.current.rd_val {
```
davidharrishmc: delete

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+
+    lr_w_instr: coverpoint ins.current.insn {
+        wildcard bins lr_w = {LR_W};
+    }
+
+    sc_w_instr: coverpoint ins.current.insn {
+        wildcard bins sc_w = {SC_W};
+    }
+
+    lr_w_base_aligned: coverpoint ins.current.rs1_val[5:0] {
```
davidharrishmc: prev

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+    sc_offset_fail: coverpoint ins.current.rs1_val[6:0] {
+        bins offset_64 = {7'd64};
+    }
+
+    sc_fail: coverpoint ins.current.rd_val {
+        bins fail = {[1:$]};
+    }
+
+    cp_za64rs_lr:      cross priv_mode_u, lr_w_instr, lr_w_base_aligned;
+    cp_za64rs_success: cross priv_mode_u, sc_w_instr, sc_offset_success;
```
davidharrishmc: rename to cp_za64rs to match testplan.  Delete priv_mode_u because it applies to any priv mode.  Add lr_w_instr and lr_w_base_aligned to the coverpoint.

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,71 @@
...
+
+    sc_w_instr: coverpoint ins.current.insn {
+        wildcard bins sc_w = {SC_W};
+    }
+
+    lr_w_base_aligned: coverpoint ins.current.rs1_val[5:0] {
+        bins aligned_64 = {6'd0};
+    }
+
+    sc_offset_success: coverpoint ins.current.rs1_val[6:0] {
```
davidharrishmc: Delete _success.  Test both succeed and fail in one.  Anyway, reservation set could be < 64 bytes so these aren't guaranteed to succeed.

coverpoints/priv/Zic64bzicboz_coverage.svh
```diff
@@ -0,0 +1,53 @@
...
+//
+// Description: Coverage for RVA23U64 profile - Zic64bzicboz extension
+///////////////////////////////////////////////
+
+`define COVER_ZIC64BZICBOZ
+
+covergroup Zic64bzicboz_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+    `ifdef XLEN64
```
davidharrishmc: Not needed.  applies to all xlen.

coverpoints/priv/Zic64bzicboz_coverage.svh
```diff
@@ -0,0 +1,53 @@
...
+    `ifdef ZICBOZ_SUPPORTED
```
davidharrishmc: Not needed.  Checked by test.

coverpoints/priv/Zic64bzicboz_coverage.svh
```diff
@@ -0,0 +1,53 @@
...
+        bins offset_40 = {7'd40};
+        bins offset_44 = {7'd44};
+        bins offset_48 = {7'd48};
+        bins offset_52 = {7'd52};
+        bins offset_56 = {7'd56};
+        bins offset_60 = {7'd60};
+        bins offset_64 = {7'd64};
+    }
+
+    cp_zic64bzicboz: cross cbo_zero, cbo_zero_offset, priv_mode_u;
```
davidharrishmc: no priv_mode_u

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -74,6 +74,40 @@ def _generate_scause_tests(test_data: TestData) -> list[str]:
 
     lines.append(f"\nCSRW(scause, x{save_reg})       # restore CSR")
 
+    lines.extend(
```
davidharrishmc: Already tested; delete this new code.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -126,7 +160,52 @@ def _generate_sstatus_sd_tests(test_data: TestData) -> list[str]:
...
+                "cp_sxlen_ge_uxlen",
+                "Ss1p13: from S-mode attempt to set sstatus.UXL = 1 and UXL = 2.\n"
+                "UXL=2 must be silently rejected when SXLEN=32 (UXLEN <= SXLEN).",
+            ),
+            f"CSRR(x{save_reg}, sstatus)        # save sstatus before UXL tests",
+            "",
+            "# Clear UXL field [33:32] to start from a known state",
+            f"LI(x{reg2}, 3)                    # x{reg2} = 0b11",
+            f"slli x{reg2}, x{reg2}, 32         # x{reg2} = UXL mask (bits [33:32])",
+            f"CSRC(sstatus, x{reg2})             # clear UXL bits",
```
davidharrishmc: I don't think this works.  00 is not a legal UXL value, so the WARL field won't clear.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -126,7 +160,52 @@ def _generate_sstatus_sd_tests(test_data: TestData) -> list[str]:
...
+
+    coverpoint = "cp_sxlen_ge_uxlen"
+    for uxl, label in ((1, "uxlen32"), (2, "uxlen64")):
+        lines.extend(
+            [
+                "",
+                f"# Testcase: Ss1p13 attempt to set sstatus.UXL = {uxl} ({label})",
+                f"LI(x{check_reg}, {uxl})            # UXL value to attempt",
+                f"slli x{check_reg}, x{check_reg}, 32 # shift into UXL field [33:32]",
+                f"CSRS(sstatus, x{check_reg})         # attempt to set UXL={uxl}",
```
davidharrishmc: Again, I don't think this works because it won't clear both bits.  I think you will need to read sstatus into a GPR, mask out bits 33:32 with an AND, set bit 33 or 32 with an OR, and CSRW back to sstatus.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -126,7 +160,52 @@ def _generate_sstatus_sd_tests(test_data: TestData) -> list[str]:
                             gen_csr_write_sigupd(check_reg, "sstatus", test_data),
                         ]
                     )
+
     lines.append(f"\nCSRW(sstatus, x{save_reg})    # restore CSR")
+
+    lines.extend(
+        [
+            "",
+            "#ifdef SS1P13_SUPPORTED",
```
davidharrishmc, line 136: This also relies on xlen=64.

coverpoints/priv/S_coverage.svh
davidharrishmc, line 108: This also depends on XLEN>32.  If XLEN=32, there is no UXL field.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -126,7 +160,52 @@ def _generate_sstatus_sd_tests(test_data: TestData) -> list[str]:
...
+    coverpoint = "cp_sxlen_ge_uxlen"
+    for uxl, label in ((1, "uxlen32"), (2, "uxlen64")):
+        lines.extend(
+            [
+                "",
+                f"# Testcase: Ss1p13 attempt to set sstatus.UXL = {uxl} ({label})",
+                f"LI(x{check_reg}, {uxl})            # UXL value to attempt",
+                f"slli x{check_reg}, x{check_reg}, 32 # shift into UXL field [33:32]",
+                f"CSRS(sstatus, x{check_reg})         # attempt to set UXL={uxl}",
+                test_data.add_testcase(f"uxl_attempt_{uxl}", coverpoint, covergroup),
```
davidharrishmc, line 158: test case should go right before the CSRW being tested, not right before the sigupd.

coverpoints/priv/Ssu64xl_coverage.svh
```diff
@@ -0,0 +1,34 @@
...
+//
+// Description: Coverage for RVA23 profile - Ssu64x extension
+//
+///////////////////////////////////////////////
+
+`define COVER_SSU64XL
+
+covergroup Ssu64xl_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+
```
davidharrishmc, line 17: Needs XLEN64

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+"""Ssu64xl privileged extension test generator."""
+
+from testgen.asm.helpers import comment_banner, write_sigupd
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_ssu64xl_tests(test_data: TestData) -> list[str]:
+    covergroup = "Ssu64xl_cg"
+    coverpoint = "cp_ssu64xl_UXLEN"
```
davidharrishmc: This coverpoint should match the name in the testplan.

coverpoints/priv/Ssu64xl_coverage.svh
```diff
@@ -0,0 +1,34 @@
...
+    sstatus_uxl: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "sstatus", "uxl") {
+        bins uxl_sstatus = {2'b10};
+    }
+
+    gpr_bit63: coverpoint ins.current.rd_val[63] {
+        bins bit63_set = {1'b1};
+    }
+
+    cp_ssu64xl_UXLEN: cross priv_mode_u, sstatus_uxl, gpr_bit63;
```
davidharrishmc: should be called cp_ssu64xl to match testplan.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+
+    save_reg, val_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        comment_banner(
+            "cp_ssu64xl_UXLEN",
```
davidharrishmc: Use f{coverpoint} so it automatically tracks coverpoint name.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
davidharrishmc: I agree these extensions are already in required, don't need to be in march.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+        ),
+        "#if __riscv_xlen == 64",
+    ]
+
+    lines.extend(
+        [
+            "RVTEST_GOTO_LOWER_MODE Smode",
+            f"CSRR(x{save_reg}, sstatus)",
+            f"LI(x{val_reg}, {3 << 32})",
+            f"CSRC(sstatus, x{val_reg})",  # clear UXL bits [33:32]
```
davidharrishmc: As above, can't write UXL this way because 00 is not a legal value.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+        ]
+    )
+
+    lines.extend(
+        [
+            "RVTEST_GOTO_MMODE",
+            "RVTEST_GOTO_LOWER_MODE Umode",
+            f"LI(x{val_reg}, 1)",
+            f"slli x{val_reg}, x{val_reg}, 63",
+            f"addi x{check_reg}, x{val_reg}, 0",  # rd_val[63]=1
```
davidharrishmc: Not needed.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+            test_data.add_testcase("uxlen64_gpr_bit63", coverpoint, covergroup),
```
davidharrishmc, line 50: Global change: test case goes right before the statement being tested.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+            f"LI(x{val_reg}, {2 << 32})",
+            f"CSRS(sstatus, x{val_reg})",  # set UXL=2 (UXLEN=64)
+        ]
+    )
+
+    lines.extend(
+        [
+            "RVTEST_GOTO_MMODE",
+            "RVTEST_GOTO_LOWER_MODE Umode",
+            f"LI(x{val_reg}, 1)",
```
davidharrishmc: Replace this with 
load_int_reg("value", check_reg, 0xFEDCBA9876543210, test_data)

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+    coverpoint = "cp_ssu64xl_UXLEN"
+
+    save_reg, val_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        comment_banner(
+            "cp_ssu64xl_UXLEN",
+            "sstatus.UXL=2 (UXLEN=64 for User mode)\nSet bit 63 in a GPR in U-mode and read it back.",
+        ),
+        "#if __riscv_xlen == 64",
```
davidharrishmc: No need for #if because SSu64xl implies RV64

generators/testgen/src/testgen/priv/extensions/Za64rs.py
davidharrishmc: Agreed not U-dependent.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+
+from testgen.asm.helpers import comment_banner, write_sigupd
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_za64rs_tests(test_data: TestData) -> list[str]:
+    covergroup = "Za64rs_cg"
+    coverpoint_success = "cp_za64rs_success"
```
davidharrishmc: adjust to match updated coverpoints.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+    coverpoint_fail = "cp_za64rs_fail"
+
+    base_reg, src_reg, dest_reg, temp_reg = test_data.int_regs.get_registers(4, exclude_regs=[0, 5, 6, 7, 9])
```
davidharrishmc: don't need these exclude_regs.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,79 @@
...
+from testgen.asm.helpers import comment_banner, write_sigupd
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+
+def _generate_ssu64xl_tests(test_data: TestData) -> list[str]:
+    covergroup = "Ssu64xl_cg"
+    coverpoint = "cp_ssu64xl_UXLEN"
+
+    save_reg, val_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
```
davidharrishmc: don't need the exclude_reg

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    lines = [
+        comment_banner(
+            "cp_za64rs",
+            "Za64rs: Reservation set size test",
+        ),
+        "#if __riscv_xlen == 64",
```
davidharrishmc: Don't have this #if.  Test should run on any xlen.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+        f"LI(x{src_reg}, 0xA5A5A5A5)",
+        "",
+    ]
+
+    for offset in range(0, 64, 4):
+        binname = f"offset_{offset}"
+        lines.extend(
+            [
+                f"# Testcase: offset={offset} — expect sc_success (within reservation set)",
+                "RVTEST_GOTO_LOWER_MODE Umode",
```
davidharrishmc: Delete.  These should run in any mode.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+                test_data.add_testcase(binname, coverpoint_success, covergroup),
+                f"addi x{temp_reg}, x{base_reg}, {offset}",
+                f"lr.w x{dest_reg}, (x{base_reg})",
+                f"sc.w x{dest_reg}, x{src_reg}, (x{temp_reg})",
+                "RVTEST_GOTO_MMODE",
```
davidharrishmc: Delete.  These should run in any mode.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+    base_reg, src_reg, dest_reg, temp_reg = test_data.int_regs.get_registers(4, exclude_regs=[0, 5, 6, 7, 9])
+
+    lines = [
+        comment_banner(
+            "cp_za64rs",
+            "Za64rs: Reservation set size test",
+        ),
+        "#if __riscv_xlen == 64",
+        "",
+        f"LA(x{base_reg}, scratch)",
```
davidharrishmc, line 30: scratch is already aligned.  No need to align it.

davidharrishmc: See RVTEST_DATA_BEGIN macro.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+        "",
+    ]
+
+    for offset in range(0, 64, 4):
+        binname = f"offset_{offset}"
+        lines.extend(
+            [
+                f"# Testcase: offset={offset} — expect sc_success (within reservation set)",
+                "RVTEST_GOTO_LOWER_MODE Umode",
+                test_data.add_testcase(binname, coverpoint_success, covergroup),
```
davidharrishmc: put test case right before sc

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+        f"LI(x{temp_reg}, 127)",
+        f"add x{base_reg}, x{base_reg}, x{temp_reg}",
+        f"not x{temp_reg}, x{temp_reg}",
+        f"and x{base_reg}, x{base_reg}, x{temp_reg}",
+        "",
+        f"LI(x{src_reg}, 0xA5A5A5A5)",
+        "",
+    ]
+
+    for offset in range(0, 64, 4):
```
davidharrishmc: 64->65 so all tests run in this loop

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+        binname = f"offset_{offset}"
+        lines.extend(
+            [
+                f"# Testcase: offset={offset} — expect sc_success (within reservation set)",
```
davidharrishmc: change comment about success.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+                f"addi x{temp_reg}, x{base_reg}, {offset}",
+                f"lr.w x{dest_reg}, (x{base_reg})",
+                f"sc.w x{dest_reg}, x{src_reg}, (x{temp_reg})",
+                "RVTEST_GOTO_MMODE",
+                write_sigupd(dest_reg, test_data),
+                "",
+            ]
+        )
+
+    lines.extend(
```
davidharrishmc: Delete this chunk.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    lines.append("#endif")
+
+    test_data.int_regs.return_registers([base_reg, src_reg, dest_reg, temp_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "Za64rs",
+    required_extensions=["Sm", "U", "I", "Zalrsc", "Za64rs", "Zicsr"],
```
davidharrishmc: Only need Zalrsc and Za64rs.

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,88 @@
...
+    march_extensions=["I", "Zalrsc", "Zicsr"],
```
davidharrishmc: don't need this.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
+##################################
+# priv/extensions/Ssu64xl.py
```
davidharrishmc: update name.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+#
+# Zic64bzicboz privileged extension test generator.
+# ammarahwakeel9@gmail.com UET (April 2026)
```
davidharrishmc: Give your name as well as email in comments.  Claim credit for your work!

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+
+_OFFSETS: list[int] = list(range(0, 61, 4)) + [64]
+_SAMPLE_OFFSETS: list[int] = [0, 60, 64, 124]
+
+
+def _generate_zic64b_tests(test_data: TestData) -> list[str]:
+    """Generate cp_zic64bzicboz tests for Zic64bZicboz (U-mode only)."""
+
+    covergroup = "Zic64bzicboz_cg"
+    coverpoint = "cp_zic64bzicboz"
```
davidharrishmc: Coverpoint should match testplan and coverage file name

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
davidharrishmc: Remove #if

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+    lines = [
+        comment_banner(
+            coverpoint,
+            "Zic64b + Zicboz (RVA23U64): 64-byte cache blocks.\n"
+            "Test cbo.zero (U-mode) at offsets 0..60 & 64 on 128B all-1s buffer; verify via signature.",
+        ),
+        "",
+        "#if __riscv_xlen == 64",
+        "",
+        "#ifdef ZICBOZ_SUPPORTED",
```
davidharrishmc: Not needed here.  Already in required_extensions.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
davidharrishmc: Should just be Zicboz, Zic64b.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+        "",
+    ]
+
+    lines.extend(
+        [
+            "RVTEST_GOTO_MMODE",
```
davidharrishmc: Delete this goto M-mode and setting up configuration registers.  Not necessary if you boot to M-mode.  And the new boot code that goes to lower privilege modes sets up envcfg as necessary.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+            "",
+            f"LI(x{envcfg_reg}, 128)               # bit 7 = CBZE for both CSRs",
+            f"csrw menvcfg, x{envcfg_reg}",
+            f"csrw senvcfg, x{envcfg_reg}",
+            "",
+            f"LA(x{base_reg}, scratch)",
```
davidharrishmc: scratch is already 256-byte aligned.  No need to align base_reg.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+            f"LI(x{tmp_reg}, 63)",
+            f"add  x{base_reg}, x{base_reg}, x{tmp_reg}",
+            f"not  x{tmp_reg}, x{tmp_reg}",
+            f"and  x{base_reg}, x{base_reg}, x{tmp_reg}    # base_reg now 64-byte aligned",
+            "",
+            "RVTEST_GOTO_LOWER_MODE Umode",
```
davidharrishmc: don't change priv modes

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+Zic64b : Cache blocks must be 64 bytes in size, naturally aligned.
+Zicboz : Cache-Block Zero instruction (cbo.zero).
+RVA23U64 profile: 64-bit only, user-mode execution environment.
+"""
+
+from testgen.asm.helpers import comment_banner, write_sigupd
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+_OFFSETS: list[int] = list(range(0, 61, 4)) + [64]
```
davidharrishmc: can this be range(0,65,4)?

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+            "RVTEST_GOTO_LOWER_MODE Umode",
+            "",
+        ]
+    )
+
+    for offset in _OFFSETS:
+        binname = f"offset_{offset}"
+
+        lines.extend(
+            [
```
davidharrishmc: Get rid of all this mode changing.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+
+        for word_off in range(0, 128, 4):
+            lines.append(f"sw   x{val_reg}, {word_off}(x{base_reg})")
+
+        lines.extend(
+            [
+                "",
+                f"# Step 2: rs1 = base + {offset}, execute cbo.zero",
+                f"LI(x{tmp_reg}, {offset})",
+                f"add  x{tmp_reg}, x{base_reg}, x{tmp_reg}   # tmp_reg = base + {offset}",
```
davidharrishmc: replace with addi instead of LI and add.

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+                    f"lw   x{val_reg}, {sample_off}(x{base_reg})",
+                    write_sigupd(val_reg, test_data),
+                ]
+            )
+
+        lines.append("")
+
+    lines.extend(
+        [
+            "RVTEST_GOTO_MMODE",
```
davidharrishmc: not needed

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+
+    test_data.int_regs.return_registers([base_reg, tmp_reg, val_reg, envcfg_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "Zic64bzicboz",
+    required_extensions=["I", "Zicsr", "Zicboz", "Zic64b", "U"],
+    march_extensions=["I", "Zicsr", "Zicboz"],
```
davidharrishmc: no march needed

davidharrishmc (changes requested):

coverpoints/priv/S_coverage.svh
```diff
@@ -104,6 +105,19 @@ covergroup S_sstatus_cg with function sample(ins_t ins);
     // main coverpoints
     cp_sstatus_sd_write: cross priv_mode_s, csrrw, sstatus, cp_sstatus_sd, cp_sstatus_fs, cp_sstatus_vs, cp_sstatus_xs;
 
+    `ifdef SS1P13_SUPPORTED
+    `ifdef XLEN64
```
davidharrishmc: Add tabs at each level of nesting (one tab here, two tabs on the coverpoint)

config/cores/cvw/ci.yaml
```diff
@@ -4,7 +4,7 @@
 # Requires Verilator and the CVW repository
 
 ci_enabled: true
-exclude_extensions: "Sm,S,InterruptsU,PMPSm,ExceptionsZalrsc,ExceptionsZaamo"
+exclude_extensions: "Sm,S,InterruptsU,PMPSm,ExceptionsZalrsc,ExceptionsZaamo,Za64rs"
```
davidharrishmc, line 8: Redesign Za64rs testplan to accept any reservation set that is contiguous, naturally aligned, maximum of 64 bytes.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -413,6 +414,40 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
             cp_mtimeh_write :   cross priv_mode_m, csrr,  timeh_csr; // assumes mtimeh has been written
         `endif
     `endif
+
+    `ifdef SM1P13_SUPPORTED
+
+    misa_b_bit: coverpoint ins.current.rs1_val[1] {
```
davidharrishmc: Indent inside `ifdefs.  Global change.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -413,6 +414,40 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+    mip_csr: coverpoint ins.current.insn[31:20] {
+        bins mip = {CSR_MIP};
+    }
+
+    mip_msip_after: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "msip")[0] {
+        bins msip_set   = {1'b1}; // mip.MSIP set after writing memory-mapped msip = 1
+        bins msip_clear = {1'b0}; // mip.MSIP clear after writing memory-mapped msip = 0
+    }
+
+    cp_misa_bv: cross priv_mode_m, misa, csrop, misa_b_bit, misa_v_bit;
```
davidharrishmc: This coverpoint requires all four combinations of setting and clearing the bits simultaneously.  Your test sets and clears b, and then sets and clears v, so I don't think you are getting full coverage.  Its would be fine to break this into two coverpoints to handle b and v independently.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -644,6 +644,143 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+
+    lines.extend(
+        [
+            "#ifdef RVMODEL_MSIP_ADDRESS",
+            f"LA(x{r_msipaddr}, RVMODEL_MSIP_ADDRESS)   # load address of memory-mapped msip register",
+            "",
+            "# Write 1 to msip (set MSIP) and check mip.MSIP is set",
+            f"LI(x{r_msip1}, 1)                         # value 1: assert msip",
+            f"SW x{r_msip1}, 0(x{r_msipaddr})           # write msip = 1 via memory-mapped I/O",
+            f"csrr x{r_msip2}, mip                      # read mip to flush pipeline",
```
davidharrishmc: what is this flush pipeline business?
The spec says 
When the memory-mapped msip register changes, it is guaranteed to be reflected
in mip.MSIP eventually, but not necessarily immediately. 

There is a parameter about interrupt latency used in the interrupt tests.  I think we need to wait that long here too.

Ammarahwakeel: oh like about flush pipeline i just wanted to add a comment for nop instruction, like to wait untill msip get its value, but I totally agree I will refer the macro from interrupt tests.

coverpoints/priv/S_coverage.svh
```diff
@@ -104,6 +105,19 @@ covergroup S_sstatus_cg with function sample(ins_t ins);
     // main coverpoints
     cp_sstatus_sd_write: cross priv_mode_s, csrrw, sstatus, cp_sstatus_sd, cp_sstatus_fs, cp_sstatus_vs, cp_sstatus_xs;
 
+    `ifdef SS1P13_SUPPORTED
+     `ifdef XLEN64
```
jordancarlin: We use 4 spaces of indentation in coverpoints, not 2. Look at the rest of the file and try to be consistent with that style.

coverpoints/priv/S_coverage.svh
```diff
@@ -104,6 +105,19 @@ covergroup S_sstatus_cg with function sample(ins_t ins);
...
 
+    `ifdef SS1P13_SUPPORTED
+     `ifdef XLEN64
+
+      uxl_write_attempt: coverpoint ins.current.rs1_val[33:32] {
+        bins attempt_1 = {2'b01};
+        bins attempt_2 = {2'b10};
+      }
+
+      cp_sxlen_ge_uxlen: cross priv_mode_s, csrrs, sstatus, uxl_write_attempt, sstatus_uxl_after;
```
jordancarlin: A lot of the coverpoints that are defined as part of this cross (like csrrs) are not defined in this covergroup. How is this working?

coverpoints/priv/Sm_coverage.svh
```diff
@@ -413,6 +414,42 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        bins v_set   = {1'b1};
+        bins v_clear = {1'b0};
+     }
+
+     mip_csr: coverpoint ins.current.insn[31:20] {
+        bins mip = {CSR_MIP};
+     }
+
+     mip_msip_after: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "msip")[0] {
+        bins msip_set   = {1'b1}; // mip.MSIP set after writing memory-mapped msip = 1
```
jordancarlin: The comment refers to the memory-mapped msip register, but the coverpoint does not look at the memory-mapped register, it looks directly at the CSR.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,77 @@
...
+        ]
+    )
+
+    test_data.int_regs.return_registers([save_reg, val_reg, check_reg, orig_reg])
+    return lines
+
+
+@add_priv_test_generator(
+    "Ssu64xl",
+    required_extensions=["S", "Zicsr", "Ssu64xl"],
```
jordancarlin: No need for Zicsr. It is required by S.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,77 @@
...
+            f"{coverpoint}",
+            "sstatus.UXL=2 (UXLEN=64 for User mode)\nSet bit 63 in a GPR in U-mode and read it back.",
+        ),
+    ]
+
+    lines.extend(
+        [
+            "RVTEST_GOTO_LOWER_MODE Smode",
+            f"CSRR(x{orig_reg}, sstatus)",
+            f"CSRR(x{save_reg}, sstatus)",
```
jordancarlin: The name save_reg is confusing. What is this saving? You are changing its value.

generators/testgen/src/testgen/priv/extensions/Ssu64xl.py
```diff
@@ -0,0 +1,77 @@
...
+            f"LI(x{val_reg}, {~(3 << 32) & 0xFFFFFFFFFFFFFFFF})",  # mask clears bits 33:32
+            f"and x{save_reg}, x{save_reg}, x{val_reg}",
+            f"LI(x{val_reg}, {2 << 32})",  # UXL=2 → bit 33 set
+            f"or x{save_reg}, x{save_reg}, x{val_reg}",
+            f"CSRW(sstatus, x{save_reg})",
```
jordancarlin: Should we store sstatus to the signature to ensure the UXL bits have the right value? Update the coverpoints accordingly too.

coverpoints/priv/Za64rs_coverage.svh
```diff
@@ -0,0 +1,57 @@
...
+
+    sc_w_instr: coverpoint ins.current.insn {
+        wildcard bins sc_w = {SC_W};
+    }
+
+    lr_w_base_aligned: coverpoint ins.prev.rs1_val[5:0] {
+        bins aligned_64 = {6'd0};
+    }
+
+    sc_offset: coverpoint ins.current.rs1_val[6:0] {
```
jordancarlin, line 28: There are more concise ways of expressing this using a range without needing to explicitly list every value. Look at other covergroups for an example.

coverpoints/priv/Zic64bzicboz_coverage.svh
```diff
@@ -0,0 +1,49 @@
...
+
+covergroup Zic64bzicboz_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+    cbo_zero: coverpoint ins.current.insn {
+        wildcard bins cbo_zero = {CBO_ZERO};
+    }
+
+    cbo_zero_offset: coverpoint ins.current.rs1_val[6:0] {
```
jordancarlin, line 21: There are more concise ways of expressing this using a range without needing to explicitly list every value.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -126,7 +126,49 @@ def _generate_sstatus_sd_tests(test_data: TestData) -> list[str]:
...
+                "",
+                f"# Testcase: Ss1p13 attempt to set sstatus.UXL = {uxl} ({label})",
+                f"CSRR(x{check_reg}, sstatus)                     # read current sstatus into GPR",
+                f"LI(x{reg2}, {~(3 << 32) & 0xFFFFFFFFFFFFFFFF})  # mask to clear UXL bits [33:32]",
+                f"and x{check_reg}, x{check_reg}, x{reg2}         # clear UXL bits [33:32]",
+                f"LI(x{reg2}, {uxl << 32})                        # UXL={uxl} shifted into position [33:32]",
+                f"or x{check_reg}, x{check_reg}, x{reg2}          # OR in desired UXL value",
+                test_data.add_testcase(f"uxl_attempt_{uxl}", coverpoint, covergroup),
+                f"CSRW(sstatus, x{check_reg})                     # attempt to set UXL={uxl}",
+                gen_csr_read_sigupd(check_reg, "sstatus", test_data),
```
jordancarlin: I think we have a CSR write sigupd function that you can use to combine this with the previous instruction.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -644,6 +644,145 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+    lines.extend(
+        [
+            "#ifdef RVMODEL_MSIP_ADDRESS",
+            f"LA(x{r_msipaddr}, RVMODEL_MSIP_ADDRESS)   # load address of memory-mapped msip register",
+            "",
+            "# Write 1 to msip (set MSIP) and check mip.MSIP is set",
+            f"LI(x{r_msip1}, 1)                         # value 1: assert msip",
+            f"SW x{r_msip1}, 0(x{r_msipaddr})           # write msip = 1 via memory-mapped I/O",
+            f"csrr x{r_msip2}, mip                      # read mip ",
+            "nop\nnop\nnop\nnop\nnop\nnop               # wait for msip write to propagate",
```
jordancarlin: Drop the nops now that you have RVTEST_IDLE_FOR_INTERRUPT.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -644,6 +644,145 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+    ######################################
+    lines.append(
+        comment_banner(
+            coverpoint,
+            "Sm1p13: write all 1s / all 0s to memory-mapped msip register.\n"
+            "Read back msip, wait, then read mip.MSIP; must reflect the written value.",
+        ),
+    )
+
+    r_msip1, r_msip2, r_msipaddr = test_data.int_regs.get_registers(3)
```
jordancarlin: Why do we need separate r_msip1 and r_msip2 registers? Seems like you could reuse the same register.

jordancarlin (dismissed):

coverpoints/priv/S_coverage.svh
Ammarahwakeel: oh , sorry. Made a mistake here, like i just looked for the coverpoints already present in the file but neglected the part of checking in what functions they are present. Let me fix this.

jordancarlin: Did you try running coverage? Because it seems like this current version would fail when trying to collect coverage.

Ammarahwakeel: Actually ,I did run this S test, and  was getting 100  percent coverage. So , I did not look at the logs file. Maybe it was skipping the whole SS1P13 guard because we have not a sperate  UDB with this name. Now, I will look into the log files too.

Ammarahwakeel: I’ll check the coverage logs and confirm whether these bins are actually being sampled or if the guard is skipping this section.

davidharrishmc:
@Ammarahwakeel : @jordancarlin pointed out that the sail.json has a "reservation_set_size_exp" item.  We just need to change that to match the DUT so we can pass the Za64rs tests.  We don't need the clever exploration of the size - no need to change your tests, just the config file.

Ammarahwakeel:
> @Ammarahwakeel : @jordancarlin pointed out that the sail.json has a "reservation_set_size_exp" item. We just need to change that to match the DUT so we can pass the Za64rs tests. We don't need the clever exploration of the size - no need to change your tests, just the config file.

yes, This is the same item, i was referring to earlier in my message. So, it has value of 3, making reservation set size=2^3=8, and that is why sail is saying to expect a success for sc.w when we have base address+ 4 (offset), while other metrices have reservation set size=4, as they are only passing  for offset+0.

But based on this comment (https://github.com/riscv-software-src/riscv-isa-sim/pull/2256#pullrequestreview-4027983124), SC can fail regardless of reservation set size. So would changing the "reservation_set_size_exp" =2   be good to go?

davidharrishmc:
Conveniently UDB defines a parameter called LRSC_FAIL_ON_NON_EXACT_LRSC

I don’t think Sail supports it yet, but I think we need it to agree with Spike.

Let’s check in the tests as they are, and put this parameter on this list of things Sail will need to support.

I’ll create a new tab on the Implementation Tracking spreadsheet <https://docs.google.com/spreadsheets/d/1EuG4LV1TQDx4wTzYDWEAvq0fwaWjNuxenwDARZmrN-E/edit?gid=1372897027#gid=1372897027> with parameters that we need.

jordancarlin:
@pmundkur and I discussed adding something similar to LRSC_FAIL_ON_NON_EXACT_LRSC to the Sail model. I haven't gotten around to opening an issue for it yet though.

pmundkur:
I'll add a config flag.

jordancarlin:
> I'll add a config flag.

Perfect. Thanks!

pmundkur:
See https://github.com/riscv/sail-riscv/pull/1675

config/cores/cvw/ci.yaml
davidharrishmc, line 8: Add comment
Reenable Za64rs when Sail supports UDB_LRSC_FAIL_ON_NON_EXACT_LRSC to match DUT

coverpoints/priv/S_coverage.svh
```diff
@@ -104,6 +105,23 @@ covergroup S_sstatus_cg with function sample(ins_t ins);
     // main coverpoints
     cp_sstatus_sd_write: cross priv_mode_s, csrrw, sstatus, cp_sstatus_sd, cp_sstatus_fs, cp_sstatus_vs, cp_sstatus_xs;
 
+    `ifdef SS1P13_SUPPORTED
+        `ifdef XLEN64
+            uxl_write_attempt: coverpoint ins.current.rs1_val[33:32] {
+                bins attempt_1 = {2'b01};
+                bins attempt_2 = {2'b10};
+            }
+            sstatus_uxl_after: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "sstatus", "uxl") {
```
davidharrishmc: This will only work if sstatus.UXL gets both values.  We don't care whether that is possible, only whether we have attempted.  So keep ixl_write_attempt, and drop sstatus_uxl_after.

An empty coverpoint checks all possibilities, rather than none, so it is equivalent to writing bins 1 and bins 0. in this case.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -413,6 +414,38 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
             cp_mtimeh_write :   cross priv_mode_m, csrr,  timeh_csr; // assumes mtimeh has been written
         `endif
     `endif
+
+    `ifdef SM1P13_SUPPORTED
+        misa_b_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "misa", "b")[0] {
```
davidharrishmc: Sorry for the bad advice on an earlier PR from Jordan or me.  We need to go back to the rs1_val approach like in S.  There are no guarantees whether misa.b and misa.v are writable.  All we can do is write/set/clear 0 and 1 and see what happens.  

So this line should be
misa_b_bit: coverpoint ins.current.rs1_val[1] { }

Similar for misa_v_bit.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -413,6 +414,38 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+            bins b_clear = {1'b0};
+        }
+        misa_v_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "misa", "v")[0] {
+            bins v_set   = {1'b1};
+            bins v_clear = {1'b0};
+        }
+        mip_csr: coverpoint ins.current.insn[31:20] {
+            bins mip = {CSR_MIP};
+        }
+        mip_msip_after: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "msip")[0] {
```
davidharrishmc: Realized mip.MSIP can be read-only zero, so there might not be a MSIP.

Our coverpoint should be along the line

`ifdef RVMODEL_MSIP_ADDRESS
  msip_address: coverpoint ins.current.rs1_val {
    bin msip = {RVMODEL_MSIP_ADDRESS};
  }
  msip_val: coverpoint ins.current.rs2_val {
    bin zero = {0};
    bin one = {1};
  }
  cp_msip: cross priv_mode_m, sw, msip_adddress, misp_val;
`endif

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -644,6 +644,147 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+            "Sm1p13: write all 1s / all 0s to memory-mapped msip register.\n"
+            "Read back msip, wait, then read mip.MSIP; must reflect the written value.",
+        ),
+    )
+
+    r_msip, r_msipaddr = test_data.int_regs.get_registers(2)
+
+    lines.extend(
+        [
+            "#ifdef RVMODEL_MSIP_ADDRESS",
```
davidharrishmc, line 722: Add RVMODEL_MSIP_ADDRESS to docs/ctp/src/abstraction.adoc (after RVMODEL_MTIME_ADDRESS) and to README.md (another section after timer macros).

Also make sure it is in the configs.

davidharrishmc (changes requested):

coverpoints/priv/Sm_coverage.svh
Ammarahwakeel: Noted, should I  also skip the part in python generator where I verify mip.MSIP reflects the written value, or keep it as is?

davidharrishmc: Your python test reads mip.MSIP and verifies that it matches the reference model.  It doesn't actually check that it matches the value in the memory-mapped address.  So your Python test is good.  As long as the reference model is configured to match the DUT, it will produce the right answer, even if that is to hardwire mip.MSIP to 0.

generators/testgen/src/testgen/priv/extensions/Sm.py
Ammarahwakeel: In the config files I see MSIP_ADDRESS already defined, should I rename it to RVMODEL_MSIP_ADDRESS directly, or keep it and add RVMODEL_MSIP_ADDRESS as an alias?

generators/testgen/src/testgen/priv/extensions/Za64rs.py
```diff
@@ -0,0 +1,61 @@
...
+    ]
+
+    for offset in range(0, 65, 4):
+        binname = f"offset_{offset}"
+        lines.extend(
+            [
+                f"# Za64rs: sc.w at offset={offset} within reservation set",
+                f"addi x{temp_reg}, x{base_reg}, {offset}",
+                f"lr.w x{dest_reg}, (x{base_reg})",
+                test_data.add_testcase(binname, coverpoint, covergroup),
```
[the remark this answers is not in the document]
Ammarahwakeel: I saw the rvtest_setup.h file , and it says .align 8 meaning the scratch is already 2 ^ 8 = 256 byte aligned. So, I don't think so we should apply this change .

generators/testgen/src/testgen/priv/extensions/Zic64bzicboz.py
```diff
@@ -0,0 +1,82 @@
...
+    base_reg, tmp_reg, val_reg = test_data.int_regs.get_registers(3)
+
+    lines = [
+        comment_banner(
+            f"{coverpoint}",
+            "Zic64b + Zicboz: 64-byte cache blocks.\n"
+            "Test cbo.zero at offsets 0..64 on 128B all-1s buffer; verify via signature.",
+        ),
+        "",
+        f"LA(x{base_reg}, scratch)",
```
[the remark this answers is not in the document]
Ammarahwakeel: Scratch is already 256 byte aligned.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: Let's change it because we are now asking the user to define the value for their DUT.  Also, we should indicate that this is only required for Sm version 1.13 and above.

Ammarahwakeel:
Prof. @davidharrishmc , just added RVMODEL_MSIP_ADDRESS macro across all config files, could you please review and let me know if the implementation is correct or if any changes are needed?

davidharrishmc (approved):

PR #1312, opened 2026-04-18
Remove lots of unused macros

jordancarlin:


jordancarlin:
Closing because of ongoing trap handler overhaul. Will revisit later if necessary.

PR #1319, opened 2026-04-19
ExceptionsSv, ExceptionsSvZaamo and ExceptionsSvZalrsc ACTs

Zain2050:


tests/env/rvtest_trap_handler.h
```diff
@@ -648,7 +648,7 @@
 
 #ifndef CAUSE_SPCL_GO2MMODE_OP // make sure this default can be overwritten (e.g. to illegal fetch addr)
     #define ALT_GOTO_M_CAUSE CAUSE_ILLEGAL_INSTRUCTION
-    #define ALT_GOTO_M_OP    .insn 0
+    #define ALT_GOTO_M_OP    .word 0
```
jordancarlin, line 651: What prompted this change?

jordancarlin (commented):
Looks like some of these tests are failing in CI right now. Can you investigate if it is all of the simulators or just one of them and add these to the EXCLUDED_EXTENSIONS lists as appropriate?

Zain2050:
These tests are probably failing due to difference of trap priorities between the models. In these tests, we have scenerios that can raise various types of traps according to the priorities defined in table 15 of the priv spec.  I'll investigate it soon.

tests/env/rvtest_trap_handler.h
Zain2050: These tests instantiate RVTEST_GOTO_DELEGATED_MMODE. Clang was assuming .insn 0 to be a compressed instruction and was raising this error
```
<instantiation>:9:11: error: compressed instructions are not allowed
      .insn 0
            ^
  <instantiation>:20:5: note: while in macro instantiation
      RVTEST_GOTO_DELEGATED_MMODE
```
That's why had to change it to .word

davidharrishmc (approved):

davidharrishmc (approved):

PR #1324, opened 2026-04-20
Boot to M-mode

davidharrishmc:
Start of PR to boot to M-mode.

Will follow up with S-mode where applicable and then continue to U-mode if it exists.

* Enhanced Zicsr tests to pick most suitable CSR
* Improved naming of tests/env files
* Eliminated m/s/htrap_routine
* Add U-mode and CONFORMING_SM_MODE to profiles where applicable
* Cleaned up required_extensions lists
* Deleted FP_TEST, PRIV_TEST 
* Reordered Makefile to put most interesting stuff first.
* Cleanup many places


docs/memory_map.md
```diff
@@ -67,7 +67,7 @@ Additional details on what each section contains and why each section is needed
 | `rvtest_data_end`                      | End of test specific data label.                                                                                                                                                            |
 | `begin_signature` / `rvtest_sig_begin` | Start of the signature region (aligned to 16 bytes).                                                                                                                                        |
 | _(signature data)_                     | Main signature region written by test code via `RVTEST_SIGUPD`.                                                                                                                             |
-| _(trap signature)_                     | Trap handler signature region (when `rvtest_mtrap_routine` is defined).                                                                                                                     |
+| _(trap signature)_                     | Trap handler signature region (when `CONFORMING_SM_SUPPORTED` is defined).                                                                                                                  |
```
jordancarlin: This will eventually be needed even without `CONFORMING_SM_SUPPORTED` because S mode tests will save to the trap signature and will run without `CONFORMING_SM_SUPPORTED`.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -79,6 +79,6 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     test_data.end_test_chunk()
 
     # Produce actual test file
-    extra_defines = [*get_priv_test_defines(testsuite), "#define RVTEST_PRIV_TEST"]
+    extra_defines = [*get_priv_test_defines(testsuite), ""]
```
jordancarlin: Can we drop the empty string?

```suggestion
    extra_defines = [*get_priv_test_defines(testsuite)]
```

generators/testgen/src/testgen/io/templates.py
```diff
@@ -193,29 +193,27 @@ def generate_march_string(ext_components: list[str], xlen: int) -> str:
...
+def format_params(params: list[str], ext_components: list[str]) -> str:
     """Format parameters for insertion into template."""
     if not params:
         return "# # no param constraints"  # Extra comment symbol necessary because YAML parser strips initial comment
     param_lines = ["params:"]
+    if any(
+        ext in ext_components for ext in ["Sm", "H", "S", "U"]
+    ):  # might need hack to require conforming Sm for all priv tests until nonconforming trap handler setup works
+        #    if any(ext in ext_components for ext in ["Sm"]):
+        param_lines.extend("#    CONFORMING_SM_REQUIRED")
```
jordancarlin: This needs to be reordered. Right now it will return `# # no param constraints"` if there are no params before checking for CONFORMING_SM_REQUIRED.

tests/env/rvtest_failure_code.h
```diff
@@ -56,7 +56,7 @@
         j failedtest_saveregs
 #endif
```
jordancarlin: This will eventually be needed even without `CONFORMING_SM_SUPPORTED` because S and U mode tests will trap and will run without `CONFORMING_SM_SUPPORTED`.

tests/env/failure_code.h
```diff
@@ -242,13 +242,10 @@
         sw x6, 260(DEFAULT_TEMP_REG)      # record failing_reg
 
         # Load bad FP value from scratch memory (written by FSREG in the sigupd macro)
-        # Use FP_LREG so we read exactly the CONFIG_FLEN bits FSREG stored,
-        # zero-extending on RV64+F-only where fsw wrote fewer bytes than LREG reads.
-        # See tests/env/utils.h for an explanation of CONFIG_FLEN and TEST_FLEN.
```
jordancarlin, line 466: Restore comment.

tests/env/rvtest_failure_code.h
```diff
@@ -242,13 +242,10 @@
...
-        # Use FP_LREG so we read exactly the CONFIG_FLEN bits FSREG stored,
-        # zero-extending on RV64+F-only where fsw wrote fewer bytes than LREG reads.
-        # See tests/env/utils.h for an explanation of CONFIG_FLEN and TEST_FLEN.
         la x6, scratch
-        FP_LREG x7, 0(x6)
+        LREG x7, 0(x6)
         SREG x7, 272(DEFAULT_TEMP_REG)    # failing_value (lower/only)
-    #if CONFIG_FLEN > XLEN
+    #if FLEN > XLEN
         LREG x7, REGWIDTH(x6)
```
jordancarlin, line 469: Revert changes to `FP_LREG` and `CONFIG_FLEN`.

tests/env/rvtest_failure_code.h
```diff
@@ -268,13 +265,13 @@
         # Load full expected FP value from signature
         LREG x7, 0(x6)
         SREG x7, 280(DEFAULT_TEMP_REG)    # expected_value (lower/only)
-    #if CONFIG_FLEN > XLEN
+    #if FLEN > XLEN
```
jordancarlin: Revert changes to `FP_LREG` and `CONFIG_FLEN`.

tests/env/rvtest_failure_code.h
```diff
@@ -397,13 +394,13 @@
...
-        # FP with CONFIG_FLEN > XLEN: combined hex "0xUPPER_LOWER"
+    #if defined(F_SUPPORTED) && FLEN > XLEN
+        # FP with FLEN > XLEN: combined hex "0xUPPER_LOWER"
         LREG a0, failing_value_upper
         LREG a1, failing_value
         jal failedtest_combined_hex_to_str
     #else
-        # FP with CONFIG_FLEN <= XLEN (or CONFIG_FLEN not defined): standard hex
+        # FP with FLEN <= XLEN (or FLEN not defined): standard hex
         LREG a0, failing_value
```
jordancarlin, line 1005: Revert changes to `FP_LREG` and `CONFIG_FLEN`.

tests/env/rvtest_failure_code.h
```diff
@@ -424,13 +421,13 @@
...
-        # FP with CONFIG_FLEN > XLEN: combined hex "0xUPPER_LOWER"
+    #if defined(F_SUPPORTED) && FLEN > XLEN
+        # FP with FLEN > XLEN: combined hex "0xUPPER_LOWER"
         LREG a0, expected_value_upper
         LREG a1, expected_value
         jal failedtest_combined_hex_to_str
     #else
-        # FP with CONFIG_FLEN <= XLEN (or CONFIG_FLEN not defined): standard hex
+        # FP with FLEN <= XLEN (or FLEN not defined): standard hex
         LREG a0, expected_value
```
jordancarlin, line 1032: Revert changes to `FP_LREG` and `CONFIG_FLEN`.

tests/env/rvtest_failure_code.h
```diff
@@ -445,7 +442,7 @@
         LA(a0, ascii_buffer)
         call rvmodel_io_write_str
 
-#ifdef rvtest_mtrap_routine
+#ifdef CONFORMING_SM_SUPPORTED
     failedtest_report_traphandler:
```
jordancarlin: This will eventually be needed even without `CONFORMING_SM_SUPPORTED` because S and U mode tests will trap and will run without `CONFORMING_SM_SUPPORTED`.

tests/env/rvtest_failure_code.h
```diff
@@ -558,7 +555,7 @@
         ret
 
 
-#if defined(RVTEST_FP) && CONFIG_FLEN > XLEN
+#if defined(F_SUPPORTED) && FLEN > XLEN
```
jordancarlin: Revert changes to `FP_LREG` and `CONFIG_FLEN`.

tests/env/rvtest_failure_code.h
```diff
@@ -630,7 +627,7 @@
         .fill 2, 4, 0xfeedf00dbaaaaaad
     failure_string_ptr:
         .fill 2, 4, 0xfeedf00dbaaaaaad
-#if defined(RVTEST_FP) && CONFIG_FLEN > XLEN
+#if defined(F_SUPPORTED) && FLEN > XLEN
```
jordancarlin: Revert changes to `FP_LREG` and `CONFIG_FLEN`.

tests/env/rvtest_setup.h
```diff
@@ -166,7 +134,9 @@
       LREG    T1, 0(T1)               // Trap signature pointer
       LA(     T2, mtrap_sigptr)       // Base address of trap signature region
       sub     T1, T1, T2              // Calculate offset
-      RVTEST_SIGUPD(x2, x5, x4, T1, check_trap_sig_offset, trap_sig_offset_mismatch)
+      // *** the following code uses x2 as a pointer.
+      // *** x2 has been modified by unpriv tests and thus causes a store access fault
+      // RVTEST_SIGUPD(x2, x5, x4, T1, check_trap_sig_offset, trap_sig_offset_mismatch)
```
jordancarlin: This needs to be fixed in the unpriv test generator to restore the sig pointer to x2 at the end of the test.

tests/env/rvtest_setup.h
```diff
@@ -374,7 +356,7 @@
         .fill SIGUPD_COUNT*(SIG_STRIDE>>2),4,0xdeadbeef
 
       // Signature region for trap handlers
-      #ifdef rvtest_mtrap_routine
+      #ifdef CONFORMING_SM_SUPPORTED
         tsig_begin_canary:
           TRAP_CANARY
```
jordancarlin: This will eventually be needed even without `CONFORMING_SM_SUPPORTED` because S mode tests will save to the trap signature and will run without `CONFORMING_SM_SUPPORTED`.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,302 @@
...
+    #ifdef CONFORMING_SM_SUPPORTED
+
+
+    // Disable interrupts
+      csrw mie, zero
+      csrw mip, zero
+
+      // disable trap delegation
+      csrw mideleg, zero  // don't delegate interrupts (until S-mode handler is set up)
+      csrw medeleg, zero  // don't delegate exceptions (until S-mode handler is set up)
```
jordancarlin: Needs to be gated on `S_SUPPORTED`.
> In harts without S-mode, the `medeleg` and `mideleg` registers should not exist.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,302 @@
...
+      // menvcfg.LPE = 0: Disable Zicfilp landing pads
+      // menvcfg.FIOM = 0: Disable Fence of I/O Implies Memory
+      // menvcfg.CBZE = 1: Enable Zicboz cache block zero instructions
+      // menvcfg.CBCFE = 1: Enable Zicbom cache block clean/flush instructions
+      // menvcfg.CBIE = 11: Enable Zicbom cache block invalidate instructions to perform invalidate operation
+      li t0, MENVCFG_CBIE | MENVCFG_CBCFE | MENVCFG_CBZE
+      csrw menvcfg, t0
+      #if __riscv_xlen == 32
+        csrw menvcfgh, zero // Clear upper bits if they exist
+      #endif
```
jordancarlin: Needs to be gated on `U_SUPPORTED`.
> If U-mode is not supported, then registers menvcfg and menvcfgh do not exist.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,302 @@
...
+      #else    // RV32
+        csrw mstateen0h, zero
+        li t0, MSTATEEN0_JVT
+        csrw mstateen0, t0
+      #endif
+      #ifdef ZFINX_SUPPORTED
+        li t0, MSTATEEN0_FCSR
+        csrs mstateen0, t0 // Set mstateen0.FCSR
+        li t0,
+      #endif
```
jordancarlin: Gate on `SMSTATEEN_SUPPORTED`.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,302 @@
...
+        csrw mhpmevent23, zero
+        csrw mhpmevent24, zero
+        csrw mhpmevent25, zero
+        csrw mhpmevent26, zero
+        csrw mhpmevent27, zero
+        csrw mhpmevent28, zero
+        csrw mhpmevent29, zero
+        csrw mhpmevent30, zero
+        csrw mhpmevent31, zero
+      #endif
```
jordancarlin, line 575: Presumably these should all be `mhpmevent#h`.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,302 @@
...
+    #if defined(F_SUPPORTED) || defined(ZFINX_SUPPORTED)
+      li t0, MSTATUS_FS
+      csrs mstatus, t0 // Set FS to dirty to enable floating-point
+      csrw fcsr, zero // Initialize fcsr
+    #endif
+    #ifdef V_SUPPORTED
+      li t0, MSTATUS_VS
+      csrs mstatus, t0 // Set VS to dirty to enable vector
+      // csrr VLENB_CACHE, vlenb // carryover from RVTEST_V_ENABLE; delete when not needed anywhere
+    #endif
```
jordancarlin: Presumably this should only require the most minimal vector extension. All of the other vector extensions (including V) in turn depend on this one.
```suggestion
    #ifdef ZVE32X_SUPPORTED
      li t0, MSTATUS_VS
      csrs mstatus, t0 // Set VS to dirty to enable vector
      // csrr VLENB_CACHE, vlenb // carryover from RVTEST_V_ENABLE; delete when not needed anywhere
    #endif
```

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,302 @@
...
+    // Enable necessary state for access from lower privilege modes
+    // mstateen0.SE0 = 1: enable access to hststateen0, hstatene0h, ssstateen0
+    // mstateen0.ENVCFG = 1: enable access to henvcfg, henvcfgh, senvcfg
+    #if __riscv_xlen == 64
+      li t0, MSTATEEN_HSTATEEN | MSTATEEN0_HENVCFG
+      csrs mstateen0, t0  // Set these fields
+    #else    // RV32
+      li t0, MSTATEENH_HSTATEEN | MSTATEEN0H_HENVCFG
+      csrs mstateen0h, t0 // Set these fields
+    #endif
```
jordancarlin: Gate of `SMSTATEEN_SUPPORTED`.

jordancarlin (changes requested):
First pass feedback.

tests/env/rvtest_setup.h
```diff
@@ -135,24 +103,24 @@
   // Switch to M-mode
   // ***dh 4/8/26 is this still the right thing to do if there is no conforming M-mode?
   rvtest_code_end:
-    RVTEST_GOTO_MMODE // If only M-mode used by tests, this has no effect
+    #ifdef CONFORMING_SM_SUPPORTED
+      RVTEST_GOTO_MMODE // *** dh 4/18/26 should this still be here?
```
davidharrishmc: Test if necessary

tests/env/rvtest_setup.h
```diff
@@ -135,24 +103,24 @@
...
-        #ifdef rvtest_htrap_routine
           RVTEST_TRAP_EPILOG H        // actual h-mode prolog/epilog/handler code
         #endif
         RVTEST_TRAP_EPILOG S          // actual s-mode prolog/epilog/handler code
       #endif
       RVTEST_TRAP_EPILOG M            // actual m-mode prolog/epilog/handler code
     #endif
 
-  #ifdef rvtest_mtrap_routine
+  #ifdef CONFORMING_SM_SUPPORTED
```
davidharrishmc, line 98: Only run this if mtvec is nonzero

davidharrishmc: instead of checking conforming sm mode

tests/env/rvtest_setup.h
```diff
@@ -201,6 +171,20 @@
...
       RVMODEL_IO_INIT(T1, T2, T3)
     #endif
+    // always boot to at least M-mode
+    RVTEST_BOOT_TO_MMODE
+    #ifndef BOOT_TO_MMODE
+      // continue to a lower privilege mode depending on the type of test
+      #ifdef S_SUPPORTED
+        RVTEST_BOOT_TO_SMODE
+      #endif
+      #ifndef S_REQUIRED
```
davidharrishmc: commented better

davidharrishmc: should there be a boot to SMODE/UMODE

tests/env/rvtest_setup.h
```diff
@@ -218,7 +202,7 @@
     j . // Explicit non-returning tail if the macro returns (it should not)
 
   // ***DH 4/8/26 check this is proper gating
-  #ifdef rvtest_mtrap_routine
+  #ifdef CONFORMING_SM_SUPPORTED
```
davidharrishmc, line 188: Instantiate S-mode interrupts even if SM is not supported

tests/env/rvtest_setup.h
davidharrishmc: I think we can get rid of CONSORMING_SM_SUPPORTED entirely.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,304 @@
...
+      #endif
+
+      // initialize trap CSRs to known values
+      csrw mepc, zero
+      csrw mtval, zero
+      csrw mcause, zero
+
+      // Set up trap handlers for all modes
+      // *** this code needs a close review
+      // *** preferably move the S and H stuff to RVTEST_BOOT_TO_SMODE, but it is giving linker problems right now
```
davidharrishmc: Umer ok with keeping it here.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,304 @@
...
+        csrs mstateen0, t0 // Set mstateen0.FCSR
+        li t0,
+      #endif
+
+      // Enable all performance counters if they exist
+      // This is reserved if mcountinhibit is not implemented, and might trap or have unspecified behavior
+      //   *** define a UDB parameter to determine whether mcountinhibit is implemented
+      csrw mcountinhibit, zero
+
+      // Initialize counter event selectors to 0.  They must be implemented.
```
davidharrishmc, line 515: Update HPM tests to expect them to not count.

davidharrishmc: mhpmevent may be read-only zero.  Need a UDB parameter to know if it is safe to write to mhpmevent.

tests/env/rvtest_setup.h
```diff
@@ -455,3 +429,304 @@
...
+  #else
+    // Default implementation assumes conforming M-mode
+    // We are in M-mode now at initial boot time
+
+    // Set up trap handler for S-mode
+    // *** uncomment this when rvtest_strap_routine etc. is replaced with S/H_SUPPORTED
+    // RVTEST_TRAP_PROLOG S
+    // if Hypervisor supported, also set up HS and VS-mode trap handlers
+    // *** how does h differ from S?
+    #ifdef H_SUPPORTED
```
davidharrishmc: remove this.  leave in M-mode boot

tests/env/rvtest_trap_handler.h
```diff
@@ -355,15 +355,13 @@
 /**** PROLOG/HANDLER/EPILOG/SAVEAREA depending on test type & mode support ****/
 /******************************************************************************/
 .macro INSTANTIATE_MODE_MACRO MACRO_NAME
-  #ifdef rvtest_mtrap_routine
+  #ifdef CONFORMING_SM_SUPPORTED
```
davidharrishmc: May be able to remove CONFORMING_SM_MODE many places in this file because nonconforming will instantiate its own trap handler.  Check this works for RV32I with no M-mode.

tests/env/rvtest_trap_handler.h
```diff
@@ -355,15 +355,13 @@
 /**** PROLOG/HANDLER/EPILOG/SAVEAREA depending on test type & mode support ****/
```
davidharrishmc, line 358: Comment that users with nonconforming M-mode may have to rewrite this entire file to be compatible with their limitations.

tests/env/rvtest_trap_handler.h
```diff
@@ -355,15 +355,13 @@
...
 /******************************************************************************/
 .macro INSTANTIATE_MODE_MACRO MACRO_NAME
-  #ifdef rvtest_mtrap_routine
```
davidharrishmc, line 358: remove xtrap routines from all of Sv and PMP tests.

tests/env/rvtest_setup.h
davidharrishmc: I think ZVL32B is even more minimal, not requiring any particular ELENs.

tests/env/rvtest_setup.h
davidharrishmc: Later

tests/env/rvtest_setup.h
davidharrishmc: No, it may be WARL 0, but is always writable even if the write is ignored.

tests/env/rvtest_setup.h
davidharrishmc: I believe you've done that in a separate PR.

config/sail/sail-RVI20U32/sail.json
jordancarlin, line 1: Why are the reservation settings being changed here?

config/spike/spike-rv32-max/sail.json
```diff
@@ -271,7 +271,7 @@
       // This specifies both the size and alignment of the reservation set, specified as a power of 2.
       // It must be at least 2 on RV32 and 3 on RV64 to support Zalrsc.  It must not be more than
       // 6 for Za64rs and 7 for Za128rs.  In all cases, it must be less than or equal to 12.
-      "reservation_set_size_exp": 3,
+      "reservation_set_size_exp": 4,
```
jordancarlin: Why is this being changed if require_exact_reservation_address is set below?

docs/ctp/src/abstraction.adoc
```diff
@@ -1,6 +1,121 @@
...
+
+* Nevertheless, most users will implement a conforming M-mode on their DUT.  Therefore, the abstraction layer should provide default implementations of boot macros, trap handler, and SBI calls that assume a conforming M-mode, and users can override these implementations as needed for their DUT.
+
+* To reduce the risk of bugs from uninitialized state and the burden on test developers, the boot process places known values in all applicable GPRs and CSRs.
+
+The REQUIRED_EXTENSIONS list in each test implies the privilege mode into which the test boots.  If REQUIRED_EXTENSIONS includes ConformingSm, the test boots into M-mode. Otherwise, if REQUIRED_EXTENSIONS includes S, the test boots into S-mode.  This is the case even for hypervisor tests, which require S and boot into S (effectively HS)-mode, and then are responsible for setting up and switching to VS/VU mode if desired.  Otherwise, the test boots into U-mode if the DUT supports it, and M-mode (possibly nonconforming) if the DUT lacks U-mode.
```
jordancarlin:
```suggestion
The REQUIRED_EXTENSIONS list in each test implies the privilege mode into which the test boots.  If REQUIRED_EXTENSIONS includes Sm, the test boots into M-mode. Otherwise, if REQUIRED_EXTENSIONS includes S, the test boots into S-mode.  This is the case even for hypervisor tests, which require S and boot into S (effectively HS)-mode, and then are responsible for setting up and switching to VS/VU mode if desired.  Otherwise, the test boots into U-mode if the DUT supports it, and M-mode (possibly nonconforming) if the DUT lacks U-mode.
```

docs/ctp/src/abstraction.adoc
```diff
@@ -1,6 +1,121 @@
...
+
+The implication of these rules is that privileged tests that specify ConformingSm, S, or U will boot into those modes, respectively, and unprivileged tests will run in U-mode if supported and M-mode (possibly nonconforming) otherwise.
```
jordancarlin:
```suggestion
The implication of these rules is that privileged tests that specify Sm, S, or U will boot into those modes, respectively, and unprivileged tests will run in U-mode if supported and M-mode (possibly nonconforming) otherwise.
```

docs/ctp/src/abstraction.adoc
```diff
@@ -1,6 +1,121 @@
...
+* menvcfg (enable unpriv features if U supported)
+* mstateen0 (enable unpriv features if SMSTATEN_SUPPORTED)
+* mcountinhibit = 0 *** if exists
+* mhpmenvent3...31/h = 0
+* mseccfg (disable features if supported)
+* mnstatus (disable resumable NMI fields if SMRNMI_SUPPORTED)
+* pmpaddr0/pmpcfg0 if PMP_ENTRIES > 0
+* turn off virtual memory and perform sfence.vma to synchronize MMU if SV32_SUPPORTED or SV39_SUPPORTED
+
+Even if conforming M-mode is not supported or RVMODEL_BOOT_TO_MMODE is defined, the boot code still attempts to turn on mstatus.FS/VS if float or vector is supported, because these are needed for unprivileged tests, even on a RVI32 DUT that has no other M-mode features.
```
jordancarlin:
```suggestion
Even if conforming M-mode is not supported or RVMODEL_BOOT_TO_MMODE is defined, the boot code still attempts to turn on mstatus.FS/VS if float or vector is supported, because these are needed for unprivileged tests, even on a RV32I DUT that has no other M-mode features.
```

docs/ctp/src/abstraction.adoc
jordancarlin, line 1: This file (especially the beginning) is largely notes (many of which are done) and should probably be cleaned up at least a little bit before merging since this is part of the CTP.

generators/testgen/src/testgen/formatters/types/csr_type.py
```diff
@@ -13,6 +13,26 @@
 csr_config = InstructionTypeConfig(required_params={"rd", "rs1", "rs1val", "rs2", "rs2val"})
 
 
+def zicsr_acccess(instr_name: str, rd: int, rs1: int) -> str:
```
jordancarlin, line 16: Could we simplify this by defining TEST_CSR at the beginning of the test and then using that in all of the CSR ops? Right now this adds ifdefs to every testcase and makes the generated assembly a lot messier.

tests/env/rvtest_failure_code.h
```diff
@@ -377,13 +375,13 @@
     failedtest_saveresults:
         # Dispatch based on failure type
         lw x9, failure_type
-#ifdef RVTEST_FP
-        li x10, 1
+  #if defined(F_SUPPORTED) || defined(ZFINX_SUPPORTED)
+          li x10, 1
```
jordancarlin: Revert whitespace change.

tests/env/rvtest_failure_code.h
jordancarlin: This was not addressed.

tests/env/rvtest_setup.h
```diff
@@ -166,7 +109,9 @@
       LREG    T1, 0(T1)               // Trap signature pointer
       LA(     T2, mtrap_sigptr)       // Base address of trap signature region
       sub     T1, T1, T2              // Calculate offset
-      RVTEST_SIGUPD(x2, x5, x4, T1, check_trap_sig_offset, trap_sig_offset_mismatch)
+      // *** the following code uses x2 as a pointer.
+      // *** x2 has been modified by unpriv tests and thus causes a store access fault
+      // RVTEST_SIGUPD(x2, x5, x4, T1, check_trap_sig_offset, trap_sig_offset_mismatch)
```
jordancarlin: This has been fixed in the unpriv generator. You should be able to restore the check.

tests/env/rvtest_setup.h
```diff
@@ -166,7 +109,9 @@
...
   #endif
```
jordancarlin, line 113: I think we need this code whenever we have a trap handler, not just when `CONFORMING_SM_SUPPORTED` is true. Notably we need this for `S_SUPPORTED` so that we check the right number of traps for profile certification.

tests/env/rvtest_setup.h
```diff
@@ -133,26 +74,28 @@
   /**** MPRV must be clear here !!! ****/
 
   // Switch to M-mode
-  // ***dh 4/8/26 is this still the right thing to do if there is no conforming M-mode?
+  // The following epilog and checks are needed if there is any trap handler.  Right now, it is not
+  // invoked unless there is CONFORMING_SM_SUPPORTED.  A user with nonconforming M-mode will need
+  // to reimplement many parts of this macro.
```
jordancarlin, line 79: If non-conforming M-mode needs to reimplement these, then they need to be in the `.text.rvmodel` section.

tests/env/rvtest_setup.h
```diff
@@ -181,7 +126,12 @@
     j       cleanup_epilogs
 
   // Instantiate trap handlers for each priv mode
+  // Guard matches the RVTEST_TRAP_EPILOG guard above: rvtest_Mend (and sibling
+  // labels) are defined by RVTEST_TRAP_EPILOG, so the handler that references
+  // them must only be emitted when the epilog is also emitted.
+  #ifdef CONFORMING_SM_SUPPORTED
   INSTANTIATE_MODE_MACRO RVTEST_TRAP_HANDLER
+  #endif
```
jordancarlin, line 132: Shouldn't this create the S-mode trap handler even without CONFORMING_SM_SUPPORTED?

tests/env/rvtest_setup.h
```diff
@@ -303,7 +274,12 @@
   .align 4
 
   // Create separate save areas for each priv mode trap handler
+  // Guard matches RVTEST_TRAP_HANDLER guard: RVTEST_TRAP_SAVEAREA references
+  // Mtrampoline (and sibling labels) which are only defined when RVTEST_TRAP_HANDLER
+  // is instantiated.
+  #ifdef CONFORMING_SM_SUPPORTED
   INSTANTIATE_MODE_MACRO RVTEST_TRAP_SAVEAREA
+  #endif
```
jordancarlin, line 280: Don't we need the S and VS save areas even without CONFORMING_SM_SUPPORTED?

tests/env/rvtest_setup.h
```diff
@@ -375,13 +349,11 @@
...
-      #ifdef rvtest_mtrap_routine
-        tsig_begin_canary:
-          TRAP_CANARY
+      tsig_begin_canary:
+        TRAP_CANARY
 
-        mtrap_sigptr:
-            .fill TRAP_SIGUPD_COUNT*(SIG_STRIDE>>2),4,0xdeadbeef
-      #endif
+      mtrap_sigptr:
```
jordancarlin: Should we rename this `trap_sigptr`? I think it is used as the signature region for all priv mode trap handlers.

tests/env/rvtest_setup.h
```diff
@@ -411,35 +383,422 @@
         or      \dstreg\(), \dstreg\(), \tmpreg\()
 .endm
```
jordancarlin: Either move `RVTEST_INIT_REGS` back to this section or move this comment and the `DBLSHIFTR` macro down to `RVTEST_INIT_REGS` so that they stay together.

tests/env/rvtest_setup.h
```diff
@@ -411,35 +383,422 @@
...
+    rvtest_boot_to_mmode:
+    // Default implementation assumes conforming M-mode or no M-mode registers
+    // We are in M-mode now at initial boot time
+
+    // Do setup that requires a conforming M-mode
+    #ifdef CONFORMING_SM_SUPPORTED
+
+    // Disable interrupts
+      csrw mie, zero
+      csrw mip, zero
```
jordancarlin: Indent comment to match code.

tests/env/rvtest_setup.h
```diff
@@ -411,35 +383,422 @@
...
+
+      // disable trap delegation
+      #ifdef U_SUPPORTED
+        csrw mideleg, zero  // don't delegate interrupts (until S-mode handler is set up)
+        csrw medeleg, zero  // don't delegate exceptions (until S-mode handler is set up)
+      #endif
```
jordancarlin: Should this be `S_SUPPORTED`?

tests/env/rvtest_setup.h
```diff
@@ -411,35 +383,422 @@
...
+        #if defined(SV32_SUPPORTED) || defined(SV39_SUPPORTED)
+          csrw satp, zero // Sv BARE mode
+          sfence.vma
+        #endif // SV32 or SV39
+      #endif // PMP
+    #endif // CONFORMING_M_MODE
+
+  #endif // !RVMODEL_BOOT_TO_MMODE
+  // Init floating-point and vector state if necessary, even if the rest of M-mode is not implemented
+  INIT_FLOAT_VECTOR_STATE
```
jordancarlin, line 614: Shouldn't this be outside `CONFORMING_SM_SUPPORTED` but still inside the check for `RVMODEL_BOOT_TO_MMODE`? How are DUTs with custom M-mode supposed to override it in the current implementation?

tests/env/rvtest_setup.h
```diff
@@ -411,35 +383,422 @@
...
+    csrw mideleg, t0
+
+    // Enable necessary state for access from lower privilege modes
+    // mstateen0.SE0 = 1: enable access to hststateen0, hstatene0h, ssstateen0
+    // mstateen0.ENVCFG = 1: enable access to henvcfg, henvcfgh, senvcfg
+    // sstateen0.JVT = 1: Enable jvt for Zcmt
+    // sstateen0.FCSR = 1: Enable fcsr access for Zfinx only if supported ZFINX_SUPPORTED (to avoid conflicts with F)
+    // sstateen0.C = 0: Disable custom state
+
+    #ifdef MSTATEEN_SUPPORTED
```
jordancarlin:
```suggestion
    #ifdef SMSTATEEN_SUPPORTED
```

tests/env/rvtest_setup.h
```diff
@@ -411,35 +383,422 @@
...
+      #if __riscv_xlen == 64
+        li t0, MSTATEEN_HSTATEEN | MSTATEEN0_HENVCFG
+        csrs mstateen0, t0  // Set these fields
+      #else    // RV32
+        li t0, MSTATEENH_HSTATEEN | MSTATEEN0H_HENVCFG
+        csrs mstateen0h, t0 // Set these fields
+      #endif
+    #endif
+    #ifdef SSTATEEN_SUPPORTED
```
jordancarlin:
```suggestion
    #ifdef SSSTATEEN_SUPPORTED
```

tests/env/rvtest_trap_handler.h
```diff
@@ -355,15 +358,11 @@
...
-    \MACRO_NAME M       // actual m-mode prolog/epilog/handler code
-  #endif
-  #ifdef rvtest_strap_routine
+  \MACRO_NAME M       // actual m-mode prolog/epilog/handler code
```
jordancarlin, line 361: Should this be conditioned on `CONFORMING_SM_SUPPORTED`?

.gitignore
```diff
@@ -42,6 +42,9 @@ tests/priv/**/*.S
 # docs
 docs/ctp/mismatch_report.txt
 docs/crd/build
+docs/ctp/src/norm
+docs/ctp/src/param
+docs/ctp/src/testplans
```
jordancarlin: Why are these necessary? I don't see anything that generates these directories.

jordancarlin (changes requested):

tests/env/rvtest_setup.h
allenjbaum: If you're changing it, and it applies to all, maybe xtrap_sigptr might be more appropriate.

config/spike/spike-rv32-max/sail.json
davidharrishmc: restored

generators/testgen/src/testgen/formatters/types/csr_type.py
davidharrishmc: No, because we want the same .S to work on many configs.

.gitignore
davidharrishmc: I get these when I run make in the ctp directory.  Do you not get them?

generators/testgen/src/testgen/formatters/types/csr_type.py
jordancarlin: I was suggesting a `#define` in the generated assembly, so the same .S would still work.

.gitignore
jordancarlin: Strange. No. I get `docs/ctp/build/generated/{norm,param,testplans}` and the `build` directory is already excluded.

generators/testgen/src/testgen/formatters/types/csr_type.py
davidharrishmc: Ok.  Fine optimization.  I don't want to hold up M-mode for now, but feel free to add this separately.

tests/env/rvtest_setup.h
davidharrishmc: Right now this needs to be called even with nonconforming M mode because RVMODEL_BOOT_TO_MMODE is defined but blank for configs such as RVI20.  Let's cross that override issue later if it ever comes up.

tests/env/rvtest_setup.h
davidharrishmc: Let's review these CONFORMING_SM_SUPPORTED in person.

tests/env/rvtest_setup.h
jordancarlin: Couldn't RVI20 leave RVMODEL_BOOT_TO_MMODE undefined and just not define CONFORMING_SM_SUPPORTED?

jordancarlin (approved):

PR #1337, opened 2026-04-21
docs: fix Sail installation instructions in README

### LINKED ISSUE #1335: README: Sail installation section missing path placeholder explanation and has typo in no-sudo command
>>Prob 1  Missing placeholder explanation
The Sail installation command uses "/path/to/install" with no explanation:

curl --location https://github.com/riscv/sail-riscv/releases/download/0.10/sail-riscv-$(uname)-$(arch).tar.gz | sudo tar xvz --directory=/path/to/install --strip-components=1

Unlike the toolchain section which clearly explains to replace "</path/to/install>"with a real path (e.g. "$HOME/riscv"), the Sail section has no such note before the command. New users will run this literally and get a error like : tar: /path/to/install: Cannot open: No such file or directory

>>Prob 2 Typo in no-sudo alternative command
The NOTE below the command has a missing space: curl --location <url> | tar xvz --directory=$HOME/riscv/--strip-components=1
it should be written as: curl --location <url> | tar xvz --directory=$HOME/riscv/ --strip-components=1

without the space tar is seeing this as a single argument resulting to command not found when you run "sail_riscv_sim --version". Because it is litteraly extracting files into a directory named "riscv/--strip-components=1" instead of "$HOME/riscv/"
THIS MAY CONFUSE A NEW USER.

Manmryt:
- Add note clarifying /path/to/install placeholder in Sail install command
- Fix missing space in no-sudo alternative command that caused tar to treat directory and strip-components as a single argument, resulting in sail_riscv_sim not being found after installation

Fixes #1335
Closes #1350

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,11 +150,14 @@ To install the sail model:
 curl --location https://github.com/riscv/sail-riscv/releases/download/0.11/sail-riscv-$(uname)-$(arch).tar.gz | sudo tar xvz --directory=/path/to/install --strip-components=1
 ```
 
+> [!NOTE]
+> Replace `/path/to/install` with your desired directory location.
+> This is simpler if it matches the directory used for the `riscv-gnu-toolchain` so both tools share the same `bin` folder.
+
 Add `/path/to/install/bin` to your `PATH` if you used a different directory than for the `riscv-gnu-toolchain`.
 
 > [!NOTE]
-> If you don't have `sudo` access, you can extract the Sail model into your home directory by specifying a local path:
-> `curl --location <url> | tar xvz --directory=$HOME/riscv/--strip-components=1`
+> If you don't have sudo access, you can extract the Sail model into your home directory by specifying a local path: `curl --location <url> | tar xvz --directory=$HOME/riscv/ --strip-components=1`
 
 Verify the installation:
 


README.md
```diff
@@ -146,6 +146,9 @@ The ACTs use the RISC-V Sail model to generate expected results. It is currently
 
 To install the sail model:
 
+> [!NOTE]
+> Replace `/path/to/intall` with your desired directory location
+>this should match the directory used for the `riscv-gnu-toolchain` so both tools share the same `bin` folder.
```
jordancarlin: Let’s put the note after the command to be consistent with the gcc install.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## [4.0.1] - 2026-04-21
```
jordancarlin: No need for a changelog entry for this small documentation fix. We’ll do the next release when enough changes have accumulated.

jordancarlin (changes requested):

CHANGELOG.md
Manmryt: ok i'll take it out

README.md
Manmryt: yesssssss

README.md
```diff
@@ -146,15 +146,20 @@ The ACTs use the RISC-V Sail model to generate expected results. It is currently
...
 
 Add `/path/to/install/bin` to your `PATH` if you used a different directory than for the `riscv-gnu-toolchain`.
+```bash
+`curl --location <url> | tar xvz --directory=$HOME/riscv/ --strip-components=1`
+```
 
 > [!NOTE]
-> If you don't have `sudo` access, you can extract the Sail model into your home directory by specifying a local path:
-> `curl --location <url> | tar xvz --directory=$HOME/riscv/--strip-components=1`
+> If you don't have sudo access, you can extract the Sail model into your home directory by specifying a local path using the command above.
```
jordancarlin: Revert this change. The extra command should still be in the note. Otherwise it is not clear what that command is for.

README.md
jordancarlin: This comment still needs to be resolved. Let's move this note about replacing `path/to/install` right after the curl command to download the sail model that is on the next line.

jordancarlin (changes requested):

README.md
```diff
@@ -149,12 +149,18 @@ To install the sail model:
...
 curl --location https://github.com/riscv/sail-riscv/releases/download/0.10/sail-riscv-$(uname)-$(arch).tar.gz | sudo tar xvz --directory=/path/to/install --strip-components=1
 ```
+> [!NOTE]
+> Replace `/path/to/intall` with your desired directory location
+>this should match the directory used for the `riscv-gnu-toolchain` so both tools share the same `bin` folder.
 
 Add `/path/to/install/bin` to your `PATH` if you used a different directory than for the `riscv-gnu-toolchain`.
+```bash
+`curl --location <url> | tar xvz --directory=$HOME/riscv/ --strip-components=1`
+```
```
jordancarlin: Why is this command being added twice? Here and also two lines below. Drop this one and just keep the one below.

Manmryt: just deleted the upper comment but i feel like 2 notes coming back to back feels a bit awkward, please let me know what can i change

README.md
```diff
@@ -149,12 +149,16 @@ To install the sail model:
 ```bash
 curl --location https://github.com/riscv/sail-riscv/releases/download/0.10/sail-riscv-$(uname)-$(arch).tar.gz | sudo tar xvz --directory=/path/to/install --strip-components=1
 ```
+> [!NOTE]
+> Replace `/path/to/intall` with your desired directory location
+>this should match the directory used for the `riscv-gnu-toolchain` so both tools share the same `bin` folder.
```
jordancarlin: We provide instructions right below this on what to do if you use a different path (and there are several good reasons for using different paths, especially with CI), so I'd rather something like this.

```suggestion
> Replace `/path/to/intall` with your desired directory location.
> This is simpler if it matches the directory used for the `riscv-gnu-toolchain` so both tools share the same `bin` folder.
```

jordancarlin (approved):

jordancarlin:
@Manmryt Please update this to pass the pre-commit checks. You can run them manually using `mise run prek`.

Manmryt:
i'm getting all the checks correct in my device with mise run prek, and here it also shows that i fail pretier even though in my device i pass that, someone please help me here understand what is even going on

jordancarlin:
> i'm getting all the checks correct in my device with mise run prek, and here it also shows that i fail pretier even though in my device i pass that, someone please help me here understand what is even going on

Looks like it was a transient network error. I restarted it.

PR #1343, opened 2026-04-22
Vector Failure Code

georgiatai:
Failure code now compliant with different SEW and XLEN, structure shouldn't change much so some reviews on structure/code style would be helpful.

Still fixing details like how the mask-producing instruction reports should look like (each element is one bit). 
Realized that failure code now uses call instead of RVTEST_WRITE_TO_STR, will fix that.


tests/env/signature.h
```diff
@@ -475,12 +475,37 @@
...
     11:                                                                                                             \
-        jal         _LINK_REG, failedtest_##_LINK_REG##_##_TEMP_REG ;                                               \
+        jal         _LINK_REG, failedtest_vec_active_##_LINK_REG##_##_TEMP_REG ;                                    \
+        RVTEST_WORD_PTR _INST_PTR            ;                                                                      \
+        RVTEST_WORD_PTR _STR_PTR             ;                                                                      \
+    20:                                                                                                             \
+        /* tail region FAIL path */                                                                                 \
+        vsetvl      _TEMP_REG, _TEMP_REG, _TEMP_REG2 ;  /* Restore original vl and vtype for failure reporting */   \
+        addi        _TEMP_REG2, _LINK_REG, 0 ;   /* Copy mismatch index to temporary register 2 */                  \
+        vmv.v.v     _VTMP, _VTMP             ;   /* Copy mismatch mask to temporary register 2 */                   \
```
jordancarlin: Maybe I'm missing something, but what is the point of moving _VTMP to _VTMP?

tests/env/signature.h
```diff
@@ -475,12 +475,37 @@
...
         beq         _TEMP_REG, _TEMP_REG, 11f;   /* Unconditional branch to failure label (mirror SIGUPD) */        \
     11:                                                                                                             \
-        jal         _LINK_REG, failedtest_##_LINK_REG##_##_TEMP_REG ;                                               \
+        jal         _LINK_REG, failedtest_vec_active_##_LINK_REG##_##_TEMP_REG ;                                    \
+        RVTEST_WORD_PTR _INST_PTR            ;                                                                      \
+        RVTEST_WORD_PTR _STR_PTR             ;                                                                      \
+    20:                                                                                                             \
+        /* tail region FAIL path */                                                                                 \
+        vsetvl      _TEMP_REG, _TEMP_REG, _TEMP_REG2 ;  /* Restore original vl and vtype for failure reporting */   \
+        addi        _TEMP_REG2, _LINK_REG, 0 ;   /* Copy mismatch index to temporary register 2 */                  \
```
jordancarlin: Use the `mv` pseudoinstruction to make the intent more obvious.

tests/env/signature.h
```diff
@@ -475,12 +475,37 @@
     9:                                                                                                              \
         vfirst.m    _LINK_REG, _VTMP         ;   /* Find first active mismatch index; -1 if none */                 \
         blt         _LINK_REG, x0, 12f       ;   /* If no mismatch found → PASS ALL */                              \
+    30:                                                                                                             \
+        /* mask region FAIL path, has to come right after mask region checks */                                     \
+        vsetvl      _TEMP_REG, _TEMP_REG, _TEMP_REG2 ;  /* Restore original vl and vtype for failure reporting */   \
+        addi        _TEMP_REG2, _LINK_REG, 0 ;   /* Copy mismatch index to temporary register 2 */                  \
+        vmv.v.v     _VTMP, _VTMP             ;   /* Copy mismatch mask to temporary register 2 */                   \
+        LREG        _TEMP_REG, 0(_SIG_PTR)   ;   /* Load first reference word (for debug context) */                \
+        beq         _TEMP_REG, _TEMP_REG, 31f;   /* Unconditional branch to failure label (mirror SIGUPD) */        \
```
jordancarlin: If you want an unconditional branch, just use a jump (`j`).

jordancarlin: Also, what is the point of branching to the next line of code? That is what will be executed anyway.

tests/env/failure_code.h
```diff
@@ -48,8 +48,8 @@
         SREG DEFAULT_TEMP_REG, 32(x9)  # save DEFAULT_TEMP_REG
         SREG DEFAULT_LINK_REG, 40(x9)  # save DEFAULT_LINK_REG
         SREG x1, 8(x9)                 # save x1 early
-        li x1, 3
-        sw x1, 0(x9)                   # failure_type = 3 (trap handler)
+        li x1, 4
+        sw x1, 0(x9)                   # failure_type = 4 (trap handler)
```
jordancarlin: I'm slightly hesitant to change the trap handler failure type in case this is used elsewhere. Could we just make vector type 4 to avoid changes to the other failure code?

tests/env/failure_code.h
```diff
@@ -159,6 +324,73 @@
         SREG x30, 240(DEFAULT_TEMP_REG)
         SREG x31, 248(DEFAULT_TEMP_REG)
 
+    #ifdef RVTEST_VECTOR
+        addi x6, DEFAULT_TEMP_REG, 452     # vecreg_scratch base address
+        vs1r.v v0, (x6)
+        addi x6, x6, 128                   # assume max VLEN of 1024 bits (128 bytes) for each vector reg
```
jordancarlin: Could we base this on VLEN instead? Maybe something like `add x6, x6, VLEN_BYTES` where `VLEN_BYTES` is defined to be `VLEN >> 2`.

tests/env/failure_code.h
```diff
@@ -124,10 +124,175 @@
...
+        SREG x8, 64(x7)
+        SREG DEFAULT_TEMP_REG, 32(x7)
+        SREG DEFAULT_LINK_REG, 40(x7)
+        SREG x1, 8(x7)
+        li x1, 3
+        sw x1, 0(x7)                                # failure_type = 3 (vector)
+        mv DEFAULT_TEMP_REG, x7
+        mv DEFAULT_LINK_REG, x8
+        SREG x2, 16(DEFAULT_TEMP_REG)
+        li x2, 3                                    # vector mismatch region = 3 (base)
```
jordancarlin: Could we simplify this to stick with a single failure_type register and have 4 = vector active, 5 = vector tail, 6 = vector mask, etc. instead of adding another mismatch type register?

tests/env/failure_code.h
jordancarlin: Update comment to account for vector.

tests/env/failure_code.h
```diff
@@ -615,27 +1205,43 @@
...
 #if defined(RVTEST_FP) && CONFIG_FLEN > XLEN
     failing_value_upper:
         .fill 2, 4, 0xfeedf00dbaaaaaad
     expected_value_upper:
         .fill 2, 4, 0xfeedf00dbaaaaaad
 #endif
+#ifdef RVTEST_VECTOR
+    failing_region:                              # 0=active, 1=tail, 2=mask (296)
+        .fill 1, 4, 0xfeedf00d
+    failing_index:                               # element index of first mismatch (300)
```
jordancarlin: This isn't going to work. Depending on if `#if defined(RVTEST_FP) && CONFIG_FLEN > XLEN` evaluates to true or not, all of the scratch addresses within the vector block will change offsets. You use the offsets directly when storing, so when that happens your reporting will break. This worked before because they were the last region that was accessed by offset so the offset after them didn't matter.

jordancarlin (changes requested):
Generally looking good. Several comments below. Once they are addressed, can you please post an example of each of the different failure outputs (active, mask, tail, etc.)? Also double check that the failure output for each of the other types (integer, fp, fflags, and trap) still print the correct information and didn't change.

tests/env/signature.h
georgiatai: this copy would make failure code easier to track what vector register contains the mismatch mask, since the number of instructions to traceback is not constant for every path (as in active/tail/mask)

jordancarlin: Got it. It is worth explaining that in the comment. Not sure what it means by temporary register 2 right now.

tests/env/failure_code.h
georgiatai: Ah, I see. I was a little concerned about that too. I think I will just have a begin_failure_scratch_vector, and load from it when needed then. Does that sound like a plan?

jordancarlin: That should work. I am concerned that all of this failure code that uses hardcoded offsets is very brittle. I wonder if we should be loading addresses from labels more often and storing to them instead of using offsets.

georgiatai: I would agree. For the sake of consistency, I will still use offset from vector scratch for now, unless you want me to try out labels instead.

jordancarlin: Since you're rewriting this part anyway to use new offsets, might be worth just switching to labels.

tests/env/failure_code.h
georgiatai: would it make sense to make a CONFIG_VLEN (which I am assuming would just be DUT_VLEN) mirroring CONFIG_FLEN then?

And just to be clear on organization wise, this VLEN_BYTES would better be defined in utils.h?

tests/env/failure_code.h
georgiatai: I am fine with that, but there are more place which I branch off of failure_type rather than failing_region. This would make the failure type branches more complicated, so I personally would prefer to keep them separate for clarity and simplicity. But if you have a strong preference on making it so i would be happy with changing that too.

jordancarlin: I’m ok either way. If you think this is simpler, then feel free to leave it.

tests/env/failure_code.h
jordancarlin: Hmm. I didn’t think about that. If you can make it work with just one VLEN that would be better.

georgiatai:
@jordancarlin Ready for another review. Verified with spike-rv64-max on I, F, and ExceptionsSm that their failure code still works.

tests/env/failure_code.h
```diff
@@ -124,10 +124,175 @@
...
+        j failedtest_saveregs
+
+#endif // RVTEST_VECTOR
+
     # for the rest of this code, DEFAULT_LINK_REG contains return address of jal from the failure, DEFAULT_TEMP_REG points to scratch space
     failedtest_saveregs:
         # x1 has already been saved by all entry points
+    #ifdef RVTEST_VECTOR
+        # SREG x2, 16(DEFAULT_TEMP_REG)
+        # x2 has already been saved by vector failure entry points to indicate mismatch region (active/tail/mask)
```
jordancarlin: What if there is a mismatch on a non-vector sigupd on a vector test? For example, when reading a CSR. This now skips saving x2 in that case.

jordancarlin (changes requested):
Looks pretty good. One more comment. I’d also still like to see an example of the output.

georgiatai:
Screenshot 2026-04-27 161510
Screenshot 2026-04-27 122244
Screenshot 2026-04-27 123057
Screenshot 2026-04-27 132249

Failure code updated again. Here's screenshot for each region (base dose not print mismatch mask since it's only one element). The ACTIVE is a widening instruction on Vx16, so it prints 32 bits for bad and expected values

jordancarlin (approved):
This looks great!

PR #1345, opened 2026-04-22
Test for Svinval

Zain2050:
```
 RVCP COVERAGE COMPLETE: sail-rv64-max
  1 covergroups all with 100% coverage

 RVCP COVERAGE COMPLETE: sail-rv32-max
  1 covergroups all with 100% coverage
```


tests/priv/Svinval/Svinval.S
```diff
@@ -0,0 +1,155 @@
...
+// 5. mstatus.TVM=1, execute sfence.w.inval, sinval.vma, sfence.inval.ir & sfence.vma in S-Mode.
+//        Expected: 2 Illegal instruction exceptions
+// 6. mstatus.TVM=1, execute sfence.w.inval, sinval.vma, sfence.inval.ir & sfence.vma in U-Mode.
+//        Expected: 4 Illegal instruction exceptions
+//
+// Total Expected Faults :: 10
+//---------------------------------------------------------------------------------------------------------------------------------
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, S, Svinval]
```
davidharrishmc: I shouldn't be required.  This test should run on E as well.  Global change for all Sv tests if needed.

tests/priv/Svinval/Svinval.S
```diff
@@ -0,0 +1,155 @@
...
+
+  RVTEST_GOTO_MMODE
+
+//---------------------------------------------------------------------------------------------------------------------------------
+RVTEST_CODE_END
+RVTEST_DATA_BEGIN
+
+//---------------------------------------------------------------------------------------------------------------------------------
+
+tvm_str:                    .string "\"Mismatch in mstatus value while setting TVM bit!\""
```
davidharrishmc, line 128: I've forgotten how the trap handler knows to print the appropriate string.  If it knows, you can close this comment.

davidharrishmc (changes requested):

tests/priv/Svinval/Svinval.S
Zain2050: This `tvm_str` isn't for the trap handler. It's being used in the SIGUPD on line 98.

tests/priv/Svinval/Svinval.S
Zain2050: Changed it to E and got these errors
```
  /home/mzain/cvw/addins/riscv-arch-test-cvw/tests/env/rvtest_setup.h:411: Error: illegal operands `or x17,x17,x15'
  /home/mzain/cvw/addins/riscv-arch-test-cvw/tests/env/rvtest_setup.h:427:  Info: macro invoked from here
  /home/mzain/cvw/addins/riscv-arch-test-cvw/tests/env/rvtest_setup.h:51:   Info: macro invoked from here
  /home/mzain/cvw/addins/riscv-arch-test-cvw/tests/priv/Svinval/Svinval.S:37:    Info: macro invoked from here
  /home/mzain/cvw/addins/riscv-arch-test-cvw/tests/env/rvtest_setup.h:409: Error: illegal operands `slli x15,x17,32-7'
```
Looks like `rvtest_setup.h` uses registers > 15. Therefore, I currently can't change it to E.

Zain2050:
QEMU is not trapping on `sfence.w.inval `and `sfence.inval.ir` in U-Mode. This is a bug, therefore excluded Svinval from QEMU's CI jobs.

tests/priv/Svinval/Svinval.S
jordancarlin: Let’s still drop I from the list of required extensions since there is nothing about this one in particular that requires I, and we can fix the trap handler separately. Can you open an issue for that?

davidharrishmc:
Great to find a real bug in QEMU!  Can you open an issue against QEMU and reference it here and link it to the Bugs found page https://docs.google.com/spreadsheets/d/1EuG4LV1TQDx4wTzYDWEAvq0fwaWjNuxenwDARZmrN-E/edit?gid=1586739087#gid=1586739087?

davidharrishmc:
I think my new M-mode boot PR will resolve the register use, so no need to open an issue about E.  Just delete I.

tests/priv/Svinval/Svinval.S
jordancarlin, line 1: This test doesn’t have any page tables or anything, so it seems like it would be easy to generate using the Python framework. It might be worth switching to that to make maintenance easier as the test suite evolves.

jordancarlin (dismissed):

tests/priv/Svinval/Svinval.S
Zain2050: I thought of this, but it's just a single test. We have about 200 handwritten tests for Sv. I guess it's okay to have one more :)

davidharrishmc:

davidharrishmc (approved):

PR #1352, opened 2026-04-23
ExcpetionsV*

jacassidy:


generators/coverage/src/covergroupgen/templates/vector/priv/cp_exceptionsv_address_fault.sv
```diff
@@ -0,0 +1,38 @@
...
+    vtype_valid: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vill") {
+        bins valid = {1'b0};
+    }
+
+    // Main condition: instruction trapped (load/store access or page fault detected via mcause).
+    // For indexed instructions with unsupported EEW (exceeds MAXINDEXEEW), the instruction
+    // traps at decode with illegal instruction (mcause=2), not during memory access.
+    // Accept mcause=2 for these cases. mop field insn[27:26]: 01=indexed-unordered, 11=indexed-ordered.
+    trap_occurred: coverpoint (
+        get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mcause", "int") == 5 |   // load access fault
```
jacassidy, line 15: want to make sure this structure is necessary, trap == 1 should be working

config/sail/sail-RVA23S64-optional/rvtest_config.svh
```diff
@@ -14,6 +14,9 @@
 // Maximum supported index EEW for indexed load/store
 `define MAXINDEXEEW64
 
+// Maximum supported index EEW for indexed load/store
+`define MAXINDEXEEW64
```
davidharrishmc: duplicate several places

davidharrishmc (changes requested):

generators/coverage/src/covergroupgen/templates/vector/priv/cp_exceptionsv_address_fault.sv
davidharrishmc: Not certain.  Does trap == 1 do what you want?

coverpoints/priv/ExceptionsVf_coverage.svh
```diff
@@ -0,0 +1,42 @@
...
+
+    vector_fp_instruction : coverpoint {ins.current.insn[14:12], ins.current.insn[6:0]} {
+        bins fp_arithmetic_vv_opcode = {10'b001_1010111};
+    }
+
+    mstatus_vs_inactive : coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "vs") {
+        bins inactive = {0};
+    }
+
+    cp_mstatus_vs_off : cross misa_v_active, mstatus_vs_inactive, vector_fp_instruction;
```
jordancarlin: Why does this require misa_v_active? What about a core that implements misa as all read-only zeros (a legal choice) or a core that only implements the Zve32f extension and not the full V extension?

coverpoints/priv/ExceptionsVf_coverage.svh
jordancarlin, line 1: Should there be another coverpoint for mstatus.FS?

> If the floating-point unit status field mstatus.FS is Off then any attempt to execute a vector floating-point instruction will raise an illegal-instruction exception. Any vector floating-point instruction that modifies any floating-point extension state (i.e., floating-point CSRs or f registers) must set mstatus.FS to Dirty.

docs/DeveloperGuide.md
```diff
@@ -691,6 +691,22 @@ There are a few important gotchas to keep in mind when writing privileged tests:
 - The trap handler skips 4 bytes when returning to the test. This means that every instruction that could trap must be followed by a `nop` (or two `c.nop` if compressed instructions are supported). Alternatively, this skipped instruction can be used to change a counter/indicator of some kind to detect if a trap was taken. This is generally not necessary because the total number of traps is always checked at the end of a test.
 - Different implementations may trap on different CSRs, so always assume a CSR access could trap. The `CSRRW`, `CSRRS`, `CSRR`, etc. macros include a `nop` after the CSR access and should always be used in place of raw CSR instructions.
 
+#### Vector Privileged Tests (ExceptionsV*, SsstrictV*)
```
jordancarlin, line 694: This doesn't seem like it belongs here. First of all, this was inserted right in the middle of the general priv test explanation and the content currently before it and after it should stay together. It also talks about sharing the vector unpriv generator, but the vector unpriv generator is not described in the developer guide. If you want to add a dedicated section on vector to the developer guide, it needs to start higher level with the unpriv generator and should not interfere with any of the other content already in the guide.

generators/testgen/scripts/vector-testgen-unpriv.py
```diff
@@ -1420,6 +1420,12 @@ def getExtensions():
 
         # Finish
         f.close()
+        # Resolve @SIGUPD_COUNT_FROM_TESTGEN@ placeholder now that the final
+        # signatureWords for this test is known (header was emitted before tests).
+        with open(tempfname, "r") as _rf:
+          _src = _rf.read()
+        with open(tempfname, "w") as _wf:
+          _wf.write(_src.replace("@SIGUPD_COUNT_FROM_TESTGEN@", str(signatureWords)))
```
jordancarlin: These can be combined so you only open the file once.

generators/testgen/scripts/vector_testgen_common.py
```diff
@@ -1251,8 +1251,9 @@ def insertTemplate(test, signatureWords, name, sew=0, vdsew=0, test_data=""):
     with open(f"{ARCH_VERIF}/generators/testgen/src/testgen/templates/{name}") as h:
         template = h.read()
 
-    if (test == "ExceptionsV"):
+    if (test == "ExceptionsV") or test.startswith("ExceptionsV"):
```
jordancarlin: I think this is equivalent.
```suggestion
    if test.startswith("ExceptionsV"):
```

generators/testgen/src/testgen/templates/testgen_header_vector_priv.S
```diff
@@ -4,29 +4,25 @@
...
-#ifdef TEST_CASE_1
-
-RVTEST_CASE(1,"//check ISA:=regex(.*I.*M.*V.*Zicsr.*); def rvtest_mtrap_routine=True; def rvtest_strap_routine=True; def rvtest_dtrap_routine=True; def TEST_CASE_1=True",ZicsrV)
-
-RVTEST_V_ENABLE()
-RVTEST_SIGBASE(SIG_BASE, signature_base)
+RVTEST_BEGIN
 
 #define SEWMIN 8
 #define ELEN 64
```
jordancarlin: Why are these defined here? Shouldn't they come from the config?

generators/testgen/src/testgen/templates/testgen_header_vector_priv.S
```diff
@@ -4,29 +4,25 @@
...
 
-#include "test_macros_vector.h"
+// Define the name of this test file for reporting and the size of the signature region
+#define TEST_FILE "@TEST_FILE_NAME@"
+#define SIGUPD_COUNT @SIGUPD_COUNT_FROM_TESTGEN@
 
-RVTEST_ISA("RV32IMV_Zicsr, RV64IMV_Zicsr")
+@EXTRA_DEFINES@
+#define rvtest_mtrap_routine
+#define rvtest_strap_routine
```
jordancarlin: Can't these come from `EXTRA_DEFINES`?

generators/testgen/src/testgen/templates/testgen_header_vector_priv.S
```diff
@@ -4,29 +4,25 @@
...
+#define rvtest_dtrap_routine
```
jordancarlin: What is `rvtest_dtrap_routine`?

generators/testgen/src/testgen/templates/testgen_header_vector_priv.S
jordancarlin, line 1: Is a separate template really needed? Seems like this could work with the regular template if you add a few things to `EXTRA_DEFINES` and move the `SEWMIN` logic into one of the header files.

generators/coverage/src/covergroupgen/generate.py
```diff
@@ -562,6 +562,100 @@ def write_instruction_sample_file(
     (coverage_dir / "RISCV_instruction_sample.svh").write_text("".join(lines))
 
 
+def write_priv_covergroups(
```
jordancarlin, line 599: There is a lot of duplicate logic in this function. Can you refactor it to resume more of the existing logic? For example, `read_covergroup_templates`?

coverpoints/unpriv/ExceptionsVx8_coverage.svh
jordancarlin, line 1: I'm confused. Why do we have `priv/ExceptionsVx` but also `unpriv/ExceptionsVx{8/16/32/64}`? At a minimum, these should all be in priv. It is not clear to me why we have a version that is not SEW dependent and a version that is with (as far as I can tell) the same coverpoints.

jordancarlin (changes requested):

coverpoints/priv/ExceptionsVf_coverage.svh
jacassidy: I was under the impression if misa.V is inactive all vector instructions trap, cant seem to find a spec quote explaining misa. We want to make sure the reason it traps is from mstatus and not the misa setting

coverpoints/unpriv/ExceptionsVx8_coverage.svh
jacassidy: artifact of committing generated files I didnt make those

coverpoints/priv/ExceptionsVf_coverage.svh
jordancarlin: Unfortunately, misa is quite troublesome to work with. A legal implementation is to make misa read-only 0 times indicate that you didn't implement misa, so it doesn't reliably tell you which extensions are supported. Furthermore, misa.V would only be set if the full V extension is supported, but our tests need to work on DUTs that implement the smaller extensions like Zve32f.

coverpoints/unpriv/ExceptionsVx8_coverage.svh
jordancarlin: It's not about what is committed. My question is why these are being generated in the first place. Whether they were committed or not, they were generated and should not be.

jacassidy: ExceptionsV* testplans used to be in the unpriv directory which would have caused this appended generation file to be generated

coverpoints/priv/ExceptionsVf_coverage.svh
jacassidy: Should this have a test but not a coverpoint to try to exercise it or just remove it and mark as untestable

jordancarlin: We should definitely test mstatus.VS. Just pull the misa.V part out of this coverpoint.

docs/DeveloperGuide.md
jordancarlin: This comment still needs to be resolved.

coverpoints/priv/ExceptionsVf_coverage.svh
```diff
@@ -0,0 +1,87 @@
...
+    cp_mstatus_vs_off : cross mstatus_vs_inactive, mstatus_fs_active, vector_fp_instruction;
+    cp_mstatus_fs_off : cross mstatus_fs_inactive, mstatus_vs_active, vector_fp_instruction;
+
+    //////////////////////////////////////////////////////////////////////////////////
+    // cp_vectorfp_mstatus_fs_state
+    // Sample a vector-FP instruction under each non-Off mstatus.FS starting state
+    // (Initial=01, Clean=10, Dirty=11) with VS != Off. We cannot observe the FS
+    // state transition directly (legal impls may hard-wire FS=Dirty), so this
+    // coverpoint only ensures the test exercises every FS state. Combined with
+    // signature-based cross-model checking, it catches transitioning-model bugs.
```
jordancarlin: You say the bits can be hardwired so we can't see all the transitions, but then the coverpoint checks that we see all possible values in the CSR (which would not be possible if the bits were hardwired).

generators/coverage/src/covergroupgen/templates/vector/cmp_vs1_vs2_sew16.sv
jordancarlin: What makes this coverpoint sew16 specific? Seems to just be a standard `cmp_vs1_vs2` coverpoint.

PR #1357, opened 2026-04-24
Add InterrutpsSstc and improve other interrupt suites

SadhviNarayanan:
This is the python generator for tests for InterruptsSstc with modified coverpoints as well to ensure reachable cross products.
The coverage is currently at 100%.
Added more helper functions into interrupts.py for more modular interrupt test generations.


generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,586 @@
...
+    test_data.int_regs.return_registers([r_scratch])
+    return lines
+
+
+# ---------------------------------------------------------------------------
+# Top-level generator
+# ---------------------------------------------------------------------------
+
+
+@add_priv_test_generator("InterruptsSstc", required_extensions=["Sm", "S", "Sstc", "I", "Zicsr"])
```
davidharrishmc: Drop I and Zicsr.  Zicsr is automatic given Sm.  Eventually the tests will work on E as well.

davidharrishmc (changes requested):

Makefile
```diff
@@ -21,9 +21,9 @@ WORKDIR     ?= work
 #  - Sv,Svade,Svadu,SvaduPMP,SvPMP,SvZicbo: sail-riscv missing support for Svade/Svadu causes mismatches. Resolved in upcoming sail-riscv release.
 #  - ExceptionsZalrsc: See sail-riscv issue 1574. Resolved in upcoming sail-riscv release.
 #  - ExceptionsZaamo: Configuration needed between access and misaligned faults
-#  - InterruptsSm,InterruptsS,InterruptsU,PMPSm,PMPZca,PMPmisaligned: Additional testing needed on a wider range of configs. Some missing config options to match ref model.
+#  - InterruptsSm,InterruptsS,InterruptsU,InterruptsSstc,PMPSm,PMPZca,PMPmisaligned: Additional testing needed on a wider range of configs. Some missing config options to match ref model.
 EXTENSIONS  ?=
-EXCLUDE_EXTENSIONS ?= Sm,S,InterruptsSm,InterruptsS,InterruptsU,ExceptionsZalrsc,ExceptionsZaamo,PMPSm,PMPZca,PMPmisaligned,Sv,Svade,Svadu,SvaduPMP,SvPMP,SvZicbo,SvPMPZicbo
+EXCLUDE_EXTENSIONS ?= Sm,S,InterruptsSm,InterruptsS,InterruptsSstc,InterruptsU,ExceptionsZalrsc,ExceptionsZaamo,PMPSm,PMPZca,PMPmisaligned,Sv,Svade,Svadu,SvaduPMP,SvPMP,SvZicbo,SvPMPZicbo
```
davidharrishmc: Determine whether InterruptsSstc really needs to be excluded, and either document why or remove exclusion.

davidharrishmc:
All copilot feedback and comments above are addressed.  I owe a careful review now.

coverpoints/priv/InterruptsSstc_coverage.svh
```diff
@@ -18,16 +18,13 @@ covergroup InterruptsSstc_cg with function sample(ins_t ins);
...
     }
-    mstatus_mie: coverpoint ins.current.csr[12'h300][3]  {
+    mstatus_mie: coverpoint ins.prev.csr[12'h300][3]  {
         // autofill 0/1
     }
-    mstatus_mie_one: coverpoint ins.current.csr[12'h300][3] {
+    mstatus_mie_one: coverpoint ins.prev.csr[12'h300][3] {
         bins one = {1};
     }
     mstatus_sie: coverpoint ins.current.csr[12'h300][1] {
```
davidharrishmc: Switch to use CSR names instead of numbers, such as CSR_MSTATUS

davidharrishmc: Please add comments explaining why mstatus_mie is prev, while mstatus_sie is current, and generally which ones need to be which.

coverpoints/priv/InterruptsSstc_coverage.svh
```diff
@@ -49,29 +52,52 @@ covergroup InterruptsSstc_cg with function sample(ins_t ins);
...
+        ignore_bins unreachable = binsof(mideleg_sti) intersect {0} && binsof(mie_stie) intersect {1};
+    }
     cp_supervisor_tm:   cross priv_mode_s, csrr, read_stimecmp, mcounteren_tm;
     cp_supervisor_stce: cross priv_mode_s, csrr, read_stimecmp, menvcfg_stce;
-    cp_user_sti:        cross priv_mode_u, menvcfg_stce, mstatus_mie, mstatus_sie, mideleg_sti, mie_stie, sip_stip_one;
+
+    cp_user_sti:        cross priv_mode_u, menvcfg_stce, mstatus_mie, mstatus_sie, mideleg_sti, mie_stie, sip_stip_one {
+        // If menvcfg.STCE=0, timer doesn't fire so sip.STIP can never be 1
+        ignore_bins stce_disabled = binsof(menvcfg_stce) intersect {0} && binsof(sip_stip_one) intersect {1};
+        // When mideleg.STI=0, interrupt not delegated to U-mode, so mie.STIE is irrelevant
```
davidharrishmc: U->S

coverpoints/priv/InterruptsSstc_coverage.svh
```diff
@@ -49,29 +52,52 @@ covergroup InterruptsSstc_cg with function sample(ins_t ins);
...
+    }
 
     // main coverpoints
     cp_machine_sti:     cross priv_mode_m, menvcfg_stce_one, mstatus_mie_one, mideleg_sti, mie_stie, stimecmp_zero;
     cp_machine_tm:      cross priv_mode_m, csrr, read_stimecmp, mcounteren_tm;
     cp_machine_stce:    cross priv_mode_m, csrr, read_stimecmp, menvcfg_stce;
-    cp_supervisor_sti:  cross priv_mode_s, menvcfg_stce, mstatus_mie, mstatus_sie, mideleg_sti, mie_stie, stimecmp_zero;
+
+    cp_supervisor_sti_deleg: cross priv_mode_s, menvcfg_stce, mstatus_mie, mstatus_sie, mideleg_sti, mie_stie, stimecmp_zero {
+        // When mideleg.STI=0, the interrupt is not delegated to S-mode, so mie.STIE is irrelevant
```
davidharrishmc: Could you explain this more?  Even if STI is not delegated, why wouldn't we want to test it enabled and disabled? with mie.STIE?

coverpoints/priv/InterruptsSstc_coverage.svh
davidharrishmc: explain again

coverpoints/priv/InterruptsSstc_coverage.svh
```diff
@@ -49,29 +52,52 @@ covergroup InterruptsSstc_cg with function sample(ins_t ins);
...
     cp_supervisor_tm:   cross priv_mode_s, csrr, read_stimecmp, mcounteren_tm;
     cp_supervisor_stce: cross priv_mode_s, csrr, read_stimecmp, menvcfg_stce;
-    cp_user_sti:        cross priv_mode_u, menvcfg_stce, mstatus_mie, mstatus_sie, mideleg_sti, mie_stie, sip_stip_one;
+
+    cp_user_sti:        cross priv_mode_u, menvcfg_stce, mstatus_mie, mstatus_sie, mideleg_sti, mie_stie, sip_stip_one {
+        // If menvcfg.STCE=0, timer doesn't fire so sip.STIP can never be 1
+        ignore_bins stce_disabled = binsof(menvcfg_stce) intersect {0} && binsof(sip_stip_one) intersect {1};
+        // When mideleg.STI=0, interrupt not delegated to U-mode, so mie.STIE is irrelevant
+        ignore_bins no_deleg = binsof(mideleg_sti) intersect {0} && binsof(mie_stie) intersect {1};
+        // When MIE=0 and STIE=0, interrupt cannot fire at all so STIP cannot be 1
```
davidharrishmc: explain why interrupt cannot be pending when it is not enabled?  They seem like separate things.

coverpoints/priv/InterruptsSstc_coverage.svh
```diff
@@ -49,29 +52,52 @@ covergroup InterruptsSstc_cg with function sample(ins_t ins);
...
+        ignore_bins all_masked = binsof(mstatus_mie) intersect {0} && binsof(mie_stie) intersect {0} && binsof(sip_stip_one) intersect {1};
+        // When MIE=0 and SIE=0, neither M-mode nor S-mode will take interrupt
```
davidharrishmc: Explain again.  sip.STIP is about pending, not whether it is taken.  If this ignore is legit, explain better.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -305,3 +304,161 @@ def clr_stimer_mmode(r_scratch: int) -> list[str]:
...
+        f"{INDENT}# Assert Sstc timer: stimecmp = 0  (interrupt fires at next boundary)",
+        "CSRW(stimecmp, zero)",
+        "#if __riscv_xlen == 32",
+        "    CSRW(stimecmph, zero)",
+        "#endif",
+    ]
+
+
+def set_stimecmp_soon(r_scratch: int, r_time: int, r_hi: int) -> list[str]:
+    """Write stimecmp = TIME + 500 from M-mode so the interrupt fires after entering U-mode.
```
davidharrishmc: Should this be RVMODEL_TIMER_INT_SOON_DELAY rather than 100?

davidharrishmc: > 

Firing after entering U-mode is a common use case, but not fundamental to this method.  Clarify the comment:
from M-mode.  For example, this may be used to delay an interrupt firing until after switching to U-mode.

davidharrishmc (changes requested):

coverpoints/priv/InterruptsSstc_coverage.svh
davidharrishmc: update author

coverpoints/priv/InterruptsSstc_coverage.svh
davidharrishmc: This was resolved without any changes. 

Suppose we are in S-mode. mencfg.STCE=1. mstatus.mie doesn't matter because we are in S-mode.  mstatus.SIE = 1.  mideleg.STI = 0. mie.STIE=1. stimecmp = becomes 0.  Then a timer interrupt will fire and we will switch to M-mode because it was not delegated.  

Why is this unreachable?

davidharrishmc: I'm not seeing any of your changes on the resolved items.  Maybe you haven't pushed yet and all this is addressed in a pending push?  I'll stop reviewing old comments.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+
+def _generate_machine_sti_tests(test_data: TestData) -> list[str]:
+    """cp_machine_sti: iterate mideleg x mie_stie; sample at ``csrw stimecmp, zero`` in M-mode.
+
+    STCE=1 and MIE=1 are fixed (required by menvcfg_stce_one and mstatus_mie_one).
+    The interrupt fires at the countdown loop after the sample.
+    """
+    covergroup = "InterruptsS_S_cg"
+    coverpoint = "cp_machine_sti"
+    r_scratch, r_stce = test_data.int_regs.get_registers(2, exclude_regs=[0])
```
davidharrishmc: No need for exclude_regs[0] anymore.  Global change.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+
+    lines = [
+        comment_banner("cp_machine_sti", "M-mode STI: mideleg x mie_stie (STCE=1, MIE=1 fixed)"),
```
davidharrishmc: replace "cp_machine_sti" with f{coverpoint}.  Global change to make code more maintainable.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+    for mideleg_sti in [0, 1]:
+        for mie_stie in [0, 1]:
+            binname = f"{'d' if mideleg_sti else 'nd'}_stie{mie_stie}"
+            lines += [
+                "",
+                f"# cp_machine_sti: mideleg={mideleg_sti} stie={mie_stie}",
+                # --- setup: STCE=1, mideleg, mie.STIE, stimecmp=-1, MIE/SIE=0 ---
+                *mmode_sti_setup(r_scratch, r_stce, mideleg_sti, mie_stie),
+                # enable MIE=1; interrupt won't fire yet because stimecmp=-1
+                "csrsi mstatus, 8",
```
davidharrishmc, line 53: Setting stimecmp to 0 is not atomic on RV32. It probably works fine as is because our counters aren't running fast in simulation and were initialized to 0, but the most robust thing for arbitrary DUTs would be to enable MIE after setting stimecmp instead of before.  Put set_stimecmp_zero before the add_testcase and csrs i mstatus 8 after.  Global change.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                test_data.add_testcase(binname, coverpoint, covergroup),
+                *set_stimecmp_zero(),
+                # --- wait for interrupt to be handled ---
+                f"LI(x{r_scratch}, 2500)",
```
davidharrishmc: Should this be  RVMODEL_INTERRUPT_LATENCY?  Global change, might affect other Interrupt tests.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Machine-mode TM / STCE read tests
+# ---------------------------------------------------------------------------
+
+
+def _generate_machine_tm_tests(test_data: TestData) -> list[str]:
+    """cp_machine_tm: M-mode csrr stimecmp with mcounteren.TM={0,1}."""
+    covergroup = "InterruptsS_S_cg"
```
davidharrishmc: InterruptsSstc.  global check

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+    ]
+    for tm_val in [0, 1]:
+        lines += [
+            "",
+            f"# cp_machine_tm: TM={tm_val}",
+            "RVTEST_GOTO_MMODE",
+            # set or clear mcounteren.TM to control stimecmp visibility
+        ]
+        if tm_val:
+            lines += [f"LI(x{r_scratch}, 0x2)", f"CSRS(mcounteren, x{r_scratch})"]
```
davidharrishmc, line 88: LI scratch could go in the shared block.  Or you could do a csrsi mcounteren, 2

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+        else:
+            lines += [f"LI(x{r_scratch}, 0x2)", f"CSRC(mcounteren, x{r_scratch})"]
+
+        lines.append(test_data.add_testcase(f"tm{tm_val}", coverpoint, covergroup))
+        lines += [f"CSRR x{r_scratch}, stimecmp", "nop"]
```
davidharrishmc, line 93: When you have multiple lines in one array, put them on separate lines for readability, as you did in liens 83-86.  Global change.

coverpoints/priv/InterruptsSstc_coverage.svh
davidharrishmc: In all your Interrupts tests!

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+    return lines
+
+
+# ---------------------------------------------------------------------------
+# Supervisor-mode STI tests
+# ---------------------------------------------------------------------------
+
+
+def _generate_supervisor_sti_tests(test_data: TestData) -> list[str]:
+    """cp_supervisor_sti_deleg / cp_supervisor_sti_nodeleg: all 32 bins.
```
davidharrishmc: There is no cp_supervisor_sti_nodeleg, so it's confusing to refer to it here.  Is it necessary?

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+        ),
+        "",
+    ]
+
+    for stce in [0, 1]:
+        for mie in [0, 1]:
+            for sie in [0, 1]:
+                for deleg in [0, 1]:
+                    # when deleg==1, generate both real-deleg (mideleg=1) and the architecturally-possible
+                    # but atypical case (mideleg=0) so coverage can observe mideleg==0 under the deleg path.
```
davidharrishmc: I don't understand what is atypical about mideleg=0 or why we line 165 can't just be for mideleg_val in [0, 1].

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+# ---------------------------------------------------------------------------
+# Supervisor-mode STI tests
+# ---------------------------------------------------------------------------
+
+
+def _generate_supervisor_sti_tests(test_data: TestData) -> list[str]:
+    """cp_supervisor_sti_deleg / cp_supervisor_sti_nodeleg: all 32 bins.
+
+    deleg=1 (cp_supervisor_sti_deleg, priv_mode_s):
+      Enter S-mode via mret. For STCE=1: stimecmp=0 is written in M-mode
```
davidharrishmc: I don't understand why we need separate paths for deleg0 and 1.  In all of these, it seems we could set up the scenario in machine mode with stimecmp = 1s, switch to S mode, and set stimecmp = 0s. If mie.STIE=0, nothing happens. If the interrupt is delegated and mstatus.SIE=1, the interrupt will trap horizontally to S-mode.  If delegated and mstatus.SIE=0, nothing happens.  If the interrupt is not delegated, the interrupt will trap to M-mode no matter what.  Am I missing something?

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+            for sie in [0, 1]:
+                for deleg in [0, 1]:
+                    # when deleg==1, generate both real-deleg (mideleg=1) and the architecturally-possible
+                    # but atypical case (mideleg=0) so coverage can observe mideleg==0 under the deleg path.
+                    mideleg_values = [1, 0] if deleg == 1 else [0]
+                    for mideleg_val in mideleg_values:
+                        for stie in [0, 1]:
+                            deleg_name = ["no_deleg", "deleg"][deleg]
+                            midname = f"mideleg{mideleg_val}"
+                            binname = f"stce{stce}_mie{mie}_sie{sie}_{deleg_name}_{midname}_stie{stie}"
```
davidharrishmc: midname could just be merged in here the way you do for the other bins.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                                "CSRW(mie, zero)",
+                                "csrci mstatus, 8",  # MIE=0
+                                "csrci mstatus, 2",  # SIE=0
+                                f"LI(x{r_scratch}, 0x20)",
+                                f"CSRC(mip, x{r_scratch})",  # clear any pending STIP
+                            ]
+
+                            # configure STCE and stimecmp
+                            if stce:
+                                lines += set_stimecmp_max(r_scratch)  # stimecmp=-1 (no pending STI yet)
```
davidharrishmc: Change stimecmp after switching to S-mode to trigger the interrupt while in S-mdoe if applicable.

davidharrishmc: Only set STCE at this point.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                                lines += set_menvcfg_stce(r_stce, True)  # STCE=1: enable Sstc
+                            else:
+                                lines += set_menvcfg_stce(r_stce, False)  # STCE=0: disable Sstc
+                                lines += set_stimecmp_zero()  # stimecmp=0; clear both halves on RV32 too
+
+                            # configure delegation and interrupt enable
+                            if deleg:
+                                if mideleg_val:
+                                    lines += [f"LI(x{r_scratch}, 0x20)", f"CSRW(mideleg, x{r_scratch})"]  # STI → S-mode
```
davidharrishmc: Can pu put 0x20 in scratch earlier so you don't have to mess with it here and again at 206?

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                                else:
+                                    # override: keep the 'deleg' test path but set mideleg=0 in CSR
+                                    lines.append("CSRW(mideleg, zero)")
+                            else:
+                                lines.append("CSRW(mideleg, zero)")  # STI → M-mode
+
+                            if stie:
+                                lines += [f"LI(x{r_scratch}, 0x20)", f"CSRW(mie, x{r_scratch})"]  # STIE=1
+                            else:
+                                lines.append("CSRW(mie, zero)")  # STIE=0
```
davidharrishmc: You already set mie to 0 above.  No need to repeat.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+
+                            # use the existing SV coverpoint name for testcase attribution
+                            test_name = "cp_supervisor_sti_deleg"
```
davidharrishmc: Usually you use coverpoint and define it at the start of the function.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
davidharrishmc: I've clarified in the test plan that mcounteren.TM should be 1 to allow accessing stimercmp in S-mode.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                            # If STCE=1, write stimecmp=0 in M-mode
+                            if stce:
+                                lines += set_stimecmp_zero()
+                            lines.append("RVTEST_GOTO_LOWER_MODE Smode")
+                            if sie:
+                                lines.append("    csrsi sstatus, 2")  # SIE=1
+                            else:
+                                lines.append("    csrci sstatus, 2")  # SIE=0
+                            lines.append(f"    {test_data.add_testcase(binname, test_name, covergroup)}")
+                            lines.append("    nop")
```
davidharrishmc: This is the place you should write stimecmp=0.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                            lines += [
+                                f"    LI(x{r_scratch}, 2500)",
```
davidharrishmc: RVMODEL

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+    covergroup = "InterruptsS_S_cg"
+    coverpoint = "cp_user_sti"
+    r_scratch, r_stce, r_hi = test_data.int_regs.get_registers(3, exclude_regs=[0])
+
+    lines = [
+        comment_banner(
+            "cp_user_sti",
+            "U-mode STI cross (32 bins)\n"
+            "menvcfg_stce x mstatus_mie x mstatus_sie x mideleg_sti x mie_stie\n"
+            "STIMECMP=TIME+500: interrupt fires after sample nop.\n"
```
davidharrishmc: RVMODEL global change

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                            # Reset stimecmp to max so no spurious interrupt
+                            *set_stimecmp_max(r_scratch),
+                            *set_menvcfg_stce(r_stce, bool(stce)),
+                        ]
+
+                        # ---- mideleg: route STI to S-mode or keep in M-mode ----
+                        if deleg:
+                            lines += [
+                                "# mideleg.STI=1 → delegate STI to S-mode",
+                                f"LI(x{r_scratch}, 0x20)",
```
davidharrishmc: similar changes of sharing 0x20 as in s-mode version

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                        if stie:
+                            lines += [
+                                "# mie.STIE=1",
+                                f"LI(x{r_scratch}, 0x20)",
+                                f"CSRW(mie, x{r_scratch})",
+                            ]
+                        else:
+                            lines += [
+                                "# mie.STIE=0",
+                                "CSRW(mie, zero)",
```
davidharrishmc: already 0

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                            ]
+
+                        # ---- Schedule the interrupt (only meaningful when STCE=1) ----
+                        # By default set stimecmp = TIME+500 so the interrupt fires during the
+                        # spin loop after entering U-mode. However, in the specific case
+                        # where the interrupt is delegated to S-mode (deleg=1) and both
+                        # MIE and SIE are 0, ensure the pending bit is observable at the
+                        # sample point by asserting stimecmp=0 immediately.
```
davidharrishmc: I don't understand why this is needed.  When you go to U-mode, the timer will eventually fire.  From U-mode, MIE and SIE don't matter, so a trap will go to M or S depending on delegation.

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                        if deleg and (mie == 0) and (sie == 0):
+                            lines += [
+                                "# Special-case: ensure STIP is asserted at sample for deleg=1, MIE=0, SIE=0",
+                                *set_stimecmp_zero(),
+                            ]
+                        else:
+                            lines += [
+                                "# Set stimecmp=TIME+500 (interrupt fires after sample when STCE=1)",
```
davidharrishmc: RVMODEL

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
```diff
@@ -0,0 +1,588 @@
...
+                        ]
+
+                        # ---- Enter U-mode, execute sample + wait loop ----
+                        lines += [
+                            "RVTEST_GOTO_LOWER_MODE Umode",
+                            # Sample instruction — coverpoint is hit here
+                            f"    {test_data.add_testcase(binname, coverpoint, covergroup)}",
+                            # Short spin loop so the interrupt has time to arrive (stce=1)
+                            # and so we don't exit U-mode immediately for stce=0 bins.
+                            f"    LI(x{r_scratch}, 2500)",
```
davidharrishmc: Need to think about delay more carefully.  We need to wait until the timer fires.  We know it is going to happen iff menvcfg.STCE = 1. The time will be RVMODEL_TIMER_INT_SOON_DELAY timer ticks, which might be more (maybe lots more) cycles because they don't count in the same units.  Puzzle about what to do here.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/InterruptsSstc.py
SadhviNarayanan: The reordering broke the mideleg=0, stie=1 bin — with stimecmp=0 already pending when MIE is enabled, the interrupt fires immediately after csrsi mstatus, 8 before any subsequent instruction runs, so we never reach a point where both stimecmp_zero=1 and prev.MIE=1 are observable in the same instruction. I could have also minterpreted this as well, so I can re-evaluate

davidharrishmc (approved):

coverpoints/priv/InterruptsS_coverage.svh
```diff
@@ -2,7 +2,7 @@
 //
 // RISC-V Architectural Functional Coverage Covergroups
 //
-// Written: Corey Hickson chickson@hmc.edu 18 Feb 2025
+// Written: Sadhvi Narayanan sanarayanan@hmc.edu April 2026
```
jordancarlin, line 5: Don't remove the existing author. You can add another line that says `Edited: ...`.

coverpoints/priv/InterruptsSstc_coverage.svh
```diff
@@ -2,40 +2,51 @@
 //
 // RISC-V Architectural Functional Coverage Covergroups
 //
-// Written: Corey Hickson chickson@hmc.edu 4 Mar 2025
+// Written: Corey Hickson chickson@hmc.edu, Sadhvi Narayanan sanarayanan@hmcedu April 2026
```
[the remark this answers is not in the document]
jordancarlin: This was marked as resolved but has not been fixed.

coverpoints/priv/InterruptsSstc_coverage.svh
```diff
@@ -2,40 +2,51 @@
...
 
+
 `define COVER_INTERRUPTSSSTC
 
+
 covergroup InterruptsSstc_cg with function sample(ins_t ins);
-    option.per_instance = 0;
-    `include "general/RISCV_coverage_standard_coverpoints.svh"
+   option.per_instance = 0;
+   `include "general/RISCV_coverage_standard_coverpoints.svh"
```
jordancarlin: You made some strange whitespace changes in this file. A lot of the indentation is switching from 4 spaces of indentation (which is the standard) to 3 spaces for some reason.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -305,3 +314,167 @@ def clr_stimer_mmode(r_scratch: int) -> list[str]:
...
+        f"    {op}(menvcfg, x{r_scratch})",
+        "#else",
+        f"    LI(x{r_scratch}, 0x80000000)",
+        f"    {op}(menvcfgh, x{r_scratch})",
+        "#endif",
+    ]
+
+
+def set_stimecmp_max(r_scratch: int) -> list[str]:
+    """Write stimecmp = -1 (max) to disable Sstc timer interrupt. M-mode only."""
```
jordancarlin: Why is this M-mode only? It is writing an S-mode CSR.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -305,3 +314,167 @@ def clr_stimer_mmode(r_scratch: int) -> list[str]:
...
+    return [
+        f"{INDENT}# Assert Sstc timer: stimecmp = 0  (interrupt fires at next boundary)",
+        "CSRW(stimecmp, zero)",
+        "#if __riscv_xlen == 32",
+        "    CSRW(stimecmph, zero)",
+        "#endif",
+    ]
+
+
+def set_stimecmp_soon(r_scratch: int, r_time: int, r_hi: int, time: int | None = None) -> list[str]:
```
jordancarlin: Should this be `delay` instead of `time` to be consistent with `set_mtimer_int_soon`?

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -305,3 +314,167 @@ def clr_stimer_mmode(r_scratch: int) -> list[str]:
...
+
+    Args:
+        r_scratch: Scratch register to hold the MPIE bitmask
+        enable: True → MPIE=1 (MIE=1 in lower mode), False → MPIE=0 (MIE=0)
+    """
+    op = "CSRS" if enable else "CSRC"
+    comment = "MPIE=1 -> MIE=1 after mret" if enable else "MPIE=0 -> MIE=0 after mret"
+    return [
+        f"LI(x{r_scratch}, 0x80)  # {comment}",
+        f"{op}(mstatus, x{r_scratch})",
```
jordancarlin, line 429: Can this use CSRSI/CSRCI to avoid needing a scratch register?

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -305,3 +314,167 @@ def clr_stimer_mmode(r_scratch: int) -> list[str]:
...
+        f"LI(x{r_scratch}, 0x20)",
+        f"CSRC(mip, x{r_scratch})",  # clear any pending STIP
+        *set_stimecmp_max(r_scratch),
+        *set_menvcfg_stce(r_stce, True),  # STCE=1
+    ]
+    # mideleg
+    if mideleg_sti:
+        lines += [f"LI(x{r_scratch}, 0x20)", f"CSRW(mideleg, x{r_scratch})"]
+    else:
+        lines.append("CSRW(mideleg, zero)")
```
jordancarlin, line 460: Inconsistent use of `+=` vs `.append`. Prefer `.append` because it is more efficient.

Makefile
```diff
@@ -13,9 +13,18 @@ COVERAGE_CONFIG_FILES ?= config/sail/sail-rv64-max/test_config.yaml config/sail/
...
+#  - InterruptsSm,InterruptsS,InterruptsU,PMPSm,PMPZca,PMPmisaligned: Additional testing needed on a wider range of configs. Some missing config options to match ref model.
 EXTENSIONS  ?=
-EXCLUDE_EXTENSIONS ?= Sm,S,InterruptsSm,InterruptsS,InterruptsU,Sv,SvaduPMP,SvPMP,SvPMPZicbo,Svade,Svadu,Svinval,SvZicbo,Svnapot,Svpbmt,ExceptionsSvZalrsc,ExceptionsSvZaamo,ExceptionsZalrsc,ExceptionsZaamo,PMPZca,PMPS,PMPU #PMPF,PMPS,PMPSm,PMPU,PMPZaamo,PMPZalrsc,PMPZca,PMPZicbo,PMPmisaligned,,
+EXCLUDE_EXTENSIONS ?= Sm,S,InterruptsSm,InterruptsS,InterruptsU,InterruptsSstc,ExceptionsZalrsc,ExceptionsZaamo,PMPSm,PMPZca,PMPmisaligned,Sv,Svade,Svadu,SvaduPMP,SvPMP,SvZicbo,SvPMPZicbo
+
+# Strip spaces from comma-separated lists so shell word-splitting doesn't break CLI arguments
+empty :=
+space := $(empty) $(empty)
+override EXTENSIONS := $(subst $(space),$(empty),$(EXTENSIONS))
+override EXCLUDE_EXTENSIONS := $(subst $(space),$(empty),$(EXCLUDE_EXTENSIONS))
```
jordancarlin: Seems like some merge issues here? This is reverting a bunch of recent changes.

jordancarlin (changes requested):

generators/testgen/src/testgen/asm/interrupts.py
SadhviNarayanan: I don't think I can use CSRSI/CSRCI here — MPIE is bit 7 (0x80 = 128), which exceeds the 5-bit unsigned immediate limit of CSRxI instructions. Added this in the docstring

jordancarlin (approved):
LGTM once the remaining Copilot comments are resolved.

PR #1370, opened 2026-04-26
Touchups on RVA23 profile coverpoints

davidharrishmc:


generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -724,6 +727,7 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
             "",
             "# Write 1 to msip (set MSIP) and check mip.MSIP is set",
             f"LI(x{r_msip}, 1)                         # value 1: assert msip",
+            test_data.add_testcase("msip_mmio_1", coverpoint, covergroup),
```
Ammarahwakeel, line 730: I see it now, since our coverage cross is  "cp_msip: cross priv_mode_m, sw, msip_address, msip_val; " , it makes a lot more sense to have the test case before sw . Then either we should drop the this later test case   " test_data.add_testcase("msip_set", coverpoint, covergroup), "  or add its relevant readback cross in the coverage.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -379,6 +379,9 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
 
     lines.append("\n#ifdef MSECCFG_SUPPORTED")
     lines.extend(csr_access_test(test_data, "mseccfgh", covergroup, coverpoint))
+    lines.append("#endif")
+    lines.append("\n#ifdef SM1P13_SUPPORTED")
+    lines.extend(csr_access_test(test_data, "medelegh", covergroup, coverpoint))
```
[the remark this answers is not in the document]
Ammarahwakeel: well, either to add the comment or not, it's fine both ways.

Ammarahwakeel:
Neglected the correct naming of Zic64bZicboz, apologies.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -222,11 +221,13 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
         `endif
         `ifdef XLEN32
             bins mstatush = {CSR_MSTATUSH};
-            // bins medelegh = {12'h312}; // move this to Sm1p13 coverpoints
             bins menvcfgh = {CSR_MENVCFGH};
             `ifdef MSECCFG_SUPPORTED // update this in four places when UDB gives a name to this parameter
                 bins mseccfgh = {CSR_MSECCFGH};
             `endif
+            `ifdef SM1P13_SUPPORTED
+                bins medelegh = {CSR_MEDELEGH};
```
Ammarahwakeel, line 229: Ahh, I see now this was the cleanest way to add it here.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -157,8 +157,6 @@ def _generate_sstatus_sd_tests(test_data: TestData) -> list[str]:
                 f"or x{check_reg}, x{check_reg}, x{reg2}          # OR in desired UXL value",
                 test_data.add_testcase(f"uxl_attempt_{uxl}", coverpoint, covergroup),
                 gen_csr_write_sigupd(check_reg, "sstatus", test_data),
-                f"CSRR(x{check_reg}, sstatus)",  # read back what DUT actually stored
-                write_sigupd(check_reg, test_data),
             ]
```
Ammarahwakeel, line 160: Though I removed the coverpoint of sstatus_uxl_after, sampling the uxl bits after the instruction. But I thought just for the sake of a sanity check, we may have the actual value of UXL in our signature region written.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: It's good checking both.  The coverpoint checks the write to msip.  Then we read back mip.MSIP and check it is reflected there.  Often we write coverpoints only verifying the write occurred, and rely on the test writer to do the read back.

generators/testgen/src/testgen/priv/extensions/S.py
davidharrishmc: The gen_csr_write_sigupd is a nifty call that does a write, then a read back, then a sigupd of the value read back, so it captures these extra lines.  Good trick to know about.

jordancarlin (approved):

PR #1371, opened 2026-04-27
Update many coverpoints to use named localparams

jordancarlin:


davidharrishmc (approved):
Approved pending Copilot suggestions.  Asked Claude to double-check all of the encodings and it didn't see any other issues.

PR #1373, opened 2026-04-27
Remove `NORUN` from many tests now that Sail 0.11 supports them

jordancarlin:


davidharrishmc (approved):
LGTM but lots of overlap with 1372.  Presume you'll merge 1372 first, and then sanity check this and then can merge.

PR #1385, opened 2026-04-27
Cv32e40p sail 0.11

karabambus:
Verified by running on DUT.


karabambus:
Added comments for parameters that are schema-required but are not relevant to our core. I see there are multiple issues open across ACT4 and Sail repo. If you find it useful and not redundant I can open issue where I can track this parameters.

jordancarlin (approved):
LGTM, but I would like a review from @MikeOpenHWGroup before merging since he is more familiar with these cores.

jordancarlin:
> Added comments for parameters that are schema-required but are not relevant to our core. I see there are multiple issues open across ACT4 and Sail repo. If you find it useful and not redundant I can open issue where I can track this parameters.

This is a limitation of the Sail JSON schema (it requires all keys to be defined), and I don't think this is something that will be changing any time soon, so probably no need for an issue.

config/cores/cve4/cv32e40p-v2-rv32imc/sail.json
```diff
@@ -6,14 +6,26 @@
     "xlen": 32,
     "E": false,
     "writable_misa": false,
+    // cv32e40p has no menvcfg register (M-mode only, no S/U-mode).
+    // SCHEMA-REQUIRED: no RTL counterpart; value has no effect on DUT comparison.
     "writable_fiom": true,
     "writable_hpm_counters": {
       "len": 32,
-      // Only HPM counter 3 is active (NUM_MHPMCOUNTERS=1 default)
+      // Only HPM counter 3 is active (NUM_MHPMCOUNTERS=1 default, cv32e40p_core.sv:40).
```
MikeOpenHWGroup: You are right that the default value of `NUM_MHPMCOUNTERS` is 1, but this can be easily overridden in the testbench.  It would be better to reference the "Released core versions" sub-section of the User Manual that states:
```
Verification of cv32e40p_v1.0.0 has been done with only following value for NUM_MHPMCOUNTERS parameter: NUM_MHPMCOUNTERS == 1.
```

config/cores/cve4/cv32e40p-v2-rv32imc/sail.json
```diff
@@ -30,60 +42,148 @@
       "reserved_exceptions": false
     },
     "reserved_behavior": {
+      // SCHEMA-REQUIRED: cv32e40p has no Zacas extension; AMO instructions decode as
+      // illegal (cv32e40p_decoder.sv:450). Value has no effect on DUT comparison.
       "amocas_odd_register": "AMOCAS_Illegal",
+      // SCHEMA-REQUIRED: cv32e40p has no FPU (F: false). Value has no effect on DUT comparison.
```
MikeOpenHWGroup: `cv32e40p has no FPU` should be, `this configuration of the cv32e40p has no FPU`.

config/cores/cve4/cv32e40p-v2-rv32imc/sail.json
```diff
@@ -30,60 +42,148 @@
...
       "fcsr_rm": "Fcsr_RM_Illegal",
+      // SCHEMA-REQUIRED: cv32e40p has no PMP (USE_PMP=0, cv32e40p_core.sv:108).
```
MikeOpenHWGroup: Referencing the RTL here is especially dangerous as it provides no context for the user.  Please reference the RV32A Extensions, Security and Memory Protection sub-section of the User Manual.

config/cores/cve4/cv32e40p-v2-rv32imc/sail.json
```diff
@@ -30,60 +42,148 @@
...
       "byte_by_byte": false,
-      "order_decreasing": false,
-      "allowed_within_exp": 0
+      "order_decreasing": false
     },
     "translation": {
+      // cv32e40p has no MMU (S: false, no virtual memory).
       "dirty_update": false
     },
+    // Address to write DTB to (if provided).
```
MikeOpenHWGroup, line 100: Question for @jordancarlin - what is `DTB` in this context?

config/cores/cve4/cv32e40p-v2-rv32imc/sail.json
karabambus: Ok I will substitute RTL with refereence in manual, for all cases where I reference RTL.

config/cores/cve4/cv32e40p-v2-rv32imc/sail.json
jordancarlin: Device Tree Blob (or Device Tree Binary, depending on who you ask).

MikeOpenHWGroup (changes requested):

MikeOpenHWGroup:
For the most part, the technical content LGTM as well.  I have made a couple of minor comments on specific files, please have a look.

I also have a few comments about style:
1. There are multiple comment lines in the core json file that follow this pattern:
```
// SCHEMA-REQUIRED: cv32e40p has no <feature xxx>
// Value has no effect on DUT comparison.
```
It is not clear exactly what `SCHEMA-REQUIRED` means.  I _suspect_ it means, "this field in the json file is required by the ACT4 schema".  Since this term is not mentioned in the ACT4 documentation it would be a good idea to add a comment field at the top of the json file explaining what `SCHEMA-REQUIRED`  means and why it is needed.

2. Unless it is absolutely unavoidable, it is dangerous to explicitly reference RTL source code.  Instead, a reference to the User Manual is more appropriate.  For example, instead of:
```
// mtval is hardwired to 0 in cv32e40p — cv32e40p_cs_registers.sv:346,516.
```
it would be more appropriate to point the reader to the User Manual:
```
// mtval is hardwired to 0 in cv32e40p, see the CSR Descriptions chapter in the User Manual.
```

config/cores/cve4/cv32e40p-v2-rv32imc/sail.json
MikeOpenHWGroup: Why on earth would this be needed when certifying a core?  The DTB describes platform resources which are not present here.

MikeOpenHWGroup: Why would this be required in a JSON that describes a core?  The DTB describes platform resources which are out of scope.

jordancarlin: It is not used for certification, but is part of the Sail config file. Sail requires that all of its config keys have values, and this is one of them. It dictates where the DTB is loaded if one is provided (for certification, a DTB is not provided).

MikeOpenHWGroup (approved):
Thanks for resolving my comments and answering my questions @karabambus and @jordancarlin.  LTGM.

PR #1404, opened 2026-04-29
Replaced nops in background region with return instruction

hamza-1821:
I have updated PMPSm and PMPZca tests only. Now there is a label TEST_FOR_EXECUTION_0/TEST_FOR_EXECUTION_X just before the PMP Region having a simple jr ra instruction.
PMPZca coverage : 100%
PMPSm coverage : 100%
Also, the address for PMP region has been changed from 80006000 to 80005004.


davidharrishmc:
Does it run on all of the simulators without waivers?  If so, can you take it off the exclusions list and the CI lists?

davidharrishmc (approved):

PR #1415, opened 2026-05-02
fix(testgen): initialize fcsr.frm before dynamic rounding mode test cases

dev-aditya-hub:
## Summary

The `cp_frm` and `cr_*_edges_frm` coverpoint generators include `dyn` (`rm=111`) in their rounding mode test lists, but never write a non-default value into `fcsr.frm` before the instruction executes. Since `fcsr.frm` is `0` (RNE) at test entry, every `dyn` test case runs identically to the explicit `rne` case — a DUT that ignores `fcsr.frm` and hard-wires `rm=111` to always use RNE will pass all of these tests without ever exercising the CSR-read path.

The RISC-V Unprivileged ISA specification (§11.2) is explicit: when the instruction `rm` field is `111`, the FPU must source its rounding mode from `fcsr.frm`. Conformance tests for this encoding must verify that path by placing a non-default mode in `fcsr.frm` before the instruction executes.

## Root Cause

```asm
# cp_frm (dyn) — before fix
fsflagsi  0b00000
fadd.s    fd, fs1, fs2, dyn   # fcsr.frm is still 0 (RNE) — same result as rne
```

The formatter clears `fflags` between test cases but does not touch `frm`. The `cp_csr_frm` coverpoint already calls `fsrmi` before each test to exercise the CSR-write path; `cp_frm` and `cr_*_edges_frm` were missing the equivalent setup for the dynamic rm decode path.

## Fix

Wrap each generated `dyn` test case with `fsrmi` to load a non-default rounding mode and restore it afterward:

```asm
fsrmi     0x2                  # set fcsr.frm = RDN
fsflagsi  0b00000
fadd.s    fd, fs1, fs2, dyn   # rm=111 now reads fcsr.frm=2 (RDN)
RVTEST_SIGUPD_F(...)
fsrmi     0x0                  # restore fcsr.frm = RNE
```

RDN (round toward negative infinity) produces a different bit-exact result from RNE on any inexact arithmetic or conversion operation, which covers the edge-value inputs present in these coverpoints.

## Impact

Affects all FP instructions that accept an `rm` field — `fadd`, `fsub`, `fmul`, `fdiv`, `fsqrt`, `fmadd`, `fmsub`, `fnmadd`, `fnmsub`, and all `fcvt.*` variants — across RV32/RV64 F, D, Zfh, and Zbf16 test files. Explicit-mode test cases (`rne`, `rtz`, `rdn`, `rup`, `rmm`) encode the rounding mode directly in the instruction and are unaffected.

## Files Changed

- `generators/testgen/src/testgen/coverpoints/cp_frm.py`
- `generators/testgen/src/testgen/coverpoints/cr_fp_reg_edges.py`

## Test Plan

- Regenerated affected test files and confirmed each `dyn` test case is wrapped with `fsrmi 0x2` / `fsrmi 0x0`
- Verified explicit-mode and `cp_csr_frm` test cases are unchanged
- Confirmed correct pass/fail behavior on Spike for rv32gc and rv64gc configurations


jordancarlin:
@dev-aditya-hub I've updated this PR, but in the future, please make sure to create your PRs against the `act4` branch, not the `main` branch.

generators/testgen/src/testgen/coverpoints/cp_frm.py
```diff
@@ -27,6 +27,12 @@ def make_frm(instr_name: str, instr_type: str, coverpoint: str, test_data: TestD
         params = generate_random_params(test_data, instr_type, exclude_regs=[0], frm=frm_mode)
         desc = f"{coverpoint} (Test frm, mode = {frm_mode})"
         tc = format_single_testcase(instr_name, instr_type, test_data, params, desc, f"b{frm_mode}", coverpoint)
+        if frm_mode == "dyn":
+            # Set fcsr.frm to a non-default value (RDN=2) so that rm=111 (dyn) is forced
+            # to read a non-RNE frm from the CSR. Without this, fcsr.frm=0 at test time,
+            # making the dyn test identical to the rne test and allowing a DUT that
+            # hard-wires dyn=RNE to produce a false PASS.
+            tc.code = "fsrmi 0x2 # set fcsr.frm to RDN before dyn test\n" + tc.code + "\nfsrmi 0x0 # restore fcsr.frm to RNE"
```
jordancarlin: Can we have this use a random rounding mode instead of hardcoding 0x2? Applies to all three spots that are changed in this PR.

jordancarlin: I think this is cleaner if we add it to the instruction formatters instead of the coverpoint generators. Any of the formatters that use `frm` can have this added to their `setup` and `check` code. This also helps avoid forgotten coverpoints (like what happened with #1419).

jordancarlin (changes requested):
Great catch! Thanks for finding and fixing this. I have a few suggestions below.

dev-aditya-hub:
Thanks for the feedback @jordancarlin!

Addressed both suggestions in the latest commit:

- **Moved the fix into the formatters** (`FR`, `FR4`, `FI`, `F2X`, `X2F`) via their `setup` and `check` lists, so any coverpoint using these formatters automatically gets correct `dyn` handling without needing per-coverpoint code. Removed the now-redundant blocks from `cp_frm.py` and `cr_fp_reg_edges.py`.
- **Replaced the hardcoded `0x2`** with `random.choice([1, 2, 3, 4])` so the rounding mode (RTZ/RDN/RUP/RMM) varies across runs rather than always being RDN.

dev-aditya-hub:
> @dev-aditya-hub I've updated this PR, but in the future, please make sure to create your PRs against the `act4` branch, not the `main` branch.
Got it, thanks for fixing it! Will make sure to target `act4` from now on.

dev-aditya-hub:
> @dev-aditya-hub I've updated this PR, but in the future, please make sure to create your PRs against the `act4` branch, not the `main` branch.

Got it, thanks for fixing it! Will make sure to target `act4` from now on.

dev-aditya-hub:
While working on this I also noticed that  had the same old  patching pattern, relied on iteration order for cleanup, and didn't explicitly encode . Opened #1424 to fix that separately.

jordancarlin:
#1424 seems to include all of these changes plus additional work, so I'm going to switch over to reviewing that PR and close this one. In the future, when there is tightly coupled work like these two PRs, feel free to add on to your existing PR to make reviewing easier.

PR #1417, opened 2026-05-02
Add coverage and python generator for Ssccptr

ayeshaanwaar05:
Passing make regression


coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,143 @@
...
+covergroup Ssccptr_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+    // -----------------------------------------------------------------------
+    // Virtual memory active — satp ≠ 0 means a VM scheme is enabled.
+    // -----------------------------------------------------------------------
+
+    satp_active: coverpoint ins.current.csr[12'h180] {
+            bins vm_on = {[1:$]};
```
davidharrishmc: Need to look at MODE field, not the whole SATP.  This is different for RV32 and RV64.

coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,143 @@
...
+
+covergroup Ssccptr_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+    // -----------------------------------------------------------------------
+    // Virtual memory active — satp ≠ 0 means a VM scheme is enabled.
+    // -----------------------------------------------------------------------
+
+    satp_active: coverpoint ins.current.csr[12'h180] {
```
davidharrishmc: Use CSR_SATP rather than 12'h180

coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,143 @@
...
+            bins vm_on = {[1:$]};
+    }
+
+    // -----------------------------------------------------------------------
+    // Instruction type bins — one per load/store width tested
+    // -----------------------------------------------------------------------
+
+    lw_insn:  coverpoint ins.current.insn { wildcard bins lw  = {LW};  }
```
davidharrishmc: This is very elaborate.  I think just executing a lw from virtual memory after sfence.vma would suffice.
The rest may just make it harder to run robustly on all DUTs.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,376 @@
...
+        "    .fill 512, 8, 0",
+        ".popsection",
+        "",
+    ]
+
+
+def _setup_sv39_identity_map() -> list[str]:
+    """
+    Set up a minimal Sv39 identity map covering the 1 GiB region at
+    0x80000000 (code + data + page tables) using a LEVEL2 superpage.
```
davidharrishmc: This assumes 0x80000000, but not all DUTs have that memory map.  Could you use AUIPC to find the address of the program, and set up a superpage with identity mapping oriented around there, in case the program is based at a different address?

davidharrishmc (changes requested):

coverpoints/priv/Ssccptr_coverage.svh
ayeshaanwaar05: Will reduce the complexity once sir umer agrees too

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,422 @@
...
+
+# ---------------------------------------------------------------------------
+# Top-level test generator
+# ---------------------------------------------------------------------------
+
+
+def _generate_ssccptr_tests_rv64(test_data: TestData, covergroup: str) -> list[str]:
+    """Generate Ssccptr testcases for RV64/Sv39."""
+    coverpoint = "cp_ssccptr"
+    addr_reg, data_reg, check_reg = test_data.int_regs.get_registers(3, exclude_regs=[0])
```
davidharrishmc: no need for exclude_regs=[0] anymore.  It is always excluded.

davidharrishmc (changes requested):

coverpoints/priv/Ssccptr_coverage.svh
UmerShahidengr: Agree with Prof. Harris, a single `lw` will serve the purpose here

coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,59 @@
...
+    `else
+    satp_active: coverpoint ins.current.csr[CSR_SATP][31] {
+            bins vm_on = {1'b1};
+    }
+    `endif
+
+    lw_insn: coverpoint ins.current.insn { wildcard bins lw = {LW}; }
+
+    load_result_sentinel: coverpoint ins.current.rd_val[31:0] {
+            bins sentinel = {32'hDEADBEEF};
```
davidharrishmc: DEADBEEF is used for unwritten memory.  Let's pick a different sentinel here and in the code.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,249 @@
...
+        "sfence.vma",
+    ]
+
+
+# ---------------------------------------------------------------------------
+# Single lw testcase under VM — RV64
+# ---------------------------------------------------------------------------
+
+
+def _generate_ssccptr_lw_rv64(test_data: TestData) -> list[str]:
```
davidharrishmc: Most of this code is the same for 32 and 64.  Could you merge the rv64 and rv32 into a single function, and put the xlen_64 guard around the SATP_SETUP_RV**?  Make the rest of the comments generic about virtual memory rather than specifying Sv39 and XLEN.

davidharrishmc (commented):
Just about there.  One comment from Copilot and minor suggestions here.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
ayeshaanwaar05: Sure

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,201 @@
...
+
+# ---------------------------------------------------------------------------
+# Top-level test generator
+# ---------------------------------------------------------------------------
+
+
+@add_priv_test_generator(
+    "Ssccptr",
+    required_extensions=["Ssccptr"],
+    march_extensions=["S", "Zicsr"],
```
davidharrishmc: None of these are needed to compile.  Delete march_extensions.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,201 @@
...
+def _generate_ssccptr_main(test_data: TestData) -> list[str]:
+    """Generate all Ssccptr tests running in S-mode."""
+    lines = []
+
+    # Initialize scratch memory with a known sentinel value so the lw
+    # result is deterministic and the signature check is meaningful.
+    lines.extend(
+        [
+            "# Initialize scratch memory with known sentinel value",
+            "LA(x10, scratch)",
```
davidharrishmc: It is safer to use get_registers to get two registers than to assume x10 and x11 are free.  The register allocator might change which registers are used in the future.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,201 @@
...
+plus rvtest_slvl1_pg_tbl / rvtest_slvl0_pg_tbl injected via
+.pushsection into .data.
+"""
+
+from testgen.asm.helpers import comment_banner, write_sigupd
+from testgen.data.state import TestData
+from testgen.priv.registry import add_priv_test_generator
+
+covergroup = "Ssccptr_cg"
+_SENTINEL = "0xC0FFEE42"
```
davidharrishmc, line 36: Nice name!

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,201 @@
...
+    addr_reg, data_reg = test_data.int_regs.get_registers(2)
+
+    lines.extend(
+        [
+            comment_banner(coverpoint, "HPTW Read: lw under virtual memory"),
+            f"LA(x{addr_reg}, scratch)",
+            test_data.add_testcase("lw_under_vm", coverpoint, covergroup),
+            f"lw x{data_reg}, 0(x{addr_reg})",
+            write_sigupd(data_reg, test_data),
+            "RVTEST_GOTO_MMODE",
```
davidharrishmc: Not really necessary to go to MMODE and turn off virtual memory.  The framework does that for us.

coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,61 @@
...
+    }
+    `else
+    satp_active: coverpoint ins.current.csr[CSR_SATP][31] {
+            bins vm_on = {1'b1};
+    }
+    `endif
+
+    lw_insn: coverpoint ins.current.insn {
+            wildcard bins lw = {LW};
+   }
```
davidharrishmc: funny spacing on some of the braces.

coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,61 @@
...
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+     // -----------------------------------------------------------------------
+    // Virtual memory active — check the MODE field of satp, not the whole
+    // register.  The MODE field position differs by XLEN:
+    //   RV64: satp[63:60] — non-zero means Sv39/Sv48/Sv57 is active
+    //   RV32: satp[31]    — 1 means Sv32 is active
+    // -----------------------------------------------------------------------
+
+    `ifdef XLEN64
+    satp_active: coverpoint ins.current.csr[CSR_SATP][63:60] {
```
davidharrishmc: indent inside `if

coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,61 @@
...
+            bins vm_on = {[1:$]};
+    }
+    `else
+    satp_active: coverpoint ins.current.csr[CSR_SATP][31] {
```
davidharrishmc: indent

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,201 @@
...
+
+    Uses t0/t1/t2 (caller-saved temporaries) so it does not interfere
+    with registers allocated for the test via get_registers().
+
+    RV64: 1 GiB superpage, 8 bytes per PTE, sd to write.
+    RV32: 4 MiB superpage, 4 bytes per PTE, sw to write.
+    The xlen guard selects the correct shift, mask, and store instruction.
+    """
+    return [
+        "# AUIPC-based superpage identity map — portable to any load address",
```
davidharrishmc, line 59: allocate three registers with get_registers rather than assuming t0/1/2 are free.

davidharrishmc (changes requested):
Just about done.  It is important to use the register allocator rather than assume certain registers are free.  The other changes are cosmetic.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
ayeshaanwaar05: Without march extensions I start getting this error :
"Error output:
    riscv64-unknown-elf-gcc: error: '-march=rv64i_zicsr_ssccptr': extension 'ssccptr' starts with 's' but is unsupported standard supervisor extension"

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,178 @@
...
+            "SATP_SETUP_SV32",
+            "#endif",
+            "sfence.vma",
+        ]
+    )
+
+    addr_reg, data_reg = test_data.int_regs.get_registers(2)
+
+    lines.extend(
+    to avoid clobbering temporary registers used internally by those
```
davidharrishmc: Looks like leftover stuff from AI

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,178 @@
...
+
+# ---------------------------------------------------------------------------
+# Top-level test generator
+# ---------------------------------------------------------------------------
+
+
+@add_priv_test_generator(
+    "Ssccptr",
+    required_extensions=["Ssccptr"],
+    march_extensions=["S"],
```
davidharrishmc, line 150: harmless but unnecessary, I believe.

davidharrishmc: Fine to leave it if needed.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,178 @@
...
+            f"LA(x{scratch_reg}, scratch)",
+            f"LI(x{val_reg}, {_SENTINEL})",
+            f"sw x{val_reg}, 0(x{scratch_reg})",
+            "",
+        ]
+    )
+    test_data.int_regs.return_registers([scratch_reg, val_reg])
+
+    # Must be in M-mode for page-table setup.
+    lines.append("RVTEST_GOTO_MMODE")
```
davidharrishmc: Why does it need to be in M-mode for page table setup?

davidharrishmc (changes requested):

ayeshaanwaar05:
Regression is passing even ran prek before committing don't know why CI is failing

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Top-level test generator
+# ---------------------------------------------------------------------------
+
+
+@add_priv_test_generator(
+    "Ssccptr",
+    required_extensions=["Ssccptr"],
```
davidharrishmc: Perhaps S should be in required_extensions instead of march_extensions.  It may be needed to set up the trap handler.

davidharrishmc (changes requested):

davidharrishmc:
See pecommit log
https://github.com/riscv/riscv-arch-test/actions/runs/25436747672/job/74616848253?pr=1417
You can click on the ... next to the failing CI test to get details.

coverpoints/priv/Ssccptr_coverage.svh
```diff
@@ -0,0 +1,61 @@
...
+// Copyright (C) 2024 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Description:
+//   Coverage for the Ssccptr extension.
+//
+//   Ssccptr: "Main memory supports hardware page-table walker (HPTW) reads."
+//
+//   When virtual memory is active (satp ≠ 0), the HPTW reads PTEs from
```
davidharrishmc: safer to use standard ASCII text rather than a not equal symbol.

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
ayeshaanwaar05: But then again, I’ll get that GCC error because Ssccptr will also be passed to GCC

generators/testgen/src/testgen/priv/extensions/Ssccptr.py
ayeshaanwaar05: We can do that either in M mode or S mode. I agree comment is misleading

davidharrishmc (approved):

PR #1418, opened 2026-05-03
Config file cleanup and early CRD draft

davidharrishmc:
Fixed inconsistencies in config files while working on getting PMPZca running on Whisper
Fixed cvw sail.json inconsistencies from #1420.

Early draft of CRD for future discussion.


davidharrishmc:
Still facing errors on Whisper running PMPZca.  Whisper gets in a trap loop but Spike does not.  Claude says this is because whisper has a misconfigured pmpaddr15.

Spike pmpaddr15 = 0x003fffffffffffff → NAPOT: base=0x0, size=0x200000000000000 — covers ALL physical memory. The store to 0x80004418 matches PMP15 (RWX, locked) → access permitted.

Whisper pmpaddr15 = 0x003ffffffffffc00 → NAPOT: base=0xfffffffffff000, size=8 bytes — covers only 8 bytes near the top of address space. 0x80004418 does not match any PMP entry.

The root cause is two compounding issues:

1. Binary link-address dependency in the catch-all PMP computation
The test binary computes pmpaddr15 using a position-relative calculation. Whisper's binary is linked 0x1000 higher internally (as we saw from the auipc 0x27 vs auipc 0x26 difference). This shifts the computed pmpaddr15 by 0x1000/4 - 4 = 0x3FF, turning 0x003fffffffffffff into 0x003ffffffffffc00. The NAPOT trailing-1s count drops from 54 → 0, shrinking the "catch-all" region from all of memory to 8 bytes. The test should use a fixed constant (e.g., li x, -1; srli x, x, 10) for pmpaddr15.

2. Whisper's M-mode PMP no-match behavior
Per the RISC-V Privileged spec §3.7.1: "If no PMP entry matches an M-mode access, the access succeeds." With no PMP entry covering 0x80004418, whisper should permit the store. Instead it faults with mcause=7 (store/AMO access fault). This is a whisper bug — it's treating M-mode no-match as deny rather than permit.

Spike doesn't hit this because its correctly-computed pmpaddr15 catches 0x80004418 with the grant entry first, so no-match never comes up.

jordancarlin (approved):

PR #1419, opened 2026-05-03
testgen: set fcsr.frm before dyn rounding mode tests in cp_fs1_edges

dev-aditya-hub:
## Summary

`cp_fs1_edges` generates `dyn` rounding mode tests without pre-setting `fcsr.frm`. At test entry `fcsr.frm = 0` (RNE), so the `dyn` test is functionally identical to the `rne` test. An implementation that hard-wires `rm=111` to always select RNE passes without exercising dynamic rounding-mode dispatch — a false PASS.

## Root Cause

`make_fs1_edges` iterates over frm modes including `dyn` but does not set `fcsr.frm` before the test case. The same class of bug was fixed in `cp_frm.py` and `cr_fp_reg_edges.py` in #1415, but `cp_fs1_edges` was missed.

## Fix

Added the same `dyn` guard used in #1415:

```python
if frm_mode == "dyn":
    tc.code = "fsrmi 0x2 # set fcsr.frm to RDN before dyn test\n" + tc.code + "\nfsrmi 0x0 # restore fcsr.frm to RNE"
```

This ensures `fcsr.frm = RDN (2)` is read by the processor's dynamic rounding path, making the test distinguishable from `rne`.

## Impact

Affects 15 coverpoints: `fcvt.{l,lu,w,wu}.s`, `fcvt.{l,lu,w,wu,s}.d`, `fcvt.{l,lu,w,wu,h}.h`, `fcvt.h.d` across F, D, Zfh, ZfhD extensions.

## Files Changed

- `generators/testgen/src/testgen/coverpoints/cp_fp_reg_edges.py`


jordancarlin:
See my comments on #1415. I think with my suggested approach there, this PR becomes unnecessary.

dev-aditya-hub:
Thanks for the pointer @jordancarlin! Since the fix has been moved into the formatters in #1415, this PR is now fully covered there. Closing to keep things clean.

PR #1420, opened 2026-05-03
config/cvw: sync sail.json extension flags with UDB YAML

dev-aditya-hub:
## Summary

- `sail.json` for `cvw-rv64gc` had 7 extensions marked `"supported": false` that `cvw-rv64gc.yaml` (the UDB ground truth) lists as implemented
- The flags have no functional impact today — Sail correctly inherits Zba/Zbb/Zbs from `B: true`, Zaamo/Zalrsc from `A: true`, Zmmul from `M: true`, and Zfhmin from `Zfh: true` — so no traps occur and no false FAILs are produced
- However the explicit `false` flags directly contradict the UDB YAML, give anyone reading the config a wrong picture of what the core supports, and would silently break if a future Sail version stops inheriting sub-extensions from parent extensions

## Fix

- Set `Zba`, `Zbb`, `Zbs`, `Zaamo`, `Zalrsc`, `Zmmul`, and `Zfhmin` to `"supported": true` in `sail.json`
- Brings the explicit flags into sync with `cvw-rv64gc.yaml` and both `rvtest_config.h` / `rvtest_config.svh` which already had the correct defines

## Test plan

- [x] All 7 extensions present in `implemented_extensions` of `cvw-rv64gc.yaml`
- [x] `rvtest_config.h` and `rvtest_config.svh` already consistent — sail.json now matches
- [x] No functional change to current Sail behaviour — parent extension inheritance already covers these
- [x] No other config files changed

### DIFF
--- a/config/cores/cvw/cvw-rv64gc/sail.json
+++ b/config/cores/cvw/cvw-rv64gc/sail.json
@@ -266,10 +266,10 @@
       "supported": false
     },
     "Zmmul": {
-      "supported": false
+      "supported": true
     },
     "Zaamo": {
-      "supported": false
+      "supported": true
     },
     "Zabha": {
       "supported": false
@@ -278,7 +278,7 @@
       "supported": false
     },
     "Zalrsc": {
-      "supported": false
+      "supported": true
     },
     "Zawrs": {
       "supported": false,
@@ -299,7 +299,7 @@
       "supported": true
     },
     "Zfhmin": {
-      "supported": false
+      "supported": true
     },
     "Zfinx": {
       "supported": false
@@ -323,13 +323,13 @@
       "supported": false
     },
     "Zba": {
-      "supported": false
+      "supported": true
     },
     "Zbb": {
-      "supported": false
+      "supported": true
     },
     "Zbs": {
-      "supported": false
+      "supported": true
     },
     "Zbc": {
       "supported": true


davidharrishmc (approved):

davidharrishmc:
This is a great catch.  

Before merging it, I'd like to understand why you observed the traps and I don't.  

I agree that config/cores/cvw/cvw-rv64gc/sail.json has Zba supported: false. Nevertheless, in work/cvw-rv64gc/build/rv64i/Zba/Zba-sh1add.uw-00.sig.log, I see the sh1add.uw run without trapping.

I'd like to get to the bottom of this in case there are other extensions that should be trapping but are not in my environment.

```
mem[X,0x0000000080000050] -> 0x20F028BB
[309] [M]: 0x0000000080000050 (0x20F028BB) sh1add.uw x17, x0, x15                   Zba_sh1add_uw_cg_cp_rs1_b0+0
x17 <- 0x89DBFA8D013F227D
```

dev-aditya-hub:
Thanks for checking this and for the detailed trace you're right, and I should clarify.  
Sail derives Zba/Zbb/Zbs from B: true, Zaamo/Zalrsc from A: true, Zmmul from M: true, and Zfhmin from Zfh: true so the individual false flags are silently overridden and no traps actually occur. The false-FAIL scenario I described in the PR doesn't happen in practice, and I overstated the severity.
That said, the explicit flags still directly contradict cvw-rv64gc.yaml (the UDB ground truth), which lists all seven as implemented. The fix is a config consistency correction rather than a critical bug fix the config as written gives anyone reading it a wrong picture of what the core supports, and a future Sail version that doesn't inherit sub-extensions from parents would break immediately.
Happy to update the PR description to reflect the actual impact if that helps. Up to you whether that's worth merging as-is

davidharrishmc (approved):

dev-aditya-hub:
Also, would you mind taking a look at my other open PRs when you get a chance?
#1415 fixes missing fsrmi wrapper for dyn rounding mode in cp_frm and cr_fp_reg_edges (the original fix for this class of bug)
#1419  same fix applied to cp_fs1_edges which was missed in #1415
#1421  catches subprocess.TimeoutExpired per-test in run_tests.py so a hung ELF doesn't crash the entire pool run and silently skip remaining tests
#1413  earlier related work
No rush at all, just flagging them in case they are useful. Thanks again for the detailed feedback on this one.

davidharrishmc:
Thank you for the root cause.  
Perhaps the Sail JSON validator should be checking these things too.

I was about to merge, and now it looks like there is a a new requirement for signing commits that breaks this PR.  The changes also apply to cvw-rv32gc.  I've added all of them to my open PR, so we can close this.

Very good to have your other PRs.  @jordancarlin was the original author of those feature and will be the best one to review.

Thank you for sweating the details to improve the quality of ACT4.

dev-aditya-hub:
Thanks for picking it up and extending it to cvw-rv32gc good catch, I missed that. Closing this PR since you've got it covered. Looking forward to @jordancarlin's review on the others. Thanks again for the thorough feedback.

jordancarlin:
> Thank you for the root cause.
> Perhaps the Sail JSON validator should be checking these things too.

There are subtle differences. For example, B in the config file sets misa.B while Zba/Zbb/Zbs do not. Same applies to Zaamo/Zalrsc and misa.A.

davidharrishmc:
If misa.B is set, should Zba also be set?

jordancarlin:
If I recall, the distinction right now is that if you say B is supported but Zba is not, then writing a 0 to misa.B will disable the Zba instructions. If Zba is marked as supported, then it does not depend on misa.B.

PR #1421, opened 2026-05-03
run_tests: catch TimeoutExpired per-test instead of crashing the pool

dev-aditya-hub:
## Summary

- `subprocess.run(timeout=5*60)` in `run_test()` had no `try/except`, so when a DUT hung and the timeout fired, `subprocess.TimeoutExpired` propagated uncaught through `pool.imap_unordered` back to the main process
- This crashed the entire for-loop in `main()`, terminated all remaining workers via `pool.terminate()`, and left `summary.log` unwritten — every ELF still in the queue was silently skipped with no record
- Fixed by wrapping `subprocess.run` in `except subprocess.TimeoutExpired`, writing an `RVCP-SUMMARY: TEST FAILED` line to the log, printing a red FAIL to the terminal, and returning a normal failure tuple so the rest of the run continues unaffected

## Fix

- Added `try/except subprocess.TimeoutExpired` around `subprocess.run` in `run_test()`
- On timeout: the ELF is marked FAIL, its log and `summary.log` entry are written correctly, and `pool.imap_unordered` continues with the next ELF
- No changes to `main()`, `Makefile`, or any other file

## Test plan

- [x] A hung ELF now appears as `FAIL (timeout after 300s)` in `summary.log` instead of crashing the run
- [x] All ELFs queued after a timeout still execute and are reported correctly
- [x] `summary.log` is always written even when one or more ELFs time out
- [x] No behaviour change for ELFs that complete within the timeout


jordancarlin:
@dev-aditya-hub Looks like this is fixed in #1409 as well, so I am going to go ahead and close this. It was opened first and should have better robustness for avoiding zombie processes. Thanks for working on this though!

dev-aditya-hub:
> @dev-aditya-hub Looks like this is fixed in #1409 as well, so I am going to go ahead and close this. It was opened first and should have better robustness for avoiding zombie processes. Thanks for working on this though!

Thanks @jordancarlin! Makes sense — #1409 was opened earlier and the `os.killpg()` approach to kill the entire process group is clearly the better fix. I hadn't thought about orphaned QEMU child processes. Good learning. I'll find something else to 
contribute

PR #1422, opened 2026-05-03
Zkr 100% coverage

juliacygong:


coverpoints/priv/Zkr_coverage.svh
```diff
@@ -47,14 +46,14 @@ covergroup Zkr_cg with function sample(ins_t ins);
 
     // Main coverpoints
     cp_zkr_seed_csrrw_M: cross csrrw, seed_csr, rs1_imm_0_1, priv_mode_m, mseccfg_sseed, mseccfg_useed;
-    cp_zkr_seed_illegal_csr_op_M: cross csrops_illegal, seed_csr, rs1_imm_0_1, priv_mode_m;
```
davidharrishmc, line 50: Why did this need to be dropped?

coverpoints/priv/Zkr_coverage.svh
```diff
@@ -22,7 +22,6 @@ covergroup Zkr_cg with function sample(ins_t ins);
     csrops_illegal: coverpoint ins.current.insn {
```
davidharrishmc, line 22: This PR has lost a fair amount of testing.

It no longer checks that csrrwi is legal.
It also doesn't test csrs/c/si/ci with zero and nonzero arguments.

Can it work with all of those?

generators/testgen/src/testgen/priv/extensions/Zkr.py
```diff
@@ -132,67 +132,54 @@ def _generate_seed_illegal_csr_op_tests(test_data: TestData) -> list[str]:
...
-    # (op, is_immediate) for each CSR op to test on seed
-    csr_ops: list[tuple[str, bool]] = [
-        ("csrrs", False),
-        ("csrrc", False),
-        ("csrrwi", True),
-        ("csrrsi", True),
-        ("csrrci", True),
-        ("csrrw", False),
+    # Each entry: (op, instruction with x0/imm=0, tag)
+    csr_ops: list[tuple[str, str]] = [
```
davidharrishmc: These four raise the illegal instruction exception.  Can we test them with different x / immediate that they don't raise exceptions, and test csrwi doesn't raise an exception?

davidharrishmc (changes requested):

PR #1424, opened 2026-05-04
fix(testgen): route cp_csr_frm through formatter pipeline

dev-aditya-hub:
## Summary

`cp_csr_frm` had three issues that this PR addresses:

- **Old `tc.code` patching** — it was still using the direct `tc.code` prepend approach that was flagged in #1415, instead of going through the formatter pipeline
- **Ordering dependency for cleanup** — the frm modes tuple was intentionally ordered to end at `rne=0` as a cleanup hack; if the order ever changed, `fcsr.frm` would be left dirty
- **Implicit `rm=dyn` encoding** — the instruction had no explicit rm field, relying on the GAS assembler defaulting to `rm=111` when rm is omitted, which is a convention not a guarantee

## Fix

Add `csr_frm_val: int | None` to `InstructionParams`. When set, the formatter uses that specific value for `fsrmi` instead of `random.choice([1, 2, 3, 4])`. This lets `cp_csr_frm` pass `frm="dyn"` (explicit `rm=111` encoding) alongside a specific `csr_frm_val` for each of the five fcsr.frm values it needs to test.

The formatter now handles both the `fsrmi <val>` setup and `fsrmi 0x0` restore in check — no ordering dependency, no `tc.code` patching.

## Files Changed

- `data/params.py` — add `csr_frm_val` field
- `formatters/types/fr_type.py`, `fr4_type.py`, `fi_type.py`, `f2x_type.py`, `x2f_type.py` — use `csr_frm_val` when set
- `coverpoints/cp_csr_frm.py` — rewrite to use formatter pipeline


generators/testgen/src/testgen/data/params.py
```diff
@@ -61,6 +61,7 @@ class InstructionParams:
 
     # Flags
     frm: str | None = None  # Floating-point rounding mode tests
+    csr_frm_val: int | None = None  # Explicit fcsr.frm value for dyn tests; None means random non-RNE
```
jordancarlin: No need for random to exclude RNE. It should be able to randomly select any rounding mode. Please update the comment here and the actual random selection logic in the other files.

generators/testgen/src/testgen/formatters/types/f2x_type.py
```diff
@@ -38,4 +40,8 @@ def format_f2x_type(
         write_sigupd(params.rd, test_data, "int"),
         write_sigupd(None, test_data, "fflags"),
     ]
+    if params.frm == "dyn":
+        fcsr_frm = params.csr_frm_val if params.csr_frm_val is not None else random.choice([1, 2, 3, 4])
```
jordancarlin: Rather than adding random logic into all of the formatters, let's add it to the `generate_random_params` function in `generators/testgen/src/testgen/formatters/params.py`. Something like the following should work:
```py
if params.frm == "dyn" and params.csr_frm_val is None:
    params.csr_frm_val = random.random_range(0, 5)
```

generators/testgen/src/testgen/coverpoints/cp_csr_frm.py
```diff
@@ -21,15 +21,16 @@ def make_frm(instr_name: str, instr_type: str, coverpoint: str, test_data: TestD
     if coverpoint != "cp_csr_frm":
         raise ValueError(f"Unknown cp_csr_frm coverpoint variant: {coverpoint} for {instr_name}")
 
-    # csr frm modes 0-4, end at 0 so the rest of the test continues in rne
-    frm_modes = (("rmm", 4), ("rup", 3), ("rdn", 2), ("rtz", 1), ("rne", 0))
+    # Test each valid fcsr.frm value (0-4) via the dynamic rounding mode path (rm=111).
+    # Passing frm="dyn" encodes rm=111 in the instruction; csr_frm_val tells the formatter
+    # which exact value to write into fcsr.frm before the instruction executes, and the
+    # formatter restores fcsr.frm=0 in check — no ordering dependency needed.
```
jordancarlin: Simplify the comment to just explain the code here as needed. There’s no need to compare it to what we had before (like the reference to the ordering dependency).

jordancarlin (changes requested):
Nice improvement. One main suggestion to simplify the code a bit and avoid duplication.

dev-aditya-hub:
Thanks for the review @jordancarlin All three points addressed in the latest commit moved the random csr_frm_val assignment into generate_random_params so it's in one place, updated the comment to allow RNE in the random selection, and trimmed the cp_csr_frm comment down. Makes the code much cleaner.

jordancarlin (approved):
LGTM!

generators/testgen/src/testgen/formatters/params.py
```diff
@@ -69,6 +69,10 @@ def generate_random_params(
     """
     params = InstructionParams(**fixed_params)
 
+    # Assign a random fcsr.frm value for dynamic rounding mode tests if not already set
+    if params.frm == "dyn" and params.csr_frm_val is None:
+        params.csr_frm_val = random_range(0, 4)
+
```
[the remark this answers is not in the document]
jordancarlin: Disregard this comment. We want to verify that RNE can be used in dynamic mode as well, so the current behavior is correct.

PR #1427, opened 2026-05-04
Fix orphaned simulator processes on Ctrl+C or task failure

### LINKED ISSUE #1238: Orphaned `sail_riscv_sim` processes on Ctrl+C or task failure
**Describe the bug**
When a `act` run is cancelled (Ctrl+C) or a task failure causes early exit, any `sail_riscv_sim` processes already launched by worker threads are **not terminated**. They become orphaned OS processes and continue consuming CPU until the OS is rebooted or they are killed manually. If the sail processes are themselves hung, they run indefinitely and saturate all cores.

**To Reproduce**
For this to have real effect - test should hang and run indefinitely. Canceling this tests with CTRL + C reproduces this behavior. 
Good target is removing F_SUPPORTED from rvtest_config.h and running F tests then canceling during hang.

**Additional context**
`KeyboardInterrupt` is delivered only to the **main thread**. The worker threads are unaffected - they continue blocking on their `subprocess.run()` call. When the `ThreadPoolExecutor` context manager exits, it calls `executor.shutdown(wait=False, cancel_futures=True)`, which cancels **not-yet-started** futures but cannot stop threads already running.

Jjateen:
When a task failed early (`keep_going=False`) or the user pressed Ctrl+C, any `sail_riscv_sim` processes already launched by `ThreadPoolExecutor` worker threads were left alive. Workers blocked in `proc.communicate()` kept the executor alive and the simulator ran indefinitely.

Three fixes in `build.py`:

- Switch `SubprocessAction` execution from `subprocess.run()` to `Popen` with `start_new_session=True` so each simulator gets its own process group, allowing `os.killpg()` to reach any children it spawns.
- Add a thread-safe registry of running process group IDs (`active_pgids`, protected by `pgids_lock`). Workers register their pgid on launch and deregister on completion. A `shutdown_event` (`threading.Event`) is set atomically under the same lock before snapshotting active pgids, so there is no race between registration and shutdown: any process that starts after `_kill_active()` runs sees the event already set and kills itself immediately without being added to the set.
- Call `_kill_active()` on early task failure (before `executor.shutdown()`) and inside a `KeyboardInterrupt` handler so cleanup runs on Ctrl+C. When killing, both `os.killpg()` and `proc.kill()` are attempted so the process is terminated even if the group kill fails.

Fixes #1238.


framework/src/act/build.py
```diff
@@ -217,6 +260,23 @@ def build(
...
+    # Install a SIGINT handler so Ctrl+C kills running subprocesses before the
+    # ThreadPoolExecutor waits for worker threads to finish.
+    # signal.signal() requires the main thread; skip silently if called from a worker.
+    _prev_sigint: signal._HANDLER = None
+    try:
+
+        def _sigint_handler(signum: int, frame: object) -> None:
+            tracker.kill_all()
+            signal.signal(signal.SIGINT, signal.SIG_DFL)
+            os.kill(os.getpid(), signal.SIGINT)
```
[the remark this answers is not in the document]
Jjateen: this suggestion introduces `frame: Any` which is breaking CI right now (`ANN401` - `Any` is banned in this project's ruff config)

apart from that I've applied rest of the suggestions

Jjateen:
@jordancarlin can you review/approve this further?

jordancarlin (changes requested):
Thanks for working on this. This seems like a very complex fix for this problem. Could we simplify by keeping `start_new_session=True`, tracking active child PGIDs locally in `build()` (maybe as a set or something), and cleaning them up on task failure / `KeyboardInterrupt` rather than introducing `_ProcTracker` and a custom `SIGINT` handler? I think that would accomplish almost the same thing in a much more maintainable way. If that isn't sufficient, can you add some comments explaining why all of this is necessary?

Jjateen:
Good point. The SIGINT handler was needed because catching KeyboardInterrupt outside the with executor block deadlocks: __exit__ calls shutdown(wait=True) while workers are stuck in communicate(). But we can avoid it entirely by catching KeyboardInterrupt inside the with executor block, killing processes there so communicate() unblocks, then re-raising.

Updated: replaced _ProcTracker with a local active_pgids: set[int] + threading.Lock in build(), a _kill_active() closure, and a try/except KeyboardInterrupt inside the executor block. One minor trade-off: the old _shutting_down flag handled the narrow race where a process starts after _kill_active() runs; the simplified version does not. I think that's an acceptable trade-off.

framework/src/act/build.py
```diff
@@ -129,29 +132,46 @@ def is_stale(task: BuildTask) -> bool:
...
+            pgid = proc.pid  # start_new_session=True makes the child its own group leader
+            if active_pgids is not None and pgids_lock is not None:
+                with pgids_lock:
+                    active_pgids.add(pgid)
+            try:
+                stdout, stderr = proc.communicate()
+            finally:
+                if active_pgids is not None and pgids_lock is not None:
+                    with pgids_lock:
+                        active_pgids.discard(pgid)
```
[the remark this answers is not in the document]
Jjateen: Fixed: `shutting_down: list[bool]` flag added to `build()`, set under `pgids_lock` in `_kill_active()`. Registration and shutdown-check now happen atomically under the same lock, so late-starting processes are killed immediately on registration if shutdown is already in progress.

framework/src/act/build.py
```diff
@@ -129,29 +132,46 @@ def is_stale(task: BuildTask) -> bool:
...
 
 
-def execute_task(task: BuildTask, *, verbose: bool = False) -> BuildError | None:
+def execute_task(
+    task: BuildTask,
+    *,
+    verbose: bool = False,
+    active_pgids: set[int] | None = None,
+    pgids_lock: threading.Lock | None = None,
+) -> BuildError | None:
```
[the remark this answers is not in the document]
Jjateen: `threading.Lock` is defined as a class in Python's typeshed, so `threading.Lock | None` is a valid instance-type annotation and pyright accepts it.

framework/src/act/build.py
```diff
@@ -217,6 +244,18 @@ def build(
     if not tasks:
         return result
 
+    active_pgids: set[int] = set()
+    pgids_lock = threading.Lock()
+    shutting_down: list[bool] = [False]
```
jordancarlin: Looks like this is only ever used as a one-element list. Can we make this `shutting_down = False` instead of needing the list?

framework/src/act/build.py
```diff
@@ -129,29 +132,53 @@ def is_stale(task: BuildTask) -> bool:
...
 
 
-def execute_task(task: BuildTask, *, verbose: bool = False) -> BuildError | None:
+def execute_task(
+    task: BuildTask,
+    *,
+    verbose: bool = False,
+    active_pgids: set[int] | None = None,
+    pgids_lock: threading.Lock | None = None,
+    shutting_down: list[bool] | None = None,
```
jordancarlin: We always pass these arguments, so let's make them required, drop the `| None` type annotation, and simplify the code that uses them to drop the `None` check.

framework/src/act/build.py
```diff
@@ -129,29 +132,53 @@ def is_stale(task: BuildTask) -> bool:
...
+            pgid = proc.pid  # start_new_session=True makes the child its own group leader
+            if active_pgids is not None and pgids_lock is not None:
+                kill_now = False
+                with pgids_lock:
+                    active_pgids.add(pgid)
+                    if shutting_down is not None and shutting_down[0]:
+                        kill_now = True
+                if kill_now:
+                    with contextlib.suppress(ProcessLookupError, PermissionError):
+                        os.killpg(pgid, signal.SIGKILL)
```
jordancarlin: What is the point of adding the new pgid to the set if we are going to immediately kill it when `kill_now` is True?

jordancarlin (changes requested):
A few minor simplifications, but otherwise this looks like a good improvement.

framework/src/act/build.py
Jjateen: Replaced with `threading.Event` (`shutdown_event = threading.Event()`). `shutdown_event.set()` / `shutdown_event.is_set()` is the idiomatic pattern for this and removes the need for the list wrapper entirely.

framework/src/act/build.py
Jjateen: Done. Made all three required params, dropped the `| None` annotations, and removed the `if x is not None` guards throughout.

framework/src/act/build.py
Jjateen: Good point. Restructured the lock block to only call `active_pgids.add(pgid)` in the `else` branch, so processes being immediately killed are never added to the set.

Jjateen:
Hey @jordancarlin, the `markdown-link-check` pre-commit hook is failing on `https://tech.riscv.org/calendar/` (404). Do you know the replacement URL, or should we just remove the hyperlink and leave "ACT SIG meetings" as plain text?

framework/src/act/build.py
```diff
@@ -129,29 +132,52 @@ def is_stale(task: BuildTask) -> bool:
...
+            pgid = proc.pid  # start_new_session=True makes the child its own group leader
+            kill_now = False
+            with pgids_lock:
+                if shutdown_event.is_set():
+                    kill_now = True
+                else:
+                    active_pgids.add(pgid)
+            if kill_now:
+                with contextlib.suppress(ProcessLookupError, PermissionError):
+                    os.killpg(pgid, signal.SIGKILL)
```
jordancarlin: I think this can be simplified to the following:
```suggestion
            if shutdown_event.is_set():
                with contextlib.suppress(ProcessLookupError, PermissionError):
                    os.killpg(pgid, signal.SIGKILL)
            else:
                with pgids_lock:
                    active_pgids.add(pgid)
```

jordancarlin (commented):
One more simplification.

jordancarlin:
> Hey @jordancarlin, the `markdown-link-check` pre-commit hook is failing on `https://tech.riscv.org/calendar/` (404). Do you know the replacement URL, or should we just remove the hyperlink and leave "ACT SIG meetings" as plain text?

Looks like that was just a transient network error. I reran the check and it passed.

jordancarlin (approved):

framework/src/act/build.py
```diff
@@ -129,29 +132,49 @@ def is_stale(task: BuildTask) -> bool:
...
+            try:
+                stdout, stderr = proc.communicate()
+            finally:
+                with pgids_lock:
+                    active_pgids.discard(pgid)
             if action.stdout_file is not None:
-                action.stdout_file.write_text(result.stderr + result.stdout)
-            if result.returncode != 0:
+                action.stdout_file.write_text(stderr + stdout)
+            if proc.returncode != 0:
```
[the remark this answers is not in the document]
Jjateen: Added `proc.kill()` as a direct fallback after `killpg` -> if `killpg` fails for any reason, the process is still killed before `communicate()` is called, so the thread can't hang.

framework/src/act/build.py
```diff
@@ -217,6 +242,18 @@ def build(
...
+    shutdown_event = threading.Event()
+
+    def _kill_active() -> None:
+        with pgids_lock:
+            shutdown_event.set()
+            pgids = set(active_pgids)
+        for pgid in pgids:
+            with contextlib.suppress(ProcessLookupError, PermissionError):
+                os.killpg(pgid, signal.SIGKILL)
+
```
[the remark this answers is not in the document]
Jjateen: The PR description is stale the `_ProcTracker` abstraction and explicit SIGINT handler were considered but dropped in favour of the simpler inline `active_pgids`/`pgids_lock`/`shutdown_event` approach. The `KeyboardInterrupt` block in `build()` already handles Ctrl+C. Updating the description now.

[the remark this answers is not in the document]
Jjateen: PR description is now updated.

PR #1430, opened 2026-05-04
S passes sail, spike, cvw

davidharrishmc:
Fixing sensitive CSRs


generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -298,53 +325,53 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+        ("mideleg", None),
+        ("mie", 0x3FFF),  # limit to implemented interrupt bits
+        #        ("mtvec", 0b1),  # tvec.MODE[0] can be 0 or 1. BASE is hard to predict with a reference model
+        ("mcounteren", None),
+        ("mscratch", None),
+        ("mepc", None),
+        #        ("mcause", None), # WLRL fields can't be handled with masks.  Use cp_mcause_* instead
+        ("mtval", None),
+        ("mip", 0x3FFF),  # limit to implemented interrupt bits
+        ("menvcfg", None),
```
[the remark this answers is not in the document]
davidharrishmc: Let this be for now.

coverpoints/priv/S_coverage.svh
```diff
@@ -39,20 +39,26 @@ covergroup S_scause_cg with function sample(ins_t ins);
...
         //bins b_17_reserved = {17};
-        bins b_18_software_check = {18};
-        bins b_19_hardware_error = {19};
-        bins b_20_instr_guest_page_fault = {20};
-        bins b_21_load_guest_page_fault = {21};
-        bins b_22_virtual_instruction = {22};
-        bins b_23_store_guest_page_fault = {23};
+        `ifdef ZICFILP_SUPPORTED
+            bins b_18_software_check = {18};
+        `endif
```
jordancarlin: Software-check is also possible if Zicfiss is supported.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -35,22 +35,34 @@ covergroup Sm_mcause_cg with function sample(ins_t ins);
...
         //bins b_17_reserved = {17};
-        bins b_18_software_check = {18};
-        bins b_19_hardware_error = {19};
-        bins b_20_instr_guest_page_fault = {20};
-        bins b_21_load_guest_page_fault = {21};
-        bins b_22_virtual_instruction = {22};
-        bins b_23_store_guest_page_fault = {23};
+        `ifdef ZICFILP_SUPPORTED
+            bins b_18_software_check = {18};
+        `endif
```
jordancarlin: Zicfiss too.

generators/testgen/src/testgen/asm/csr.py
```diff
@@ -22,18 +24,31 @@ def gen_csr_read_sigupd(check_reg: int, csr_name: str, test_data: TestData) -> s
 
     Args:
         check_reg: Register to read CSR into
-        csr_name: Name of the CSR to read
+        csr: Tuple of (csr_name, mask) where csr_name is the CSR name string and
+             mask is either None or an integer representing a binary mask of bits to ignore
         test_data: TestData object to track signature updates
+        mask_reg: Register pre-loaded with the complement of the mask (~mask), required
+                  when csr mask is not None. Supports masks of any bit width.
```
jordancarlin, line 31: Why is `csr` becoming a tuple with both the `csr_name` and `mask`? Seems like this would be simpler as two separate arguments. Then `mask` could default to `None` and you could leave most of the calling sites as they were. Also, it seems redundant to require both the mask and a register with the mask pre-loaded into it. Can this function load the mask instead?

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -406,7 +443,38 @@ def _generate_scsr_tests(test_data: TestData) -> list[str]:
...
 
     # Standard S-mode CSRs
-    csrs = ["sstatus", "scause", "sie", "stvec", "scounteren", "senvcfg", "sscratch", "sepc", "stval", "sip"]
+    # Format: (CSR Name, IgnoreMask).  IgnoreMask specifies a set of bits to ignore
+
+    csrs = [
+        (
+            "sstatus",
+            0x100000040,
+        ),  # bug in QEMU allows writing 11 to xstatus.UXLEN (bit 32)-mask off until it is fixed; sail does not yet support UBE
```
jordancarlin: I don't think we should be modifying the tests to work around bugs in other simulators like QEMU. If QEMU is wrong, then it should be failing the test.

jordancarlin (changes requested):

generators/testgen/src/testgen/asm/csr.py
davidharrishmc: In csr_access_test, there are four calls to gen_csr_read_sigupd with the same mask, so it is tighter code to load the mask once.  There are pros and cons both ways, but this implementation works, and I'd rather leave it as is.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -31,18 +31,46 @@ def _generate_scause_tests(test_data: TestData) -> list[str]:
         f"CSRR(x{save_reg}, scause)     # save CSR before testing it",
     ]
 
+    gated_exceptions = [
+        (10, "H_SUPPORTED"),  # ecall from VS-mode
+        (11, "RESERVED"),  # ecall from M-mode never delegated
+        (14, "RESERVED"),
+        (16, "RESERVED"),  # Double trap never delegated
+        (17, "RESERVED"),
+        (18, "ZICFILP_SUPPORTED"),  # software check
```
jordancarlin: Software check for Zicfiss too.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -31,18 +31,46 @@ def _generate_scause_tests(test_data: TestData) -> list[str]:
...
+        else:
+            lines.extend(
+                [
+                    "",
+                    f"# Testcase: set scause to exception cause {i}",
+                    f"LI(x{check_reg}, {i})",
+                    test_data.add_testcase(f"b_{i}", coverpoint, covergroup),
+                    gen_csr_write_sigupd(check_reg, "scause", test_data),
+                ]
+            )
```
jordancarlin: Can you change this to just have the ifdef and endif lines be in `if gated is not None` blocks? The main portion of the code is the same and I'd rather not duplicate it.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -406,7 +443,38 @@ def _generate_scsr_tests(test_data: TestData) -> list[str]:
...
 
     # Standard S-mode CSRs
-    csrs = ["sstatus", "scause", "sie", "stvec", "scounteren", "senvcfg", "sscratch", "sepc", "stval", "sip"]
+    # Format: (CSR Name, IgnoreMask).  IgnoreMask specifies a set of bits to ignore
+
+    csrs = [
+        (
+            "sstatus",
+            0x000000040,
+        ),  # TODO: sail does not yet support UBE; mask it until available to avoid mismatches.  Delete mask when Sail has UBE support.
```
jordancarlin: The formatting of this is kind of hard to follow. Can you put the comment on its own line before each relevant CSR? That should prevent the strange line wrapping.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -406,7 +443,38 @@ def _generate_scsr_tests(test_data: TestData) -> list[str]:
...
+        #        ("scause", 0x7FFFFFFFFFFFFFF0),  # WLRL fields can't be managed with masks.  Use cp_scause_* instead
+        ("sie", None),
+        #        ("stvec", 0b1),  # stvec.MODE[0] can be 0 or 1. BASE is hard to predict with a reference model
```
jordancarlin: It is worth at least checking the top bit of MODE to ensure it is zero.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -564,36 +632,55 @@ def _generate_scsr_tests(test_data: TestData) -> list[str]:
...
-            _add_shadow(r1, r2, "sstatus", "mstatus", coverpoint, covergroup, test_data),
-            _add_shadow(r1, r2, "sie", "mie", coverpoint, covergroup, test_data),
-            _add_shadow(r1, r2, "sip", "mip", coverpoint, covergroup, test_data),
+            f"LI(x{r1}, 0x007FFFBF) # skip UBE, UXL bits which would cause weird behavior",
+            _add_shadow(
+                r1, r2, rmask, rsave, "mstatus", "sstatus", 0xCFFFFFFCF, coverpoint, covergroup, test_data
+            ),  # TODO: set mask to 0 when UXL and UBE are supported by Sail
+            _add_shadow(
+                r1, r2, rmask, rsave, "sstatus", "mstatus", 0xCFFFFFFCF, coverpoint, covergroup, test_data
+            ),  # TODO: set mask to 0 when UXL and UBE are supported by Sail
```
jordancarlin: Conflicting comments. Are we skipping those bits because they cause weird behavior or because Sail doesn't have sufficient configuration to get them to match?

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -564,36 +632,55 @@ def _generate_scsr_tests(test_data: TestData) -> list[str]:
...
+) -> str:
     """Helper function to generate shadow CSR test lines for writing wreg and reading rreg."""
     return str.join(
         "\n",
         [
             "",
-            f"# Testcase: shadow CSR test for writing {wreg} and reading {rreg}",
-            f"csrw {wreg}, x{r1}       # write all 1s to {wreg}",
+            f"# Testcase: shadow CSR test for writing {wreg} and reading {rreg} with mask 0x{mask:x}",
+            f"LI(x{rmask}, 0x{mask:x}) # mask.  Note that these are bits to keep, instead of the usual bits to ignore.  Comment in sigupd mask is incorrect",
```
jordancarlin: Shouldn't we just fix the comment then?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -298,53 +325,53 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
-        "mhpmevent23",
-        "mhpmevent24",
-        "mhpmevent25",
-        "mhpmevent26",
-        "mhpmevent27",
-        "mhpmevent28",
-        "mhpmevent29",
-        "mhpmevent30",
-        "mhpmevent31",
+        ("mstatus", None),  # will need a mask for endianness, UXL/SXL until Sail can match all fields
```
jordancarlin: Comment says needs a mask but the mask is set to None.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -298,53 +325,53 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+        ("medeleg", None),
+        ("mideleg", None),
+        ("mie", 0x3FFF),  # limit to implemented interrupt bits
```
jordancarlin: The spec says bits 15:0 of mie are the standard portion. Shouldn't we include all 16 of those bits?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -298,53 +325,53 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
-        "mhpmevent25",
-        "mhpmevent26",
-        "mhpmevent27",
-        "mhpmevent28",
-        "mhpmevent29",
-        "mhpmevent30",
-        "mhpmevent31",
+        ("mstatus", None),  # will need a mask for endianness, UXL/SXL until Sail can match all fields
+        ("medeleg", None),
+        ("mideleg", None),
```
jordancarlin: `mideleg` probably needs the same mask as `mie`.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -298,53 +325,53 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
-        "mhpmevent24",
-        "mhpmevent25",
-        "mhpmevent26",
-        "mhpmevent27",
-        "mhpmevent28",
-        "mhpmevent29",
-        "mhpmevent30",
-        "mhpmevent31",
+        ("mstatus", None),  # will need a mask for endianness, UXL/SXL until Sail can match all fields
+        ("medeleg", None),
```
jordancarlin: Bits 24-31 and 48-63 are "reserved for custom use". We should mask them out.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -298,53 +325,53 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+        ("mideleg", None),
+        ("mie", 0x3FFF),  # limit to implemented interrupt bits
+        #        ("mtvec", 0b1),  # tvec.MODE[0] can be 0 or 1. BASE is hard to predict with a reference model
+        ("mcounteren", None),
+        ("mscratch", None),
+        ("mepc", None),
+        #        ("mcause", None), # WLRL fields can't be handled with masks.  Use cp_mcause_* instead
+        ("mtval", None),
+        ("mip", 0x3FFF),  # limit to implemented interrupt bits
```
jordancarlin: Same as above. Shouldn't this be 0xFFFF?

jordancarlin (changes requested):

jordancarlin (approved):

PR #1434, opened 2026-05-05
Fix vector coverage

jacassidy:


tests/env/rvtest_macros_vector.h
```diff
@@ -40,5 +40,28 @@
...
+        #ifndef TEST_LMULf4_SUPPORTED
+            #define TEST_LMULf4_SUPPORTED
+        #endif
+    #endif
+    #if (RVTEST_SEW <= ELEN / 8)
+        #ifndef TEST_LMULf8_SUPPORTED
+            #define TEST_LMULf8_SUPPORTED
+        #endif
+    #endif
+#endif
```
jordancarlin, line 64: Should we drop these from the test generator then? Seems confusing to define them twice.

jordancarlin (changes requested):

jordancarlin (approved):

PR #1441, opened 2026-05-06
Fix misleading debug output when unpriv test traps

### LINKED ISSUE #1359: Misleading debug information when an unpriv test traps
**Describe the bug**
Now that the boot sequence is setting up a trap handler for unprivileged tests when M-mode is available, the trap handler will catch illegal instructions and terminate the simulation.  However, the debug messages printed are misleading.

When a trap happens on ref on an unpriv test during build, it should be reported as a build problem before proceeding, rather than generating a bad signature file.  When a trap happens on an unpriv test on DUT, that should also be reported.

**To Reproduce**
Run a test with unsupported instructions.  For example, in Issue #1358, UDB is putting Zfh in extensions.txt so sail-RVA23S64 is attempting to run Zfh instructions.  
make sail-RVA23S64

Note that this needs the boot PR to be merged.

```
This log file generated by running:
sail_riscv_sim --config config/sail/sail-RVA23S64/sail.json /home/harris/cvw/addins/riscv-arch-test-dh/work/sail-RVA23S64/elfs/rv64i/ZfhD/ZfhD-fcvt.w.h-00.elf


RVCP-SUMMARY: TEST FAILED - Test File "ZfhD-fcvt.w.h-00.S"
RVCP: Test Info: "test: 1; cg: ZfhD_fcvt.w.h_cg; cp: cp_fs1_badNB_D_H; bin: b0xffffffff00000000"
RVCP: Instruction: 0xc408fc53
RVCP: Approximate address (failure may be slightly after this): 0x000000008000028c
RVCP: Register: x24
RVCP: Bad Value:      0x5bfddb7d5bfddb7d

HTIF located at 0x80024400
begin_signature: 0x80006d40
end_signature: 0x80024350
Entry point: 0x80000000
FAILURE: 1
```

Jjateen:
When the M-mode trap handler aborted a test due to an unsupported instruction, failure_type 3 fell through to failedtest_saveresults_int, which reconstructed a non-existent beq/ld instruction pair and filled failing_reg, failing_value, and expected_value with garbage. The reporter then printed those garbage values as "Bad Value" / "Expected Value", making the output look like a register comparison failure rather than a trap abort.

Two fixes in rvtest_failure_code.h:
- Add a dispatch for failure_type == 3 at failedtest_saveresults that branches to a new failedtest_saveresults_trap label, bypassing the integer beq/ld reconstruction entirely and jumping straight to the common string-pointer loading path.
- After printing "RVCP: Test Info:" in failedtest_report, check for failure_type == 3 and branch directly to failedtest_report_traphandler, skipping the instruction / approximate address / register / bad-value / expected-value fields that are meaningless for a trap abort.

After the fix the output for a trapped test is:

  RVCP-SUMMARY: TEST FAILED - Test File "..."
  RVCP: Address of instruction that trapped (XEPC): 0x...
  RVCP: Instruction that trapped: 0x...

This applies to both the ref-model build step (where sail now produces a clear trap message before exiting non-zero) and to the selfcheck ELF running on the DUT.

Fixes #1359.


davidharrishmc (approved):

Zain2050:
Hi @Jjateen. Thanks for proposing this fix. However, by this fix, you have broken the entire trap handler error reporting mechanism. By these changes you are successfully skipping unwanted info during test abortion. However, you are also doing the same for every mismatch in the trap handler. For example
```
RVCP-SUMMARY: TEST FAILED - Test File "sv39_zicbom_exceptions_Smode.S"
RVCP: Test Info: "Mismatch in scause value! Trap was being handled in S-Mode."
RVCP: Address of instruction that trapped (XEPC): 0x0000000180001714
RVCP: Instruction that trapped: 0x0017a00f
```
Now, we are unable to know what the mismatched scause value was. 

You also mentioned that 

> which reconstructed a non-existent beq/ld instruction pair and filled failing_reg, failing_value, and expected_value with garbage.

That beq/ld is not non-existent. It is in `rvtest_setup.h` and it intentionally has that special value `0xBAD0DEAD`, indicating test abortion. Please revert this change and propose a different solution. Thanks

+ @davidharrishmc , @jordancarlin

Jjateen:
@Zain2050 You're right, the fix was too broad. `failure_type == 3` covers both trap aborts and trap handler mismatches (e.g., scause). The beq/ld in `rvtest_setup.h` is intentional and correctly sets `failing_value = 0xBAD0DEAD` for the abort case, so the right discriminator is `failing_value == 0xBAD0DEAD`, not `failure_type`.

Fix: revert the `failedtest_saveresults_trap` dispatch so all type-3 failures go through `failedtest_saveresults_int` as before. In `failedtest_report`, replace the `failure_type == 3` check with `failing_value == 0xBAD0DEAD`. Abort cases skip the value fields; mismatch cases (actual values) still print them.

I've opened a new PR to revert this

Zain2050:
Thanks for the quick fix.

PR #1447, opened 2026-05-07
fix(testgen): add frm cross-product support to cp_fs2_edges and cp_fs3_edges

dev-aditya-hub:
## Summary

`cp_fs2_edges` and `cp_fs3_edges` were missing the frm cross-product logic that their sibling `cp_fs1_edges` already has. When a coverpoint name contains `_frm`, the generator is expected to iterate over all six rounding modes and emit one test bin per edge value per mode. Without this, any test plan referencing `cp_fs2_edges_frm` or `cp_fs3_edges_frm` would silently produce tests with `frm=None`. The assembler defaults to DYN in that case, and since `fcsr.frm` is at its reset value of RNE, a DUT that hardwires RNE computes the correct result and passes every bin without being asked to use any other rounding mode.

## Fix

Added `cross_frm = "_frm" in coverpoint` and the inner `for frm_mode in frm_modes` loop to both `cp_fs2_edges` and `cp_fs3_edges`, matching the structure already present in `cp_fs1_edges`. Bin names and description strings now include the frm suffix when a rounding mode is active, consistent with the existing pattern.

## Files Changed

- `coverpoints/cp_fp_reg_edges.py` — add frm cross-product logic to `make_fs2_edges` and `make_fs3_edges`

## Test plan

- [x] Coverpoints containing `_frm` now produce six bins per edge value from `cp_fs2_edges` and `cp_fs3_edges`, matching the output of `cp_fs1_edges` for the same input
- [x] Coverpoints without `_frm` still produce one bin per edge value with `frm=None`, identical to the previous behavior
- [x] Existing YAML test plans that use `cp_fs2_edges` without the `_frm` suffix are unaffected

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/cp_csr_frm.py
+++ b/generators/testgen/src/testgen/coverpoints/cp_csr_frm.py
@@ -21,15 +21,16 @@ def make_frm(instr_name: str, instr_type: str, coverpoint: str, test_data: TestD
     if coverpoint != "cp_csr_frm":
         raise ValueError(f"Unknown cp_csr_frm coverpoint variant: {coverpoint} for {instr_name}")
 
-    # csr frm modes 0-4, end at 0 so the rest of the test continues in rne
-    frm_modes = (("rmm", 4), ("rup", 3), ("rdn", 2), ("rtz", 1), ("rne", 0))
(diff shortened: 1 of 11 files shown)


dev-aditya-hub:
Reopening from a clean branch rebased on current upstream/act4 — the previous branch had stale commits that bloated the diff.

PR #1448, opened 2026-05-07
fix(testgen): remove duplicate sigupd_count increment in fs_type formatter

dev-aditya-hub:
## Summary

The `fs_type` formatter was incrementing `test_data.test_chunk.sigupd_count` manually on the line immediately before calling `write_sigupd` for fflags capture. `write_sigupd` already increments `sigupd_count` internally, so the explicit increment was a duplicate that left the counter one ahead of the actual number of SIGUPD entries in the output. This caused the signature buffer to be over-allocated by one slot for every float store test case.

## Fix

Remove the redundant manual `sigupd_count += 1`. The `write_sigupd` call on the next line handles the count correctly on its own, consistent with every other formatter in the codebase.

## Files Changed

- `formatters/types/fs_type.py` — remove duplicate `test_data.test_chunk.sigupd_count += 1`

## Test plan

- [x] Float store test cases now produce a `sigupd_count` that matches the number of actual `write_sigupd` calls in the generated check sequence
- [x] Behavior for all other formatter types is unaffected since none of them had this manual pre-increment pattern

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/cp_csr_frm.py
+++ b/generators/testgen/src/testgen/coverpoints/cp_csr_frm.py
@@ -21,15 +21,16 @@ def make_frm(instr_name: str, instr_type: str, coverpoint: str, test_data: TestD
     if coverpoint != "cp_csr_frm":
         raise ValueError(f"Unknown cp_csr_frm coverpoint variant: {coverpoint} for {instr_name}")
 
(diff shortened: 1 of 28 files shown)


dev-aditya-hub:
Reopening from a clean branch rebased on current upstream/act4 — the previous branch had stale commits that bloated the diff.

PR #1449, opened 2026-05-07
fix(testgen): route cp_csr_frm through formatter pipeline with explicit csr_frm_val

dev-aditya-hub:
## Summary

cp_csr_frm had three issues that this PR addresses:

- **Old tc.code patching** — it was still using the direct tc.code prepend approach that was flagged in #1415, instead of going through the formatter pipeline
- **Ordering dependency for cleanup** — the frm modes tuple was intentionally ordered to end at rne=0 as a cleanup hack; if the order ever changed, fcsr.frm would be left dirty
- **Implicit rm=dyn encoding** — the instruction had no explicit rm field, relying on the GAS assembler defaulting to rm=111 when rm is omitted, which is a convention not a guarantee

## Fix

Add `csr_frm_val: int | None` to `InstructionParams`. When set, the formatter uses that specific value for `fsrmi` instead of `random.choice([1, 2, 3, 4])`. This lets `cp_csr_frm` pass `frm="dyn"` (explicit rm=111 encoding) alongside a specific `csr_frm_val` for each of the five `fcsr.frm` values it needs to test.

The formatter now handles both the `fsrmi <val>` setup and `fsrmi 0x0` restore in check — no ordering dependency, no tc.code patching.

## Files Changed

- `data/params.py` — add `csr_frm_val` field
- `formatters/types/fr_type.py`, `fr4_type.py`, `fi_type.py`, `f2x_type.py`, `x2f_type.py` — use `csr_frm_val` when set
- `coverpoints/cp_csr_frm.py` — rewrite to use formatter pipeline

## Test plan

- [x] `cp_csr_frm` produces five bins, one per valid `fcsr.frm` value (rne=0 through rmm=4), each with `frm=dyn` and the matching `fsrmi` setup and `fsrmi 0x0` restore in the generated assembly
- [x] Passing `csr_frm_val` explicitly through `generate_random_params` prevents the random fallback, so the CSR value written before each test is always the one being tested
- [x] All five formatter types assert `csr_frm_val is not None` before using it, so a misconfigured call fails loudly rather than silently using a stale CSR value


dev-aditya-hub:
Closing — these changes (csr_frm_val field, formatter pipeline routing for cp_csr_frm) are already present in upstream/act4 via the maintainer commits. No action needed.

PR #1451, opened 2026-05-07
fix(testgen): remove duplicate sigupd_count increment in fs_type formatter

dev-aditya-hub:
## Summary

The `fs_type` formatter was incrementing `test_data.test_chunk.sigupd_count` manually on the line immediately before calling `write_sigupd` for fflags capture. `write_sigupd` already increments `sigupd_count` internally, so the explicit increment was a duplicate that left the counter one ahead of the actual number of SIGUPD entries written. This caused the signature buffer to be over-allocated by one slot for every float store test case.

## Fix

Remove the redundant manual `sigupd_count += 1`. The `write_sigupd` call on the next line handles the count correctly on its own, consistent with every other formatter in the codebase.

## Files Changed

- `formatters/types/fs_type.py` — remove duplicate `test_data.test_chunk.sigupd_count += 1`

## Test plan

- [x] Float store test cases now produce a `sigupd_count` that matches the number of actual `write_sigupd` calls in the generated check sequence
- [x] Behavior for all other formatter types is unaffected since none of them had this manual pre-increment pattern

### DIFF
--- a/generators/testgen/src/testgen/formatters/types/fs_type.py
+++ b/generators/testgen/src/testgen/formatters/types/fs_type.py
@@ -76,6 +76,5 @@ def format_fs_type(
         "#endif",
     ]
     assert test_data.test_chunk is not None
-    test_data.test_chunk.sigupd_count += 1
     check.append(write_sigupd(None, test_data, "fflags"))
     return (setup, test, check)
--- a/tests/rv32i/D/D-fsd-00.S
+++ b/tests/rv32i/D/D-fsd-00.S
@@ -17,7 +17,7 @@
 
 // Define the name of this test file for reporting and the size of the signature region
 #define TEST_FILE "D-fsd-00.S"
-#define SIGUPD_COUNT 343
+#define SIGUPD_COUNT 232
 
 // Extra defines (if needed) to enable trap handlers, fpu, vector, etc.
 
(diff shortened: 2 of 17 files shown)


generators/testgen/src/testgen/formatters/types/fs_type.py
```diff
@@ -76,6 +76,5 @@ def format_fs_type(
         "#endif",
     ]
     assert test_data.test_chunk is not None
-    test_data.test_chunk.sigupd_count += 1
     check.append(write_sigupd(None, test_data, "fflags"))
```
jordancarlin, line 79: The manual increment of `sigupd_count` is to account for the direct store to the signature region by the actual test code (where it says "perform store"). The `write_sigupd` is a second store to the signature region to check for fflags. Without the manual increment of `sigupd_count` there will not be enough spaces reserved in the signature region.

jordancarlin (changes requested):
I don't think this is correct. The extra `sigupd_count` increment is necessary. See detailed explanation below.

dev-aditya-hub:
You're right, Jordan. The manual `sigupd_count += 1` is necessary to account for the direct store by the test instruction itself, separate from the `write_sigupd` call for fflags. Closing this PR as the original code is correct.

PR #1468, opened 2026-05-09
Updated configs for InterruptsSstc and InterruptsSm running on different DUTs

SadhviNarayanan:
Updated the configs for the DUTs to get InterruptsSstc and InterruptsSm running on them.
Currently, we have - 
InterruptsSstc running in SAIL, Spike, qemu, Imperas
InterruptsSm running on SAIL, Spike, qemu


config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -171,12 +171,45 @@
...
-#define RVMODEL_CLR_MEXT_INT(_R1, _R2)
+#define PLIC_BASE_ADDRESS    0x0c000000
+#define PLIC_ENABLE_ADDRESS  0x0c002000
+#define PLIC_THRESH_ADDRESS  0x0c200000
+#define PLIC_CLAIM_ADDRESS   0x0c200004
+#define NS16550_BASE_ADDRESS 0x10000000
+#define UART_INT_SRC         10           /* NS16550 interrupt source ID in QEMU virt */
+
+/* Set machine external interrupt via PLIC + NS16550 UART transmitter interrupt */
+#define RVMODEL_SET_MEXT_INT(_R1, _R2)          \
```
davidharrishmc, line 201: Can you add some comments beforehand explaining how this works?  Along the lines of:

* Generates external interrupt by causing UART to throw an interrupt
* PLIC receives this interrupt and relays it to QEMU Machine External Interrupt input
* Configures PLIC UART_INT_SRC to priority 7
* Enables UART_INT_SRC in PLIC
* Sets priority threshold to 0
* Set UART IER.ETBEI 
*     Interrupt Enable Register Enable Transmitter Holding Register Empty Interrupt
* The UART throws an interrupt because the THR defaults to empty
* The PLIC sees that the UART source is enabled and has priority greater than the threshold
* PLIC asserts Machine External Interrupt

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -171,12 +171,45 @@
...
+  li _R2, PLIC_ENABLE_ADDRESS;                   \
+  sw _R1, 0(_R2);                                \
+  li _R2, PLIC_THRESH_ADDRESS;                   \
+  sw zero, 0(_R2);                               \
+  li _R1, 0x02;                                  \
+  li _R2, NS16550_BASE_ADDRESS;                  \
+  sb _R1, 1(_R2);
+
+/* Clear machine external interrupt: disable UART, then claim/complete in PLIC */
+#define RVMODEL_CLR_MEXT_INT(_R1, _R2)          \
```
davidharrishmc, line 217: * Turn off UART interrupt by disabling IER.ETBEI
* Read the PLIC Claim register and write it back to deassert Machine External Interrupt.

config/qemu/qemu-RVI20U64/rvmodel_macros.h
```diff
@@ -171,12 +171,45 @@
...
-#define RVMODEL_CLR_MEXT_INT(_R1, _R2)
+#define PLIC_BASE_ADDRESS    0x0c000000
+#define PLIC_ENABLE_ADDRESS  0x0c002000
+#define PLIC_THRESH_ADDRESS  0x0c200000
+#define PLIC_CLAIM_ADDRESS   0x0c200004
+#define NS16550_BASE_ADDRESS 0x10000000
+#define UART_INT_SRC         10           /* NS16550 interrupt source ID in QEMU virt */
+
+/* Set machine external interrupt via PLIC + NS16550 UART transmitter interrupt */
+#define RVMODEL_SET_MEXT_INT(_R1, _R2)          \
```
davidharrishmc: RVI20U64 doesn't test interrupts, so maybe it is better to leave this blank instead of running the risk of it getting out of sync with qemu-rv64-max.

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -102,16 +102,50 @@
 
 ##### Machine Interrupts #####
 
-#define RVMODEL_INTERRUPT_LATENCY 10
+#define RVMODEL_INTERRUPT_LATENCY 2000
 
 #define RVMODEL_TIMER_INT_SOON_DELAY 100
```
davidharrishmc, line 107: Does it need to be so long? Just needs to be long enough to get to the next thing happening.  Right now you are giving 100 * 200 =20,000 cycles.  Maybe 10 would be plenty? This applies to all of the other configs with long waits.

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -102,16 +102,50 @@
...
-
-#define RVMODEL_CLR_MEXT_INT(_R1, _R2)
+#define PLIC_BASE_ADDRESS    0x0c000000
+#define PLIC_ENABLE_ADDRESS  0x0c002000
+#define PLIC_THRESH_ADDRESS  0x0c200000
+#define PLIC_CLAIM_ADDRESS   0x0c200004
+#define NS16550_BASE_ADDRESS 0x10000000
+#define UART_INT_SRC         1            /* NS16550 interrupt source ID in Spike */
+
+/* Set machine external interrupt via PLIC + NS16550 UART transmitter interrupt */
```
davidharrishmc: Copy over same comments from QEMU.

config/spike/spike-RVI20U64/rvmodel_macros.h
```diff
@@ -106,12 +106,46 @@
...
-#define RVMODEL_CLR_MEXT_INT(_R1, _R2)
+#define PLIC_BASE_ADDRESS    0x0c000000
+#define PLIC_ENABLE_ADDRESS  0x0c002000
+#define PLIC_THRESH_ADDRESS  0x0c200000
+#define PLIC_CLAIM_ADDRESS   0x0c200004
+#define NS16550_BASE_ADDRESS 0x10000000
+#define UART_INT_SRC         1            /* NS16550 interrupt source ID in Spike */
+
+/* Set machine external interrupt via PLIC + NS16550 UART transmitter interrupt */
+#define RVMODEL_SET_MEXT_INT(_R1, _R2)          \
```
davidharrishmc: All of these could be blank because they are not used.

Makefile
```diff
@@ -15,7 +15,7 @@ COVERAGE_CONFIG_FILES ?= config/sail/sail-rv64-max/test_config.yaml config/sail/
 #  - Sm, S: Insufficient WARL configuration options.
 #  - InterruptsSm,InterruptsS,InterruptsU,PMPSm,PMPZca,SvaduPMP,SvPMP,SvPMPZicbo: Additional testing needed on a wider range of configs. Some missing config options to match ref model.
 EXTENSIONS  ?=
-EXCLUDE_EXTENSIONS ?= Sm,Sv,SvaduPMP,SvPMP,SvPMPZicbo,Svade,Svadu,Svinval,SvZicbo,Svnapot,Svpbmt,InterruptsSm,ExceptionsSvZalrsc,ExceptionsSvZaamo,PMPS,PMPU,PMPSm
+EXCLUDE_EXTENSIONS ?= Sm,S,InterruptsSm,InterruptsS,InterruptsU,InterruptsSstc,Sv,SvaduPMP,SvPMP,SvPMPZicbo,Svade,Svadu,Svinval,SvZicbo,Svnapot,Svpbmt,ExceptionsSvZalrsc,ExceptionsSvZaamo,ExceptionsZalrsc,ExceptionsZaamo,PMPS,PMPU,PMPSm
```
davidharrishmc: ExceptionsZalrsc and ExceptionsZaamo are not in the Makefile currently.  Don't add them because that will prevent important tests from running.

Although Jordan said to leave interrupts in the Makefile exclusion, they weren't there to begin with.  I don't think we should add them back in.  We want to know that they are failing on certain DUTs.  

So I would leave the Makefile as it was except delete InterruptsSm.

davidharrishmc (changes requested):
This looks really good! Ready to go with minor changes.

config/spike/spike-rv64-max/rvmodel_macros.h
SadhviNarayanan: I Reverted RVMODEL_INTERRUPT_LATENCY back to 10.

For the timer spin (RVMODEL_TIMER_INT_SOON_DELAY * 200 = 20,000 iterations), I tried reducing RVMODEL_TIMER_INT_SOON_DELAY to 10 but the timer interrupt tests failed. The 200x multiplier is
needed because I think Spike increments mtime only once every 100 instructions. With TIMER_INT_SOON_DELAY = 100 ticks as the stimecmp offset, we need at least 100 × 100 = 10,000
instructions for the timer to fire — the 20,000-iteration spin (40,000 instructions) gives a 4× margin. QEMU has a similar mismatch between instruction rate and CLINT tick rate.

davidharrishmc: Doesn't Spike have a very long interrupt latency?  You may need to leave it at 2000 to work reliably on other test suites?

config/spike/spike-RVI20U64/rvmodel_macros.h
```diff
@@ -106,12 +106,46 @@
...
 #define RVMODEL_TIMER_INT_SOON_DELAY 100
 
+// Spike ticks the CLINT timer every 100 instructions (default --insns-per-tick).
+// The default RVTEST_IDLE_FOR_TIMER_INTERRUPT spins RVMODEL_TIMER_INT_SOON_DELAY iterations
+// (200 instructions), which only advances mtime by ~2 ticks — far less than the 100-tick
+// stimecmp offset. Override with a 200x multiplier so the spin outlasts the timer delay.
+#define RVTEST_IDLE_FOR_TIMER_INTERRUPT(_R1) \
+    LI(_R1, RVMODEL_TIMER_INT_SOON_DELAY * 200); \
+    99: addi _R1, _R1, -1; \
+        bnez _R1, 99b;
```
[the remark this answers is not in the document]
davidharrishmc: Are warnings occurring?  Let's fix if they are, or skip if not.

davidharrishmc (changes requested):

davidharrishmc (approved):

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -102,16 +102,60 @@
...
+#define RVMODEL_INTERRUPT_LATENCY 2000
 
 #define RVMODEL_TIMER_INT_SOON_DELAY 100
 
+// Spike ticks the CLINT timer every 100 instructions (default --insns-per-tick).
+// The default RVTEST_IDLE_FOR_TIMER_INTERRUPT spins RVMODEL_TIMER_INT_SOON_DELAY iterations
+// (200 instructions), which only advances mtime by ~2 ticks — far less than the 100-tick
+// stimecmp offset. Override with a 200x multiplier so the spin outlasts the timer delay.
+#define RVTEST_IDLE_FOR_TIMER_INTERRUPT(_R1) \
+    LI(_R1, RVMODEL_TIMER_INT_SOON_DELAY * 200); \
```
jordancarlin, line 114: @SadhviNarayanan @davidharrishmc we don’t want users overriding RVTEST macros. Would this work if we just set RVMODEL_TIMER_INT_SOON_DELAY higher in this file? If not, this needs to become another RVMODEL macro instead.

PR #1472, opened 2026-05-09
Improve HTIF failure reporting in build error output (#1360)

### LINKED ISSUE #1360: Poor reporting in build/test.sig.log
**Describe the bug**
When the build process fails, the information in build/test.sig.log is unhelpful.

**To Reproduce**
Not easy - need a bug that prevents building.

** Report **
FAILURE: 100664112 isn't meaningful

```
HTIF located at 0x8002ff00
begin_signature: 0x80011d20
end_signature: 0x8002feb0
Entry point: 0x80000000
FAILURE: 100664112
using /home/harris/cvw/addins/riscv-arch-test-dh/work/sail-rv64-max/build/priv/UF/UF-00.sig for test-signature output.
setting signature-granularity to 8 bytes
```

Jjateen:
When the Sail reference model exited with an unexpected tohost value (e.g. "FAILURE: 100664112"), the decimal number was not meaningful to the user and the surrounding Sail boilerplate lines added noise.

Changes in build.py:

- Add "HTIF located at" and "Entry point" to the noise-strip patterns so they are omitted from the failure block in the console output (they are still preserved in the .sig.log file).
- Add _HTIF_FAILURE_RE and _annotate_htif_failure() to rewrite "FAILURE: N" as "FAILURE: N (0xHHHHHHHH)" in the displayed output, making the tohost value immediately readable in hex.
- When a HTIF FAILURE line is detected, append a hint directing the user to re-run with --debug and inspect the .sig.trace file, which is the recommended debugging path for unexpected tohost values.

The .sig.log file content is unchanged; only the console error block is affected.

Fixes #1360.


framework/src/act/build.py
```diff
@@ -418,6 +428,16 @@ def _strip_noise(output: str) -> str:
...
 
 
+def _annotate_htif_failure(output: str) -> str:
+    """Add hex representation to HTIF FAILURE lines for readability."""
+
+    def _sub(m: re.Match[str]) -> str:
+        val = int(m.group(2))
+        return f"{m.group(1)}{val} (0x{val:08x})"
+
+    return _HTIF_FAILURE_RE.sub(_sub, output)
```
jordancarlin: I'd rather not rewrite the output of Sail. If hex is more intuitive here (and I think it is), then please open a PR on https://github.com/riscv/sail-riscv to output the failure in hex instead of converting the log here.

framework/src/act/build.py
```diff
@@ -374,6 +375,11 @@ def _print_failure(console: Console, task: BuildTask, error: BuildError, verbose
             output_lines = output_lines[-max_output_lines:]
         for line in output_lines:
             console.print(f"    {line}", soft_wrap=True, highlight=False)
+        if _HTIF_FAILURE_RE.search(error.output):
+            console.print(
+                "  [dim]Hint: unexpected HTIF tohost value — re-run with --debug and check the .sig.trace file.[/]",
```
jordancarlin: When using the Makefile (which most users will be doing), you need to set `DEBUG=True` instead of adding the `--debug` flag. This suggestion is also not specific to HTIF failures, so I'd rather not print it only for that kind of failure.

jordancarlin (changes requested):
See comments below. I wonder if the better solution is to just drop this PR and improve the output from the sail model directly instead.

framework/src/act/build.py
Jjateen: noted.

framework/src/act/build.py
Jjateen: done, please check https://github.com/riscv/sail-riscv/pull/1714

jordancarlin (approved):

PR #1481, opened 2026-05-11
fix(testgen): honor tail-agnostic policy for scalar-dst vector ops

### LINKED ISSUE #1473: ACT vector self-checking does not honor tail agnostic on operations that only write element 0
The following mismatch occurs during make whisper-rv64-max for Vx64 vmv.s.x.
Confidence is low, but I think what is going on is that Whisper is legally putting 1s
in the tail after element 0, and Sail is preserving the old elements.  Both behaviors are
legal, but the self-checker doesn't accept the 1s behavior.

Same self-checker issue for reduction operations where whisper is putting 1s in
the tail (after element 0) and self-checker is not accepting them.

`vredand.vs`, `vredmax.vs`, `vredmaxu.vs`, `vredmin.vs`, `vredminu.vs`,
`vredor.vs`, `vredsum.vs`, `vredxor.vs`

Report from Claude (seems to be a partial hallucination; I can't find the quote from the
RVVI spec).

### Failing instruction

`vmv.s.x`

### Root cause

`vmv.s.x vd, rs1` writes `rs1` to element 0 of `vd`.  The RVV 1.0 spec
(§16.1) is explicit: **"Other elements of the destination are not written by
these instructions."**  This is stronger than the tail-agnostic policy: those
elements must be **undisturbed** regardless of `vta`.

Whisper incorrectly applies its `tail_agnostic_policy: "ones"` to elements
1–VLMAX-1, overwriting them with `0xffffffffffffffff`.

**Example:** `vmv.s.x v29, x25`, VL=4, VTYPE=0x58 (SEW=64, LMUL=1, vta=1).

```
vsetvli x5, x0, e64,m1,ta,mu     # VL=4, vta=1
vle64.v v29, (x4)                 # pre-load v29 with test pattern
  # v29 before = ffffffffffffffff...a69592085d3631ab (element 1 = 0xa69592085d3631ab)
vmv.s.x v29, x25                  # <-- failing instruction
```

| Element | Expected | Whisper actual |
|---|---|---|
| 0 (written) | x25 value = `0x619245c0422a2070` | `0x619245c0422a2070` ✓ |
| 1–3 (not written) | old v29[1..3] (undisturbed) | `0xffffffffffffffff` ✗ |

**RVCP failure:**
```
RVCP-SUMMARY: TEST FAILED - Test File "vmv.s.x.S"
RVCP: Test Info: "test: 81; cp: Vx64_vmv.s.x_cg/cr_vl_lmul (Test lmul = 1, vl = vlmax)"
RVCP: Instruction: 0x420ceed7          # vmv.s.x v29, x25
RVCP: Register: v29  Region: ACTIVE  Element Index: 1
RVCP: VL: 0x4  VTYPE: 0x58
RVCP: Bad Value: 0xffffffffffffffff  Expected Value: 0xa69592085d3631ab
```

### Correct result

Element 0 of v29 gets `rs1`; elements 1–3 retain their pre-instruction
values.  The `tail_agnostic_policy` setting must not override the
architectural "not written" guarantee.

Jjateen:
vmv.s.x and reduction instructions (vredand.vs, vredmax.vs, vredmaxu.vs, vredmin.vs, vredminu.vs, vredor.vs, vredsum.vs, vredxor.vs and their widening variants) only write element 0 of the destination register. When vta=1 (tail-agnostic), implementations may legally write all-1s to elements 1..VLMAX-1, but the self-checker treated those elements as "active body elements" requiring an exact match against the reference, causing false failures (e.g. Whisper writing 0xffffffffffffffff to element 1 while Sail preserved the original value).

Root cause: RVTEST_SIGUPD_V_LEN built the active-element mask as (i < original_vl), so elements 1..vl-1 were checked with strict equality even though the instruction never writes them.

Fix:
- Add a _SCALAR_DST_FLAG parameter to RVTEST_SIGUPD_V_LEN in signature.h. When the flag is 1, the saved vl is overridden to 1 immediately after the csrr, so the active mask covers only element 0. Elements 1..VLMAX-1 become "tail" and receive the existing tail-agnostic relaxation: both the reference value and all-1s are accepted when vta=1, while vta=0 still requires the reference value. The non-SELFCHECK variant accepts the new parameter as a no-op (matching nops preserve instruction-count parity).
- Add scalar_dst parameter to writeSIGUPD_V in vector_testgen_common.py and pass _SCALAR_DST_FLAG=1 for vmv.s.x and all vredins (including widening reductions) in length-suite tests.

Fixes #1473.

### DIFF
--- a/generators/testgen/scripts/vector_testgen_common.py
+++ b/generators/testgen/scripts/vector_testgen_common.py
@@ -1519,7 +1519,7 @@ def writeSIGUPD_F(fd):
 #         writeLine(f"RVTEST_SIGUPD_V(x{sigReg}, x{tempReg}, {sew}, {offset}, v{vd})", f"# stores v{vd} (sew = {sew}, AVL = {avl}) in signature with base (x{sigReg}) and helper (x{tempReg}) register")
 
 
-def writeSIGUPD_V(inst_ptr, vd, sew, avl=1, sig_lmul = None, load_testline = None, sig_whole_register_store = False, vd_mask = False, testtype = "base", masked = False, lmul = 1):
+def writeSIGUPD_V(inst_ptr, vd, sew, avl=1, sig_lmul = None, load_testline = None, sig_whole_register_store = False, vd_mask = False, testtype = "base", masked = False, lmul = 1, scalar_dst = False):
 
     global sigupd_count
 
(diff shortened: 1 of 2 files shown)


Jjateen:
@davidharrishmc can you please review this?

davidharrishmc:
@georgiatai could you review?  (Georgia is the original author).

georgiatai (approved):

georgiatai:
Clean change. Looks good to me :)

davidharrishmc (approved):

PR #1499, opened 2026-05-14
Qemu fixed catching bad tests

jacassidy:


jacassidy:
Vx8,Vx16,Vx32,Vx64,Vls8,Vls16,Vls32,Vls64,Vf16,Vf32,Vf64,ExceptionsVf16,ExceptionsVf32,ExceptionsVf64,ExceptionsVx,ExceptionsVls,Smv,SmVF,UV,MisalignedV passing on spike

UV and MisalignedV32 failing on QEMU, UV is Qemu issue, misaligned is due to missing sail parameter support for indexEEw max (sail traps with wrong mcause)

jordancarlin:
> misaligned is due to missing sail parameter support for indexEEw max (sail traps with wrong mcause)

Can you open an issue for Sail if there isn't one already?

scan_uncovered.py
jordancarlin, line 1: I assume this wasn't meant to be checked in?

jordancarlin (changes requested):
One rogue file. Otherwise, LGTM.

scan_uncovered.py
jacassidy: I quite like using it so I threw it in but i can remove it

jacassidy:
> > misaligned is due to missing sail parameter support for indexEEw max (sail traps with wrong mcause)
> 
> Can you open an issue for Sail if there isn't one already?

Just opened, please let me know if I should provide more detail

https://github.com/riscv/sail-riscv/issues/1719

scan_uncovered.py
jordancarlin: If you think it is useful then it should be integrated into the existing coverpoint reporting code and not left as a standalone script in the root of the repo. Let's drop it from this PR and you can open a follow up PR that integrates it if you think that would be worthwhile.

jacassidy:
@jordancarlin ready for merge

jordancarlin (approved):

PR #1504, opened 2026-05-15
Generate `rvtest_config.{h/svh}` from UDB config file

### LINKED ISSUE #1322: Auto-derive `RVMODEL_MTVEC_ALIGN` from `MTVEC_BASE_ALIGNMENT_*` yaml params
**Is your feature request related to a problem? Please describe.**

Alignment of `mtvec.BASE` above 64 bytes must be declared in two files, in different units, with no cross-check:

1. `<dut>.yaml`: `MTVEC_BASE_ALIGNMENT_{DIRECT,VECTORED}` in **bytes**.
2. `rvmodel_macros.h`: `#define RVMODEL_MTVEC_ALIGN` in **log2(bytes)**.

Forgetting step 2 fails silently - no build error, no runtime trap. The `Mtrampoline` is under-aligned, RTL masks the `mtvec` write, `rvtest_init` falls into the in-place copy fallback, PC-relative `jal` targets inside the relocated table break, and the next synchronous exception jumps back into init. On CV32E20 this caused `ExceptionsZc-00` to re-execute its test body and produce a bogus signature mismatch that took a full waveform-debug session to diagnose. Fix was `#define RVMODEL_MTVEC_ALIGN 8`, matching the yaml's already-declared 256.

**Describe the solution you'd like**

Have the generator emit `RVMODEL_MTVEC_ALIGN` directly from the yaml so it's
never restated manually. E.g. an autogenerated header:

```c
#define RVMODEL_MTVEC_ALIGN <ceil(log2(max(DIRECT, VECTORED)))>
```

- Omit if neither param is set (fall back to the existing 64-byte default).
- Error at gen time if the yaml value isn't a power of two, or is below schema minimum.

**Describe alternatives you've considered**

- **Validation only:** keep the manual macro but have the generator error if it disagrees with the yaml. Flags the failure instead of removing it.
- **Runtime check in `rvtest_init`:** read `mtvec` back, halt with a distinctive signature on mismatch. Turns silent into loud but keeps the duplication.
- **Docs-only:** promote the existing note in `docs/crd/src/old/general.adoc:27`. Cheapest, but this incident shows docs alone aren't enough.

**Additional context**

If `rvmodel_macros.h` autogeneration is already planned, the ask is simply that `RVMODEL_MTVEC_ALIGN` is covered by it and validated (power-of-two, ≥ schema minimum) at gen time.

### LINKED ISSUE #1339: ExceptionsSm and ExceptionsZc Hang: mstatush (CSR 0x310) Not Implemented
**Test name**
ExceptionsSm, ExceptionsZc

**Describe the bug**
`rvtest_trap_handler.h` line 1267 unconditionally reads mstatush (CSR 0x310) on every RV32 M-mode trap. On CV32E40P, CSR 0x310 is not implemented — the decoder raises an illegal instruction exception on any access to it. This causes an infinite loop: first trap fires the handler, handler hits the csrr, takes another illegal instruction exception, re-enters the handler, repeat. The simulation times out.

**Expected behavior**
A guard macro (e.g. `RVMODEL_NO_MSTATUSH` in rvmodel_macros.h) would let per-core configs skip the read, defaulting to current behavior for cores that do implement it.

**Additional context**
Confirmed via waveform: mepc is stuck at 0x2ecc (`csrrs x9, 0x310, x0`) from the first trap entry and never changes. CV32E20 does not hang because its RTL handles `CSR_MSTATUSH` as read-only zero.

jordancarlin:
- Update UDB configs to be compatible with the latest version of UDB.
- Use the `udb-gen` gem to automatically generate `rvtest_config.{h/svh}` from the UDB config file. Resolves part of #1141.
  - Also generate `rvmodel_macros.svh` based on `rvmodel_macros.h` so that `RVMODEL_ACCESS_FAULT_ADDRESS` (and similar defines) are available in the coverpoints).
- Update many custom defines to use the UDB generated parameter versions. Resolves #1322.
- Gate use of `mstatush` on `#ifndef SM1P11P0_SUPPORTED`. Resolves #1339.
- Add an option to use a local version of UDB instead of the default released gem using an environment variable. This is useful for UDB development and will also help when testing unratified extensions that are not yet merged into UDB.


jordancarlin:
Draft until the next UDB release is published (should be in the next few days).

tests/rv64e/Zba/Zba-add.uw-00.S
```diff
@@ -12,7 +12,7 @@
 # REQUIRED_EXTENSIONS: ['E', 'Zba']
 # params:
 #   MXLEN: 64
-# MARCH: rv64e_zicsr_zba
+# MARCH: rv64e_zicsr_zifencei_zba
```
davidharrishmc, line 15: Why is zifencei required on unpriv tests?

tests/rv64e/Zbb/Zbb-max-00.S
```diff
@@ -12,7 +12,7 @@
 # REQUIRED_EXTENSIONS: ['E', 'Zbb']
 # params:
 #   MXLEN: 64
-# MARCH: rv64e_zicsr_zbb
+# MARCH: rv64e_zicsr_zifencei_zbb
```
davidharrishmc, line 15: Why are zicsr and zifencei required on unpriv tests?

config/cores/cve2/cv32e20/rvmodel_macros.h
```diff
@@ -7,6 +7,8 @@
 
 #define RVMODEL_DATA_SECTION
 
+#define STANDARD_SM_SUPPORTED
```
davidharrishmc, line 10: Is this becoming part of UDB (earlier you called it CONFORMING_SM_SUPPORTED)?

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -12,6 +12,8 @@
         .align 8; .global fromhost; fromhost: .dword 0;     \
         .popsection
 
+#define STANDARD_SM_SUPPORTED
```
davidharrishmc, line 15: If we have to put this in the rvmodel_macros, there should be good documentation about what it means, what effect it has on tests, and for what files it should be included.

config/cores/cve2/cv32e20/rvmodel_macros.h
jordancarlin: Eventually, yes (I hope), but that PR has not been merged yet. https://github.com/riscv/riscv-unified-db/pull/1808

tests/rv64e/Zba/Zba-add.uw-00.S
jordancarlin: The code that loads the trap handler (which is included in unpriv tests when `STANDARD_SM_SUPPORTED` is defined) uses a `fence.i` as part of the sequence when `mtvec` is not writable. This was incorrectly replaced with a `nop` before because `ZIFENCEI_SUPPORTED` was not defined. Note that this is just in the march string, not in the list of required extensions.

tests/rv64e/Zbb/Zbb-max-00.S
jordancarlin: Zicsr is needed in all of the march strings (not as a required extension) so that the boot code is able to run. We added this when you did the boot code refactor. See above comment for Zifencei.

config/spike/spike-rv64-max/rvmodel_macros.h
jordancarlin: We previously had it in `rvtest_config.h` and this PR just moves it to `rvmodel_macros.h` temporarily until UDB merges the PR linked above. I think whatever existing documentation we had for it should be sufficient for now. This PR does not change how it is used.

jordancarlin:
Added Zicsr, PMPS, and PMPU to the cvw exclusions. The problem with Zicsr is tracked in #1538. The PMPS/PMPU issue is not immediately obvious and will be resolved later. I don't want it to block getting this in.

jordancarlin:
@davidharrishmc this should be good to go once CI passes.

davidharrishmc (commented):
I approve but the approve box for this PR (and none others) is grayed out for me.

UmerShahidengr (approved):
LGTM

davidharrishmc:

davidharrishmc:
@UmerShahidengr  resolving merge conflicts.

jordancarlin:
Merge conflicts resolved. Should be good to go once CI passes.

UmerShahidengr:
Thank you @jordancarlin , I was also in the middle of updating it, but you were quicker

jordancarlin:
> Thank you @jordancarlin , I was also in the middle of updating it, but you were quicker

Figured it would be easier since I am more familiar with the rest of the changes and what would need to be kept for each conflict. Thanks for working on it though.

PR #1506, opened 2026-05-15
Fix RVCP-SUMMARY parsing with VERBOSE logs

### LINKED ISSUE #1490: make cvw fails when VERBOSE=True
VERBOSE puts stuff in the log that confuses the checker and prints fail on every simulation.

somiljain2006:
Fixes #1490 

Hardened RVCP summary parsing in run_tests.py by relaxing the summary regex and using the last RVCP-SUMMARY occurrence found in the log. This prevents false failures when verbose/debug simulator output adds
additional log content before or around the summary line while VERBOSE=True is enabled.

### DIFF
--- a/run_tests.py
+++ b/run_tests.py
@@ -21,7 +21,10 @@ from functools import partial
 from multiprocessing import Pool
 from pathlib import Path
 
-_SUMMARY_RE = re.compile(r'RVCP-SUMMARY: TEST (PASSED|FAILED|SIGRUN) - Test File ".*"')
+_SUMMARY_RE = re.compile(
+    r"^RVCP-SUMMARY: TEST (PASSED|FAILED|SIGRUN)\b.*$",
+    re.MULTILINE,
+)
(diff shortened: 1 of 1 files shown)


jordancarlin:
I am unable to reproduce the bug mentioned in #1490. Unless we are able to reproduce it, I'd rather avoid changing this code. I will leave this open until we confirm that bug.

davidharrishmc:
I don't think this will fix the root cause of #1490  when ImperasDV and Questa are running, and I'm not sure what other issue it fixes.  Can you provide logs with and without this code showing how this code fixes a specific problem? Otherwise, suggest closing.

somiljain2006:
Thanks for the feedback.

You're right I wasn't able to reproduce the ImperasDV/Questa failure locally, so I can't confidently show before/after logs demonstrating that this fixes the actual root cause of issue.

The change was based on hardening the parsing logic against additional verbose/debug log content and multiple RVCP-SUMMARY occurrences, but without access to a working ImperasDV/Questa setup, I can't properly validate it against the reported failure mode.

PR #1510, opened 2026-05-15
Scan uncovered added to makefile

jacassidy:


jordancarlin (changes requested):
Rather than adding all of this infrastructure, can we just integrate this as a function in the existing build pipeline? Add it to coverreport.py and have it produce _overall_uncovered.txt similar to the existing _overall_summary.txt. Unless it is slow, I don’t see a reason to guard this behind a flag. We can always produce the report.

davidharrishmc:
I'm not certain we actually want this script in the flow.  It seems like a very useful feature in Kaden's AI-driven coverage of the complex vector coverpoints.  Once that is complete, I'm not sure anyone else will be doing similar work and will need this report. I'm concerned about whether it adds to runtime for features nobody else will use, interferes with using VCS for coverage reports, or has any long-term maintainability costs.  Kaden has a lot on his plate about finishing vector, and I'm also not sure this is the most important use of time.  Willing to be convinced otherwise.

davidharrishmc:
@jacassidy suggest closing this PR and keeping it in your personal files.  If you think it is general enough that everyone ought to be running, say so, otherwise please close.

jacassidy:
I personally think it is, but this is not a difficult script to reproduce if it becomes more clear this report would be useful in the future

PR #1513, opened 2026-05-16
fix: use 0xBAD0DEAD sentinel to suppress abort-only debug fields (#1359)

Jjateen:
The fix in #1441 was too broad: it dispatched ALL failure_type==3 cases away from failedtest_saveresults_int, losing the register/value data that the trap handler correctly sets for real mismatches (e.g., scause mismatches in sv39 tests). The same over-broad check in failedtest_report caused those mismatch values to be silently dropped.

failure_type==3 covers two distinct sub-cases:
- Trap abort: abort_test sets T5=0xBAD0DEAD before branching into the fail path. failedtest_saveresults_int extracts T5 as rs2 of the beq, so failing_value is 0xBAD0DEAD.
- Trap handler mismatch (e.g., scause): the trap handler calls the fail macro with actual register values; failing_value is the real bad value.

Fix: revert the failedtest_saveresults_trap dispatch so all type-3 failures go through failedtest_saveresults_int (restoring mismatch data). In failedtest_report, replace the failure_type==3 guard with a failing_value==0xBAD0DEAD check. Abort cases skip the instruction/ register/value fields cleanly; mismatch cases continue to print them.


Jjateen:
@Zain2050 can you review this fix?

Zain2050 (approved):

davidharrishmc (approved):

PR #1518, opened 2026-05-17
Removed fixed suites from Whisper CI exclusion

davidharrishmc:
Many whisper fixes today now can pass in CI.


jordancarlin (approved):
LGTM once Copilot comments are addressed.

PR #1520, opened 2026-05-17
fix(tests): replacing nop with with a 100-iteration retry loop

aclfe:
Addresses issue https://github.com/riscv/riscv-arch-test/issues/1399

SC tests before lacks a retry loop, the RISC-V spec only guarantees a LR/SC sequence will eventually succeed if you actually retry. this means any failure, for exapmle interrupt, cache coherence event, etc, would silently record a wrong value with no recovery path. The sc_type.py nop workaround for a Spike bug is also resolved by the proper retry loop.

This address problems by same retry loop, fix is simple

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/special/cp_custom_sc.py
+++ b/generators/testgen/src/testgen/coverpoints/special/cp_custom_sc.py
@@ -37,14 +37,24 @@ def make_custom_sc(instr_name: str, instr_type: str, coverpoint: str, test_data:
             and params.rs2val is not None
             and params.temp_reg is not None
         )
+        label_line = test_data.add_testcase(suffix, "cp_custom_aqrl")
+        label = test_data.current_testcase_label
+        retry_label = f"{label}_retry"
+        success_label = f"{label}_success"
         test_lines.extend(
             [
                 f"# Testcase: cp_custom_aqrl with suffix '{suffix}'",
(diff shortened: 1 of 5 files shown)


aclfe:
Open to any discussion or changes, do let me know thanks!

coverpoints/norm/Zalrsc.yaml
```diff
@@ -78,13 +78,13 @@ normative_rule_definitions:
     coverpoint: ["Untested"]
 
   - name: constrained_lrsc_instruction_set
-    coverpoint: ["Untested"]
+    coverpoint: ["Zalrsc_sc_w_cg/cp_custom_sc_lr, Zalrsc_sc_d_cg/cp_custom_sc_lr; zero instructions between LR and SC in all constrained sequences"]
```
davidharrishmc: The loop does not actually fully test this normative rule.  Better to leave rule listed as untested, unless adding a test that actually fully exercises it.  There are a lot of legal instructions that could go in the loop, so it is hard to test.

davidharrishmc (changes requested):

davidharrishmc:
This is a good change to add the loop.  Minor feedback about the normative rule.

aclfe:
Made changes

davidharrishmc (approved):

davidharrishmc (approved):

jordancarlin (approved):

PR #1525, opened 2026-05-18
Fix Imperas 32-bit S tests

davidharrishmc:
Exclude mscontext from cp_csr_insufficient_priv because it is s-accessible if implemented


Teemooooooooo (approved):

coverpoints/priv/S_coverage.svh
```diff
@@ -251,7 +251,9 @@ covergroup S_scsr_cg with function sample(ins_t ins);
 
     csr_machine: coverpoint ins.current.insn[31:20]  {
         bins machine_0[] = {[12'h300:12'h3FF]};
-        bins machine_1[] = {[12'h700:12'h7FF]};
+        bins machine_1a[] = {[12'h700:12'h7A9]};
+        // ignore mscontext at 0x7aa, which is readable from S mode and tested in Sstrict and Sdtrig
+        bins machine_1b[] = {[12'h7AB:12'h7FF]};
```
jordancarlin: I think you can put two ranges in the same bins to avoid needing different names.

jordancarlin (approved):
One minor improvement, but LGTM.

PR #1528, opened 2026-05-18
Integrate Cover-Float Tests in ACT4

### LINKED ISSUE #390: FP floating point corner case isn't tested
There is a strange corner case in FP arithmetic where a result can be normal, but the underflow flag is set.

This occurs for any result value that has exp=0 and mantissa 1.FFFFFF.000000 (through 1.FFFFF.7FFFFF ( I think).
When that result is rounded, ( depending on the rounding mode for the larger values), no rounding happens.
At this point, exponent is zero, so  it's a denorm, and (IEEE and RV spec) underflow flag is set
But now the mantissa must be to be converted to a denorm.
The mantissa is shifted by one (so the "hidden bit is no longer hidden), and that shifts out the LSB - which is a "1"
Now it must round again, and adding back rounding bit (again depending on rounding mode) kicks the result back into normalized territory (the smallest normal number)

Multiply doesn't cover this case, and (because it was found in the wild) divide doesn't either.
I have no clue whether add, subtract, FMA, convert, or sqrt  tests do, but I suspect not.

These test values must be added to all FP op test suites, single, double and half precision, for all rounding modes. (we don't need to test all possible sign variations, though)

fourth-bit:
This PR adds a cover-float as a submodule, and enables generations of tests in the F, D, Zfh, Zfhmin, Zfbfmin, and Zfa* testplans. 

Currently, the cover-float submodule is not in a working state (there is a runtime error in one of the generators), and there explicit exclusion for the cp_ibm_b14 case, so this pr is a draft for now. But, feedback on the the integration outside of the cover-float related issues would be appreciated.


generators/testgen/src/testgen/cover-float
jordancarlin, line 1: @coreyqh and I have discussed moving `cover-float` out of his personal account and into a `cover-float` organization for better long term maintainability. I'd like to see that happen before this is merged to avoid the difficulties of changing a submodule.

Makefile
```diff
@@ -50,11 +50,13 @@ override EXCLUDE_EXTENSIONS := $(subst $(space),$(empty),$(EXCLUDE_EXTENSIONS))
 # Setting to 1 is helpful for debugging test hangs so that only a single test runs at a time.
 JOBS ?= $(or $(patsubst -j%,%,$(filter -j%,$(MAKEFLAGS))),0)
 
+# Use this flag to enable the generation of floating-point tests from the cover-float test suite
+COVER_FLOAT ?=
```
jordancarlin: Move this up to the section with the other flags that users change (somewhere before the `ifneq ($(VERBOSE),)` line).

jordancarlin: Add this flag to the help entry in the Makefile.

testplans/F.csv
```diff
@@ -1,31 +1,31 @@
...
+Instruction,Type,RV32,RV64,cp_asm_count,cp_rs1,cp_rd,cp_rs1_edges,cp_imm_edges,cp_fs1,cp_fs2,cp_fs3,cp_fd,cp_fs1_edges,cp_fs2_edges,cp_fs3_edges,cmp_fd_fs1,cmp_fd_fs2,cmp_fd_fs3,cp_frm,cp_csr_fflags,cp_csr_frm,cr_fs1_fs2_edges,cr_fs1_fs3_edges,cp_fclass,cp_ibm
+fadd.s,FR,x,x,x,,,,,x,x,,x,x,x,,x,x,,3,von,x,frm,,,x
+fclass.s,FIX,x,x,x,,x,,,x,,,,x,,,,,,,,,,,x,x
+fcvt.l.s,F2X,,x,x,,x,,,x,,,,frm,,,,,,2,vn,x,,,,x
+fcvt.lu.s,F2X,,x,x,,x,,,x,,,,frm,,,,,,2,vn,x,,,,x
+fcvt.s.l,X2F,,x,x,x,,x,,,,,x,,,,,,,2,n,x,,,,
+fcvt.s.lu,X2F,,x,x,x,,x,,,,,x,,,,,,,2,n,x,,,,
+fcvt.s.w,X2F,x,x,x,x,,x,,,,,x,,,,,,,2,n,x,,,,
+fcvt.s.wu,X2F,x,x,x,x,,x,,,,,x,,,,,,,2,n,x,,,,
+fcvt.w.s,F2X,x,x,x,,x,,,x,,,,frm,,,,,,2,vn,x,,,,
```
jordancarlin: Why don't these converts have ibm tests?

generators/testgen/src/testgen/cli.py
```diff
@@ -124,10 +129,21 @@ def generate_all_tests(
...
                 if E_ext and testsuite not in E_EXTENSION_TESTS:
                     continue
-                tasks.append(UnprivTask(xlen, E_ext, testsuite, testplan_dir, output_test_dir))
+                tasks.append(UnprivTask(xlen, E_ext, testsuite, testplan_dir, output_test_dir, with_cover_float))
 
     tasks.extend(PrivTask(testsuite, output_test_dir) for testsuite in sorted(priv_ext_list))
 
+    # If we need the cover-float tests, build them now, we cannot do them in the individual cp_ibm
+    # calls because we lose all benefits of parallelization, and we run into dangers with multiple
+    # calls to cover-float attempting to generate tests at the same time due to ACT4 parallelization
```
jordancarlin: Clarify in this comment that it is only building the cover-float vectors, not the actual tests.

generators/testgen/src/testgen/cli.py
```diff
@@ -124,10 +129,21 @@ def generate_all_tests(
...
+    if with_cover_float:
+        cover_float_dir = Path(__file__).parent / "cover-float"
+        subprocess.run(
+            ["make", "-s", "-C", str(cover_float_dir), "AGGRESSIVENESS=0", "PROCESSED_ONLY=1", "processed-tests-only"],
```
jordancarlin: Prefer long forms of flags in scripts for better readability (`--silent` instead of `-s`).

generators/testgen/src/testgen/cli.py
```diff
@@ -124,10 +129,21 @@ def generate_all_tests(
...
+            stderr=subprocess.DEVNULL,
+            check=False,
+        )
```
jordancarlin: Even though it is fairly simple, can we move the cover-float code out of the top-level `cli.py` into a dedicated file since it is not part of the top-level flow? Then just import and call the relevant function (`generate_coverfloat` or something similar).

jordancarlin (changes requested):
Overall, this looks pretty good. Several comments below.

testplans/F.csv
fourth-bit: They should have cp_ibm tests. Fixed

generators/testgen/src/testgen/coverpoints/cp_ibm.py
```diff
@@ -101,14 +103,14 @@ def make_cp_ibm(instr_name: str, instr_type: str, coverpoint: str, test_data: Te
                     continue
                 values[col] = int(cell, 0)
 
-            params = generate_random_params(test_data, instr_type, exclude_regs=[0], frm=frm_mode, **values)
+            params = generate_random_params(test_data, instr_type, exclude_regs=[0], **values)
```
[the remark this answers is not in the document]
fourth-bit: Added frm by keyword argument instead

generators/testgen/src/testgen/cli.py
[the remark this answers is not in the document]
fourth-bit: Removed /dev/null redirect and added check=True instead

fourth-bit:
I'm not sure if coverfloat should be enabled by default. Right now it is, but it looks like CI needs a change so it pulls the submodule

jordancarlin:
I've been thinking more about this one, and I am leaning towards not using a submodule but instead pulling `cover-float` in as a dependency in the `pyproject.yaml` using uv. Happy to set up a meeting to sort out the details of what that would look like if that would be helpful. Sorry for all of the changes this PR has had to go through.

fourth-bit:
Looks like DCO and signed commits might prevent this from merging. Do I just need to rebase the PR?

davidharrishmc:
Claude is great at fixing those problems. If it can’t, then rebase.

Issue #1260, opened 2026-04-15
Add Zcmp/Zcmt tests


nazar-pc:
Right now Zca/Zcb tests are present, but Zcmp/Zcmt are missing.

davidharrishmc:
These are on the development plan, but not until after RVA23 certification is complete, so Fall 2026 at the earliest.
If you would like to create tests sooner, we would welcome more contributors.

PR #1535, opened 2026-05-19
feat(testgen): add Zcmp extension test infrastructure (#1260)

aclfe:
We've added test generation infrastructure for the Zcmp extension, covering six Zcmp instructions with rlist and s-register testcases. 

Commands run: `testgen testplans -o tests --extensions Zcmp`

Partially addresses #1260, zcmt can be a seperate PR

Open to any discussion of the following changes


davidharrishmc:
This is exciting.  Does it also run with 
make spike EXTENSIONS=Zcmt
Does it pass Whisper and QEMU as well?

I don't believe Sail has Zcmt support yet, so I presume it doesn't pass Sail or get coverage?

coverpoints/unpriv/Zcmp_coverage.svh
```diff
@@ -0,0 +1,143 @@
+///////////////////////////////////////////
```
davidharrishmc, line 1: Zcmp can also produce faults, so a full solution will need ExceptionsZcmp as well to test faults.  There could be some complex interactions with PMP and virtual memory to consider.

coverpoints/unpriv/Zcmp_coverage.svh
```diff
@@ -0,0 +1,143 @@
...
+// Copyright (C) 2024 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+////////////////////////////////////////////////////////////////////////////////////////////////
+
+`define COVER_ZCMP
+covergroup Zcmp_cm_mva01s_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    cp_zcmp_mva01s : coverpoint ins.current.insn[9:7] iff (ins.trap == 0) {
```
davidharrishmc, line 17: This covers r1s'. Seems another coverpoint is needed for r2s'.

coverpoints/unpriv/Zcmp_coverage.svh
davidharrishmc: There should also be cases added to Ssstrict (PR open, not yet merged):
cm.mva01s with all 8 values of r1s' and 8 values of r2s' (values > 1 should be reserved in RVE).

coverpoints/unpriv/Zcmp_coverage.svh
```diff
@@ -0,0 +1,143 @@
...
+        bins s0 = {3'd0};
+        bins s1 = {3'd1};
+        bins s2 = {3'd2};
```
davidharrishmc, line 20: Bins 2-7 should be gated by I_SUPPORTED because they are reserved for E.

coverpoints/unpriv/Zcmp_coverage.svh
```diff
@@ -0,0 +1,143 @@
...
+        bins s5 = {3'd5};
+        bins s6 = {3'd6};
+        bins s7 = {3'd7};
+    }
+
+endgroup
+// ---------------------
+covergroup Zcmp_cm_mvsa01_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    cp_zcmp_mvsa01 : coverpoint ins.current.insn[9:7] iff (ins.trap == 0) {
```
davidharrishmc, line 32: Again, another coverpoint is needed for r2s'.  Rs1' and rs2' must be different. Bins s2-s87 should be gated by I_SUPPORTED.

Ssstrict should add add the cross-product of all 8*8 values of the two registers to check that r1s=r2s are reserved and that rs > 1 is reserved for E.

coverpoints/unpriv/Zcmp_coverage.svh
```diff
@@ -0,0 +1,143 @@
...
+        bins s5 = {3'd5};
+        bins s6 = {3'd6};
+        bins s7 = {3'd7};
+    }
+
+endgroup
+// ---------------------
+covergroup Zcmp_cm_pop_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    cp_zcmp_pop_rlist : coverpoint ins.current.insn[7:4] iff (ins.trap == 0) {
```
davidharrishmc, line 47: Needs coverpoint for spimm.
Gate 7-15 with I_SUPPORTED.

Add Sstrict cases with all 16 values of list.

davidharrishmc: Also applies to popret and popretz and push.

tests/rv32i/Zcmp/Zcmp-cm.mva01s-00.S
```diff
@@ -0,0 +1,128 @@
...
+  mv x28, x2 # switch signature pointer register to avoid conflict with test
+  LI(x8, 0x8080808)
+  LI(x9, 0x9090909)
+Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s0:
+  cm.mva01s s0, s1
+  # Check if x10 contains the expected result. x28 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x28, x5, x4, x10, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s0, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s0_str)
+  # Check if x11 contains the expected result. x28 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x28, x5, x4, x11, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s0, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s0_str)
+  LI(x9, 0x9090909)
```
davidharrishmc, line 46: Add a blank line between test for readability.

tests/rv32i/Zcmp/Zcmp-cm.mva01s-00.S
```diff
@@ -0,0 +1,128 @@
...
+  RVTEST_SIGUPD(x28, x5, x4, x11, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s6, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s6_str)
+  LI(x23, 0x17171717)
+  LI(x8, 0x8080808)
+Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s7:
+  cm.mva01s s7, s0
+  # Check if x10 contains the expected result. x28 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x28, x5, x4, x10, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s7, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s7_str)
+  # Check if x11 contains the expected result. x28 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x28, x5, x4, x11, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s7, Zcmp_cm_mva01s_cg_cp_zcmp_mva01s_s7_str)
+
```
davidharrishmc, line 102: This will hit all coverpoints, but it would be nice to to have a set of tests where r1 and r2 are not consecutive, just in case there is something weird about the implementation.

tests/rv32i/Zcmp/Zcmp-cm.mvsa01-00.S
```diff
@@ -0,0 +1,128 @@
...
+RVTEST_BEGIN
+
+
+
+/////////////////////////////////
+// cp_zcmp_mvsa01
+/////////////////////////////////
+
+  mv x29, x2 # switch signature pointer register to avoid conflict with test
+  LI(x10, 0xa0a0a0a0)
```
davidharrishmc, line 38: Same comments as in mva01s.

tests/rv32i/Zcmp/Zcmp-cm.pop-00.S
```diff
@@ -0,0 +1,444 @@
...
+
+
+
+/////////////////////////////////
+// cp_zcmp_pop_rlist
+/////////////////////////////////
+
+  mv x29, x2 # switch signature pointer register to avoid conflict with test
+  LI(x1, 0x1010101)
+  LA(x15, scratch)
```
davidharrishmc, line 39: Some comments in code would be useful here.

tests/rv32i/Zcmp/Zcmp-cm.pop-00.S
```diff
@@ -0,0 +1,444 @@
...
+  addi x2, x15, 240
+  cm.push {ra}, -16
+  li x1, 0
+Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist4:
+  cm.pop {ra}, 16
+  # Check if x1 contains the expected result. x29 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x29, x5, x4, x1, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist4, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist4_str)
```
davidharrishmc, line 46: This would still pass if push and pop wrote to the wrong address.  Suggest adding some extra checking to LREG values from memory and SIGUPD them to ensure the right things are in the right address.

davidharrishmc (changes requested):
Partial review

aclfe:
@davidharrishmc Thanks for the review! I think before addressing the changes, let me make an issue and see if we can get Zcmp support on sail. I can work on it if allowed, then I'll connect both of these. without a reference model, this might be hard, what do you think?

davidharrishmc:
Zcmp is in development. 
https://github.com/riscv/sail-riscv/pull/730

The Sail team appreciates more contributors, though some of the work is pretty subtle.

Sounds good to put this PR on idle until there is Sail support to test it.

tests/rv32i/Zcmp/Zcmp-cm.popret-00.S
```diff
@@ -0,0 +1,432 @@
...
+/////////////////////////////////
+// cp_zcmp_popret_rlist
+/////////////////////////////////
+
+  mv x17, x2 # switch signature pointer register to avoid conflict with test
+  LA(x1, Zcmp_cm_popret_cg_cp_zcmp_popret_rlist_rlist4_ret)
+  LA(x6, scratch)
+  addi x2, x6, 240
+Zcmp_cm_popret_cg_cp_zcmp_popret_rlist_rlist4:
+  cm.push {ra}, -16
```
davidharrishmc, line 42: Doesn't ra need to be loaded with some value to return to, such as the start of the next test?  And add a SIGUPD after the popret to write a bad value that would be caught if the popret doesn't jump.

tests/rv32i/Zcmp/Zcmp-cm.popret-00.S
```diff
@@ -0,0 +1,432 @@
...
+  li x1, 0
+  cm.popret {ra}, 16
```
davidharrishmc, line 44: A SIGUPD is needed to check ra got the right value in each of these tests.

tests/rv32i/Zcmp/Zcmp-cm.popret-00.S
davidharrishmc: Disregard - I see it is loaded above.  It is confusing to use ra in some places and x1 in others - please be consistent, with ra preferred here.

tests/rv32i/Zcmp/Zcmp-cm.popret-00.S
```diff
@@ -0,0 +1,432 @@
...
+// Set up test environment and declare main entry point
+RVTEST_BEGIN
+
+
+
+/////////////////////////////////
+// cp_zcmp_popret_rlist
+/////////////////////////////////
+
+  mv x17, x2 # switch signature pointer register to avoid conflict with test
```
davidharrishmc, line 37: x17 is in the upper half, so this wouldn't work with E.  Prefer to use only the lower half registers to work with E unless there is a compelling reason otherwise.

tests/rv32i/Zcmp/Zcmp-cm.pop-00.S
```diff
@@ -0,0 +1,444 @@
...
+  li x1, 0
+Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist6:
+  cm.pop {ra,s0-s1}, 16
+  # Check if x9 contains the expected result. x29 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x29, x5, x4, x9, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist6, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist6_str)
+  # Check if x8 contains the expected result. x29 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x29, x5, x4, x8, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist6, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist6_str)
+  # Check if x1 contains the expected result. x29 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x29, x5, x4, x1, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist6, Zcmp_cm_pop_cg_cp_zcmp_pop_rlist_rlist6_str)
+  LI(x18, 0x12121212)
```
davidharrishmc, line 77: All of these tests using s2 and above should be gated by #ifdef I_SUPPORTED so they aren't run on E.

tests/rv32i/Zcmp/Zcmp-cm.popretz-00.S
```diff
@@ -0,0 +1,468 @@
...
+  LA(x1, Zcmp_cm_popretz_cg_cp_zcmp_popretz_rlist_rlist4_ret)
+  LA(x11, scratch)
+  addi x2, x11, 240
+Zcmp_cm_popretz_cg_cp_zcmp_popretz_rlist_rlist4:
+  cm.push {ra}, -16
+  li x1, 0
+  LI(x10, 0xDEADBEEF)
+  cm.popretz {ra}, 16
+Zcmp_cm_popretz_cg_cp_zcmp_popretz_rlist_rlist4_ret:
+  # Check if x10 contains the expected result. x15 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
```
davidharrishmc, line 47: Using a0 instead of x10, s0, instead of x8, etc. would be easier to read in accordance with the spec that refers to the named registers.

tests/rv32i/Zcmp/Zcmp-cm.push-00.S
```diff
@@ -0,0 +1,444 @@
...
+
+
+/////////////////////////////////
+// cp_zcmp_push_rlist
+/////////////////////////////////
+
+  mv x31, x2 # switch signature pointer register to avoid conflict with test
+  LI(x1, 0x1010101)
+  LA(x13, scratch)
+  addi x2, x13, 240
```
davidharrishmc, line 40: Calling x2 sp would be easier to follow.

tests/rv32i/Zcmp/Zcmp-cm.push-00.S
```diff
@@ -0,0 +1,444 @@
...
+Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist4:
+  cm.push {ra}, -16
+  lw x11, 12(x2)
+  # Check if x11 contains the expected result. x31 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x31, x5, x4, x11, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist4, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist4_str)
```
davidharrishmc, line 45: x2 (sp) is also changed, so might as well sigupd it.

tests/rv32i/Zcmp/Zcmp-cm.pop-00.S
davidharrishmc: sp is also changed.  Might as well sigupd it at the end of test to make sure it has the expected value.

tests/rv32i/Zcmp/Zcmp-cm.push-00.S
```diff
@@ -0,0 +1,444 @@
...
+  # Check if x11 contains the expected result. x31 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x31, x5, x4, x11, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist6, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist6_str)
+  lw x11, 8(x2)
+  # Check if x11 contains the expected result. x31 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x31, x5, x4, x11, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist6, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist6_str)
+  lw x11, 4(x2)
+  # Check if x11 contains the expected result. x31 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x31, x5, x4, x11, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist6, Zcmp_cm_push_cg_cp_zcmp_push_rlist_rlist6_str)
+  addi x2, x2, 16
+  LI(x18, 0x12121212)
```
davidharrishmc, line 77: gate everything with s2 and above by I_SUPPORTED.

tests/rv64i/Zcmp/Zcmp-cm.mva01s-00.S
```diff
@@ -0,0 +1,128 @@
...
+RVTEST_BEGIN
+
+
+
+/////////////////////////////////
+// cp_zcmp_mva01s
+/////////////////////////////////
+
+  mv x28, x2 # switch signature pointer register to avoid conflict with test
+  LI(x8, 0x8080808)
```
davidharrishmc, line 38: How about loading 64-bit values on RV64?

davidharrishmc (changes requested):
This is great stuff.  A few more comments on the last part.  When Sail supports Zcmp, let's test before merging.

aclfe:
https://github.com/riscv/riscv-unified-db/blob/72971dcc13f52ba737135bfe36cc68dc2b615fca/spec/std/isa/ext/Zcmp.yaml#L12-L16

Since they're incompatible, we'll need seperate spike/whisper/qemu configs I believe. I'll work on getting those added

PR #1537, opened 2026-05-19
Corrected pmpaddr1 construction

henrikg-qc:
Adding PMP grain to REGIONSTART by immediate loaded register (t0) rather than immediate

### DIFF
--- a/tests/priv/pmp/pmp64/PMPSm/pmpsm_access_double_region.S
+++ b/tests/priv/pmp/pmp64/PMPSm/pmpsm_access_double_region.S
@@ -98,7 +98,7 @@ main:
 
     LA(x4, REGIONSTART)
     li t0, g
-    add x4, x4, g
+    add x4, x4, t0
     srl x4, x4, PMP_SHIFT
     csrw pmpaddr1, x4
 


henrikg-qc:
I'd also like to review the PMP_GRAIN setup from rvtest_config. Is it supposed to be the actual granularity in bits?
The g MACRO:
#define g   (1 << (RVMODEL_PMP_GRAIN **_+ 2_**))

henrikg-qc:
This is a widespread issue in the PMP tests with g being used as an immdiate to addi. 
I'll create an issue

jordancarlin:
> I'd also like to review the PMP_GRAIN setup from rvtest_config. Is it supposed to be the actual granularity in bits? The g MACRO: #define g (1 << (RVMODEL_PMP_GRAIN **_+ 2_**))

`rvtest_config` will be autoderived from the UDB YAML file soon (hopefully within the next few hours, see #1504), so no need to worry about defining it.

> This is a widespread issue in the PMP tests with g being used as an immdiate to addi. I'll create an issue

Thanks for catching this. Most of the PMP tests have only been tested with relatively small granularities, so this problem slipped through. If you are willing to update all of the uses, a PR would be great. Otherwise, please open an issue.

jordancarlin:
Closing in favor of #1642.

PR #1540, opened 2026-05-19
Fix/cv32e20 correct extensions

karabambus:
- added U mode extension. 
- all test passing (except InterruptsSm and InterruptsU. tracked issue)


config/cores/cve2/cv32e20/sail.json
```diff
@@ -286,7 +286,7 @@
       "supported": false
     },
     "Zicntr": {
-      "supported": true
+      "supported": false
```
davidharrishmc, line 289: It looks like cycle, instret, time are supported.  Why is this false?

https://docs.openhwgroup.org/projects/cve2-user-manual/en/latest/01_specification/index.html

karabambus: I thought so too, that why I opened this [issue]( https://github.com/openhwgroup/cve2/issues/324), upon further examination of rtl - while this CSRs are implemented for M mode, there is no U mode aliases. I suppose that is why manual page does not claim to support them. Running the tests on DUT would produce hang if we dont leave out U mode extension (and manual explicitly notes that cv32e20 supports U mode).

@MikeOpenHWGroup can you confirm this? Also I might reopen this issue and rephrase it, since  I am not only one that made this mistake, maybe manual really needs clarification?

MikeOpenHWGroup: It is my understanding that the `hpmcounter` CSRs (should) exist and are readable in user-mode if the appropriate bit is set in `mcounteren`.  As such, both Zicntr and Zihpm should be true.

karabambus: Enabling Zicntr and Zihpm causes illegal instructions since U-mode aliases for counter are not implemented for this core - causing decoder too fall trough to default case for this addresses. We believe this core has compliant Zicntr and Zihpm extensions even though M mode counters are implanted while U mode counter CSRs are not. This was concluded according to section 3.1.11 of the privilege spec:

> In harts with U-mode, the mcounteren must be implemented, but all fields are WARL and may be read-only zero, indicating reads to the corresponding counter will cause an illegal-instruction exception when executing in a less-privileged mode. In harts without U-mode, the mcounteren register should not exist.

since we can have mcounteren as permanently disabled it is implicit that U mode counters CSR implementation can be left out.

I am working on more detailed report.

davidharrishmc: Ah, I see.  The docs say cycle, instret are implemented, but do not mention `time`.  Therefore the core is not Zicntr-compatible.

davidharrishmc (approved):

config/cores/cve2/cv32e20/sail.json
karabambus: There is also no cycle or instret in RTL, this is manual gap.

Issue #1440, opened 2026-05-06
Add RVMODEL_SET/CLR_MTIMER_INT macros for timer interrupt testing


karabambus:
**Is your feature request related to a problem? Please describe.**
Cores that do not have a CLINT cannot hit `InterruptsSm` coverpoints `cp_trigger_mti` and `cp_wfi`. The test generator triggers the machine timer interrupt (MTIP) exclusively by reading `RVMODEL_MTIME_ADDRESS` and writing `RVMODEL_MTIMECMP_ADDRESS`. These are CLINT registers - a platform peripheral, not part of the core under test.

The core being tested only sees `irq_timer_i` going high. It does not care whether that signal came from a CLINT, a custom timer, or a testbench virtual peripheral. The test should reflect this - it is testing how the core responds to a timer interrupt, not testing the platform timer itself.

MSIP and MEIP already work this way. They are triggered via `RVMODEL_SET_MSW_INT` and `RVMODEL_SET_MEXT_INT` - macros that each platform implements for its own hardware. `sail_macros.h` overrides them with Sail-compatible implementations so the reference ELF fires the interrupt via Sail's platform, while the DUT ELF fires it via the DUT's platform. Both produce the same core behaviour and matching signatures. 

**Describe the solution you'd like**
Add `RVMODEL_SET_MTIMER_INT(_R1, _R2)` and `RVMODEL_CLR_MTIMER_INT(_R1, _R2)` macros, completing the pattern already established for MSIP and MEIP.

Update `generators/testgen/src/testgen/asm/interrupts.py` to replace the read-MTIME / write-MTIMECMP sequence with a macro call when the macro is defined, keeping the existing MTIME address path as fallback for platforms that have a CLINT:

```c
// current:
LA(x, RVMODEL_MTIME_ADDRESS)
lw x_lo, 0(x)
LA(x, RVMODEL_MTIMECMP_ADDRESS)
sw x_lo, 0(x)

// proposed:
#ifdef RVMODEL_SET_MTIMER_INT
  RVMODEL_SET_MTIMER_INT(T2, T5)   // each platform provides its own implementation
#elif defined(RVMODEL_MTIME_ADDRESS)
  // existing path unchanged
#endif
```

[...]

PR #1544, opened 2026-05-20
Update RVMODEL interrupt macros explanation

MikeOpenHWGroup:
As discussed in #1440.  Clarified the requirements for RVMODEL interrupt macros and added details about certification for timer interrupts.


davidharrishmc (approved):

allenjbaum:
Do we need to test that the physical MMIO addresses of mtime nd mtimecmp
are inaccessible to U and S modes?

To do that,  a A or U mode test would have to have page tables set up to
access it, or execute in BARE mode, and demonstrate that they trap on read
from U or S mode(s).
Huge security hole otherwise....

davidharrishmc:
When we fully switch over to the T-SBI, we need to make all accesses to mtime/mtimecmp through the T-SBI.

However, there are no normative rules that restrict the access.  For example, these registers may be in a CLINT.  I don’t see privilege mode as one of the explicit PMA criteria in the first paragraph of 3.6, so perhaps
the PMA will allow access.  In such case, the system may need PMP to prohibit access to peripheral space
from lower privilege modes.  That would be a firmware behavior and should not be certified.

MikeOpenHWGroup:
Hi @allenjbaum and @davidharrishmc, it is generally considered risky business to continue a discussion in a PR that is already merged.  If this topic still requires discussion, let's open a new Issue.

PR #1546, opened 2026-05-20
testgen: dedicated cp_fpr_hazard generator

dev-aditya-hub:
## Summary

`cp_fpr_hazard` was registered against the GPR hazard generator. The GPR
generator only pins integer registers (`rs1`/`rs2`/`rd`), which are not
required params for FP instruction types, so the pins were silently dropped
and the emitted test pair carried no real FP register dependency.

This PR introduces a dedicated `make_cp_fpr_hazard` generator and unbinds
`cp_fpr_hazard` from the GPR path.

## Problem

For an FP consumer (e.g. an `FR`-type `fmul.s`), the existing flow:

1. produced an integer `add` as instruction A,
2. tried to pin instruction B's `rs1`/`rs2`/`rd` to A's destination,
3. those fields were not in B's `required_params` set, so
   `generate_random_params` ignored them and randomly assigned B's
   `fs1`/`fs2`/`fs3`/`fd`.

Net effect: zero RAW/WAW/WAR coverage on the floating-point register file.
A DUT with broken FP forwarding would still pass `cp_fpr_hazard`.

## Fix

A new `make_cp_fpr_hazard` generator that:

- Selects an `fadd.{s,d,h,q}` producer matching the consumer instruction's
  precision suffix, so the producer writes a value-compatible `fd`.
- Introspects the consumer's `required_params` via
  `get_instr_type_config(instr_type)` and pins the appropriate FP fields:
  - **RAW** — pin one of the consumer's `fs1` / `fs2` / `fs3` to the
    producer's `fd`.
  - **WAW** — pin the consumer's `fd` to the producer's `fd`.
  - **WAR** — pin the consumer's `fd` to one of the producer's source regs.
- Skips a hazard variant when the consumer has no compatible FP field
  (e.g. RAW against an `FL` load, WAW against an `FS` store) rather than
  emitting a misleading no-op pair.

Both integer and float registers used by the two instructions are returned
to the pool after each iteration.

## Risk

Latent — no testplan in `coverpoints/` currently references
`cp_fpr_hazard`. The coverpoint is, however, advertised in
`docs/ctp/src/unpriv.adoc:76` as a 4-bin coverpoint (RAW / WAR / WAW /
nohaz). Without this fix, adding it to any Zfh/F/D testplan would have
silently generated dependency-free pairs.

## Test plan

- [ ] Build a small Zfh testplan referencing `cp_fpr_hazard_rw` and confirm
      the emitted assembly carries genuine FP register dependencies between
      the two instructions of each pair.
- [ ] Confirm `cp_gpr_hazard` output is unchanged (the GPR generator is now
      registered for a single coverpoint and otherwise untouched).

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/hazards.py
+++ b/generators/testgen/src/testgen/coverpoints/hazards.py
@@ -120,24 +120,22 @@ def make_cp_fpr_hazard(instr_name: str, instr_type: str, coverpoint: str, test_d
 
     for haz_type in hazard_types:
         for i in range(2):
+            # Bail out before allocating the producer if the consumer can't form this
+            # hazard, otherwise params_a's regs would leak from the pool on continue.
+            if haz_type == "raw" and not consumer_fp_srcs:
+                continue
+            if haz_type in ("waw", "war") and not consumer_has_fd:
+                continue
+
             params_a = generate_random_params(test_data, producer_type)
             assert params_a.fs1 is not None and params_a.fs2 is not None and params_a.fd is not None
 
             if haz_type == "raw":
-                if consumer_fp_srcs:
-                    src_field = consumer_fp_srcs[i % len(consumer_fp_srcs)]
-                    params_b = generate_random_params(test_data, instr_type, **{src_field: params_a.fd})
-                else:
-                    # Consumer reads no FP source register — skip this iteration's hazard
-                    # rather than emit a no-op test that pretends to cover RAW.
-                    continue
+                src_field = consumer_fp_srcs[i % len(consumer_fp_srcs)]
+                params_b = generate_random_params(test_data, instr_type, **{src_field: params_a.fd})
             elif haz_type == "waw":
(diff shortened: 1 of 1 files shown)


davidharrishmc:
@DymShanks is working on a general solution to hazards.  Let's give him some time to get that in so we don't duplicate effort.

dev-aditya-hub:
Thanks for the heads up about @DymShanks  work happy to defer to that and I won't push for merge here.

Just pushed a small follow-up (93abc57) to address the bot's note about register pool leaks: moved the compatibility check above the producer allocation so incompatible variants short-circuit cleanly instead of leaking regs on continue. Figured it was worth fixing either way in case any of this is useful as a reference later. Feel free to close if it ends up being redundant once the general solution lands.

jordancarlin:
Thanks for your work on this. As previously mentioned, there is a more comprehensive hazard refactor happening, so I am closing this in favor of #1614.

PR #1548, opened 2026-05-20
Fix CSR testgen register collisions in Zicsr validation paths

### LINKED ISSUE #1538: Zicsr tests use `instret` without setting it first
**Test name**
Zicsr-***

**Describe the bug**
The Zicsr tests select the CSR to access using the following logic:
```c
#if defined(F_SUPPORTED)
  csrrw x0, fflags, x17
#elif defined(V_SUPPORTED)
  csrrw x0, vxsat, x17
#elif !defined(U_SUPPORTED)
  csrrw x0, mepc, x17
  csrrw x0, instret, x17
#else
#endif
```

When it falls through to the `instret` case, the SIGUPD mismatches because the number of instructions executed in the boot code is different.

**Expected behavior**
We either need to pick a different CSR, set `instret` before executing the CSR instructions, or mask the final result.

somiljain2006:
Fixed #1538 

Fix CSR generator scratch register collisions in Zicsr tests. Use dynamically allocated non-overlapping scratch registers in CSR and CSRI formatters to avoid collisions with architectural operands during runtime validation. The previous implementation could alias validation scratch registers with rd/rs1/rs2, corrupting architectural state and causing invalid
runtime behavior in generated Zicsr tests.

This change:
- allocates safe temporary registers dynamically
- validates allocator correctness explicitly
- restores allocated registers back to the pool
- preserves deterministic generator state
- regenerates affected rv32i/rv64i Zicsr tests

Validated by:
- successful ELF generation
- successful Spike execution
- reaching write_tohost_pass without traps


davidharrishmc:
This seems overly complicated to fix a simple issue.  I think the test could just read instret twice, subtract the two, and record that difference (presumably 1) as the signature.

somiljain2006:
@davidharrishmc Thanks for the suggestion. I simplified the logic accordingly. Instead of validating both the architectural return value and post-state using multiple dynamically allocated temporaries, the ZICNTR path now just:

1. reads instret before the CSR instruction,
2. executes the instruction,
3. reads instret again,
4. records the delta between the two samples in the signature.

This keeps the validation logic much smaller. I also fixed the copilot reported aliasing issue by storing the initial instret sample in a dedicated scratch register instead of rs2, avoiding collisions with rd/rs1/rs2 or x0.

davidharrishmc:
I think this is still way too complicated.   I don't understand why registers need to be initialized here.  The only difficult case is instret, so within that #ifdef, can't you just have

csrr <somereg1>, instret
csrr <somereg2>, instret
sub somreg1, somereg2, somereg1

and then seep the existing sigupd.

Also, there is a standard get_registers function (see, for example, ExceptionsSm.py) so you don't have to roll your own if you need a register.

David

somiljain2006:
@davidharrishmc I've removed the redundant CSR setup logic entirely and switched to using get_registers(1). The ZICNTR validation is now just a simple back-to-back csrrs -> csrrs -> sub-sequence followed by the existing sigupd handling.

generators/testgen/src/testgen/formatters/types/csr_type.py
```diff
@@ -41,21 +39,51 @@ def format_csr_type(
...
-        "// read back CSR to check updated value",
-        zicsr_acccess("csrrs", params.rs2, 0),
-        write_sigupd(params.rs2, test_data, "int"),
-    ]
-    return (setup, test, check)
+
+    allocated = []
+    
+    try:
+        allocated = test_data.int_regs.get_registers(1)
```
davidharrishmc: scratch_reg = test_data.int_regs.get_register()

generators/testgen/src/testgen/formatters/types/csr_type.py
```diff
@@ -41,21 +39,51 @@ def format_csr_type(
...
+        ]
+        test = [
+            "// perform operation",
+            zicsr_acccess(instr_name, params.rd, params.rs1),
+        ]
+        check = [
+            "#if defined(F_SUPPORTED)",
+            write_sigupd(params.rd, test_data, "int"),
+            "// read back CSR to check updated value",
+            f"csrrs x{params.rs2}, fflags, x0",
```
davidharrishmc: "csrr x{params.rs2}, fflags" is more succinct.

davidharrishmc:
This isn't converging as fast as I hoped, so I've taken a stab at the simplest approach I can think of to avoid traps and get a consistent signature on instret.  Could you review and see if it accomplishes what you are aiming for?

https://github.com/riscv/riscv-arch-test/pull/1561

somiljain2006:
Yeah, the PR looks good. Reusing rs1/rs2 to avoid the allocator is nice, and the x0 checks handle the trap cases cleanly. I'll close this PR in favor of yours.

Issue #1516, opened 2026-05-16
Vx mask instructions need to accept writing tail with value op would have produced


davidharrishmc:
Affects vector SIGUPD self-checking.

See https://github.com/tenstorrent/whisper/issues/21#issuecomment-4465043401

fourth-bit:
Closed by #1549. Added another signature entry for mask producing instructions, so that we can check the value as if tail undisturbed, the value that op would have produced, and all ones

PR #1549, opened 2026-05-20
Updates to Vector Length Suite Sig Update Macro

fourth-bit:
Picks up #1183 where @georgiatai left off. 

Beyond what has already been done, this now runs mask producing operations in the length suite twice. Once with vl set to the desired value, and a second time with vl=vlmax. The second SIGUPD nops in self-checking mode and adds the value to the signature in when not in self-checking mode. Then, the first SIGUPD macro can read the signature twice, once for the value leaving the tail undisturbed, and once with the value where the tail is the result of the operation as if vl = vlmax. This addresses #1516.


tests/env/signature.h
```diff
@@ -547,17 +577,22 @@
...
         /* Save architecture state of instruction under test (vl and vtype) */                                      \
         nop                                  ;                                                                      \
         nop                                  ;                                                                      \
         nop                                  ;   /* _SCALAR_DST_FLAG: LI placeholder */                             \
-        nop                                  ;   /* _SCALAR_DST_FLAG: beqz placeholder */                          \
+        nop                                  ;   /* _SCALAR_DST_FLAG: beqz placeholder */                           \
         nop                                  ;   /* _SCALAR_DST_FLAG: li placeholder */                             \
     40:                                      ;                                                                      \
+        /* Obtain effective vl if insstruction is vcompress.m */                                                                       \
```
davidharrishmc: Spelling

tests/env/signature.h
```diff
@@ -401,20 +417,25 @@
...
+        /* For scalar-dst instructions (vmv.s.x, reductions) only element 0 is written. */                          \
+        /* Override the saved vl to 1 so the active mask covers only element 0 and     */                           \
         /* elements 1..VLMAX-1 are treated as tail, getting the vta-agnostic relaxation.*/                          \
         LI(_LINK_REG, _SCALAR_DST_FLAG)      ;                                                                      \
         beqz        _LINK_REG, 40f           ;                                                                      \
         li          _TEMP_REG, 1             ;                                                                      \
     40:                                      ;                                                                      \
+        /* Obtain effective vl if insstruction is vcompress.m */                                                                       \
+        LI(_LINK_REG, _VCOMPRESS_FLAG)       ;   /* Load whether instr is vcompress.m which changes effective vl of vd */              \
```
davidharrishmc: Does it work to do #if (_VCOMPRESS_FLAG == 1) instead of the LI and beqz?  If so, this may apply many places.

tests/env/signature.h
```diff
@@ -444,28 +465,37 @@
...
-        vmandn.mm   _MTMP2, _VTMP, _MTMP2    ;   /* MTMP2[i] = inactive && !(VR[i] == 1) → mismatch with all 1s */  \
-        j           5f                       ;   /* Unconditional skip data vec agnostic handling to tail check */  \
-    4:                                                                                                              \
         /* Data vector tail agnostic(vta == 1) handling: all 1s in agnostic element is also legal */                \
         vmseq.vi    _MTMP2, _VR, -1          ;   /* MTMP2[i] = (VR[i] == -1) */                                     \
         vmandn.mm   _MTMP2, _VTMP, _MTMP2    ;   /* MTMP2[i] = tail && !(VR[i] == -1) → mismatch with all 1s */     \
+        j           5f                       ;   /* Unconditional skip data vec agnostic handling to tail check */  \
+    4:                                                                                                              \
+        /* Mask vector tail agnostic(vta == 1) handling: all 1s in agnostic element is also legal */                \
+        vmand.mm    _MTMP2, _VR, _VR         ;   /* MTMP2[i] = (VR[i] == 1) */                                      \
```
davidharrishmc: Does this do anything?  is _VR & _VR just _VR?  If it does more useful stuff, could you point it out in a comment?

tests/env/signature.h
```diff
@@ -444,28 +465,37 @@
...
+        vmandn.mm   _MTMP2, _VTMP, _MTMP2    ;   /* MTMP2[i] = inactive && !(VR[i] == 1) → mismatch with all 1s */  \
```
davidharrishmc: Mask logical tails need to be able to preserve the old value, make all 1s, or perform the operation.  It's not very clear from the comments how all three of those situations are accepted.  Perhaps you just don't want to check anything in the tail at all.

davidharrishmc (changes requested):

tests/env/signature.h
fourth-bit: It works with conditional assembly, and has been changed now

tests/env/signature.h
fourth-bit: Good catch. This can just be a vmv.v.v

tests/env/signature.h
fourth-bit: I've added more comments now, and with the conditional assembly, hopefully its clearer.

davidharrishmc (approved):

fourth-bit:
I would like to see another round of review because the changes to mask calculation were pushed

PR #1469, opened 2026-05-09
Share cp_fs{1,2,3}_edges implementation; enable cross_frm for fs2/fs3

dev-aditya-hub:
## Summary

`make_fs1_edges`, `make_fs2_edges`, and `make_fs3_edges` in `cp_fp_reg_edges.py` were near-duplicates. The `cross_frm` branch was only ever added to fs1, so `cp_fs2_edges_frm` and `cp_fs3_edges_frm` would silently fall through to a single frm mode if any testplan ever opted into them.

Pulled the shared body into `_make_fs_edges(operand, ...)` and reduced each public generator to a one-line dispatch. The helper handles edge-set selection, the `_frm` cross, and the bin-name / description formatting.

## Why

- Three copies of the same logic were drifting. fs1 already gained `cross_frm` support; fs2 and fs3 didn't.
- One source of truth means the next change (new operand, new frm behavior, new edge category) lands in one place by construction.

## Behavior

- `cp_fs1_edges*` — output unchanged.
- `cp_fs2_edges` / `cp_fs3_edges` (no `_frm`) — output unchanged; `frm_mode=None` reproduces the same `bin_name` and `desc` as before.
- `cp_fs2_edges_frm` / `cp_fs3_edges_frm` — now functional; sweep all six rounding modes, mirroring fs1.

## Notes

- This revision drops the `csr_frm_val=4` dyn-mode pin per @jordancarlin's review comment. That change is being handled separately on #1467, where the discussion of how to address the random-rne collision belongs.
- Net diff: -42 / +19 in the only file touched.
- No public API change; `add_coverpoint_generator` registrations are unchanged.

PR #1554, opened 2026-05-21
cp_fp_reg_edges: honor _frm suffix on fs2 and fs3 generators

dev-aditya-hub:
## Summary

`make_fs1_edges` in `cp_fp_reg_edges.py` honors a `_frm` suffix on the coverpoint name and crosses each edge value with all six rounding modes. `make_fs2_edges` and `make_fs3_edges` ignore the suffix and emit one frm-unspecified test per edge value.

## Problem

Because the registry uses longest-prefix matching, a coverpoint named `cp_fs2_edges_frm` routes to `make_fs2_edges`, which silently drops the rounding-mode cross. The label says the suite covered the cross-product; the generated assembly did not. A DUT with a rounding-path bug that only triggers for specific `fs2` or `fs3` edge values (e.g., subnormal `fs2` with RTZ on `fmadd.s`) is never exercised, while the symmetric `fs1` case is.

## Fix

Mirror the `cross_frm` / `frm_modes` handling from `make_fs1_edges` into `make_fs2_edges` and `make_fs3_edges`, including the bin-name and description suffixes so the generated labels stay distinct per frm mode.

## Risk

Low. When the coverpoint does not end in `_frm`, `frm_modes = [None]` and the new loop runs exactly once per edge value with `frm=None`, producing the same output as before. Only coverpoint names containing `_frm` are affected.

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/cp_fp_reg_edges.py
+++ b/generators/testgen/src/testgen/coverpoints/cp_fp_reg_edges.py
@@ -57,13 +57,21 @@ def make_fs2_edges(instr_name: str, instr_type: str, coverpoint: str, test_data:
     else:
         edges = FLOAT_EDGES.single
 
+    cross_frm = "_frm" in coverpoint
+
+    frm_modes = ("dyn", "rdn", "rmm", "rne", "rtz", "rup") if cross_frm else [None]
+
     test_chunks: list[TestChunk] = []
     for edge_val in edges:
-        params = generate_random_params(test_data, instr_type, exclude_regs=[0], fs2val=edge_val)
-        desc = f"{coverpoint} (Test source fs2 value = {test_data.flen_format_str.format(edge_val)})"
-        tc = format_single_testcase(instr_name, instr_type, test_data, params, desc, f"b{edge_val:#x}", coverpoint)
-        test_chunks.append(tc)
-        return_test_regs(test_data, params)
+        for frm_mode in frm_modes:
+            params = generate_random_params(
+                test_data, instr_type, exclude_regs=[0], fs2val=edge_val, frm=frm_mode
+            )
+            bin_name = f"b{edge_val:#x}{f'_{frm_mode}' if frm_mode is not None else ''}"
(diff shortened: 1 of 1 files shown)


dev-aditya-hub:
Heads-up on landing order for this one. I have two other PRs touching the same file/concept:

- #1469 refactors the three `make_fs{1,2,3}_edges` generators into a single shared helper. If that lands first, this PR's diff collapses into the shared helper.
- #1467 expands the `dyn` rounding mode into 5 explicit `fcsr.frm` sweep variants. If that lands first (or the precedent it sets is accepted), the `frm_modes` tuple here should follow the same pattern for the dyn case.

Happy to sequence however you prefer. My read is #1469 first (pure refactor, no behavior change), then #1554, then #1467 on top of the deduped helper. But I'll defer to whatever order makes the review easier.

davidharrishmc:
I am concerned this change skips the tests entirely for cp_fs2_edges (without _frm) because frm_modes is set to None, and then the for loop doesn't iterate over any modes and doesn't print anything.  Could be mistaken, but hard to be certain without looking at the emitted code.

The floating-point tests are placeholders and will be superseded this summer by more comprehensive tests based on the IBM coverpoints.  Not sure that it's worth investing more time in tweaking and reviewing the placeholders.

dev-aditya-hub:
On (1): `frm_modes` is `[None]` (a 1-element list), not `None` the loop runs once with `frm_mode = None`, producing the same single test chunk per edge value as the previous `make_fs2_edges`. The `cp_fs2_edges` (no `_frm`) output is byte-identical to pre-refactor happy to attach a diff of generated tests if it'd help confirm.

On (2): fair point. #1469 (the refactor) already merged, so the shared helper is in tree. This PR is just the small follow-up that lets fs2/fs3 actually honor the `_frm` suffix (today they silently fall through to a single frm mode). If the IBM-coverpoint replacement is landing soon and you'd rather not carry even this small delta, happy to close your call.

jordancarlin:
As far as I can tell, this was fully done in #1469. Please update this PR if there is anything else that actually needs to change. For now, I’m going to close it.

PR #1556, opened 2026-05-21
sail_to_rvvi: emit the final instruction instead of dropping it

dev-aditya-hub:
## Summary

\`sailLog2Trace\` rotates \`output_line\` so each instruction is written only when the next instruction arrives — that's how RVVI's end-of-instruction mode field gets backfilled with the start mode of the following instruction (Sail logs mode at the start, RVVI expects it at the end). Once the outer loop exits, the trailing instruction is still sitting in \`output_line\` with its \`{mode_num}\` placeholder unfilled, and never reaches the trace file.

Every Sail-driven RVVI conversion silently truncates by exactly one instruction — typically the test's halt/sigupd at the tail. \`build_plan.py:320\` wires this into the live build path for every Sail config.

## Fix

After the loop, flush \`output_line\` using the last instruction's own start mode as the \`{mode_num}\` value. There is no architectural mode after the trailing instruction in the log; its own start mode is the closest defensible value rather than dropping the entry entirely.

Initialized \`prev_mode_num = None\` before the loop so the post-loop write is a clear no-op when the input contained zero parseable instruction lines.

## Risk

Low. Behavior is unchanged for every instruction *except* the last one in the log — that one was being silently dropped and now appears in the trace. No fields beyond \`{mode_num}\` change. No API change.

### DIFF
--- a/framework/src/act/sail_to_rvvi.py
+++ b/framework/src/act/sail_to_rvvi.py
@@ -34,6 +34,7 @@ def sailLog2Trace(inputLogFile: Path, outputTraceFile: Path) -> None:
     with inputLogFile.open() as f, outputTraceFile.open("w") as outfile:
         lines = f.readlines()
         output_line = ""
+        prev_mode_num: str | None = None
         for i in range(len(lines)):
             line = lines[i]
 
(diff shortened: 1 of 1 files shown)


davidharrishmc:
Best for @jordancarlin to review this one, but he may not be reviewing again until July.

dev-aditya-hub:
Ok sir

jordancarlin (approved):
The final instructions that exit the test should never be part of what is needed for coverage, and this script will be going away soon, but these changes seem correct, so I'm going to go ahead and merge.

PR #1561, opened 2026-05-22
Zicsr improvements for instret

davidharrishmc:
When U mode and Zicntr are supported, the Zicsr test suite picks instret as the CSR to test.  This CSR is not writable and the value may differ between simulators, leading to a mismatch (e.g. on cvw-rv32imc). It also can cause traps, which are not intended in unprivileged tests.

This PR fixes the problem for instret by:

1) avoiding writes to instret
2) returning the difference between two consecutive reads of instret rather than the absolute value.


dev-aditya-hub:
Nice fix, especially the two-read delta approach neatly sidesteps the writability question for U-mode.

One subtle case that the current guard doesn't catch: when `rs1 == rd` (or `rs2 == rd` for the immediate variant), the delta degenerates to zero. The `cmp_rd_rs1` coverpoint hits this systematically. For example in the regenerated `Zicsr-csrrs-00.S` around the `Test rd = rs1 = x1` case:

```
  csrrs x1, instret, x0       # first read into x1 (== rs1)
  csrrs x1, instret, x0       # second read into x1 (== rd) overwrites the first
  sub x1, x1, x1              # always 0
```

Since `rs1 == rd == x1`, the first read is clobbered by the second, and the subtraction is `xN - xN`. That happens for every register in the cmp_rd_rs1 set (and the analogous cmp_rd_rs2 set in the immediate variant).

It's not a correctness bug Sail computes the same zero, so the signature matches but the "check that instret value has incremented" intent is lost for that subset of cases, and the SIGUPD just sees zero regardless of what instret is doing.

Simplest fix would be to extend the existing guard:

```python
if instr_name in ["csrrw", "csrrwi"] or rs1 == 0 or rd == 0 or rs1 == rd:
    instret_access = f"li x{rd}, 0 # ..."
```

(and the corresponding `rs2 == rd` in `csri_type.py`). Falling back to `li xRD, 0` for those cases is consistent with the other skip cases and keeps the signature deterministic.

davidharrishmc:
Don’t care. If x0 works for every other instruction and Zicsr works for
every other register, I’m ok to let this be. More important things to spend
time on.

dev-aditya-hub:
Fair call agreed it's below the threshold worth churning on. Thanks for the quick read.

UmerShahidengr (approved):
Nice fix. Approved

PR #1564, opened 2026-05-22
RVA23 CRD Update

davidharrishmc:
Based on discussion today in Requirements WG.


docs/crd/src/rva23.adoc
allenjbaum: This should add the word "Test" to be consistent, (according to copilot)

docs/crd/src/rva23.adoc
```diff
@@ -58,13 +60,30 @@ Nevertheless, RVA23S64 runs within a supervisor execution environment.  For exam
...
-=== SBI Control of M-mode CSRs
+=== M-Mode Trap Handler Requirements
+
+Most traps default to being delegated to S-mode, with the following exceptions:
 
-RVA23 certification requires that the SBI provide methods of emulating every writable standard M-mode CSR field that can affect the behavior of lower privilege modes.  With a standard M-mode, the SBI may just write bitfields in the M-mode registers.  With custom M-mode, the SBI must make it appear to S-mode software that the field holds the value written by the SBI, even though the implementation might be different.
+* `ecall` from S-mode must not be delegated because the T-SBI must trap to M-mode to handle these calls.
+* Machine interrupts MEI, MSI, MTI are not delegated because they normally go to M-mode.
+* Any trap cause that should be handled invisibly by the firmware must not be delegated.
+* Bits of `m*deleg` that are read-only zero.
```
allenjbaum: change to "trap causes that hardware cannot delegate (e.g. bits of `m*deleg` that are read-only zero.)

allenjbaum (approved):

allenjbaum (approved):

docs/crd/src/rva23.adoc
```diff
@@ -28,13 +28,15 @@ a| * Initial draft
 
 == Introduction
 
-The RVA23 processor certificate is based on the https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva23s64_profile[RVA23S64] profile. To
+The RVA23 processor certificate is based on the https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva23s64_profile[RVA23S64] and RVA23U64 profiles. To
 qualify for a certificate, the processor must implement the RV64I base instruction set and all of the mandatory
```
kdockser, line 32: the processor must comply with the RVA23 profiles: RVA23U64 and RVA23S64.

Explanation: The profiles already explain what is required and what is optional. Retstaing the requirements here can lead to confusion.

docs/crd/src/rva23.adoc
```diff
@@ -28,13 +28,15 @@ a| * Initial draft
...
 extensions of the profile.  It may implement any of the optional features of
 the profile.  It may implement ratified extensions that are not part of the profile. It may implement other unratified extensions, but must not implement
 unratified extensions outside the custom encoding spaces if the optional Ssstrict extension is implemented. Extensions outside the profile are not tested during
 certification.
 
+Although it is theoretically possible to run RV32I instructions within user mode, the RVA23 certificate only covers RV64I within user mode, in accordance with RVA23U64.
```
kdockser: While it is permitted for an implementation to also support RV32I instructions, this is neither a mandated requirement or explicitly specified option. Therefore, there are no RV32I tests included in RVA23 certification. We encourage those who have implemented RV32I instructions in their processor to test these with the appropriate RISC-V ACTs.

docs/crd/src/rva23.adoc
kdockser, line 57: RVA23 does not place any specific requirements on M-mode implementations as it is not required that M-mode code is portable for compliant designs. Instead, in order to be able to run portable code for modes below M-mode, the processor must interface with a Supervisor Execution Environment (SEE) such as defined in the RISC-V SBI specification.

Explanation: The abstraction is not just for _custom_ M-mode **ISAs**. It allows for hiding differences in implementations as well allowing lower privileges to make requests to M-mode. Furthermore, it is common for implementations to comply with the RISC-V M-Mode ISA while adding implementation-defined features.

docs/crd/src/rva23.adoc
```diff
@@ -58,13 +60,30 @@ Nevertheless, RVA23S64 runs within a supervisor execution environment.  For exam
 
 RVA23 requires that the execution environment boot to a deterministic state that allows software to run in S-mode, and to handle traps to the execution environment that implements a test SBI (described below). If a custom M-mode execution environment does not directly implement a feature of standard M-mode in hardware, it must be able to emulate that feature in a way that is transparent to software running in lower privilege modes.  For example, if it does not support delegating interrupts to S-mode in hardware, it must accomplish the same purpose in software by trapping to the execution environment, which delivers the interrupt to S-mode.
```
kdockser, line 67: As defined in the Unprivileged architecture, the SEE, which will typically include M-mode hardware, must be able to provide certain capabilities to the underlying lower privileged implementation. For example, an SBI can be used to request that some or all traps be delegated to  be directly handled in S-Mode. Which traps can be delegated is up to the SEE.

Explanation: There is no requirement that any traps can be delegated. But if they can be, the SEE must provide a means to request that they are and a means to to determine which requested traps have been delegated.

docs/crd/src/rva23.adoc
```diff
@@ -58,13 +60,30 @@ Nevertheless, RVA23S64 runs within a supervisor execution environment.  For exam
...
 
-The execution environment must provide a Supervisor Binary Interface (SBI) that allows S-mode software to make requests of the execution environment.  These requests are made by placing commands in registers and performing an `ECALL`.
+The execution environment must provide a Test Supervisor Binary Interface (T-SBI) that allows S-mode software to make requests of the execution environment.  These requests are made by placing commands in registers and performing an `ECALL`.
 
-=== SBI Control of M-mode CSRs
+=== M-Mode Trap Handler Requirements
+
+Most traps default to being delegated to S-mode, with the following exceptions:
```
kdockser, line 77: NB: From the SBI, it appears that most traps _cannot_ be delegated to S-Mode. One could create custom (i.e., non-portable) SBI calls to delegate other traps.

davidharrishmc: We aren't testing the ratified SBI.  We are testing whether the DUT meets the requirements of the RVA23 profile.

kdockser: It is unclear to me what we mean by _most traps default_. Are we talking about in actual systems? In that case,  in order to be portable, OSs follow the RISC-V SBI; Therefore, a limited subset defaults to S-mode. Or, are we talking about in the certification tests? While the tests themselves might be able to delegate most to S-Mode, the actual testing is limited to what the DUT is able to delegate to S-mode.

We should provide further clarification.

docs/crd/src/rva23.adoc
kdockser: Just to be clear _Custom M-mode_ is _not_ a supervisor execution environment. M-mode itself is just the more privileged mode than [H]S and less than Debug.
Even a _custom M-mode ISA_ is  _not_ a supervisor execution environment. An M-mode _ISA_ can be a _part_ of the supervisor execution environment.

I think that the current text is confusing and misleading. CHERI will soon be ratified and we will then have a second _standard_ M-mode _ISA_ which can be used in an RVA23 compliant implementation.

PR #1566, opened 2026-05-23
Gate MTI testcase generation on MTIME support

### LINKED ISSUE #1543: MTIME `#ifdef` guards in InterruptsSm testgen are incomplete - testcase execute even when MTIME is absent
**Describe the bug**
`InterruptsSm` testcases that depend on the machine timer interrupt (MTI) are supposed toself-disable on CTPs that leave `RVMODEL_MTIME_ADDRESS` undefined. The `#ifdef` guards in `generators/testgen/src/testgen/asm/interrupts.py` only wrap the reads/writes accesses inside the `set_mtimer_int` / `clr_mtimer_int` helpers. The surrounding testcase structure - CSR writes, `wfi` instructions, signature capture - is emitted unconditionally. After `cpp` strips the reads/writes, what remains is a testcase shell that either hangs (`cp_wfi_*`, because the `wfi` has no wake source) or mismatches the Sail reference (`cp_trigger_mti_*`, because Sail's built-in CLINT fires MTI from cycle 0 regardless of arming, while the DUT never sees `irq_timer_i` go high). Every MTI leg of the cross-product coverpoints (`cp_interrupts`, `cp_priority`, `cp_vectored`) hits the same divergence.

**To Reproduce**
On core where `RVMODEL_MTIME_ADDRESS` is **not** defined in `rvmodel_macros.h`  e.g. cv32e20 run ACT4's https://github.com/openhwgroup/cv32e20-dv (sim/core/ make gen-certify)

The `cp_wfi_*` testcases will timeout (verilator wall-time cap); `cp_trigger_mti_*` and any cross-product MTI will produce a signature mismatch - `mcause`/`mip` entries present in Sail's output, absent in the DUT's. Only the helper reads/writes bodies were gated; the test scaffolding that calls them was not.

`cp_wfi_*` testcases are expected to hit the Verilator wall-time cap (simulator hangs, no output). `cp_trigger_mti_*` testcases are expected to produce a signature mismatch in the RVCP-SUMMARY log - Sail records an MTI trap entry, DUT produces none.

**Additional context**
- guards were added in PR #1002 
- fix must gate the full testcase block, not just the helpers.
- MSI and MEI coverpoints in the same run pass cleanly after the macro fixes, confirming the issue is specific to MTI arming.

somiljain2006:
Fixes #1543 

Fix MTI test case generation on platforms without RVMODEL_MTIME_ADDRESS by guarding entire MTI-dependent test case blocks instead of only timer helper accesses. This prevents hangs and Sail/DUT signature mismatches on targets such as CV32E20 while preserving MSI/MEI coverage generation.


generators/testgen/src/testgen/priv/extensions/InterruptsS.py
```diff
@@ -960,6 +966,9 @@ def _generate_vectored_s_tests(test_data: TestData) -> list[str]:
         for int_name, int_bit, int_set, int_clr, uses_timer in interrupts:
             binname = f"{stvec_mode_name}_{int_name}"
 
+            if int_name == "mtip":
+                lines.append("#if defined(RVMODEL_MTIME_ADDRESS) || defined(RVMODEL_SET_MTIMER_INT)")
```
davidharrishmc: There is no RVMODEL_SET_MTIMER_INT in the certification test plan, so lets drop that in several places.

davidharrishmc (commented):
Looks good with minor change.  Please resolve the merge conflict as well.

generators/testgen/src/testgen/priv/extensions/InterruptsS.py
```diff
@@ -1759,6 +1777,7 @@ def _generate_wfi_timeout_s_tests(test_data: TestData) -> list[str]:
     r_temp, r_stimecmp, r_scratch = test_data.int_regs.get_registers(3, exclude_regs=[])
 
     lines = [
+        "#if defined(RVMODEL_MTIME_ADDRESS) || defined(RVMODEL_SET_MTIMER_INT)",
```
davidharrishmc: This one doesn't involve a timer interrupt, so no need to exclude the whole thing.  Just exclude the part setting mtimecmp to all 1s.

generators/testgen/src/testgen/priv/extensions/InterruptsU.py
```diff
@@ -262,6 +266,7 @@ def _generate_user_wfi_timeout_tests(test_data: TestData) -> list[str]:
     r_temp, r_mtimecmp, r_scratch = test_data.int_regs.get_registers(3, exclude_regs=[])
 
     lines = [
+        "#if defined(RVMODEL_MTIME_ADDRESS) || defined(RVMODEL_SET_MTIMER_INT)",
```
davidharrishmc: Timer never fires here, so just gate part about clr_mtimer_interrupt, not entire test.

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -272,6 +279,10 @@ def _generate_vectored_tests(test_data: TestData) -> list[str]:
 
         # Raise each interrupt type
         for int_pending in ["meip", "mtip", "msip"]:
+
+            if int_pending == "mtip":
+                lines.append("#if defined(RVMODEL_MTIME_ADDRESS) || defined(RVMODEL_SET_MTIMER_INT)")
```
davidharrishmc: No such RVMODEL macro.  Global change.

davidharrishmc (changes requested):

davidharrishmc:
AI review

PR #1566 Review — "Gate MTI testcase generation on MTIME support"

Overview

The PR wraps MTI (Machine Timer Interrupt) test-generation blocks in #ifdef RVMODEL_MTIME_ADDRESS / #endif guards across three files (InterruptsS.py, InterruptsSm.py, InterruptsU.py). The goal is to prevent hangs and signature mismatches on targets like CV32E20 that lack a memory-mapped MTIME peripheral. The approach (guarding entire test blocks rather than individual timer-helper calls) is the right design; the concern is completeness.

---
Bugs: Unguarded clr_mtimer_int Calls

The following locations still call clr_mtimer_int() — which accesses RVMODEL_MTIME_ADDRESS — without an #ifdef guard, defeating the purpose of this PR for those targets.

1. InterruptsSm.py — _generate_priority_tests, cleanup block

The inner loop wraps the block in #ifdef RVMODEL_MTIME_ADDRESS only when mip_bits & 2 is set. But the unconditional cleanup at the bottom always emits clr_mtimer_int:

lines.extend([
    ...
    "RVTEST_CLR_MEXT_INT",
    *clr_mtimer_int(r_temp, r_mtimecmp),  # ← unguarded for mip_bits = 0,1,4,5
    "RVTEST_CLR_MSW_INT",
    ...
])
if mip_bits & 2:
    lines.append("#endif // RVMODEL_MTIME_ADDRESS")

For mip_bits ∈ {0, 1, 4, 5} (no MTIP bit), the generated assembly accesses MTIMEn.

Fix: Separate the timer clear from the other cleanup and guard it independently:

"RVTEST_CLR_MEXT_INT",
"#ifdef RVMODEL_MTIME_ADDRESS",
*clr_mtimer_int(r_temp, r_mtimecmp),
"#endif // RVMODEL_MTIME_ADDRESS",
"RVTEST_CLR_MSW_INT",

---
2. InterruptsU.py — _generate_user_wfi_timeout_tests

The function clears the timer to prevent interference during TW=1 WFI tests, but does so without a guard:

lines.append("# Clear timer (ensure no interrupt)")
lines.extend(clr_mtimer_int(r_temp, r_mtimecmp))   # ← unguarded

On a platform without MTIME, this hangs before any test case runs.

Fix: Wrap in #ifdef RVMODEL_MTIME_ADDRESS.

---
3. InterruptsU.py — make_interruptsu initial setup

The top-level generator clears any pending timer interrupt unconditionally:

lines.append("CSRW(mideleg, zero)")
lines.extend(clr_mtimer_int(r_temp, r_mtimecmp))   # ← unguarded

This fires before any test function is called, so even a platform that skips allhang here.

Fix: Wrap in #ifdef RVMODEL_MTIME_ADDRESS.

---
Smaller Issues

4. InterruptsSm.py _generate_vectored_tests — #endif comment missing

The outer MTVEC-mode guards close with a bare #endif, but the inner MTIME guards correctly close with #endif // RVMODEL_MTIME_ADDRESS. This is fine for the MTIME guards specifically, but the bare #endif for the mode guards is easy to misread in deeply nested code. Consid (e.g., #endif // UDB_MTVEC_MODES_<N>).

davidharrishmc:
Has this been tested on a system with no MTIME, and does it run?

jordancarlin:
@somiljain2006 are you still working on this? If so, please address the latest round of feedback from @davidharrishmc. Otherwise, please close this PR and someone else will address the issue.

PR #1567, opened 2026-05-24
Updated fast handler for Ssstrict tests to resolve dependency on mtval valid read

### LINKED ISSUE #1560: SsstrictSm trap loop on cores with mtval=0 for illegal instruction
**Test name**
tests/priv/SsstrictSm/SsstrictSm-00..10.s
**Describe the bug**
Every SsstrictSm test file injects its own fast illegal-instruction trap handler that overwrites `mtvec`. The handler uses `mtval & 3` to determine whether the faulting instruction was 2 bytes or 4 bytes before advancing `mepc`. On cores where `mtval` is hardwired to 0, the handler always treats every illegal instruction as 2 bytes, advancing `mepc` by 2 instead of 4. This leads to consequent traps, Sail detects this as a trap loop and aborts.

**Expected behavior**
The norm rules (`c-addi16sp_rsv`, `c-lwsp_rsv`, etc.) require both 2-byte and 4-byte reserved encodings to trap as illegal instruction. The handler must advance mepc correctly for both widths. mtval is not required to carry the encoding; these norm rules are fully testable on mtval=0 cores - the only broken piece is handler's size detection.
**Suggested change**
The fix has two parts: a generator (1) change and a handler change (2).
1. Before each illegal instruction, emit a `li a5, N` that loads the instruction's size in bytes into `a5`. The `li` is a valid, non-trapping instruction; `a5` holds the correct size when the illegal instruction traps immediately after it.
2. Replace the `andi t0, mtval, 3` / `beq` size branch with a direct add using `a5`

**Additional context**
There were some other options considered that didnt work:
- skipping tests (skips coverpoints that can be exercised)
- reading from memory instruction size (can break with PMP)
- fixed 4-byte advance (what to do with 2-byte illegal instructions)

UmerShahidengr:


generators/testgen/src/testgen/generate/priv.py
```diff
@@ -37,88 +37,64 @@
...
-#     Clobbers t0 and t1 only — acceptable for Ssstrict.
-#
-# The handler is defined before ssstrict_test_body: so that the LA() forward
-# reference resolves within this single translation unit.
+#     "// ── Fast illegal-instruction handler (prepended to every Ssstrict file) ────",
+#     "// Handles ALL illegal instruction traps — writes mcause, mepc and mtval to",
+#     "// signature on each trap, then advances mepc by 2 or 4.",
+#     "//",
+#     "// Instruction-width detection reads bits[1:0] from mepc directly.",
+#     "// PMP read access on the instruction region is guaranteed because the core",
```
davidharrishmc: This is not true.  There are some assumptions that should be mentioned here:

// This trap handler assumes that there is machine mode read access to the trapping instruction
// and that address translation is disabled so that the trap handler does not need to manipulate
// mstatus.MPRV to read the instruction.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -37,88 +37,64 @@
...
-    "\tSREG t0, 0(x2)          # store mtval to signature",
+    "\tcsrr t0, mtval",
+    "\tSREG t0, 0(x2)                  # store mtval to signature",
     "\taddi x2, x2, SIG_STRIDE",
-    "\tandi t0, t0, 3          # extract bits[1:0] into t0 (t0 still holds mtval)",
-    "\tli t1, 3                # uncompressed marker = 0b11",
-    "\tbeq t0, t1, uncompressedillegalinstructionreturn  # bits[1:0]==11 → uncompressed",
-    "compressedillegalinstructionreturn:",
+    "    # Branchless mepc advance — reads faulting instruction from *mepc.",
+    "    # PMP access is safe: the fetch already succeeded.",
```
davidharrishmc: Might not be safe if PMP has X=1, R=0.  Suggest deleting this line.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -37,88 +37,64 @@
...
-    "\taddi t0, t0, 2          # compressed: skip 2 bytes",
-    "\tj doneillegalinstructionreturn",
-    "uncompressedillegalinstructionreturn:",
-    "\tcsrr t0, mepc",
-    "\taddi t0, t0, 4          # uncompressed: skip 4 bytes",
-    "doneillegalinstructionreturn:",
-    "\tcsrw mepc, t0",
+    "\tlw   t0, 0(t0)                  # load faulting instruction word from *mepc",
+    "\tandi t0, t0, 3                  # t0 = bits[1:0]",
+    "\tsrli t0, t0, 1                  # t0 = 1 (uncompressed 0b11) or 0 (compressed)",
```
davidharrishmc: This doesn't work reliably because 0b10 will also be treated as uncompressed.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -37,88 +37,64 @@
...
+    "\tslli t0, t0, 1                  # t0 = 4 (uncompressed) or 2 (compressed)",
+    "\tcsrr t1, mepc                   # t1 = mepc  (t1 first written here)",
+    "\tadd  t1, t1, t0                 # t1 = mepc + advance",
+    "\tcsrw mepc, t1",
     "\tmret",
     "",
     "othertrap:",
     "\tcsrr t1, mtval",
-    "\tbgez t0, Mtrampoline    # msb clear = exception, jump to full handler",
+    "\tbgez t0, Mtrampoline            # msb clear = exception, hand off to full handler",
```
davidharrishmc: Why are interrupts ignored?  May not matter for these tests, but couldn't we just unconditionally jump to Trampoline?

davidharrishmc (changes requested):

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Fair point — the original comment implied PMP was sufficient to explain the direct memory read, but it missed the address-translation assumption. Added an explicit "Assumptions:" block covering both: (1) M-mode has read access to the faulting PC, and (2) address translation is disabled so the handler can read the instruction word at mepc without toggling mstatus.MPRV.

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: The concern is valid in general. However, this handler is only used by Ssstrict tests, which configure full PMP permissions (R=1, W=1, X=1), so the load is always safe in practice. Updated the comment to document this assumption rather than deleting the line.

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Good catch — Fixed by replacing the single right-shift with addi +1 followed by `srli` by 2: (bits[1:0] + 1) >> 2 is 1 only when bits[1:0] == 0b11, correctly identifying all other values as compressed. Updated the formula comment accordingly.

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Updated

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -37,88 +37,64 @@
...
+    "    # advance = ((bits[1:0] >> 1) + 1) << 1  →  4 if uncompressed, 2 if compressed.",
     "\tcsrr t0, mepc",
-    "\taddi t0, t0, 2          # compressed: skip 2 bytes",
-    "\tj doneillegalinstructionreturn",
-    "uncompressedillegalinstructionreturn:",
-    "\tcsrr t0, mepc",
-    "\taddi t0, t0, 4          # uncompressed: skip 4 bytes",
-    "doneillegalinstructionreturn:",
-    "\tcsrw mepc, t0",
+    "\tlw   t0, 0(t0)                  # load faulting instruction word from *mepc",
```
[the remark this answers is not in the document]
UmerShahidengr: Fixed the alignment hazard by switching from lw to lhu: mepc is always at least 2-byte aligned, so a 16-bit load is never misaligned. Since bits[1:0] are fully contained in the lower 16 bits, lhu gives us everything needed for width detection.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -37,88 +37,75 @@
...
-    "\taddi t0, t0, 2          # compressed: skip 2 bytes",
-    "\tj doneillegalinstructionreturn",
-    "uncompressedillegalinstructionreturn:",
-    "\tcsrr t0, mepc",
-    "\taddi t0, t0, 4          # uncompressed: skip 4 bytes",
-    "doneillegalinstructionreturn:",
-    "\tcsrw mepc, t0",
+    "\tlhu  t0, 0(t0)                  # load lower 16 bits from *mepc (always 2-byte aligned)",
+    "\tandi t0, t0, 3                  # t0 = bits[1:0]",
+    "\taddi t0, t0, 1                  # t0 = bits[1:0]+1; equals 4 only when was 0b11",
```
davidharrishmc, line 90: Other handler uses xor.  Make same between the two.

davidharrishmc (approved):

PR #1571, opened 2026-05-25
fix: guard UDB mtvec balign directives for single-mode cores

### LINKED ISSUE #1565: `UDB_MTVEC_BASE_ALIGNMENT_DIRECT` is used unconditionally in `rvtest_trap_handler.h`
**Describe the bug**
Commit 8112c2ad replaced the old single-line alignment:
```asm
.align MTVEC_ALIGN   // had a #ifndef fallback default of 6
```

with three unconditional `.balign` directives:

```asm
.balign 64
.balign UDB_MTVEC_BASE_ALIGNMENT_VECTORED
.balign UDB_MTVEC_BASE_ALIGNMENT_DIRECT   
```

For cv32e20, which declares `MTVEC_MODES: [1]` (vectored only), UDB correctly does not generate UDB_MTVEC_BASE_ALIGNMENT_DIRECT. Since this macro is missing from UDB-generated rvtest_config.h for vectored-only cores, but used unconditionally in rvtest_trap_handler.h, assembler fails with `bad or irreducible absolute expression` when .balign has undefined symbol.

**To Reproduce**
```
make CONFIG_FILES=config/cores/cve2/cv32e20/test_config.yaml
```
 or any core with MTVEC_MODES: [1]

tests/env/rvtest_trap_handler.h:1012: Error: bad or irreducible absolute expression
tests/env/rvtest_trap_handler.h:346:  Info: macro invoked from here
tests/env/rvtest_setup.h:131:        Info: macro invoked from here
tests/priv/ExceptionsZc/ExceptionsZc-00.S:3503: Info: macro invoked from here

**Additional context**
- `tests/env/rvtest_trap_handler.h` is invoked for every test via `rvtest_setup.h` under `#ifdef STANDARD_SM_SUPPORTED
- fix is simple add guards (same as for direct)

karabambus:
fixes #1565

### DIFF
--- a/tests/env/rvtest_trap_handler.h
+++ b/tests/env/rvtest_trap_handler.h
@@ -1008,8 +1008,12 @@ rvtest_\__MODE__\()prolog_done:
 .option rvc             // temporarily allow compress to allow c.nop alignment
 // Ensure that trampoline is on a boundary that is the max of 64 bytes, UDB_MTVEC_BASE_ALIGNMENT_VECTORED, and UDB_MTVEC_BASE_ALIGNMENT_DIRECT
 .balign 64
+#ifdef UDB_MTVEC_BASE_ALIGNMENT_VECTORED
 .balign UDB_MTVEC_BASE_ALIGNMENT_VECTORED
+#endif
+#ifdef UDB_MTVEC_BASE_ALIGNMENT_DIRECT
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
I thought I added ifdefs to these, but must have accidentally dropped them during one of the many refactors that the PR went through. Thanks for the fix.

PR #1573, opened 2026-05-25
fix: zicsr mepc fallback u mode

### LINKED ISSUE #1572: Zicsr tests fail to compile for U-mode cores without F, V, or Zicntr
**Test name**

**Describe the bug**

```c
#else
```

**Expected behavior**

**Additional context**

Fixes considered and gaps:

karabambus:
fixes #1572 by implementing proposed fix (note: there are still gaps that this fix doesn't cover see issue)


davidharrishmc (changes requested):
The boot code is being modified soon to run unprivileged tests in the lowest privilege mode available.  Therefore, Zicsr tests will run in U mode if supported. That would break this proposed change because accesses to mepc or sepc will fail.  

Maybe it would be easier to gate these tests with REQUIRED_EXTENSIONS to only run if F_SUPPORTED or V_SUPPORTED or ZICNTR_SUPPORTED or not U_SUPPORTED?

If only SM and U are supported, the boot code already will do basic manipulation of CSRs on the way to reach U-mode.

jordancarlin:
We don't have a way to say that one of a list of extensions is required for a test right now. We can only say that all of the extensions are required. I could add this feature if needed (it would probably have to wait until after the summit though).

Alternatively, what if we added something like the following to one of the header files. Then if there are no unpriv CSRs to use, we fall back to testing a priv CSR in a higher priv mode.

```c
#ifdef RVMODEL_TEST_CSR
    #define RVTEST_TEST_CSR RVMODEL_TEST_CSR
#else
    #if defined(F_SUPPORTED)
        #define RVTEST_TEST_CSR fflags
    #elif defined(ZVE32X_SUPPORTED)
        #define RVTEST_TEST_CSR vxsat
    #elif !defined(U_SUPPORTED)
        #define RVTEST_TEST_CSR mepc
        #define RVTEST_TEST_CSR instret
    #elif defined(S_SUPPORTED)
        #define RVTEST_TEST_CSR sepc
        #define BOOT_TO_S_MODE
    #elif defined(STANDARD_SM_SUPPORTED)
        #define RVTEST_TEST_CSR mepc
        #define BOOT_TO_M_MODE
    #else
        #error no CSR known for testing. Zicsr testing requires F, V, Zicntr, S, or STANDARD_SM_SUPPORTED. If you only have custom CSRs, define RVMODEL_TEST_CSR.
    #endif
#endif
```

jordancarlin:
@karabambus I'm considering a larger refactor of the syntax behind our YAML header soon to more closely match UDB's constraint system, so I would rather avoid adding the `REQUIRED_EXTENSIONS_ANY_OF` key. Can you update this to use the list of ifdefs that I described in my previous comment? I'm also happy to create a new PR myself if that is easier. Sorry for the back and forth and thanks for all your work on this!

karabambus:
@jordancarlin no problem. I am currently working on this ready to push.

tests/env/rvtest_setup.h
```diff
@@ -3,6 +3,29 @@
...
+  #elif defined(ZICNTR_SUPPORTED)
+    #define RVTEST_TEST_CSR instret
+  #elif defined(S_SUPPORTED)
+    #define RVTEST_TEST_CSR sepc
+    #define BOOT_TO_SMODE
+  #elif defined(STANDARD_SM_SUPPORTED)
+    #define RVTEST_TEST_CSR mepc
+    #define BOOT_TO_MMODE
+  #endif
+#endif
```
jordancarlin: No need for the outer ifndef wrapper. The user should only ever define `RVMODEL_TEST_CSR`, so `RVTEST_TEST_CSR` will never be defined before getting to this point.

tests/env/rvtest_setup.h
```diff
@@ -3,6 +3,29 @@
...
 
+/****************************** Zicsr test CSR selection *********************************/
+// Pick a CSR for Zicsr tests. Set here, before the boot cascade, so BOOT_TO_*MODE pins
+// the test to a mode where the CSR is writable. DUT may override via RVMODEL_TEST_CSR.
+#ifndef RVTEST_TEST_CSR
+  #ifdef RVMODEL_TEST_CSR
+    #define RVTEST_TEST_CSR RVMODEL_TEST_CSR
+  #elif defined(F_SUPPORTED)
+    #define RVTEST_TEST_CSR fflags
+  #elif defined(V_SUPPORTED)
```
jordancarlin: Let's change this to ZVE32X instead of V. Any vector implementation (even the more minimal extensions) will support the vxsat CSR.

tests/env/rvtest_setup.h
jordancarlin: This might fit better in `utils.h`.

generators/testgen/src/testgen/formatters/types/csri_type.py
```diff
@@ -31,16 +31,10 @@ def zicsr_acccessi(instr_name: str, rd: int, rs2: int, immval: int) -> str:
...
 
     return (
-        "#if defined(F_SUPPORTED)\n"
-        f"{instr_name} x{rd}, fflags, {immval}\n"
-        "#elif defined(V_SUPPORTED)\n"
-        f"{instr_name} x{rd}, vxsat, {immval}\n"
-        "#elif !defined(U_SUPPORTED)\n"
-        f"{instr_name} x{rd}, mepc, {immval}\n"
-        "#elif defined(ZICNTR_SUPPORTED)\n"
+        "#if defined(ZICNTR_SUPPORTED) && defined(U_SUPPORTED) && !defined(F_SUPPORTED) && !defined(V_SUPPORTED)\n"
```
jordancarlin: Instead of duplicating this selection logic, what about adding a `#define RVTEST_READ_ONLY_TEST_CSR` (or something similar) and using that as the `#ifdef` here?

jordancarlin (changes requested):
A few small tweaks, but this looks close.

tests/env/rvtest_setup.h
karabambus: Had to change include ordering in utils.h, afaik it shouldn't break anything

tests/env/utils.h
```diff
@@ -3,6 +3,30 @@
...
 
+// Zicsr test CSR selection. Set before rvtest_setup.h's boot cascade so BOOT_TO_*MODE
+// pins the test to a mode where the chosen CSR is writable. A DUT may override the
+// choice via RVMODEL_TEST_CSR. RVTEST_READ_ONLY_TEST_CSR marks a non-writable choice.
+#ifdef RVMODEL_TEST_CSR
+  #define RVTEST_TEST_CSR RVMODEL_TEST_CSR
+#elif defined(F_SUPPORTED)
+  #define RVTEST_TEST_CSR fflags
+#elif defined(ZVE32X_SUPPORTED)
+  #define RVTEST_TEST_CSR vxsat
```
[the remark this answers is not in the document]
karabambus: I have verified ZVE32X_SUPPORTED is emitted for vector cores

tests/env/utils.h
```diff
@@ -3,6 +3,30 @@
...
+  #define RVTEST_READ_ONLY_TEST_CSR
+#elif defined(S_SUPPORTED)
+  #define RVTEST_TEST_CSR sepc
+  #define BOOT_TO_SMODE
+#elif defined(STANDARD_SM_SUPPORTED)
+  #define RVTEST_TEST_CSR mepc
+  #define BOOT_TO_MMODE
+#else
+  #error no CSR known for testing. Zicsr testing requires F, V, Zicntr, S, or STANDARD_SM_SUPPORTED. If you only have custom CSRs, define RVMODEL_TEST_CSR.
+#endif
```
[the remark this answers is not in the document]
karabambus: Gated the whole selection on RVTEST_SELECT_TEST_CSR, which the generator now emits only in Zicsr test headers (templates.py).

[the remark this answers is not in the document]
jordancarlin: Rather than introducing yet another macro, how about we use ZICSR_SUPPORTED?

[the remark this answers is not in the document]
karabambus: I need a per-test-file signal, not a per-DUT capability flag. I don't think using this macro would solve the problem - we want to gate selection block for only Ziscr tests. Maybe we should move this out of utils.h?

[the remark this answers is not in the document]
jordancarlin: I don’t think we actually need it to be per test. If the DUT supports ZICSR, then it will run the ZICSR test, so there needs to be a valid CSR to pick. There is no harm in the selection happening in all of the tests.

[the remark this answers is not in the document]
karabambus: okay, this should be resolved. Thank you for saying with me.

tests/env/utils.h
```diff
@@ -3,6 +3,34 @@
 # Jordan Carlin jcarlin@hmc.edu November 2025
 # SPDX-License-Identifier: BSD-3-Clause
 
+// Zicsr test CSR selection. Set before rvtest_setup.h's boot cascade so BOOT_TO_*MODE
+// pins the test to a mode where the chosen CSR is writable. A DUT may override the
+// choice via RVMODEL_TEST_CSR. RVTEST_READ_ONLY_TEST_CSR marks a non-writable choice.
```
davidharrishmc: I feel that RVMODEL_TEST_CSR is overkill.  I'd rather not complicate the RVMODEL macros by introducing this, which every user will need to understand and likely none will ever need to use.  If there are no standard CSRs supported, how about just not declaring ZICSR_SUPPORTED?

Also, I'm concerned about mixing BOOT_TO_SMODE and BOOT_TO_MMODE here, which won't be defined if, for example, !U_SUPPORTED.  Those should be in their own #if group if they belong here at all.

davidharrishmc (changes requested):

tests/env/utils.h
karabambus: okay, I have made requested changes. 

This PR has been result of misconfiguring cv32e20, so I guess safe option is to also close it.

PR #1432, opened 2026-05-04
updated trap handler and updated trap failure code

UmerShahidengr:
### Summary
 
This PR adds the **Test Supervisor Binary Interface (T-SBI)** dispatch to `rvtest_trap_handler.h` and enhances trap failure reporting in `failure_code.h` with per-field mismatch identification, xcause name decoding, and actionable diagnostic hints.
 
### T-SBI Implementation (`rvtest_trap_handler.h`)
 
Implements the T-SBI ecall-based interface for S/U-mode tests to request services from the execution environment. Uses `a0` for the operation code and `a1` for an optional argument, with results returned in `a0`. 
 
**Operations implemented:**
 
| Operation | a0 Value | Description |
|-----------|----------|-------------|
| `GOTO_MMODE` | 1 | Switch to M-mode (sets MPP=11, clears MPV) |
| `GOTO_SMODE` | 2 | Switch to S/HS-mode (sets MPP=01) |
| `GOTO_UMODE` | 3 | Switch to U-mode (sets MPP=00) |
| `GOTO_VSMODE` | 4 | Switch to VS-mode (MPP=01, MPV=1) |
| `GOTO_VUMODE` | 5 | Switch to VU-mode (MPP=00, MPV=1) |
| `ECALL_TEST` | 0x73 | Returns xEPC in a0 to verify trap path |
| `CSR_ACCESS` | CSR encoding | Executes CSR instruction dynamically via scratch memory |
| Reserved | other | Returns -1 |
 
**M-mode handler** services all operations directly. **S-mode handler** handles `ECALL_TEST`, `GOTO_SMODE`, `GOTO_UMODE`, and S-mode `CSR_ACCESS` locally; forwards `GOTO_MMODE`, `GOTO_VSMODE`/`VUMODE`, and M-mode `CSR_ACCESS` to M-mode via ecall with `a0`/`a1` passthrough.
 
**CSR_ACCESS mechanism:** Writes the caller-supplied 32-bit CSR instruction encoding to scratch memory (reuses `rvmodel_sv` area in the save area — no layout changes), appends a `ret` instruction, executes `fence.i`, then `jalr` into the scratch to execute the CSR instruction dynamically. Requires `rd=a0` in the encoding for reads and `rs1=a1` for writes.
 
**Key design constraints preserved:**
- `actual_tramp_sz` and `sv_area_sz` unchanged (no trampoline or save area layout changes)
- Legacy `x3==0` GOTO_MMODE/GOTO_SMODE paths fully preserved and functional
[...]

PR #1575, opened 2026-05-25
Fix/rvtest trap honor mtvec modes

### LINKED ISSUE #1574: Trap handler default setup assumes direct mtvec mode, vectored-only cores get mismatched trap-vector signatures
**Describe the bug**

The trap framework sets up `mtvec` in direct mode by default (and the InterruptsSm cross-product section additionally selects direct via `csrci mtvec, 3`) without honoring the DUT's declared `MTVEC_MODES`. For a vectored-only core (`MTVEC_MODES: [1]`, e.g. cv32e20), the core's WARL `mtvec.MODE` legally forces vectored mode, while the Sail reference - whose config does not restrict `mtvec` to vectored-only - runs in direct mode.

As a result, the DUT and the reference take the same interrupt to different trap-vector entries.

This is the **run-time counterpart of #1565**: both stem from the framework not fully handling vectored-only cores. #1565 was the assemble-time facet (missing `UDB_MTVEC_BASE_ALIGNMENT_DIRECT` guard); this is the run-time facet (direct-mode trap vectoring assumed reachable).

Note: this is independent of the timer/CLINT path. The same divergence appears on any interrupt (MSI/MEI) because the reference vectors in direct mode regardless of cause.

**To Reproduce**

Branches (forks) to reproduce:
- Test framework + cv32e20 config:
  https://github.com/karabambus/riscv-arch-test/tree/fix/cv32e20-interruptssm-enable
- DUT (cv32e20-dv with a CLINT mtime/mtimecmp so timer interrupts can fire):
  https://github.com/karabambus/cv32e20-dv/tree/ci/act4-interruptssm-enable

Steps:
1. On `act4` @ `48332172` (the base of the branches above), use a vectored-only core config (`MTVEC_MODES: [1]`), e.g. `config/cores/cve2/cv32e20` with InterruptsSm enabled.
2. Generate the test:
   `make CONFIG_FILES=config/cores/cve2/cv32e20/test_config.yaml` (InterruptsSm).
3. Inspect the generated reference signature `work/cv32e20/build/priv/InterruptsSm/InterruptsSm-00.sig`. The first trap-signature word (vec-offset/#entries/mode, per `rvtest_trap_handler.h`) is **cause-independent**: the MTI trap (`mcause 0x80000007`) and the MSI trap (`mcause 0x80000003`) both yield   `0x03101b53` / `0x03101b4f` - vector field = 13 in both - i.e. direct-mode vectoring.
4. Run `InterruptsSm-00.elf` on the vectored-only DUT. From the cv32e20-dv branch above: `cd sim/core && make gen-certify FILTER=InterruptsSm`. The core forces vectored mode, and the self-check halts with "Mismatch in trap signature".

```
RVCP-SUMMARY: TEST FAILED - Test File "InterruptsSm-00.S"
RVCP: Test Info: "Mismatch in trap signature! Trap was being handled in M-Mode."
RVCP: Register: x8
RVCP: Bad Value:      0x03101d13     # vector field = 20  (vectored, DUT)
RVCP: Expected Value: 0x03101b53     # vector field = 13  (direct,   reference)
RVCP: Address of instruction that trapped (XEPC): 0x00004238
```

Reference `.sig` (cause-independent vector = direct mode):
```
03101b53   # word0  vector=13
80000007   # mcause = MTI  (cause 7)
00000080   # mip    = MTIP (bit 7)
deadbeef
03101b4f   # word0  vector=13   (same vector as MTI)
80000003   # mcause = MSI  (cause 3)
00000008   # mip    = MSIP (bit 3)
```

**Additional context**

- cv32e20 forcing vectored mode on a write of `MODE=0` is spec-compliant (`mtvec.MODE`
  is WARL; a core may implement vectored-only).
- The framework has the information it needs - `MTVEC_MODES` is present in `rvtest_config.h` as `UDB_MTVEC_MODES_1` - but does not use it to keep the DUT and the reference in the same mode.
- Candidate fix: gate direct-mode trap setup and the `MODE=00` cross-product on `MTVEC_MODES` containing 0, so vectored-only cores never exercise direct-mode paths. (with #1571)
- Same family as #1565 (`UDB_MTVEC_BASE_ALIGNMENT_DIRECT` used unconditionally).

### LINKED ISSUE #1590: Trap handler relies on reset mtvec.MODE; cores that reset to vectored (cv32e40p) get mismatched trap-vector signatures in InterruptsSm
**Describe the bug**

`priv/InterruptsSm/InterruptsSm-00` fails on cv32e40p with a trap-signature mismatch (`Vector+Mode+Status word`):

- Expected (reference): `0x03101b53` - vector field = 13 (direct-mode entry)
- Actual (DUT):          `0x03101d13` - vector field = 20 (vectored-mode entry)

The first interrupt-trigger sub-tests (`trigger_mti/msi/mei`) run before the test writes `mtvec.MODE`, so they take their interrupts in whatever mode `mtvec` reset to. cv32e40p resets `mtvec.MODE` to vectored (`2'b01`), while the Sail reference resets `mtvec` to `0` = direct (no reset override in its config). The trap framework's prolog preserves the boot-time MODE bits (`rvtest_trap_handler.h`: `andi T2, T3, WDBYTMSK`), so the trampoline is installed vectored on the DUT and direct on the reference. The same interrupt is therefore taken to two different trampoline entries, and the recorded vector field diverges.

**To Reproduce**

1. On https://github.com/karabambus/cv32e40p-dv-review/tree/feat/mm_ram-interrupt-support set in Act4 head in external repos. Edit configs to add interrupts (copying from cv32e20 rvmodel_macros machine interrupts section, and uncommeting and setting default address for mtime and mtimecmp
2. Generate + run InterruptsSm:
3. Inspect the reference signature
   `work/cv32e40p-v2-rv32imc/build/priv/InterruptsSm/InterruptsSm-00.sig`
4. The DUT, taking the first interrupt at its reset MODE = vectored, vectors to `BASE + cause*4` and the self-check halts with "Mismatch in trap signature".

```
RVCP-SUMMARY: TEST FAILED - Test File "InterruptsSm-00.S"
RVCP: Test Info: "Mismatch in trap signature! Trap was being handled in M-Mode."
RVCP: ===== TRAP FAILURE DIAGNOSTICS =====
RVCP: Failure: "Mismatch in trap signature! Trap was being handled in M-Mode."
RVCP: Trap handler mode: M-mode
RVCP: Mismatching field: Vector+Mode+Status word (trap signature word 0)
RVCP: Expected value: 0x03101b53     # vector field = 13 (direct,   reference)
RVCP: Actual value:   0x03101d13     # vector field = 20 (vectored, DUT)
RVCP: HINT: Vector+Mode word mismatch may indicate: trap handled in wrong
RVCP:       privilege mode (check medeleg/mideleg), incorrect mstatus fields
RVCP:       (MPP/SPP/MPV), or wrong vectored interrupt entry.

[tb_top] @ t=87730ns: TEST(S) FAILED!
```

Reference `.sig` (cause-independent vector = direct mode):
```
03101b53   # word0  vector=13
80000007   # mcause = MTI  (cause 7)
00000080   # mip    = MTIP (bit 7)
deadbeef
03101b4f   # word0  vector=13   (same vector as MTI)
80000003   # mcause = MSI  (cause 3)
00000008   # mip    = MSIP (bit 3)
```

**Additional context**

- cv32e40p forcing vectored vectoring on its reset MODE is spec-compliant (`mtvec.MODE` reset value is implementation-defined; the field is WARL R/W and the core honors explicit writes — `cv32e40p_core.sv:361`).
- The framework has the information it needs (`MTVEC_MODES` / `UDB_MTVEC_MODES_*` in `rvtest_config.h`) but does not keep the DUT and the reference in the same mode for the `trigger_*` sections, which run before any `mtvec.MODE` write.
- I have two fixes in mind: 
    1. Setting the mode explicitly i.e. rvtest_trap_handler.h prolog.
    2. Adding Sail parameter for default mtvec reset value

karabambus:
Stacked on  #1571
fixes #1574 
fixes #1590


UmerShahidengr:
Better hold it till #1432 gets merged.

jordancarlin:
@karabambus can you extend this to also have ifdefs that would work for a core that only supports direct mode?

karabambus:
@jordancarlin I extended to direct mode. I really want CI to get working it was shortsighted of me for covering only my case.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -257,8 +260,11 @@ def _generate_vectored_tests(test_data: TestData) -> list[str]:
         )
     ]
 
-    # Test both mtvec modes
+    # Test both mtvec modes; gate each mode on its own MTVEC_MODES so single-mode
+    # cores (direct-only or vectored-only) only exercise the mode they implement.
     for mode, mode_name in [(0, "direct"), (1, "vectored")]:
+        mode_guard = "UDB_MTVEC_MODES_0" if mode == 0 else "UDB_MTVEC_MODES_1"
+        lines.append(f"#ifdef {mode_guard}")
```
jordancarlin: Can be simplified to the following so the variable can be dropped.
```suggestion
        lines.append(f"#ifdef UDB_MTVEC_MODES_{mode}")
```

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -257,8 +260,11 @@ def _generate_vectored_tests(test_data: TestData) -> list[str]:
         )
     ]
 
-    # Test both mtvec modes
+    # Test both mtvec modes; gate each mode on its own MTVEC_MODES so single-mode
```
jordancarlin:
```suggestion
    # Test both mtvec modes; gate each mode on its own MTVEC_MODES parameter so single-mode
```

tests/env/rvtest_trap_handler.h
jordancarlin, line 1: There are a lot of unnecessary white space and comment changes in this file. Please revert them (many seem unintentional) to make it easier to review.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
```diff
@@ -257,8 +260,11 @@ def _generate_vectored_tests(test_data: TestData) -> list[str]:
         )
     ]
 
-    # Test both mtvec modes
+    # Test both mtvec modes; gate each mode on its own MTVEC_MODES parameter so single-mode
+    # cores (direct-only or vectored-only) only exercise the mode they implement.
     for mode, mode_name in [(0, "direct"), (1, "vectored")]:
+        mode_guard = "UDB_MTVEC_MODES_0" if mode == 0 else "UDB_MTVEC_MODES_1"
```
jordancarlin: Looks like this is unused and can be dropped.

jordancarlin (approved):
One minor nit. Otherwise, LGTM.

generators/testgen/src/testgen/priv/extensions/InterruptsSm.py
karabambus: done

karabambus:
@davidharrishmc @jordancarlin  can we merge, it should unblock CI for cv32e20

karabambus:
I added fix to #1590 - force deterministic supported mtvec.MODE in trap prolog.

davidharrishmc (approved):

PR #1576, opened 2026-05-25
cv32e20 interrupts enable

karabambus:


karabambus:
InterruptsU tests signature mismatch, cause is bug in RTL - https://github.com/openhwgroup/cve2/issues/327

davidharrishmc (approved):

PR #1577, opened 2026-05-25
SmV/SmVF: capture vector CSR access and fflags in the signature

dev-aditya-hub:
## Summary

`SmV` and `SmVF` emit testcases but never write their results into the signature, so those checks are coverage-only and can't fail a DUT. This adds signature readbacks for the parts that are safe to check across all reference models, matching what `SmF` (scalar FP CSRs) and `UV` (the same vector CSRs from U-mode) already do.

## What it checks now

- `SmV` `cp_vcsrrswc`: after csrrs/csrrc/csrrw on each vector CSR (`vstart`/`vxsat`/`vxrm`/`vcsr`/`vl`/`vtype`/`vlenb`), read the CSR back into the signature.
- `SmV` `cp_vcsrs_walking1s`: after each walking-1s `CSRW`, read the written CSR back.
- `SmV` `cp_vstart_out_of_bounds`: read `vstart` back after the out-of-bounds write.
- `SmVF` `cp_vectorfp_mstatus_fs_state_affecting_csr`: read `fflags` back after the `vfdiv`/`vfadd` that raise DZ/NV.

These mirror `UV`'s `csr_access_test`/`csr_walk_test` and `SmF`'s FP-CSR readbacks, which already pass on every model.

## Deliberately not captured

An earlier revision of this branch also read back `vl`/`vtype` after every `vsetvl*` and the `mstatus.FS`/`VS` dirty bits. CI showed why those don't belong in a cross-model conformance signature:

- `mstatus.FS`/`VS` dirty transitions: the reference simulators disagree. After a `vfdiv` that raises DZ under `FS=Initial`, sail reports `FS=Dirty` but qemu's `max` CPU leaves it `Initial`, so the test can't pass on every model.
- The full `vsetvl*` `vl`/`vtype` sweep added ~2.6k signature writes, which overflowed the signature region and aborted qemu.

This revision keeps only the deterministic, model-agreed readbacks. The `mstatus`-dirty and `vsetvl`-result coverpoints stay coverage-only, which is the right call until those behaviours are pinned down.

## Notes

- The readbacks are appended after the instruction under test (read-only `CSRR`), so they don't disturb coverage sampling.
- Removed a dead `.rept __riscv_xlen` stub in `cp_vcsrs_walking1s` that was sliced out of the line list and never emitted.
- `ruff format`/`check` pass; both generators run through `generate_priv_test`. SmV now emits 278 signature writes, comparable to `UV`'s 924.
- Reference signatures for `SmV`/`SmVF` need regenerating when merged.

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/SmV.py
+++ b/generators/testgen/src/testgen/priv/extensions/SmV.py
@@ -8,6 +8,7 @@
 
 """SmV privileged test generator: vector CSRs and vtype/vl/vstart behavior in M-mode."""
 
+from testgen.asm.csr import gen_csr_read_sigupd
 from testgen.asm.helpers import comment_banner
 from testgen.data.state import TestData
 from testgen.priv.registry import add_priv_test_generator
@@ -51,14 +52,14 @@ def _gen_vcsrrswc(test_data: TestData, temp_reg: int) -> list[str]:
     ]
     lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
     lines.extend(_vector_setup(temp_reg))
-    save_reg = test_data.int_regs.get_register()
+    save_reg, check_reg = test_data.int_regs.get_registers(2)
     lines.append(f"LI(x{save_reg}, -1)  # all 1s mask for csr ops")
     for csr in _VECTOR_CSRS:
         for op_name, op in (("csrrs", "CSRS"), ("csrrc", "CSRC"), ("csrrw", "CSRW")):
             lines.append(test_data.add_testcase(f"{csr}_{op_name}", coverpoint, _CG))
             lines.append(f"{op}({csr}, x{save_reg})  # {op_name} {csr}")
-            lines.append("nop")
-    test_data.int_regs.return_registers([save_reg])
+            lines.append(gen_csr_read_sigupd(check_reg, (csr, None), test_data))
+    test_data.int_regs.return_registers([save_reg, check_reg])
     return lines
 
 
@@ -72,26 +73,10 @@ def _gen_vcsrs_walking1s(test_data: TestData, temp_reg: int) -> list[str]:
     ]
     lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
     lines.extend(_vector_setup(temp_reg))
-    walk_reg, mask_reg = test_data.int_regs.get_registers(2)
-    for csr in _VECTOR_CSRS_WR:
-        lines.append(f"# walking-1s on {csr}")
-        lines.append(f"LI(x{mask_reg}, -1)  # all 1s")
-        lines.append(f"LI(x{walk_reg}, 1)   # one-hot starting at bit 0")
-        # We need to walk all XLEN bits; emit for both 32 and 64 via #if
-        lines.append(".rept __riscv_xlen")
-        # Note: .rept doesn't access __riscv_xlen as a number; emit per-bit explicitly with #if guard
-        lines.append(".endr")
-    # Replace the .rept stub with explicit unroll: emit bit-by-bit with #if for RV64 high bits
-    lines = lines[
-        : lines.index(
-            comment_banner(
-                coverpoint, "csrrw walking-1s into vstart/vxsat/vxrm/vcsr (covers walking_ones_rs1 bins 0..XLEN-1)"
-            )
-        )
-        + 1
-    ]
-    lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
-    lines.extend(_vector_setup(temp_reg))
+    # Walk one bit at a time, reading the CSR back after each write so the
+    # signature captures how the DUT latched it. RV64 high bits are guarded by
+    # #if so the RV32 build only emits the low 32.
+    walk_reg, mask_reg, check_reg = test_data.int_regs.get_registers(3)
     for csr in _VECTOR_CSRS_WR:
         lines.append(f"# walking-1s on {csr}")
         lines.append(f"LI(x{mask_reg}, -1)  # all 1s")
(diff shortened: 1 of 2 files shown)


jordancarlin:
How does this relate to #1553? It seems like this supersedes all of the changes in that PR. If so, please close that PR. In the future, for a small set of related changes, please just update the existing PR. It makes it harder to review when there are many similar PRs that all include some of the same changes.

dev-aditya-hub:
Superseded by #1577, 
which has the complete cleaned-up version of these changes closing this to keep everything consolidated in one place. Thanks!

jordancarlin:
@dev-aditya-hub I think you closed the wrong PR. This one looks more complete than #1553.

dev-aditya-hub:
Sorry sir I will reopen it and close the other one

jacassidy:
From my understanding the issue you raised for not including readback for mstatus is a bug in QEMU. Per the spec

> Any vector floating-point instruction that modifies any floating-point extension state (i.e., floating-point CSRs or f registers) must set mstatus.FS to Dirty.

You stated that when comparing mstatus.FS after vfdiv and the the DZ flag qemu fails to set mstatus.FS to dirty

dev-aditya-hub:
Thanks @jacassidy — that's a really helpful clarification, and you're right. The spec is unambiguous: a vector-FP op that raises a flag is modifying FP state, so `mstatus.FS` must transition to Dirty. So omitting the `mstatus.FS` readback because QEMU didn't set it was me treating a QEMU bug as if it were a spec gray area — that was the wrong call on my part.

Given that, I think the right thing is to add the `mstatus.FS == Dirty` readback after the `vfdiv`/`vfadd` cases (it's deterministic and spec-mandated) and treat `-cpu max` not setting it as a QEMU bug. Would you prefer I (a) add the check now and file/track the QEMU issue so the qemu reference run is expected to fail until it's fixed, or (b) hold the check until QEMU is patched so CI stays green in the meantime? Happy to go whichever way the project prefers.

jacassidy:
I think a sounds great, we also found a qemu bug in UV so we added it to the extensions list to skip in CI, if we can do that and document the bug, as well as add a comment above the excluded extension explaining what the exclusion is holding off on that would be great.


If you go to config/qemu/ci.yaml you'll see a list of extesnions to skip for CI, I would add it there as well as a comment so we know why its currently excluded

dev-aditya-hub:
While debugging a vector self-check failure here, I hit what looks like a framework-side issue in the self-check failure path that's independent of my test changes.

When a `RVTEST_SIGUPD` comparison fails in a vector test, control jumps to `failedtest_vec_*`, which dumps all 32 vector registers via whole-register stores:

```asm
// tests/env/rvtest_failure_code.h:308
#ifdef RVTEST_VECTOR
    la x6, vecreg_scratch
    vs1r.v v0, (x6)
    addi x6, x6, VLEN_BYTES
    vs1r.v v1, (x6)
    ...   // v2..v31
```

On `qemu-rv32-max` (`-cpu max -icount shift=1`), this aborts QEMU instead of producing a signature mismatch:

```
ERROR: ../accel/tcg/cputlb.c:1464: probe_access_flags:
assertion failed: (-(addr | TARGET_PAGE_MASK) >= size)
Bail out!
```

The process dies with SIGABRT (`exit -6`) and never emits an `RVCP-SUMMARY` line, so the runner reports a hard failure with no diff the actual mismatch that triggered the handler is lost.

**Impact:** any vector self-check failure on this config turns into an opaque abort rather than an actionable mismatch report, which makes vector tests much harder to debug (and can mask a real DUT divergence as a "crash").

**Confirmed:**
- The only `vs1r.v` instructions in the failing binary are in this handler; there are no vector loads/stores in the test body itself.
- The same test passes on `act4` simply because that build never enters the failure handler.

**Likely cause (not fully confirmed):** the handler runs with whatever `vtype`/`vill` state the test left set, so QEMU may compute a whole-register access that straddles a page on the `probe_access_flags` path. Resetting vector state (e.g. a known-good `vsetvli`) before the dump, and/or page-aligning `vecreg_scratch`, would likely avoid it but I'd defer to you on the right fix since this is core framework code.

Happy to provide the full ELF, the `SmV-00.log`, or a minimal repro if useful.

jacassidy:
@georgiatai

dev-aditya-hub:
hie @everyone is there something i can do on this now i am happy to do the changes

davidharrishmc:
@fourth-bit can you give feedback?

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -495,11 +482,12 @@ def _gen_vstart_oob(test_data: TestData, temp_reg: int) -> list[str]:
     coverpoint = "cp_vstart_out_of_bounds"
     lines = [comment_banner(coverpoint, "csrrw vstart with rs1 = 2^16 (above max VLEN-1)")]
     lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
-    rs1_reg = test_data.int_regs.get_register()
+    rs1_reg, check_reg = test_data.int_regs.get_registers(2)
     lines.append(f"LI(x{rs1_reg}, 0x10000)  # 2^16")
     lines.append(test_data.add_testcase("vstart_oob", coverpoint, _CG))
     lines.append(f"CSRW(vstart, x{rs1_reg})")
-    test_data.int_regs.return_registers([rs1_reg])
+    lines.append(gen_csr_read_sigupd(check_reg, ("vstart", None), test_data))
```
fourth-bit, line 489: I'm not sure if sigupd is a good idea here because the value written to vstart is out of bounds and has an implementation defined result. This test should be structured to show that we read back a legal result instead of only sail's result

fourth-bit:
Per the spec (vill section), "vset_i_vl_i_ and whole register loads and stores do not depend upon vtype." As such, they execute even when vill is set. In the whole register load store section, it says: "The instructions operate with an effective vector length, evl=NFIELDS*VLEN/EEW, regardless of current settings in vtype and vl." This means that execution of the vs1r.v is completely independent of vl and vtype, so resetting the vector state should have no impact on these tests.

This also seems like a qemu bug because the simulator is crashing rather than handling what it thinks is an invalid state. I'd rather have this reported to qemu and fixed than work around it in the failure handler.

davidharrishmc:
Agreed that we should file an issue against QEMU instead of working around a bug in QEMU.  @dev-aditya-hub are you able to do that, with a gzip'd version of the ELF that triggers the bug and the exact command to replicate it?  Put a link to the issue in this issue so we can track it.

davidharrishmc:
Looks like somebody else filed a similar issue on QEMU.  https://lists.gnu.org/archive/html/qemu-devel/2026-06/msg02712.html @dev-aditya-hub can you tell if that is the same issue, or if there is another issue remaining to be filed?  This PR has been open for a while.  Can you make sure the necessary issues are filed on QEMU, put these tests in the ci waiver files for QEMU with a link to the open issue explaining what fix is necessary to lift the waiver, make sure CI passes, and indicate this PR is ready for review?

jordancarlin:
@dev-aditya-hub are you still working on this PR? It would be good to get this merged soon. If not, please let us know and someone else will take this over.

jordancarlin:
Closing this PR due to lack of activity.

@fourth-bit can you take a look at this PR and make sure that anything in it that we want to keep works its way into a new PR?

PR #1578, opened 2026-05-26
Updated scripts to run SsstrictS and SsstrictU tests

UmerShahidengr:


generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+#
+# t1 (x6) is deliberately never touched by either handler: with
+# rvtest_strap_routine defined, x6 is the Mtrampoline save-area pointer.
+# ---------------------------------------------------------------------------
+_FAST_SMODE_HANDLER_PREFIX: list[str] = [
+    "",
+    "// ── Fast M-mode handler (mtvec) — non-delegated traps ──────────────────",
+    "// Routes non-illegal traps (fetch faults, RVTEST_CODE_END ecall) to Mtrampoline.",
+    "// Illegal instructions are delegated to S-mode via medeleg and never reach here.",
+    "// ── Fast S-mode handler (stvec) — delegated illegal instructions ────────",
```
davidharrishmc: Should this move down to near strap_handler_fastillegalinstr?

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+    "// scause==2 → write scause/sepc/stval to signature, advance sepc, sret.",
+    "// Any other S-mode trap → Strampoline (real framework S-mode handler).",
+    "// Neither handler touches t1 (x6) — preserved as Mtrampoline save-area pointer.",
+    "\tj ssstrict_test_body",
+    "",
+    "\t.align 4",
```
davidharrishmc: This now should be UDB_MTVEC_BASE_ALIGNMENT_DIRECT.  Global change.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+    "trap_handler_fastillegalinstr:",
+    "\tj    Mtrampoline         # non-delegated traps go directly to framework handler",
+    "",
+    "\t.align 4",
```
davidharrishmc: This eventually should be UDB_STVEC_BASE_ALIGNMENT_DIRECT but that hasn't been defined yet.  Leave a comment to change?

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+# RVTEST_GOTO_LOWER_MODE Smode runs from M-mode here, writing a valid M-mode sp
+# into the framework save area so RVTEST_CODE_END's epilog can restore it.
+_SPLIT_FILE_SMODE_GPR_INIT: list[str] = (
+    [
+        "",
+        "# Switch to S-mode and re-initialize GPRs for this split file.",
+        "\tRVTEST_GOTO_LOWER_MODE Smode",
+        "\tcsrw    mie, x0",
+        "\t# x8 = permanent scratch base, 8-byte aligned for atomics",
+        "\tnop",
```
davidharrishmc: What are these two nops for?

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+        "\tnop",
+        "\tla x8, scratch",
+    ]
+    + [f"\tmv x{r}, x8" for r in range(7, 32) if r != 8]
```
davidharrishmc: Is there a reason for this now?  If we are trying to initialize with dummy values, how about DEADBEEF instead of scratch, so we don't accidentally have a load succeed?

Same comments apply to U mode.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -218,34 +325,38 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
...
+        #   to S-mode.
+        #   Each file gets _FAST_SMODE_HANDLER_PREFIX + _SPLIT_FILE_SMODE_GPR_INIT
+        #   (GOTO Smode from M-mode + scratch reload).
         #
-        # Do NOT append a RVTEST_GOTO_LOWER_MODE Mmode suffix here.
-        # On some configs (including RV32 sail-rv32-max) that macro is a
-        # preprocessor no-op that generates zero machine code.  Appending it
-        # does nothing useful, and the accompanying `csrw mie, x0` becomes
-        # unreachable dead code that clutters the generated files.
+        # TODO: SsstrictU: all code runs from U-mode; illegal instructions delegated (same as SsstrictS).
```
davidharrishmc, line 363: Is TODO done now?

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
     + ["", ""]
 )
 
+# ---------------------------------------------------------------------------
+# Fast S-mode illegal-instruction handler — used by SsstrictS and SsstrictU.
+#
+# Design:
+#   mtvec → M-mode fast handler (identical to SsstrictSm, handles non-delegated
```
davidharrishmc, line 131: This isn't identical to SsstrictM right now.

What if there is always a fast S and M-mode handler set up with stvec and mtvec?  Use the same M-mode fast handler here as in SsstrictSm.  Then the code will work whether or not illegal instructions are delegated (or if medeleg is read only zero), and will also work in a system with no S mode where they can't be delegated.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+    "// Illegal instructions are delegated to S-mode via medeleg and never reach here.",
+    "// ── Fast S-mode handler (stvec) — delegated illegal instructions ────────",
+    "// scause==2 → write scause/sepc/stval to signature, advance sepc, sret.",
+    "// Any other S-mode trap → Strampoline (real framework S-mode handler).",
+    "// Neither handler touches t1 (x6) — preserved as Mtrampoline save-area pointer.",
+    "\tj ssstrict_test_body",
+    "",
+    "\t.align 4",
+    "trap_handler_fastillegalinstr:",
+    "\tj    Mtrampoline         # non-delegated traps go directly to framework handler",
```
davidharrishmc: Suggest keep same M-mode handler here as for SsstrictSm.  Then it won't need to be special cased for S/U either.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+    "\taddi t0, t0, 4          # 32-bit instruction: advance sepc by 4",
+    "s_done:",
+    "\tcsrw sepc, t0",
+    "\tsret",
+    "",
+    "sothertrap:",
+    "\tj    Strampoline         # hand off non-illegal S-mode traps to framework",
+    "",
+    "ssstrict_test_body:",
+    "\tLA(t0, strap_handler_fastillegalinstr)",
```
davidharrishmc, line 234: Also write mtvec here regardless of delegation.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -218,34 +325,38 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
...
+        #   _SPLIT_FILE_GPR_INIT (scratch reload, no mode switch).
         #
-        # Invariant enforced by SsstrictS.py's batch boundary layout:
-        #   The splitter blank line is placed AFTER RVTEST_GOTO_LOWER_MODE Smode,
-        #   not before it.  This guarantees that every split file's first code
-        #   line (after the fast-handler prefix) is GOTO Smode executing from
-        #   M-mode, which writes a valid M-mode sp into the save area.  When
-        #   RVTEST_CODE_END later runs from S-mode, rtn_fm_mmode restores that
-        #   valid sp and the epilog succeeds.
+        # SsstrictS: all code runs from S-mode; illegal instructions delegated
```
davidharrishmc, line 358: Delegated to S mode if possible.  TODO: rvtest_boot_to_smode does not yet delegate exceptions.

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -218,34 +325,38 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
...
+        # TODO: SsstrictU: all code runs from U-mode; illegal instructions delegated (same as SsstrictS).
+        #   Each file gets _FAST_SMODE_HANDLER_PREFIX + _SPLIT_FILE_UMODE_GPR_INIT
+        #   (GOTO Umode from M-mode + scratch reload).
+        if testsuite == "SsstrictS":
+            handler_prefix = _FAST_SMODE_HANDLER_PREFIX
+            gpr_init = _SPLIT_FILE_SMODE_GPR_INIT
+        # TODO: Commented for now, will be uncommented once RVTEST_SETUP starts working completely
+        # Traps are not delegated to S-mode for SsstrictU, so the handler is the same as SsstrictSm's.
+        # The GPR init does not need to switch to U-mode, for now.
+        # elif testsuite == "SsstrictU":
```
davidharrishmc: Now I think the TODO can be resolved and this can match SsstrictS so long as both M and S fast handlers are installed.

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
davidharrishmc: These are part of S.py now and don't need to be duplicated here.  I don't think this is Ssstrict behavior, so I'd suggest dropping this unless there is a reason for it to go in Ssstrict that I'm overlooking.

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
```diff
@@ -258,7 +217,7 @@ def _generate_shadow_csr(test_data: TestData) -> list[str]:
 
 @add_priv_test_generator(
     "SsstrictS",
-    required_extensions=["Sm", "S", "Zicsr"],
+    required_extensions=["S", "Zicsr"],
     march_extensions=[
         # Zcf excluded — RV32-only.
         # Vector excluded — covered by SsstrictV.
```
davidharrishmc: What is the significance of these comments?  I think all of these things are tested, whether or not a DUT supports them, because that is the point of Ssstrict.  Could you add a comment on the V line of why it is necessary (to set vl for vector loads and stores).

generators/testgen/src/testgen/priv/extensions/SsstrictU.py
```diff
@@ -9,19 +9,22 @@
...
+_FAST_SMODE_HANDLER_PREFIX (which installs strap_handler_fastillegalinstr at
+stvec and delegates illegal instructions to S-mode via medeleg) plus
+_SPLIT_FILE_UMODE_GPR_INIT (which issues RVTEST_GOTO_LOWER_MODE Umode) to
+every split file.
 
 Structure
 ---------
-1. Switch to U-mode via RVTEST_GOTO_LOWER_MODE Umode.
+1. Per-split-file prefix switches to U-mode; the body stays in U-mode.
 2. CSR sweep from U-mode (user-privilege CSRs only: 0x000-0x0FF,
```
davidharrishmc: Why user mode only?

davidharrishmc: In SsstrictU at line 51, only non-custom U-mode CSRs are tested.  Don't we also want to test higher priv CSRs to show that they are inaccessible?

davidharrishmc: Alternatively, we have done that in S and U, so it is likely redundant here.  If we only test U here, we need to adjust the comments, and then remove M/Debug tests from SsstrictS.

davidharrishmc (changes requested):

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -134,6 +134,113 @@
...
+#            medeleg, using scause/sepc/stval/sret).
+#   medeleg bit 2 is set to delegate illegal-instruction traps to S-mode so the
+#   strap handler fires for every illegal encoding during the CSR sweep and the
+#   encoding sweeps — both of which run entirely from S-mode (or U-mode for
+#   SsstrictU, in which case the delegated trap also goes to S-mode).
+#
+# t1 (x6) is deliberately never touched by either handler: with
+# rvtest_strap_routine defined, x6 is the Mtrampoline save-area pointer.
+# ---------------------------------------------------------------------------
+_FAST_SMODE_HANDLER_PREFIX: list[str] = [
```
jordancarlin, line 143: At this point it feels like it might be easier to pull the fast trap handler out of the Python script, move it over to `rvtest_trap_handler.h` and have a `#define RVTEST_FAST_TRAP_HANDLER` at the beginning of these tests that indicates that it should be included for a particular test. Doesn't need to be part of this PR, but I think it is worth doing in the future.

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
```diff
@@ -44,6 +44,8 @@
     [0x180]  # satp  — skip: TLB flush / address-translation mode change
```
davidharrishmc: Not sure why we are testing most of the standard S-mode CSRs but not the H-mode CSRs here.  Similarly in SsstrictM, we are testing most of the M-mode CSRs.  We need to find a pattern and be consistent.

I suggest we strike H-mode std0, std1, std2, std3 from this list and add back individual exclusions if need be.

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
```diff
@@ -44,6 +44,8 @@
...
     + [0x104]  # sie   — skip: all-ones write enables S-mode interrupts; covered in shadow test
     + [0x144]  # sip   — skip: all-ones write asserts SSIP software interrupt; covered in shadow test
+    + [0x105]  # stvec   — skip: stvec not to update itself
```
davidharrishmc: strike all tests of M-mode because they are already covered by S_coverage.svh

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Good point but fast handler only supports illegal instruction exceptions, if we include everything into it (access faults, page faults, virtualization, masking etc.), then it will no remain fast enough. This trap handler is good enough for only Ssstrict, bringing it into trap handler can complicate other tests too.

jordancarlin: To clarify, I agree that it should only support illegal instruction exceptions. What I'm suggesting is keep the same fast, limited trap handler but move it into the `rvtest_trap_handler.h` file. It would only get included if the test defines `#define RVTEST_FAST_TRAP_HANDLER` (wrap the fast trap handler in an ifdef), so it doesn't matter if it can't handle other types of traps. This also makes the splitting process much simpler for long Ssstrict tests.

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: It is done in #1608 for Mmode fast trap handler, I will do it for Smode too after #1608 gets merged

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: It was to entertain old trap handler, it will be removed now

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: There is no reason of it now, will update this section to initialize gprs with 0xDEADBEEF

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done that already

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
UmerShahidengr: Good idea, will drop that one

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
UmerShahidengr: Yes, it is an old comment, will delete that

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Agreed. Lets do it in the next iteration of trap handler update

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Done

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Yes, done

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Not yet, will be done once SBI calls will be fully tested and Umode tests start running in Umode. It is blocked by tvtest_setup work

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: Agreed. Will uncomment it

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
UmerShahidengr: Done

generators/testgen/src/testgen/priv/extensions/SsstrictS.py
UmerShahidengr: Done

generators/testgen/src/testgen/priv/extensions/SsstrictU.py
UmerShahidengr: Done

davidharrishmc (approved):

PR #1579, opened 2026-05-26
Adding coverage files for all extensions of pointer masking

Ammarahwakeel:
Adds functional coverage for RISC-V pointer masking extensions (Ssnpm, Smmpm, SmnpmS, SmnpmU)  covering PMM=00/10/11 configurations with all supported instruction types.
Common instruction coverpoints are factored into a shared include file to eliminate duplication across the four covergroup files.


coverpoints/general/RISCV_coverage_pmm_instruction_coverpoints.svh
```diff
@@ -0,0 +1,190 @@
...
+//     - Zicfiss shadow-stack instructions    [ZICFISS_SUPPORTED]
+//     - RVV 1.0 vector loads/stores          [ZVL32B_SUPPORTED]
+//
+///////////////////////////////////////////
+
+    // ---- Base integer STORE instructions ----
+    sb_insn:  coverpoint ins.current.insn { wildcard bins sb  = {SB};  }
+    sh_insn:  coverpoint ins.current.insn { wildcard bins sh  = {SH};  }
+    sw_insn:  coverpoint ins.current.insn { wildcard bins sw  = {SW};  }
+    sd_insn:  coverpoint ins.current.insn { wildcard bins sd  = {SD};  }
```
davidharrishmc: We can't test illegal instructions, so all the RV64 stuff needs to be gated, including sd, leu, ld, amp*_d, amocas_q, c_*d

coverpoints/general/RISCV_coverage_pmm_instruction_coverpoints.svh
```diff
@@ -0,0 +1,190 @@
...
+//     - Zicboz cache-block zero              [ZICBOZ_SUPPORTED]
+//     - Zicbom cache-block maintenance       [ZICBOM_SUPPORTED]
+//     - Zicbop prefetch hints                [ZICBOP_SUPPORTED]
+//     - Zicfiss shadow-stack instructions    [ZICFISS_SUPPORTED]
+//     - RVV 1.0 vector loads/stores          [ZVL32B_SUPPORTED]
+//
+///////////////////////////////////////////
+
+    // ---- Base integer STORE instructions ----
+    sb_insn:  coverpoint ins.current.insn { wildcard bins sb  = {SB};  }
```
davidharrishmc: All of these should be bins in a pointer_insn coverpoint.  Then the main S*pm coverpoints just need to cross with pointer_insn rather than each of these individually.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,408 @@
+///////////////////////////////////////////
```
davidharrishmc, line 1: Most of this is repetitive with the other coverage files.  Please factor out the common stuff into RISCV_coverage_pmm_coverpoints and then just cross each of them with the privilege mode.

coverpoints/general/RISCV_coverage_pmm_instruction_coverpoints.svh
```diff
@@ -0,0 +1,190 @@
+///////////////////////////////////////////
+//
+// RISC-V Architectural Functional Coverage — Shared Coverpoints
+//
+// PMM instruction coverpoints — common to Ssnpm, Smmpm, Smnpm(S), Smnpm(U).
```
davidharrishmc: I think you can drop the word instruction and have all of the share coverpoints in here that will be used in each privilege mode.

davidharrishmc (changes requested):

coverpoints/priv/SmnpmS_coverage.svh
```diff
@@ -0,0 +1,448 @@
...
+            //bins stvec    = {CSR_STVEC}; //// warl field has complex write restrictions and is not easy to test
+            bins sscratch = {CSR_SSCRATCH};
+        }
+
+        // =======================================================================
+        // cp_pmlen_masking_write
+        // =======================================================================
+
+        // ---- Base integer stores ----
+        cp_pmlen_masking_write_sb: cross priv_mode_s, pmm_active, satp_mode, a_upper_bits, sb_insn;
```
davidharrishmc: Define in RISCV_coverage_pmm_coverpoints: 
pm_masking_write: cross pmm_active, satp_mode, a_upper_bits_pm_insn;

in this file
cp_pm_len_masking_write: cross priv_mode_s, pm_masking_write;

coverpoints/general/RISCV_coverage_pmm_instruction_coverpoints.svh
Ammarahwakeel: pointer masking (Ssnpm/Smmpm/SmnpmS/smnpmU) is itself an RV64-only feature. Each extension coverage file is already guarded with `ifdef XLEN64, so these instructions will only be compiled and exercised in a 64-bit configuration where they are legal.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,168 @@
...
+    // cp_hardware_csr_writes
+    // Fault crosses confirm lw/sw executed in M-mode at the illegal address.
+    // Trap crosses confirm mtval holds the correctly masked address after the fault.
+    cp_hardware_csr_writes_read_fault:   cross priv_mode_u,pm_read_fault;
+    cp_hardware_csr_writes_write_fault:  cross priv_mode_u, pm_write_fault;
+
+    // PMM=11, Bare
+    cp_hardware_csr_writes_read_pmm11_bare_trap:   cross priv_mode_m, mtval_upper_pmm11_bare, exception_occurred, lw_insn;
+    cp_hardware_csr_writes_write_pmm11_bare_trap:  cross priv_mode_m, mtval_upper_pmm11_bare, exception_occurred, sw_insn;
+    // PMM=10, Bare
```
Ammarahwakeel: Prof. @davidharrishmc, I was thinking of removing one of the lw/sw variants from cp_hardware_csr_writes_* trap and fault coverpoints across all files, since it feels like duplication. The core thing we want to verify is that when a fault occurs on an illegal address under pointer masking, hardware writes the correctly masked address into mtval/stval and that masking result depends only on PMM, satp.MODE, and the address sign bit, not on whether the faulting instruction was a load or a store. So keeping both lw and sw doubles the coverpoint count . so should i drop one of them ?

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,168 @@
...
+    // PMLEN bits; M-mode always runs bare so upper bits are always zeroed.
+    // NOTE: RV64A / Zacas / Zabha have no standalone load instructions; their
+    // read-backs use base lw/ld, already covered by the crosses above.
+    // The Zicbom and Zicbop read-backs are already implied by the existing base load crosses — no duplication needed.
+    cp_pmlen_masking_read : cross priv_mode_m, pmm_active, pm_load_insn ;
+
+    // cp_pmlen_disabled[]
+    cp_pmlen_disabled_lw: cross priv_mode_m, pmm_disabled, lw_insn;
+    cp_pmlen_disabled_sw: cross priv_mode_m, a_upper_bits, pmm_disabled, sw_insn;
+
```
Ammarahwakeel, line 70: For cp_pmlen_disabled_sw, since pointer masking is disabled (PMM=00), the upper bits of the A address will not be masked. Will this result in a fault because the upper bits are non-zero and the address is illegal, or will it just land on some other scratch memory location depending on what the upper bits resolve to? Unlike the JALR case where we explicitly said no masking is applied to implicit accesses so it faults, here the access is explicit but masking is simply off.  I just had this question in my mind, probably we will only know for sure once we actually run the test suite against this coverpoint ?

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -0,0 +1,271 @@
...
+        `endif
+        `ifdef SV48_SUPPORTED
+            bins sv48 = {4'b1001};
+        `endif
+        `ifdef SV57_SUPPORTED
+            bins sv57 = {4'b1010};
+        `endif
+    }
+    a_upper_bits: coverpoint (ins.current.rs1_val + ins.current.imm)[63:48] {
+        bins upper_0000 = {16'h0000};
```
davidharrishmc: Could you add comments about why each of these cases are interesting?

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -0,0 +1,271 @@
...
+    }
+    // ---- JALR instruction ----
+    jalr_insn: coverpoint ins.current.insn {
+        wildcard bins jalr = {JALR};
+    }
+    // for CBO.ZERO readback value
+    zero_loaded: coverpoint ins.current.rd_val[63:0] {
+        bins zero = {64'h0000_0000_0000_0000};
+    }
+    // For hardware csr writes, we raise an exception.
```
davidharrishmc: Comment unclear

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -0,0 +1,271 @@
...
+        bins upper_8000 = {16'h8000};
+        bins upper_FFFF = {16'hFFFF};
+        bins upper_FE00 = {16'hFE00};
+        bins upper_FF00 = {16'hFF00};
+        // upper_0000 excluded — would mean masking was incorrectly applied
+    }
+    // ---- Misaligned instruction ----
+    // Misaligned address (e.g. scratch+1); upper 7 bits = 0x01 or 0x00
+    misaligned_addr: coverpoint (ins.current.rs1_val + ins.current.imm)[1:0]
+        iff (ins.current.insn inside {LW, SW}) {
```
davidharrishmc: why is the insn part needed?  Isn't this later crossed with lw/sw anyway?

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -0,0 +1,222 @@
...
+        bins upper_FF00 = {16'hFF00};
+    }
+    // ---- Misaligned instruction ----
+    // Misaligned address (e.g. scratch+1); upper 7 bits = 0x01 or 0x00
+    misaligned_addr: coverpoint (ins.current.rs1_val + ins.current.imm)[1:0]
+        iff (ins.current.insn inside {LW, SW}) {
+        bins misaligned = {[2'b01:2'b11]};
+    }
+    misaligned_a_upper7: coverpoint (ins.current.rs1_val + ins.current.imm)[63:57] {
+        bins upper_zero = {7'b0000001}; // upper_0200 = {16'h0200};
```
davidharrishmc: The names don't agree with the numbers.  Wouldn't hurt to adjust test plan to have the all zeros case as well to show pointer masking doesn't impact misaligned when nothing is masked.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+//   effect on M-mode memory accesses.  M-mode always operates in bare (PA)
+//   mode — no virtual addressing, no sign-extension — so masked upper bits
+//   are always forced to zero.
+//
+//       PMM = 00 → PMLEN =  0  (masking disabled)
+//       PMM = 10 → PMLEN =  7  (upper  7 bits ignored, bits [63:57])
+//       PMM = 11 → PMLEN = 16  (upper 16 bits masked, bits [63:48])
+///////////////////////////////////////////
+
+`ifdef XLEN64
```
davidharrishmc: No ifdef needed.  Pointer masking only supported on RV64, so we know it must be RV64.  Global change.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+    `define COVER_SMMPM
+
+        covergroup Smmpm_cg with function sample(ins_t ins);
+        option.per_instance = 0;
+        `include "general/RISCV_coverage_standard_coverpoints.svh"
+        `include "general/RISCV_coverage_pmm_coverpoints.svh"
+
+        pmm_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mseccfg", "pmm") {
```
davidharrishmc: why is it called pmm_bit.  There are two bits.  I'd suggest pmm.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -0,0 +1,222 @@
...
+        wildcard bins jalr = {JALR};
+    }
+    `ifdef RVMODEL_ACCESS_FAULT_ADDRESS
+        // For hardware csr writes, we raise an exception.
+        illegal_addr: coverpoint (ins.current.rs1_val + ins.current.imm)[47:0] {
+            bins is_illegal_base = {`RVMODEL_ACCESS_FAULT_ADDRESS[47:0]};
+        }
+        pm_fault : cross pmm_bit, illegal_addr, a_upper_bits, sw_lw_insn ;
+    `endif
+    iff (pmm_bit == 2'b10) {
```
davidharrishmc: I don't think this is legal to have a coverpoint that dynamically exists or not depending on the state of a CSR.  I'd suggest amending the testplan to work for all pmm_bit values.  Otherwise, you'll need ignore_bins.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+        mpp_field: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "mpp") {
+            `ifdef U_SUPPORTED
+                bins mpp_u = {2'b00};   // effective privilege = U-mode
+            `endif
+            `ifdef S_SUPPORTED
+                bins mpp_s = {2'b01};   // effective privilege = S-mode
+            `endif
+        }
+        // ---- CSR write/read instructions ----
+        csr_rw_insn: coverpoint ins.current.insn {
```
davidharrishmc: This is used in several test files.  How about moving everything used in more than one file to RISCV_coverage_pmm_coverpoints.svh?  Set the weight to 0 (see other shared RISCV_* files) so it doesn't matter if it isn't used in every file.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+            wildcard bins csrrw = {CSRRW};
+            wildcard bins csrr = {CSRR};
+        }
+        // ---- Target CSR address (mepc, mtvec, mscratch) ----
+        csr_target: coverpoint ins.current.insn[31:20] { //excluding read-only csrs
+            bins mepc     = {CSR_MEPC};
+            //bins mtvec    = {CSR_MTVEC}; //// warl field has complex write restrictions and is not easy to test
```
davidharrishmc: Ok to skip, but then remove from test plan spreadsheet.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+            bins mscratch = {CSR_MSCRATCH};
+        }
+        // cp_pmlen_masking
+        cp_pmlen_masking : cross priv_mode_m, pmm_bit, a_upper_bits, pm_insn ;
+        // cp_pmlen_misaligned_word
+        // One misaligned store and one misaligned load, PMLEN=7 only, upper 7 bits = 7'b1000000 or 7'b0000001
+        cp_pm_misaligned_word: cross priv_mode_m, pm_misalign;
+        `ifdef S_SUPPORTED  //As MXR is read-only zero if S mode is not supported
+            // cp_pmm_mxr
+            cp_pmm_mxr: cross priv_mode_m, pmm_bit,   mxr_bit, a_upper_bits, sw_lw_insn;
```
davidharrishmc: extra space.

coverpoints/priv/SmnpmS_coverage.svh
```diff
@@ -0,0 +1,93 @@
...
+//   Covers menvcfg.PMM configuration (PMM=00/10/11 → PMLEN=0/7/16) in M-mode
+//   and its effect on S-mode memory accesses across all supported  modes
+//   (Bare, Sv39, Sv48, Sv57).  When PMM is active, hardware strips the upper
+//   PMLEN bits of the effective address before the access reaches PMP/MMU,
+//   so a tagged address A and its masked form  alias to the same
+//   physical location.
+//
+///////////////////////////////////////////
+
+`ifdef XLEN64
```
davidharrishmc: DOn't need XLEN64 or S_SUPPORTED.  Tests should have a REQUIRED_EXTENSIONS with S and SMNPM_SUPPORTED so they only run in these conditions.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+            wildcard bins csrr = {CSRR};
+        }
+        // ---- Target CSR address (mepc, mtvec, mscratch) ----
+        csr_target: coverpoint ins.current.insn[31:20] { //excluding read-only csrs
+            bins mepc     = {CSR_MEPC};
+            //bins mtvec    = {CSR_MTVEC}; //// warl field has complex write restrictions and is not easy to test
+            bins mscratch = {CSR_MSCRATCH};
+        }
+        // cp_pmlen_masking
+        cp_pmlen_masking : cross priv_mode_m, pmm_bit, a_upper_bits, pm_insn ;
```
davidharrishmc: no space before ;

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+            `ifdef U_SUPPORTED
+                bins mpp_u = {2'b00};   // effective privilege = U-mode
+            `endif
+            `ifdef S_SUPPORTED
+                bins mpp_s = {2'b01};   // effective privilege = S-mode
+            `endif
+        }
+        // ---- CSR write/read instructions ----
+        csr_rw_insn: coverpoint ins.current.insn {
+            wildcard bins csrrw = {CSRRW};
```
davidharrishmc: CSRW is sufficient (and no csrrw needed).  The test should contain a sigupd that reads back the written value and confirms it is correct.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+        `ifdef S_SUPPORTED  //As MXR is read-only zero if S mode is not supported
+            // cp_pmm_mxr
+            cp_pmm_mxr: cross priv_mode_m, pmm_bit,   mxr_bit, a_upper_bits, sw_lw_insn;
+        `endif
+        // cp_pmm_jalr
+        cp_pmm_jalr: cross priv_mode_m, pmm_bit, a_upper_bits, jalr_insn ;
+        // cp_hardware_csr_writes_fault
+        `ifdef RVMODEL_ACCESS_FAULT_ADDRESS
+            // Fault crosses confirm lw/sw executed in M-mode at the illegal address.
+            cp_hardware_csr_writes_fault:   cross priv_mode_m,pm_fault;
```
davidharrishmc: extra space after :, missing space after ,.  Please review spacing everywhere in these files.

coverpoints/priv/SmnpmS_coverage.svh
```diff
@@ -0,0 +1,93 @@
...
+            // ---- MXR bit from sstatus ----
+            mxr_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "mxr") {
+                bins mxr_1 = {1'b1};   // MXR=1: execute-only pages readable
+                bins mxr_0 = {1'b0};   // MXR=0: normal permission checks
+            }
+            sxl_rv32: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "sxl") {
+                bins sxl_01 = {2'b01};
+            }
+            // menvcfg.PMM must have been cleared to 00 by hardware after SXL=01.
+            pmm_after_clear: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "menvcfg", "pmm") {
```
davidharrishmc: This is an output, not an input.  Drop the coverpoint, and make sure your test sigupd's menvcfg after changing SXL.

coverpoints/priv/SmnpmS_coverage.svh
```diff
@@ -0,0 +1,93 @@
...
+                bins pmm_cleared = {2'b00};
+            }
+            // medeleg delegation witness — precondition, not a cross dimension
+            medeleg_delegated: coverpoint ins.current.csr[CSR_MEDELEG][15:5] {
+                // [10]=medeleg[15] [8]=medeleg[13] [2]=medeleg[7] [0]=medeleg[5]
```
davidharrishmc: Comment not useful.  If you want to keep it, replace with "delegate load and store access and page faults."

coverpoints/priv/SmnpmS_coverage.svh
```diff
@@ -0,0 +1,93 @@
...
+            // cp_pmlen_misaligned_word
+            // One misaligned store and one misaligned load, PMLEN=7 only, upper 7 bits = 7'b1000000 or 7'b0000001
+            cp_pm_misaligned_word: cross priv_mode_s, pm_misalign;
+            // cp_pmm_mxr
+            cp_pmm_mxr: cross priv_mode_s, pmm_bit, mxr_bit, satp_mode, a_upper_bits, sw_lw_insn;
+            // cp_pmm_addr_mode_jalr — not guarded by S_SUPPORTED; implicit fetch is
+            // never pointer-masked regardless of PMM or MXR availability.
+            cp_pmm_jalr: cross priv_mode_s, pmm_bit, a_upper_bits, mxr_bit, satp_mode, jalr_insn;
+            // cp_pmm_sxl_clear
+            iff (pmm_bit != 2'b00) {
```
davidharrishmc: not legal.  Coverpoints can't wink in and out of existence based on the value of a register.

coverpoints/priv/SmnpmU_coverage.svh
```diff
@@ -0,0 +1,67 @@
...
+            pmm_after_clear: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "menvcfg", "pmm") {
+                bins pmm_cleared = {2'b00};
+            }
+            // cp_pmlen_masking
+            cp_pmlen_masking : cross priv_mode_u, pmm_bit, a_upper_bits, pm_insn ;
+            // cp_pmlen_misaligned_word
+            // One misaligned store and one misaligned load, PMLEN=7 only, upper 7 bits = 7'b1000000 or 7'b0000001
+            cp_pm_misaligned_word: cross priv_mode_u, pm_misalign;
+            // cp_pmm_uxl_clear
+            iff (pmm_bit != 2'b00) {
```
davidharrishmc: illegal

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,76 @@
...
+        cp_pmlen_masking : cross priv_mode_u, pmm_bit, satp_mode, a_upper_bits, pm_insn;
+        // cp_pmlen_misaligned_word
+        // One misaligned store and one misaligned load, PMLEN=7 only, upper 7 bits = 7'b1000000 or 7'b0000001
+        cp_pmlen_misaligned_word: cross priv_mode_u, pm_misalign;
+        // cp_pmm_mxr
+        cp_pmm_mxr: cross priv_mode_u, pmm_bit,   mxr_bit, satp_mode, a_upper_bits, sw_lw_insn;
+        // cp_pmm_jalr
+        cp_pmm_jalr:      cross priv_mode_u, pmm_bit, a_upper_bits, mxr_bit, satp_mode, jalr_insn;
+        // cp_pmm_uxl_clear
+        iff (pmm_bit != 2'b00) {
```
davidharrishmc: illegal syntax

coverpoints/priv/SmnpmU_coverage.svh
```diff
@@ -0,0 +1,67 @@
...
+            pmm_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "menvcfg", "pmm") {
+                bins pmm_00_disabled = {2'b00};  // PMLEN = 0, no masking
+                bins pmm_10_pmlen7  = {2'b10};   // PMLEN =  7, upper  7 bits masked
+                bins pmm_11_pmlen16 = {2'b11};   // PMLEN = 16, upper 16 bits masked
+            }
+            uxl_rv32: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "uxl") {
+                bins uxl_01 = {2'b01};
+            }
+            // menvcfg.PMM must have been cleared to 00 by hardware after UXL=01.
+            pmm_after_clear: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "menvcfg", "pmm") {
```
davidharrishmc: not needed

coverpoints/priv/SmnpmU_coverage.svh
```diff
@@ -0,0 +1,67 @@
...
+            // cp_pmm_uxl_clear
+            iff (pmm_bit != 2'b00) {
+                cp_pmm_uxl_clear: cross pmm_bit, uxl_rv32, pmm_after_clear;
+            }
+            // cp_hardware_csr_writes
+            `ifdef RVMODEL_ACCESS_FAULT_ADDRESS
+                // Fault crosses confirm lw/sw executed in U-mode at the illegal address.
+                cp_hardware_csr_writes_fault:   cross priv_mode_u, pm_fault;
+            `endif
+
```
davidharrishmc, line 36: does cp_pmm_jalr still apply?

coverpoints/priv/SmnpmU_coverage.svh
```diff
@@ -0,0 +1,67 @@
...
+//
+//
+//       PMM = 00 → PMLEN =  0  (masking disabled)
+//       PMM = 10 → PMLEN =  7  (upper  7 bits ignored, bits [63:57])
+//       PMM = 11 → PMLEN = 16  (upper 16 bits masked, bits [63:48])
+//
+///////////////////////////////////////////
+
+`ifdef XLEN64
+    `ifndef S_SUPPORTED
```
davidharrishmc: drop.  global change.

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,76 @@
...
+            bins mxr_0 = {1'b0};   // MXR=0: normal permission checks
+        }
+        uxl_rv32: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "sstatus", "uxl") {
+            bins uxl_01 = {2'b01};
+        }
+        // senvcfg.PMM must have been cleared to 00 by hardware after UXL=01.
+        pmm_after_clear: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "senvcfg", "pmm") {
+            bins pmm_cleared = {2'b00};
+        }
+        // medeleg delegation witness — precondition, not a cross dimension
```
davidharrishmc: The medeleg coverpoints are not used.  Looks like they could be dropped globally.

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,76 @@
...
+        //Main Crosses
+        // cp_pmlen_masking_write
+        cp_pmlen_masking : cross priv_mode_u, pmm_bit, satp_mode, a_upper_bits, pm_insn;
+        // cp_pmlen_misaligned_word
+        // One misaligned store and one misaligned load, PMLEN=7 only, upper 7 bits = 7'b1000000 or 7'b0000001
+        cp_pmlen_misaligned_word: cross priv_mode_u, pm_misalign;
+        // cp_pmm_mxr
+        cp_pmm_mxr: cross priv_mode_u, pmm_bit,   mxr_bit, satp_mode, a_upper_bits, sw_lw_insn;
+        // cp_pmm_jalr
+        cp_pmm_jalr:      cross priv_mode_u, pmm_bit, a_upper_bits, mxr_bit, satp_mode, jalr_insn;
```
davidharrishmc: pmm_xmr and _jalr have almost the same inputs.  Put them in the same order to emphasize similarities.  Global change.

davidharrishmc (changes requested):

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
Ammarahwakeel: yes, just added as extra guard here.  Will remove it

coverpoints/priv/SmnpmS_coverage.svh
Ammarahwakeel: hmm , sure. At first when i was removing the outputs, I decided to remove it too. But thought it would not be harmful. But now when I am thinking of it again. I think it was just creating a useless bin as at one instant time in the cross I would not have been able to get value of PMM before and after instruction.
That said, I will remove it too.

coverpoints/priv/SmnpmU_coverage.svh
Ammarahwakeel: yes, Infact I did add it in the Smmpm, even if there MXR bit was read-only zero ( if S mode was not supported ) . Perhaps forgot to add it here. Will update this file as well as it's testplan to add cp_pmm_jalr unconditionally.

coverpoints/priv/Ssnpm_coverage.svh
Ammarahwakeel: Yup, I agree. Previously I thought it was nice to have just it as a separate coverpoint to actually check if the traps were being delegated correctly. But yes as  we will see it eventually through the sigupd / trace file in the tests. I will drop it .

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -0,0 +1,76 @@
...
+        covergroup Ssnpm_cg with function sample(ins_t ins);
+        option.per_instance = 0;
+        `include "general/RISCV_coverage_standard_coverpoints.svh"
+        `include "general/RISCV_coverage_pmm_coverpoints.svh"
+
+        pmm_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "senvcfg", "pmm") {
+            bins pmm_00_disabled = {2'b00};  // PMLEN = 0, no masking
+            bins pmm_10_pmlen7  = {2'b10};   // PMLEN =  7, upper  7 bits masked
+            bins pmm_11_pmlen16 = {2'b11};   // PMLEN = 16, upper 16 bits masked
+        }
```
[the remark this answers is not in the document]
Ammarahwakeel: prof. @davidharrishmc , Should I update the RISCV_coverage_csr.sv file to include pmm as a field to the respective csrs for each s*pm extension ?

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -0,0 +1,226 @@
...
+    misaligned_addr: coverpoint (ins.current.rs1_val + ins.current.imm)[1:0] {
+        type_option.weight = 0;
+        bins misaligned = {[2'b01:2'b11]};
+    }
+    // ---- JALR instruction ----
+    jalr_insn: coverpoint ins.current.insn {
+        type_option.weight = 0;
+        wildcard bins jalr = {JALR};
+    }
+    // ---- CSR write/read instructions ----
```
davidharrishmc: Comment still says read/write, csr_rw, and name is still csrrw.  change all to write.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,92 @@
...
+///////////////////////////////////////////
+
+
+`define COVER_SMMPM
+
+    covergroup Smmpm_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+    pmm: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mseccfg", "pmm") {
```
davidharrishmc, line 30: See open copilot feedback that 

framework/src/act/fcov/coverage/RISCV_coverage_csr.svh needs to define mseccfg pmm entry around line 451.

get_csr_val(..., "mseccfg", "pmm") is not implemented in the CSR field decoder: RISCV_coverage_csr.svh handles mseccfg but only extracts mml/mmwp/rlb/useed/sseed (no pmm). As written, this coverpoint will likely not reflect the 2-bit PMM field and will undercount/never hit the intended bins if other MSECCFG bits are set. This needs CSR decoding support for mseccfg.pmm (or a direct slice from ins.current.csr[CSR_MSECCFG] once bit positions are confirmed).

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -0,0 +1,226 @@
...
+    csr_rw_insn: coverpoint ins.current.insn {
+        type_option.weight = 0;
+        wildcard bins csrrw = {CSRW};
+    }
+    `ifdef RVMODEL_ACCESS_FAULT_ADDRESS
+        // For hardware csr writes, we raise an exception.
```
davidharrishmc: Replace with
// Exception should write xtval with masked version of pointer.

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,92 @@
...
+        `include "general/RISCV_coverage_pmm_coverpoints.svh"
+
+    `ifdef S_SUPPORTED  //As MXR is read-only zero if S mode is not supported
+    // ---- MXR bit from mstatus ----
+        mxr_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mstatus", "mxr") {
+            bins mxr_1 = {1'b1};   // MXR=1: execute-only pages readable
+            bins mxr_0 = {1'b0};   // MXR=0: normal permission checks
+        }
+    `endif
+    // ---- MPRV bit from mstatus ----
```
davidharrishmc: When you have lines of comments associated with a coverpoint, add a blank line between unrelated coverpoints to make it easier to see where one ends and the next begins.  Global change.  If you have
many coverpoints with just // comments at the end of the line, it's fine to have them without blank lines
because there is nothing to separate.

davidharrishmc (changes requested):

coverpoints/priv/Smmpm_coverage.svh
```diff
@@ -0,0 +1,95 @@
...
+        covergroup Smmpm_cg with function sample(ins_t ins);
+        option.per_instance = 0;
+        `include "general/RISCV_coverage_standard_coverpoints.svh"
+        `include "general/RISCV_coverage_pmm_coverpoints.svh"
+
+        pmm_bit: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "mseccfg", "pmm") {
+            bins pmm_00_disabled = {2'b00};  // PMLEN = 0, no masking
+            bins pmm_10_pmlen7  = {2'b10};   // PMLEN =  7, upper  7 bits masked
+            bins pmm_11_pmlen16 = {2'b11};   // PMLEN = 16, upper 16 bits masked
+        }
```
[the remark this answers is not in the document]
davidharrishmc: Also menvcfg and senvcfg

framework/src/act/fcov/coverage/RISCV_coverage_csr.svh
```diff
@@ -385,6 +387,14 @@ function `XLEN_BITS get_csr_val_addr(int hart, int issue, int prev, int addr, st
       default: val = 0; // Todo: error
     endcase
   end
+  if (name == "menvcfg") begin
+    case(field)
+`ifdef UDB_MXLEN_64
```
davidharrishmc, line 398: Might as well add all the fields of each of these registers while you are doing any of them.

coverpoints/priv/SmnpmS_coverage.svh
```diff
@@ -0,0 +1,78 @@
...
+    }
+    // ---- Target CSR address (sepc, stvec, sscratch) ----
+    csr_target: coverpoint ins.current.insn[31:20] { //excluding read-only csrs
+        bins sepc     = {CSR_SEPC};
+        //bins stvec    = {CSR_STVEC}; //// warl field has complex write restrictions and is not easy to test
+        bins sscratch = {CSR_SSCRATCH};
+    }
+
+    // Main Crosses
+    // cp_pmlen_masking
```
davidharrishmc: Remove all the comments that are just a repetition of the name of the coverpoint; they don't add info.

framework/src/act/fcov/coverage/RISCV_coverage_csr.svh
Ammarahwakeel: okay, Let me add those too.

coverpoints/priv/SmnpmS_coverage.svh
Ammarahwakeel: hmm. Agree.

framework/src/act/fcov/coverage/RISCV_coverage_csr.svh
```diff
@@ -449,6 +470,9 @@ function `XLEN_BITS get_csr_val_addr(int hart, int issue, int prev, int addr, st
       "rlb" : val = (val >> 2) & 'h1;
       "useed" : val = (val >> 8) & 'h1;
       "sseed" : val = (val >> 9) & 'h1;
+`ifdef UDB_MXLEN_64
```
davidharrishmc, line 474: MLPE (bit 10)

davidharrishmc (approved):

PR #1581, opened 2026-05-26
InterruptsU running on Spike

Teemooooooooo:
increased interrupt latency and changed LI to LA to ensure the code length is consistent


davidharrishmc (approved):

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -104,14 +104,19 @@
...
 
 #define RVMODEL_TIMER_INT_SOON_DELAY 100
 
 // Spike ticks the CLINT timer every 100 instructions (default --insns-per-tick).
 // The default RVTEST_IDLE_FOR_TIMER_INTERRUPT spins RVMODEL_TIMER_INT_SOON_DELAY iterations
 // (200 instructions), which only advances mtime by ~2 ticks — far less than the 100-tick
 // stimecmp offset. Override with a 200x multiplier so the spin outlasts the timer delay.
+// in-case that the utils.h file gets read first and already defines RVTEST_IDLE_FOR_TIMER_INTERRUPT, we want to overwrtie it
+#ifdef RVTEST_IDLE_FOR_TIMER_INTERRUPT
+  #undef RVTEST_IDLE_FOR_TIMER_INTERRUPT
```
jordancarlin: We should never be messing with RVTEST macros in the rvmodel files. This implies we either need a new RVMODEL macro (avoid this if we can) or we need to rethink the existing RVTEST macro.

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -104,14 +104,15 @@
...
+#define RVMODEL_INTERRUPT_LATENCY 4096
 
 #define RVMODEL_TIMER_INT_SOON_DELAY 100
 
 // Spike ticks the CLINT timer every 100 instructions (default --insns-per-tick).
 // The default RVTEST_IDLE_FOR_TIMER_INTERRUPT spins RVMODEL_TIMER_INT_SOON_DELAY iterations
 // (200 instructions), which only advances mtime by ~2 ticks — far less than the 100-tick
 // stimecmp offset. Override with a 200x multiplier so the spin outlasts the timer delay.
+
 #define RVTEST_IDLE_FOR_TIMER_INTERRUPT(_R1) \
```
jordancarlin: It’s not just the undef that was the issue. We should not be defining RVTEST macros in rvmodel_macros.h. It doesn’t need to be addressed in this PR, but this will need to be resolved eventually.

Teemooooooooo: Got it - I will try to fix it in another PR. I think there are other models that also has RVTEST defined in their rvmodel_macros.h so it might take a bit to go through every model

tests/env/check_defines.h
```diff
@@ -62,6 +62,11 @@
 #ifndef RVMODEL_TIMER_INT_SOON_DELAY
   #error "RVMODEL_TIMER_INT_SOON_DELAY not defined. Make sure to define it in rvmodel_macros.h."
 #endif
+// For Spike/QEMU, which have the CLINT ticks slower than the core clock, define RVMODEL_TIMER_INT_SOON_DELAY_CYCLES with a 200x multiplier;
```
davidharrishmc: This comment is not right.  This file shouldn't refer to specific simulators.

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -176,10 +176,7 @@
 // QEMU virt CLINT runs at 10 MHz; with -icount shift=1 (2 ns/insn) that is ~50 insns/tick.
 // The default spin (DELAY iterations = 200 insns) only advances mtime ~4 ticks vs the
 // 100-tick stimecmp offset. Override with a 200x multiplier to reliably outlast the delay.
-#define RVTEST_IDLE_FOR_TIMER_INTERRUPT(_R1) \
-    LI(_R1, RVMODEL_TIMER_INT_SOON_DELAY * 200); \
-    99: addi _R1, _R1, -1; \
-        bnez _R1, 99b;
+#define RVMODEL_TIMER_INT_SOON_DELAY_CYCLES (RVMODEL_TIMER_INT_SOON_DELAY * 200)
```
davidharrishmc: This works, but would it be cleaner to define RVMODEL_MAX_CYCLES_PER_TIMER_TICK here, and then in utils.h #define RVTEST_TIMER_INT_SOON_DELAY_CYCLES (RVMODEL_TIMER_INT_SOON_DELAY * RVMODEL_MAX_CYCLES_PER_TIMER_TICK) and use it in RVTEST_IDLE_FOR_TIMER_INTERRUPT?  

Define RVMODEL_MAX_CYCLES_PER_TIMER_TICK in abstraction.adoc as the highest possible ratio of the clock frequency to the timer frequency in any mode of operation.  Give it a default value of 1 in check_defines.h.

davidharrishmc (changes requested):

allenjbaum:
That matches what I described in my (unpublished) update to the test format
spec, I think.

davidharrishmc (approved):

PR #1584, opened 2026-05-26
Imperas and Whisper Improvements

fourth-bit:
A few fixes to coverage that broke, and many fixes to the imperas configs. Currently, imperas-rv32gcv will trap on unaligned accesses. There are a few problems with testgen which means that it will not pass every test yet. There are a few changes that bring whisper to pass every test (excluding vfredusum where it differs by 1 ulp). These mostly include setting tu and mu to get deterministic behavior in load/store tests, and a fix to the scalar dst flag so that it works with vfmv. 

This improves Vf* and Vls* on whisper and imperas.


config/imperas/imperas-rv64-max/imperas.ic
```diff
@@ -51,21 +51,24 @@
...
+--override cpu*/Zve32f=F #(Boolean) (default=F) (default) Specify that Zve32f is implemented
+--override cpu*/Zve32x=F #(Boolean) (default=F) (default) Specify that Zve32x is implemented
+--override cpu*/Zve64d=F #(Boolean) (default=F) (default) Specify that Zve64d is implemented
+--override cpu*/Zve64f=F #(Boolean) (default=F) (default) Specify that Zve64f is implemented
+--override cpu*/Zve64x=F #(Boolean) (default=F) (default) Specify that Zve64x is implemented
 --override cpu*/Zvfbfmin=T #(Boolean) (default=T) (default) Specify that Zvfbfmin is implemented
 --override cpu*/Zvfbfwma=T #(Boolean) (default=T) (default) Specify that Zvfbfwma is implemented
 --override cpu*/Zvfh=T #(Boolean) (default=T) (default) Specify that Zvfh is implemented
 --override cpu*/Zvfhmin=T #(Boolean) (default=T) (default) Specify that Zvfhfmin is implemented
+--override cpu*/Zvlsseg=T #(Boolean) (default=F) (default) Specify that Zvlsseg is implemented
```
davidharrishmc, line 65: Zvlsseg is not a ratified extension.  What does it mean to be true?  Same query about Zvqmac.
Please add a comment explaining its significance if it is necessary, or turn it off if not.

davidharrishmc (changes requested):

config/imperas/imperas-rv64-max/imperas.ic
fourth-bit: Imperas will trap on the vlsseg instructions from the V extension without the the Zvlsseg flag set. The Zvqmac flag was unnecessary.

davidharrishmc (approved):

PR #1593, opened 2026-05-27
Merge vector testgen

jacassidy:


jacassidy:
@jordancarlin this seems to connect the two but I feel like wasnt your intention when you asked me to merge vector testgen, I told copilot to keep the files seperate as I assumed you didnt want a huge dump of vector into normal testgen. Please let me know how I should update

davidharrishmc:
What is the purpose of this PR?  Does it fix specific issues in vector?

jacassidy:
@davidharrishmc Jordan asked me to try to get AI to merge both test generators, so vector testgen is no longer seperate. I've given it the explicit constraint that there must be an empty diff between the current tests and the updated generator so none of the generated files change

jordancarlin:
> @jordancarlin this seems to connect the two but I feel like wasnt your intention when you asked me to merge vector testgen, I told copilot to keep the files seperate as I assumed you didnt want a huge dump of vector into normal testgen. Please let me know how I should update

This approach seems brittle and doesn’t do anything to solve the maintenance burden of having two different test generators. What I think needs to happen eventually is to actually merge vector testgen into the standard testgen package and use the existing functions/etc. The existing generator should be extended to support vector instead of calling out to a separate generator.

davidharrishmc:
It's not reasonable for Kaden to do a big refactor over the summer.  The rest of the HMC team is fully loaded trying to get to RVA23 on schedule.  Nobody else but Jordan has the knowledge to do this in a timely way. Should we close this PR, and Jordan can address when/if it bubbles to the top of his priorities?

jacassidy:
Jordan had asked me to get my copilot to try to do the refactor given all the vector specific knowledge ive taught it as well as the fact the copilot price structure is about to change. I dont think we need to merge this soon but I think it would be useful for this pr to serve as a branch of changes that could be worked from in the future

jordancarlin:
If pieces can serve as a reference, that’s great. If not, we will start the refactor from scratch down the road. The most useful thing for future work might be all of your vector-specific Claude instruction files.

jacassidy:
That can be found here, you may find a older version of the repo (before the shard refactor) more useful

https://github.com/jacassidy/riscv-arch-test-claude

jordancarlin:
Closing now that #1755 is open.

PR #1596, opened 2026-05-28
Zvkg Test Generation and Covergroup Generation

fourth-bit:
This patch adds test generation for basic egs = 4 instructions from the Zvkg extension. These tests generate and run on all simulators, but do not collect any coverage as of now. Coverage seems like it will turn out to be a bigger problem to solve as many of the utilities assume that vl = 1 to collect coverage. Each test in this suite requires vl = 4, with edge values that can potentially span multiple vector registers which something like vs_edges_check is not built for, so it will take some effort to bring up coverage for the edges coverpoints. As of now, this means that coverage is at 0% because it throws an error. The covergroup template for cr_vl_lmul_edg4_sew32 is also a work in progress.


fourth-bit:
Coverage is now fully working, so this is ready for a review.

coverpoints/unpriv/Zvkg_coverage.svh
```diff
@@ -0,0 +1,804 @@
...
+    cp_asm_count : coverpoint ins.ins_str == "vghsh.vv"  iff (ins.trap == 0 )  {
+        // Number of times instruction is executed
+        bins count[]  = {1};
+    }
+
+    //////////////////////////////////////////////////////////////////////////////////
+    // cp_vd_edges_egs4
+    //////////////////////////////////////////////////////////////////////////////////
+
+    cp_vd_edges_egs4 : coverpoint vs_edges_check_sew32_egs4(ins.hart, ins.issue, ins.get_vr_val_lmul4(ins.current.vd))
```
davidharrishmc, line 189: Seems values should be EGS*SEW in size.  How is this being handled?

generators/testgen/scripts/vector-testgen-unpriv.py
```diff
@@ -387,14 +423,61 @@ def make_fs1_edges_v(instruction, sew):
...
+  assert(all(len(val) == vedges_length for val in vs1edges_lmul_dict.values()))
+
+  for i1 in range(vedges_length):
+    for i2 in range(vedges_length):
+      instruction_data = {}
+      for vlen in [32, 64, 128, 256]:
+        lmul = math.ceil(sew * egs / vlen)
+
+        v1 = vs1edges_lmul_dict[lmul][i1]
+        v2 = vs1edges_lmul_dict[lmul][i2]
```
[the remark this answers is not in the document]
fourth-bit: This function is dead code, and has now been removed

coverpoints/unpriv/Zvkg_coverage.svh
fourth-bit: It gets the entire group of registers (lmul=4, so we know that EGS\*SEW fits), and then checks the EGS\*SEW wide edge values in the vs_edges_check_sew32_egs4 function. This function works for now, but will change to be more parameterized with the egs8 Zvksh instructions and the sew 64 Zvknhb instructions

framework/src/act/fcov/coverage/RISCV_coverage_vector.svh
```diff
@@ -232,6 +232,37 @@ function edge_vs_values_t vs_edges_check_eew_64(`VLEN_BITS val);
...
+    $error("ERROR: SystemVerilog Functional Coverage: SEW 32 Support Required for EGS 4 Edge Check");
+    $fatal(1);
+  `endif
+
+  if (sew != 32) begin
+    $error("ERROR: SystemVerilog Functional Coverage: SEW 32 EGS4 edge check requires SEW=32, but was given %d", sew);
+    $fatal(1);
+  end
+
+  casez (val)
```
davidharrishmc, line 250: Worth some thought about what edge values would be most interesting for vector crypto.  These are the same in most of the words, so they don't stress the units too much.

generators/coverage/src/covergroupgen/templates/vector/cp_vs2_edges_egs4.sv
```diff
@@ -0,0 +1,22 @@
+    //////////////////////////////////////////////////////////////////////////////////
+    // cp_vs2_edges_egs4
+    //////////////////////////////////////////////////////////////////////////////////
+
+    cp_vs2_edges_egs4 : coverpoint vs_edges_check_sew32_egs4(ins.hart, ins.issue, ins.get_vr_val_lmul4(ins.current.vs2))
+        iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vlmul") == 2)  {
+        // Edge values of vs2 (EGS=4), assuming vl = 4
+        bins zero       = {vs_zero      };   //  = {(`SEW){1'b0}},
```
davidharrishmc: Is this really SEW?  Or EGS*SEW?  In comments.

davidharrishmc (changes requested):

generators/coverage/src/covergroupgen/templates/vector/cp_vd_edges_egs4.sv
```diff
@@ -0,0 +1,22 @@
+    //////////////////////////////////////////////////////////////////////////////////
+    // cp_vd_edges_egs4
+    //////////////////////////////////////////////////////////////////////////////////
+
+    cp_vd_edges_egs4 : coverpoint vs_edges_check_sew32_egs4(ins.hart, ins.issue, ins.get_vr_val_lmul4(ins.current.vd))
+        iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vlmul") == 2)  {
+        // Edge values of vd (EGS=4), assuming vl = 1
+        bins zero       = {vs_zero      };   //  = {(`SEW){1'b0}},
```
davidharrishmc: Maybe 0s, 1s, walkover, walk odd is sufficient for egs4 cases where the other bits aren't interesting.  Then just make sure suites with sbox or other special behaviors fully exercise those relevant inputs.

fourth-bit: Changed to only be these values. Is it worth adding a cp_custom target with values from examples, e.g. for vghsh NIST provides these values https://csrc.nist.rip/groups/ST/toolkit/BCM/documents/proposedmodes/gcm/gcm-spec.pdf

davidharrishmc (approved):

PR #1597, opened 2026-05-29
Check for all possible job statuses

nazar-pc:
This stuff is under-documented and apparently there is also `skipped` value, so checking non-success is more reliable long-term

### DIFF
--- a/.github/workflows/regress.yml
+++ b/.github/workflows/regress.yml
@@ -286,9 +286,10 @@ jobs:
       ]
     if: "${{ always() }}"
     steps:
-      - name: exit failure
-        if: "${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') }}"
-        run: exit 1
-      - name: exit success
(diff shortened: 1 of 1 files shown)


jordancarlin:
I think we might eventually want to allow skipped so that we can conditionally run some checks depending on which files changed. Was there something in particular motivating this change? Or just an abundance of caution?

nazar-pc:
I had a similar code in my project and noticed a strange situation where one of the dependencies failed, but the result succeeded. It was very strange though because the job didn't fail because of my logic, but rather seemed like the worker lost (partially) network connection: https://github.com/nazar-pc/abundance/actions/runs/26595401463/job/78366644414

All that prompted me to look into the docs (which don't seem to mention available statuses) and eventually I decided to invert the logic and check for non-success status instead of capturing the complete list of possible failure states.

If you ever want to allow skipped jobs, that'd be better to do explicitly too.

jordancarlin (approved):
Thanks for the details. LGTM.

PR #1609, opened 2026-06-01
print CSR context (mepc, mcause, mtval, mstatus) on on trap signature mismatches

UmerShahidengr:
On any signature mismatch (integer, FP, fflags, vector), the failure report now prints the current values of `mepc`, `mcause`, `mtval`, and `mstatus` after the expected/actual value comparison. The same CSR context is added to all trap failure paths: field mismatches (cause/epc/tval/vect/ip/etc.), trap count mismatches (extra and missing), and the generic unknown-trap fallback.


tests/env/rvtest_failure_code.h
```diff
@@ -1642,6 +1743,39 @@
 
     trap_field_not_epc:
 
+        // Print current CSR state for all trap field mismatches
```
jordancarlin: This code is duplicated many times. Can this become a function that we call instead?

jordancarlin (changes requested):
Why are mcause and mtval relevant to an instruction that didn’t trap? Seems like reporting them would just add confusion. Also, this will break the failure reporting for an rv32i core that doesn’t have Zicsr (or any core that doesn’t have standard machine mode).

tests/env/rvtest_failure_code.h
UmerShahidengr: Sure, will do that

UmerShahidengr: Done

UmerShahidengr:
Good points @jordancarlin, thanks for the review. Three changes made:

1. Removed CSR context from non-trap failures — mepc/mcause/mtval/mstatus are all stale from a prior trap in that path and would just add noise.
2. Extracted a `failedtest_print_csr_context` helper — the repeated 28-line print block is now a single function called from the four trap failure paths.
3. Zicsr safety is automatic — the helper is only ever reached via failure_type == 3 (trap handler path), which by definition requires M-mode and Zicsr, so no extra #ifdef guard is needed.

jordancarlin (approved):
Please update the PR title and description to reflect what this actually does now (the CSRs are only printed for trap mismatches, not all mismatches). Otherwise, LGTM.

PR #1616, opened 2026-06-01
build: delete a failed task's outputs so partial writes can't be reused

dev-aditya-hub:
## Problem

The DAG executor decides whether a task is up-to-date purely by mtime (`is_stale`): a task is skipped if its outputs exist and no input is newer.

If an action crashes *after* it has started writing its output, that half-written file is left on disk. On the next run it looks newer than its inputs, so `is_stale` returns `False` and the task is silently skipped the corrupt output is treated as a finished build. This is dangerous in the signature pipeline specifically:

`*.S` → `*.sig.elf` → **`*.sig` (reference model)** → `*.results` → self-check `*.elf`

If the reference model is interrupted or exits early mid-write, the truncated `*.sig` is turned into `*.results` and baked into the self-check ELF. The DUT then only checks the answers that survived the truncation; everything past that point is never verified, so a non-conformant DUT can pass.

## Fix

Apply GNU Make's `.DELETE_ON_ERROR` semantics: when a task's action fails, delete its declared outputs. The task then stays stale and is regenerated on the next run instead of reusing a partial file.

- The action runner is split out unchanged as `_run_action`; `execute_task` just layers the cleanup policy on top, so the existing subprocess / process-group logic is untouched.
- Only `task.outputs` are removed. The log file (`stdout_file`) is not an output and is deliberately kept for debugging the failure.

## Testing

Adds `framework/tests/test_build.py` (stdlib `unittest`, no new dependencies):

- a failed task's partial output is removed
- a failed task stays stale and re-runs (does not silently skip) on the next run
- every output of a multi-output task is removed even if only one was written
- a successful task keeps its output and is still skipped when up-to-date

Run with:

    python -m unittest discover -s framework/tests

The failure-path tests were confirmed to fail without the fix and pass with it. `ruff` and `pyright` are clean on both changed files.
Also adds `extraPaths = ["framework/src"]` to `[tool.pyright]` so files under `framework/tests` can statically resolve `import act...`. Without it pyright flags the editable-installed package as missing even though it imports fine at
runtime.

## Note

This builds on the executor introduced in #1427, so `execute_task` is a thin wrapper that forwards the `active_pgids` / `pgids_lock` / `shutdown_event` arguments to `_run_action`.

### DIFF
--- a/framework/src/act/build.py
+++ b/framework/src/act/build.py
@@ -133,6 +133,20 @@ def is_stale(task: BuildTask) -> bool:
     return any(inp.exists() and inp.stat().st_mtime > oldest_output_mtime for inp in all_inputs)
 
 
+def _discard_outputs(task: BuildTask) -> None:
+    """Delete a failed task's outputs so a partial write can't pose as up-to-date.
+
+    Staleness is judged by mtime, so a half-written output left behind by a
+    crashing action looks newer than its inputs on the next run and is silently
+    skipped. In the signature pipeline that means a truncated reference signature
+    could be baked into the self-check ELF and let a broken DUT pass. Mirror
+    GNU Make's .DELETE_ON_ERROR. The task's log file is not an output and is kept.
+    """
+    for out in task.outputs:
+        with contextlib.suppress(OSError):
+            out.unlink(missing_ok=True)
+
+
 def execute_task(
     task: BuildTask,
     *,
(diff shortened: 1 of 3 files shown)


jordancarlin:
How does this relate to #1621? There seems to be significant overlap. Please try not to open multiple PRs with the same feature. Either continue to extend your existing PR, open a separate PR with distinct features, or wait for the first PR to be merged and then open another PR.

jordancarlin:
Closing this because of the overlap with #1621.

PR #1618, opened 2026-06-02
Add erb gem dependency for Ruby 3.4 compatibility

snnbyyds:
Ruby 3.4 removed erb from its standard library, making it a bundled gem that must be explicitly declared in the Gemfile. The udb_helpers gem requires erb (via udb/obj/instruction.rb), causing a LoadError on Ruby 3.4 without this declaration.

The conditional ensures:
- Ruby >= 3.4: erb is resolved from rubygems (erb 6.0.4, requires Ruby >= 3.2, compatible with the project's ruby ~> 3.2 constraint)
- Ruby < 3.4: no change; erb is still a default gem provided by Ruby

### DIFF
--- a/framework/src/act/data/Gemfile
+++ b/framework/src/act/data/Gemfile
@@ -25,5 +25,10 @@ else
   gem "idlc",        "0.1.5"
 end
 
+# erb was removed from Ruby's stdlib in 3.4, but is required by udb_helpers.
+# On Ruby < 3.4, erb is a default gem provided by the Ruby installation
+# and this declaration is a no-op.
+gem "erb", "6.0.4"
+
 # Mirrors udb's own required_ruby_version (~> 3.2).
 ruby "~> 3.2"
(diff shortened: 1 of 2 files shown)


jordancarlin (changes requested):
I am using Ruby 3.4.9 installed via mise (the recommended Ruby installation method for the project) and am not running into issues with this. Additionally, https://stdgems.org/erb/ still shows erb as a default gem that can be used without adding it to the Gemfile. I don't think this change is needed. Can you please share the error that you got that led to you adding this?

snnbyyds:
I was wrong about the root cause — thank you for pointing that out, and for the `stdgems.org` reference.

After revisiting this, the issue turns out to be an **Arch Linux packaging artifact**, not a Ruby 3.4 upstream change. ERB remains a default gem in Ruby 3.4 (as stdgems.org correctly shows). Arch's `ruby` package splits ERB into a separate `ruby-erb` package, so on a fresh Arch install `erb.rb` is absent from the stdlib path and `bundle exec` cannot find it. Installing `ruby-erb` via pacman resolves it, and the standard mise install (which builds Ruby from source?) is unaffected.

Adding ERB to the Gemfile was a workaround for this distro-specific packaging choice, not a genuine upstream dependency gap.

Closing this PR — thanks again for the review.

PR #1621, opened 2026-06-02
Build pipeline fixes: delete-on-error, dry-run purity, and cleanups

dev-aditya-hub:
## Summary

Three related fixes to the `act` build pipeline:

- **delete-on-error**: delete a failed task's outputs so a partial/truncated write can't masquerade as up-to-date on the next run (mirrors GNU Make's `.DELETE_ON_ERROR`)
- **pyright**: add `framework/src` to `extraPaths` and target the Linux platform so static analysis resolves the `act` package and the POSIX-only `os.killpg` / `signal.SIGKILL` used by the executor.
- **--dry-run + assorted cleanups**:
  - planning no longer writes `.tracelist` files during a dry run (it must not touch the filesystem)
  - `nm` is derived from the objdump *basename* only, so a parent dir containing "objdump" no longer corrupts the path
  - the coverage objdump symlink now depends on the objdump it links to, not just the elf
  - the Sail version check matches on leading `major.minor` instead of exact string equality
  - subprocess logs are ordered `stdout + stderr` to match the advertised `2>&1` reproduce hint
  - removed a dead post-install `bundle` guard (and its unused import)

## Testing

- `ruff check` passes on all touched files; Pylance/pyright clean
- verified a `--dry-run` coverage plan writes zero `.tracelist` files while a real run writes them as before


jordancarlin:
@dev-aditya-hub A few high-level notes on this. I will review the code in more detail after the PR is updated.
- Let's skip the Python unit-testing for now. Eventually we might add a more comprehensive pytest setup, but for now I would rather not have ad-hoc tests.
- I'm concerned about debugging build failures with delete-on-error. At a minimum, it should be disabled when the `--debug` flag is passed.
- The Sail version is very tightly coupled to the test suite, so I would rather we maintain the strict equality check.

jordancarlin:
@dev-aditya-hub are you still working on this? If so, please push a change addressing my comments from above.

jordancarlin:
Closing this due to lack of activity. I have pulled the useful parts of this PR into #1996.

PR #1627, opened 2026-06-02
Ssstrict coverage improvements

davidharrishmc:
SsstrictSm/S/U at 100% except for SstructU_instr_cg missing ebreak that will be fixed when #1636 is merged.  Runs on Sail but not other simulators.


generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
```diff
@@ -430,12 +430,8 @@ def generate_illegal_instr(
     emit_raw_words(
         lines,
         "cp_privileged_rd",
-        "00000000000000000000EEEEE1110011",
-        exclusion=[
-            "00000000000000000000000001110011",  # exclude ecall
-            "XXXXXXXXXXXXXXXXXXXX00010XXXXXXX",  # exclude rd=x2 (sp)
-            "XXXXXXXXXXXXXXXXXXXX01000XXXXXXX",  # exclude rd=x8 (scratch base)
-        ],
+        "RRRRRRRRRRRR00000000EEEEE1110011",
```
UmerShahidengr, line 439: Randomizing `funct12` here means the sweep can emit defined `funct3=000` SYSTEM instructions, not just illegal ones. `PRIVILEGED_000_EXCLUSIONS` only removes the rd=0 forms of `ecall/mret/sret/wfi/mnret`, so the `sfence/sinval/hfence` family (whose `rs2` is a general register) isn't covered. 
With `rd != 0` these are implementation-defined (a strict decoder traps; a lenient one executes the flush and raises nothing), so the covergroup can count the bin as hit without an illegal-instruction trap actually occurring — and it's fragile. Could we constrain funct12 to reserved values, or add the full `funct3=000` legal set to the exclusion list with X on the rd field?

generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
```diff
@@ -662,15 +641,19 @@ def generate_compressed_instr(
...
-            "1110XXXXXXXXXX10",  # c.sdsp/c.fswsp — sp-relative store
+            "111XXXXXXXXXXX10",  # c.sdsp/c.fswsp — sp-relative store
             "1010XXXXXXXXXX10",  # nop-like edge — unpredictable on some platforms
         ],
-        reinit_interval=50,
     )
     lines.append("")
 
+    emit_raw_words(lines, "illegal_c_jr", "1000000000000010", length=16)
+    emit_raw_words(lines, "illegal_c_jalr", "1001000000000010", length=16)
```
UmerShahidengr: `1001000000000010 (0x9002)` is `C.EBREAK` — a defined, legal instruction. Per the C-extension spec the `rs1=0` codepoint in the `c.jalr` slot is `c.ebreak`, so there's no illegal c.jalr to test here. This also contradicts the quadrant-10 exclusion a few lines up that drops the same encoding with # c.ebreak — legal, tested elsewhere. 
Suggest removing both this emit and the illegal_c_jalr bin in RISCV_coverage_comp_instr.svh. (illegal_c_jr = 0x8002 is correct — c.jr rs1=0 is genuinely reserved.)

coverpoints/priv/RISCV_coverage_vect_instr.svh
```diff
@@ -255,24 +253,32 @@
...
            & ins.current.insn[31:26] == 6'b101000) {
         // 2^6 = 64 bins
     }
 
     // vaes.vs: funct6=101001, funct3=010 (OPMVV), type in {vm,vs1}
-    // Generator: "101001ERRRRREEEEE010RRRRR1010111"
+    // Generator: "101001ERRRRREEEEE010RRRRR1110111"
     v_vaesvs : coverpoint {ins.current.insn[25], ins.current.insn[19:15]}
-        iff (ins.current.insn[6:0] == 7'b1010111
+        iff (ins.current.insn[6:0] == 7'b1110111
```
UmerShahidengr, line 275: Was the opcode change from `1010111` to `1110111` intentional? Zvkned vaes* encode in OP-V (1010111), and the f6 sweeps in the same per-SEW loop still emit at 1010111 — so as written these coverpoints look like they're exercising reserved-opcode space rather than the actual vaes encodings. Generator and coverpoint moved together so coverage closes either way; just want to confirm 1110111/"OPVE" is what's intended.

generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
```diff
@@ -527,49 +523,33 @@ def generate_vector_illegal_instr(
...
-    emit_raw_words(lines, "cp_vs_lumop_16", "RRR0001EEEEE01000101011RR0100111")
-    emit_raw_words(lines, "cp_vs_lumop_32", "RRR0001EEEEE01000110011RR0100111")
-    emit_raw_words(lines, "cp_vs_lumop_64", "RRR0001EEEEE01000111011RR0100111")
+    # mew=both, all load widths (funct3) — rs1=x8, vd=011RR
+    emit_raw_words(lines, "cp_vl_0_000", "RRRERRRRRRRR01000EEE011RR0100111")
+    # sumop sweep — mop=00, vm=1, rs1=x8, vd=011RR
+    emit_raw_words(lines, "cp_vl_lumop_8", "RRR000REEEEE01000000011RR0100111")
+    emit_raw_words(lines, "cp_vl_lumop_16", "RRR000REEEEE01000101011RR0100111")
+    emit_raw_words(lines, "cp_vl_lumop_32", "RRR000REEEEE01000110011RR0100111")
+    emit_raw_words(lines, "cp_vl_lumop_64", "RRR000REEEEE01000111011RR0100111")
```
UmerShahidengr: This store block (opcode 0100111) reuses the load coverpoint names — cp_vl_0_000, cp_vl_lumop_8/16/32/64 — and the comments still read "all load widths" / "vd=011RR". Coverage still closes via the store opcode gating on vs_width/vs_sumop, but the generated assembly ends up with duplicate cp_vl_* testcase labels and misleading traceability. Should these be cp_vs_* with store comments?

generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
```diff
@@ -527,49 +523,33 @@ def generate_vector_illegal_instr(
...
 
     # ── Vector arithmetic per-SEW sweeps ──────────────────────────────
     for sew in ["8", "16", "32", "64"]:
         lines.append(comment_banner(f"Vector arithmetic SEW={sew}", f"funct6 sweeps with e{sew}"))
         lines.append(f"\tvsetivli x0, 1, e{sew}, m1, ta, ma")
         lines.append("")
 
-        _emit_vector_init(lines)
+        #        _emit_vector_init(lines)
```
UmerShahidengr: Leftover commented-out _emit_vector_init — drop it.

generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
```diff
@@ -614,27 +595,28 @@ def generate_compressed_instr(
...
+    # Quadrant 00: Exclude loads and stores that could cause exceptions for bad addresses
     emit_raw_words(
         lines,
         "compressed00",
         "EEEEEEEEEEEEEE00",
         length=16,
-        exclusion=[
-            "XXXXXXXXXXX000XX",  # rd' = x8 — clobbers scratch base pointer
+        exclusion=[  # exclude load and store instructions that could cause exceptions for bad addresses
+            "001XXXXXXXXXXX00",  # c.fld
```
UmerShahidengr, line 609: Moving the compressed exclusions into the base coverpoint defs is the right call. One note: the exclusion logic now lives in two places — here and the generator's lists — with no single source of truth, and the c.sb/c.sh sub-range carving is fiddly enough (e.g. c_sb upper bound trimmed to `10001011111111`) . 
Relatedly, dropping reinit_interval and the rd=x8/x2 protections is safe only because every faulting memory op is excluded — so the "scratch base can't be clobbered" invariant now depends entirely on these lists staying complete.

UmerShahidengr (changes requested):

generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
davidharrishmc: Agreed it is touchy, and if new memory operations are added to the compressed set, they will need support in Sail or else this will break.

UmerShahidengr (approved):
Approved.

PR #1636, opened 2026-06-03
Increment xepc to the next instruction

UmerShahidengr:


davidharrishmc (approved):

allenjbaum:
The encodings for >32b ops are not ratified yet, so [Copilot] should not
assume that the draft spec is the one that will be implemented (in fact,
unlikely - there's a meeting to discuss it Weds morning pacific time)

davidharrishmc:
Claude review 


PR #1636 Review — "Increment xepc to the next instruction"

Overview

Three files changed: regress.yml (defensive --fail flag on curl), ExceptionsCommon.py (JALR misalignment test layout fixes), and rvtest_trap_handler.h (the substantive change — replacing a fixed +8 xEPC advance with a dynamic instruction-width probe via lhu). The core motivation is sound: the old handler always advanced by 8 bytes (masking the low 2 bits then adding 8), which skipped two instructions' worth of code and was blind to compressed vs. 32-bit instruction width. The new approach reads the low halfword at xEPC to detect compressed (16-bit) vs. standard (32-bit) encoding and advances by exactly 2 or 4.

---
Critical Issue: lhu Will Fault on PMP Execute-Only Regions

lhu     T2, 0(T3)          // fetch via trapped context

The comment claims "MXR lets the load read execute-only (R=0,X=1) pages." This is only half-true. MXR in mstatus/sstatus relaxes virtual-memory page-table permission checks to allow loads from executable-only virtual pages. It has no effect on PMP physical-access restrictions. If a PMP entry covers the instruction at xEPC with R=0, X=1, the lhu will raise a load-access fault regardless of MXR, because PMP's R bit is evaluated before the page table and is not overridden by MXR.

This matters because:
- The riscv-arch-test suite heavily tests PMP execute-only regions (PR #1729 adds 90+ files doing exactly this).
- When a trapped instruction is inside an execute-only PMP region (e.g., code in the region raises an exception — illegal instruction, system call, or a jalr from inside an execute-permitted region), xEPC points into that region. The lhu from the same address at the same privilege level (via MPRV) will PMP-fault.
- A fault inside the trap handler at this point will likely trigger undefined behavior or an infinite nested-trap loop, since there is no inner trap handler to catch it.

The same failure mode applies to instruction-access faults: if the original exception was because the instruction at xEPC was itself inaccessible (PMP or page table), the lhu will fail for precisely the same reason the original fetch failed.

Fix options: (a) Check xcause before entering this code path and skip the lhu if the cause is an instruction-access fault; (b) use an alternative strategy such as reading xTVAL (which carries the faulting instruction encoding for illegal-instruction exceptions) instead of re-loading from xEPC; (c) perform the lhu from M-mode with MPRV=0 (pure M-mode physical load) which bypasses PMP on locked entries only if M-mode bypasses PMP entirely — but this only works when the handler itself is M-mode and the PMP region is not locked, which is not guaranteed.

---
Bug: .align 4 Comment Is Wrong and May Be a Directive Error

".align 4",  # align to 4-byte boundary

On the RISC-V GNU assembler (GAS), .align N means "align to a 2^N byte boundary." So:
- .align 2 = 4-byte boundary (what the old code had, and what the comment here describes)
- .align 4 = 16-byte boundary (what the new directive actually does)

The comment says "4-byte boundary" but the directive produces 16-byte alignment. Either the comment is wrong (the directive is intentionally 16-byte) or the directive is wrong (it should be .align 2 for 4-byte alignment, matching the previous code). Either way this is agn 4 (16-byte) is truly the intent, a comment should explain why 16-bytealignment is needed here (it is not obvious given the 4-byte-aligned addresses already guaranteed by .align 2). If .align 2 was meant, the directive is wrong.

With 16-byte alignment, each of the 16 test-case blocks gets padded to a 32-byte slot (24 bytes of code + 8 bytes pad), wasting 128 bytes total and potentially confusing anyone reasoning
about test layout.

---
Significant: MPP-Intact Invariant Is Implicit and Fragile

The M-mode branch relies on:

// Relies on mstatus.MPP still holding the trapped privilege (don't write MPP before here).
csrr    T6, CSR_MSTATUS
li      T2, (1 << MPRV_LSB) | (1 << SUM_LSB) | (1 << MXR_LSB)
or      T2, T6, T2
csrw    CSR_MSTATUS, T2
lhu     T2, 0(T3)
csrw    CSR_MSTATUS, T6

The MPRV mechanism uses MPP to determine which privilege the load runs under. If anything earlier in the handler has written MPP (e.g., to prepare for a mode switch via mret), the
translation context for the lhu will be wrong — silently reading from the wrong ce. The comment notes this constraint, but it is buried inside a conditionalassembly block. This invariant should be enforced by a prominent comment at every earlier point in the handler where mstatus is written.

---
Correctness of the Branchless Sequence

andi    T2, T2, 3     // bits[1:0] → 0..3
addi    T2, T2, 1     // 1..4   (4 only if was 0b11)
srli    T2, T2, 2     // 0 or 1 (1 only if bits were 0b11)
addi    T2, T2, 1     // 1 or 2
slli    T2, T2, 1     // 2 or 4

The arithmetic is correct for all four cases:
- 0b11 (32-bit) → 3+1=4, 4>>2=1, 1+1=2, 2*2=4 ✓
- 0b10, 0b01, 0b00 (compressed) → ≤3 after add, >>2=0, 0+1=1, 1*2=2 ✓

This correctly advances xEPC by 4 for 32-bit instructions and 2 for compressed. s and readable with the inline comments.

---
The ExceptionsCommon.py Label Placement

The new code places the testcase label and the jalr at the same PC (PC+8):

test_data.add_testcase(...),  # PC+8  (label)
f"jalr x1, ...",              # PC+8  THE TEST

Labels have zero width, so this is architecturally correct. With the new trap ha it returns to addi x0, x2, 0 at PC+12, which is the right resume point. The oldhandler advanced by 8 from the jalr, returning to PC+16, skipping addi x0, x2, 0 entirely — this was the bug being fixed. The two nop pads at PC+16 and PC+20 provide a safe landing zone for
the specific case where jalr_off=6 causes a branch to PC+14 (upper half of the auld be treated as a 16-bit instruction on C-extension targets.

Lost comment: The three-line comment that was removed:
# jalr_off controls the offset[1:0], covering all 16 combinations of (rs1+offset)[1:0].
# JALR jumps to (rs1 + offset) with bit 0 cleared.
# Misaligned exception occurs when bit 1 of the target is set
explained the test's mathematical structure. The removal makes the offsets_for_lharder to follow without context.

---
Minor Issues

regress.yml double space: The second curl line reads curl  --fail (two spaces). Cosmetic but inconsistent with the first.

48-bit/64-bit instructions: All bits[1:0]==0b11 cases are treated as 4-byte. Allen Baum correctly notes these wider encodings are unratified, so omitting them is defensible. However, the
comment should say "32-bit" explicitly rather than implying completeness (e.g., :0]==0b11)") — the current comment already does this, which is fine.

SUM is set unconditionally in both branches. SUM enables S-mode access to U-modeit's only needed when MPP=S and the instruction is in a U-only page; in the non-M branch it's needed when the trapped code was U-mode. Setting it unconditionally is harmless but the comment overstates its necessity.

---
Summary
                                                                                                                                                                                               The branchless instruction-length detection and the MPRV/SUM/MXR translation-conuine improvement over the blunt +8 advance. The ExceptionsCommon.py layout fixcorrectly relocates the resume target from PC+16 to PC+12. However, the PMP execute-only case is a hard correctness bug: lhu with MPRV will fault on PMP R=0,X=1 regions, MXR notwithstanding, which will break a large class of PMP tests. The .align 4 directive comment is atself may be a mistake. These need to be resolved before merge.

UmerShahidengr:
Changes of this PR have been merged with PR #1837, so I am closing this one

PR #1637, opened 2026-06-03
InterruptsS working on spike

Teemooooooooo:
InterruptsS working on spike
- Spike only has one UART for external interrupt source, which does not allow MEIP and SEIP to be driven by interrupt controller at the same time
    - modified the test so that it only uses interrupt controller to drive SEIP when necessary and use mip otherwise
    - modified the trap handler: if MEIP is high, trap handler only clears SEIP by writing to mip and does not clear it from interrupt controller. If MEIP is low, it clears SEIP using both ways.


config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -155,8 +155,10 @@
   li _R2, NS16550_BASE_ADDRESS;                  \
   sb zero, 1(_R2);                               \
   li _R2, PLIC_CLAIM_ADDRESS;                    \
-  lw _R1, 0(_R2);                               \
-  sw _R1, 0(_R2);
+  lw _R1, 0(_R2);                                 \
+  sw _R1, 0(_R2);                               \
+  li _R2, PLIC_ENABLE_ADDRESS;  \
```
davidharrishmc: Why is it necessary to disable the PLIC?

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -189,15 +189,12 @@
   sb zero, 1(_R2);                               \
   li _R2, PLIC_SCLAIM_ADDRESS;                    \
   lw _R1, 0(_R2);                               \
-  sw _R1, 0(_R2);
+  sw _R1, 0(_R2);                               \
+  li _R2, PLIC_SENABLE_ADDRESS;                 \
```
davidharrishmc: Why is this needed?

generators/testgen/src/testgen/asm/interrupts.py
davidharrishmc: Please move comments inside quotes.

davidharrishmc (commented):
This is looking great!  Minor feedback and a merge conflict to reserve.

davidharrishmc (changes requested):

config/spike/spike-rv64-max/rvmodel_macros.h
Teemooooooooo: I don't think it is necessary but I just felt it is safer to disable it. Should I just keep it enabled the whole time?

davidharrishmc:
Let’s only do what is necessary.  Don’t disable unless there is a reason to
do so. To the best of my understanding, the PLIC releases interrupts using
the claim register and won’t cause them again by accident.

config/spike/spike-rv64-max/rvmodel_macros.h
Teemooooooooo: I lied it is necessary - The test breaks when removing the disable PLIC. Since we have S and M external interrupt contexts sharing the same source, keeping it enabled can cause the PLIC to trigger MEXT int when it is supposed to trigger SEXT int :( Should I just keep it then?

davidharrishmc (approved):

PR #1642, opened 2026-06-04
Updated PMP tests

UmerShahidengr:


tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_off_all.S
```diff
@@ -338,7 +368,9 @@ test_3_str: .string "\"test: 3; cp: pmpm_cfg_A_off_all_lw\""
 
 .align 12
```
davidharrishmc, line 369: Does this align 12 and rest 1024 imply any assumptions about the grain size?  Could there be a comment explaining what is happening here and why these numbers are selected?

davidharrishmc (approved):

config/whisper/whisper-rv64-max/whisper.json
```diff
@@ -61,7 +61,7 @@
 
   "misaligned_has_priority": false,
   "num_mmode_perf_regs": 8,
-  "physical_memory_protection_grain": "0x1000",
+  "physical_memory_protection_grain": "0x10",
```
jordancarlin: Rather than changing the whisper config, can we update the Sail config to match it? The tests should work with a large granularity like what Whisper currently has and it is useful to test that since our other configs have small granularity.

config/whisper/whisper-rv32-max/whisper.json
```diff
@@ -61,7 +61,7 @@
 
   "misaligned_has_priority": false,
   "num_mmode_perf_regs": 8,
-  "physical_memory_protection_grain": "0x1000",
+  "physical_memory_protection_grain": "0x10",
```
jordancarlin: Same comment. Let’s keep the whisper config as is and update the sail config to match instead.

config/cores/cve4/ci.yaml
```diff
@@ -3,7 +3,9 @@
 # CI configuration for the cve4 family: CV32E40P and CV32E40X
 
 ci_enabled: true
-exclude_extensions: "Sm,SmF,SsstrictSm" # SmF: cv32e40p-rv32imcf DUT bug - https://github.com/openhwgroup/cv32e40p/issues/1060; SsstrictSm not supported
+exclude_extensions: "Sm,SmF,SsstrictSm,PMPSm"
+# SmF: cv32e40p-rv32imcf DUT bug - https://github.com/openhwgroup/cv32e40p/issues/1060; SsstrictSm not supported
+# PMP not supported, but still getting selected. Excluded it for the time being
```
jordancarlin: Are all of the tests running or just one? If it’s just one, it may be missing the parameter constraint in the YAML header.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -27,13 +27,12 @@
 # REQUIRED_EXTENSIONS:  ['Sm']
 # params:
 #   MXLEN: 32
-#   NUM_PMP_ENTRIES: '64'
```
jordancarlin, line 30: This should at least require the number of PMP regions to be greater than 0. No need to run if PMP is not supported.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -231,8 +276,10 @@ main:
 // Test Case: 9 : PMP Region 54
     LA(x5, REGIONSTART)
     srl x5, x5, PMP_SHIFT
-    andi x5, x5, PMP_MASK
-    ori x5, x5, PMP_REGION_SIZE
+    li x6, PMP_MASK
```
jordancarlin: This should be using the LI macro everywhere.

jordancarlin (changes requested):

config/whisper/whisper-rv64-max/whisper.json
UmerShahidengr: Setting UDB_PMP_GRANULARITY to 2048 was raising error, thats why I had to reduce this value. Probably UDB is not expecting such a large grain size.

config/whisper/whisper-rv32-max/whisper.json
UmerShahidengr: Tried it, but UDB raised error

config/cores/cve4/ci.yaml
UmerShahidengr: All tests are running, probably we require a mask to avoid running PMPSm on cores supporting Sm but not PMP

davidharrishmc:
Jordan knows UDB best, but we need to test at least 4KB grains.  The lack of testing was how the last round of bugs slipped through.

davidharrishmc:
I think using UDB_NUM_PMP > 0 in the required parameters would exclude the cores that don’t have PMP.   Is that sufficient to avoid a CI exclusion for the OpenHW cores?

config/whisper/whisper-rv64-max/whisper.json
jordancarlin: The `UDB_PMP_GRANULARITY` parameter is defined as follows:

> log2 of the smallest supported PMP region.
>
 > Generally, for systems with an MMU, should not be smaller than 12,
  as that would preclude caching PMP results in the TLB along with
  virtual memory translations
>
 > Note that PMP_GRANULARITY is equal to G+2 (not G) as described in
  the privileged architecture.

So in this case I think it should be 12 (2^12 = 0x1000),

davidharrishmc:
These units are terribly confusing.  It might be worth putting a comment next to every PMP granularity mention about what units it is in (grain size in bytes, log(grain size in bytes), log(grain size in words).

jordancarlin:
Agreed. One of the parts making it particularly tricky is the presence of G and the +2. Sail uses G as the input config option, so it will always be 2 less than the UDB equivalent parameter.

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_off_all.S
UmerShahidengr: It is done to keep label at an address where it will be captured by coverpoint while running coverage. Nothing special related to grain size here

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_zero.S
```diff
@@ -17,6 +17,25 @@
...
+//   and the trap handler rather than removing the PA 0 probes.
+//     - Define RVMODEL_ACCESS_FAULT_ADDRESS in config/.../rvmodel_macros.h
+//       to the base of your platform's unmapped/faulting region (typically 0).
+//       Match this in sail.json memory.regions and link.ld.
+//     - SKIP_MEPC and SKIP_MTVAL (below) enable trap-handler support in
+//       rvtest_trap_handler.h for intentional faults at that address:
+//         * If xEPC equals RVMODEL_ACCESS_FAULT_ADDRESS (+2), EPC relocation
+//           is skipped so the handler does not abort on an out-of-segment PC.
+//         * Fetch probes set x4 = 0xACCE; on fetch faults the handler sets
+//           xEPC = ra so execution resumes after the jalr to PA 0.
```
jordancarlin: I don't think the "resolution" part of this note is relevant here. Those are more general PMP notes that might be good to capture elsewhere, but not in this specific test.

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_zero.S
```diff
@@ -17,6 +17,25 @@
 //                 Dropping the background region. With pmpcfg0.L=1, pmpcfg0.A=TOR,
 //                 pmpcfg0.XWR=111, default TOR region: {lw, sw, jalr} to {0,
 //                 pmpaddr0-4, pmpaddr0}.
+//
+// Platform note (riscv-arch-test issue#721):
+//   Access to physical address 0 is platform-dependent: Spike/Sail and many
+//   SoCs leave the first page unmapped, so load/store/fetch at PA 0 may trap
+//   even when PMP permits the access. Other platforms may have memory at 0.
+//   This test intentionally probes PA 0 because PMP TOR region 0 starts
+//   there; it is verifying PMP boundaries, not generic access-fault behavior.
```
jordancarlin: Sail has a configurable memory map, so this isn't true. It does not always leave the first page unmapped.

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_na4_all.S
```diff
@@ -97,7 +97,8 @@ main:
     //Lock the Background region.
     LI(x4, (PMP_REGION_GLOBAL << PMP3_CFG_SHIFT))
     csrw pmpcfg3, x4
-
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: Shouldn't this be `#if UDB_PMP_GRANULARITY == 2` because the parameter is "log2 of the smallest supported PMP region"?

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S
```diff
@@ -85,6 +85,11 @@ main:
     //Lock the Background region.
     LI(x4, (PMP_REGION_GLOBAL))
     csrw pmpcfg3, x4
+// For larger UDB_PMP_GRANUMARITY, it is raising
+// relocation truncation error to fit: R_RISCV_JAL
+// Regions get so large, trap handler, RVMODEL macros,
+// and failure macros gets far apart and start raising errors
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: Can we use `LA` + `jalr` instead to avoid this? Would be good to maintain the test even for larger granularities.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
```diff
@@ -45,17 +45,16 @@
...
 
     // -----------------------------------------------------------------------
     // Walking ones through even pmpcfg registers (pmpcfg0, pmpcfg2, ...).
     //
-    // The original test writes walking ones only to even-numbered pmpcfg
+    // The test writes walking ones only to even-numbered pmpcfg
     // registers (pmpcfg0, pmpcfg2, and — for 64-PMP systems — pmpcfg4 through
     // pmpcfg14).  Odd registers (pmpcfg1, pmpcfg3, ...) are RV32 aliases and
-    // are intentionally left out, matching original behaviour.
+    // are intentionally left out.
```
jordancarlin: Is there a different test that checks the odd pmpcfg registers? If so, it would be good to reference it in this comment.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
```diff
@@ -103,12 +101,7 @@ main:
     .endr
 
     // --- pmpcfg even walking ones, bit 1 ---
-    .set pmpcfgi, CSR_PMPCFG0
-    .rept UDB_NUM_PMP_ENTRIES/8
-1:  li t2, (1 << 1)
-    RVTEST_SIGUPD_CSR_WRITE(pmpcfgi, t2, 1b, test_6_str)
-    .set pmpcfgi, pmpcfgi+2
-    .endr
+    // Skipped, R=0,W=1 is reserved
```
jordancarlin, line 98: We could still test this and just use slightly more complex checking logic to make sure that we read back something other than the reserved value. It is important to test that the reserved value is not read back and that the field follows proper WARL behavior.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-6.S
```diff
@@ -105,14 +104,7 @@ main:
...
     // --- pmpcfg even walking ones, bit 9 ---
-
-    .set pmpcfgi, CSR_PMPCFG0
-    .rept UDB_NUM_PMP_ENTRIES/8
-1:  li t2, (1 << 9)
-    RVTEST_SIGUPD_CSR_WRITE(pmpcfgi, t2, 1b, test_6_str)
-    .set pmpcfgi, pmpcfgi+2
-    .endr
-
+    // Skipped, R=0,W=1 is reserved
```
jordancarlin, line 99: Same comment as on the previous test.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-8.S
```diff
@@ -61,12 +61,7 @@ main:
     // -----------------------------------------------------------------------
 
     // --- pmpcfg even walking ones, bit 17 ---
-    .set pmpcfgi, CSR_PMPCFG0
-    .rept UDB_NUM_PMP_ENTRIES/8
-1:  li t2, (1 << 17)
-    RVTEST_SIGUPD_CSR_WRITE(pmpcfgi, t2, 1b, test_1_str)
-    .set pmpcfgi, pmpcfgi+2
-    .endr
+    // Skipped, R=0,W=1 is reserved
```
jordancarlin, line 57: See comment on previous test.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-9.S
```diff
@@ -69,12 +69,7 @@ main:
     .endr
 
     // --- pmpcfg even walking ones, bit 25 ---
-    .set pmpcfgi, CSR_PMPCFG0
-    .rept UDB_NUM_PMP_ENTRIES/8
-1:  li t2, (1 << 25)
-    RVTEST_SIGUPD_CSR_WRITE(pmpcfgi, t2, 1b, test_2_str)
-    .set pmpcfgi, pmpcfgi+2
-    .endr
+    // Skipped, R=0,W=1 is reserved
```
jordancarlin, line 65: Same comment. This is still important to test.

tests/priv/pmp/pmp32/PMPSm/pmpsm_na4_legal_lwxr.S
```diff
@@ -155,6 +155,8 @@ main:
     LI(x4, PMP_REGION_GLOBAL)
     csrw pmpcfg3, x4
 
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: Should be 2.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -132,19 +147,35 @@ main:
...
+    LI(x4, PMP_REGION_GLOBAL)
     csrw pmpcfg15, x4
+#elif UDB_NUM_PMP_ENTRIES == 16
+    LA(x4, PMPADDRESS_TOP_GLOBAL)
+    csrw pmpaddr15, x4
+    LI(x4, PMP_REGION_GLOBAL)
+    csrw pmpcfg3, x4
+#else
+    #error "Unsupported UDB_NUM_PMP_ENTRIES value; only 16 and 64 are supported."
+#endif
```
jordancarlin, line 169: How does this work for DUTs that have read-only PMP regions?

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_na4_all.S
jordancarlin: Looks like several of the UDB configs are wrong and need to be updated as well because they list the granularity as 4 instead of 2.

tests/priv/pmp/pmp32/PMPSm/pmpsm_napot_legal_lwxr.S
```diff
@@ -66,6 +66,7 @@ RVTEST_BEGIN
     nop
     nop
 
+    LI(x4, 0xACCE)
```
jordancarlin, line 69: Why is this added? A comment about why the LI is needed would be helpful here since it doesn't look like the value is used anywhere.

tests/priv/pmp/pmp32/PMPSm/pmpsm_priority.S
```diff
@@ -172,10 +174,10 @@ main:
 
     VERIFICATION_RWX   (TEST_FOR_EXECUTION + 6*g - 4), test_6
 
-// Test Case: 6 : Accessing at end of region with No Permissions
+// Test Case: 7 : Accessing at end of region with No Permissions
 
     VERIFICATION_RWX   (TEST_FOR_EXECUTION + 7*g - 4), test_7
-
+*/
```
jordancarlin: Why is this whole section commented out?

tests/priv/pmp/pmp32/PMPSm/pmpsm_priority.S
```diff
@@ -188,11 +190,13 @@ test_3_str: .string "\"test: 3; cp: pmpm_priority_jalr\""
...
 .align 12
-TEST_FOR_EXECUTION_x:
+TEST_FOR_EXECUTION_0:
+    .rept 1024
     jr ra
+    .endr
 
 TEST_FOR_EXECUTION:
-    .rept (8*(1<<(UDB_PMP_GRANULARITY)))
+    .rept (3*(1<<(UDB_PMP_GRANULARITY)))
```
jordancarlin: Why *3? Where does this number come from?

tests/priv/pmp/pmp64/PMPSm/pmpsm_all_entries_check.S
```diff
@@ -497,45 +499,45 @@ main:
...
-    SET_PMP_ENTRY pmpaddr1, pmpcfg0, PMP_REGION_1
-    VERIFICATION_RWX    TEST_FOR_EXECUTION, test_14
+# // Test Case: 14 — slot 1, pmpcfg0
+#     SET_PMP_ENTRY pmpaddr1, pmpcfg0, PMP_REGION_1
+#     VERIFICATION_RWX    TEST_FOR_EXECUTION, test_14
 
-// Test Case: 15 — slot 0, pmpcfg0
-    SET_PMP_ENTRY pmpaddr0, pmpcfg0, PMP_REGION_0
-    VERIFICATION_RWX    TEST_FOR_EXECUTION, test_15
+# // Test Case: 15 — slot 0, pmpcfg0
```
jordancarlin: Why are all of these tests being commented out?

tests/priv/pmp/pmp64/PMPSm/pmpsm_cfg_A_tor_zero.S
```diff
@@ -17,6 +17,25 @@
...
+//   and the trap handler rather than removing the PA 0 probes.
+//     - Define RVMODEL_ACCESS_FAULT_ADDRESS in config/.../rvmodel_macros.h
+//       to the base of your platform's unmapped/faulting region (typically 0).
+//       Match this in sail.json memory.regions and link.ld.
+//     - SKIP_MEPC and SKIP_MTVAL (below) enable trap-handler support in
+//       rvtest_trap_handler.h for intentional faults at that address:
+//         * If xEPC equals RVMODEL_ACCESS_FAULT_ADDRESS (+2), EPC relocation
+//           is skipped so the handler does not abort on an out-of-segment PC.
+//         * Fetch probes set x4 = 0xACCE; on fetch faults the handler sets
+//           xEPC = ra so execution resumes after the jalr to PA 0.
```
jordancarlin: Same issue with this comment as mentioned above.

tests/priv/pmp/pmp64/PMPSm/pmpsm_cfg_na4_all.S
```diff
@@ -102,6 +102,8 @@ main:
     srl x5, x5, PMP_SHIFT
     csrw pmpaddr14, x5
 
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: Should be 2.

tests/priv/pmp/pmp64/PMPSm/pmpsm_cfg_tor_all.S
```diff
@@ -88,7 +88,11 @@ main:
     //Lock the Background region.
     LI(x4, (PMP_REGION_GLOBAL))
     csrw pmpcfg2, x4
-
+// For larger UDB_PMP_GRANUMARITY, it is raising
+// relocation truncation error to fit: R_RISCV_JAL
+// Regions get so large, trap handler, RVMODEL macros,
+// and failure macros gets far apart and start raising errors
+#if UDB_PMP_GRANULARITY == 4
     LA(x4, TEST_FOR_EXECUTION_0)
```
jordancarlin, line 96: See comment above. Can we use a `jalr` instead to avoid this and keep the test?

tests/priv/pmp/pmp64/PMPSm/pmpsm_cfg_tor_check-02.S
```diff
@@ -151,7 +151,8 @@ main:
 
     //addresses for a default TOR region
     LA(x4, TEST_FOR_EXECUTION)
-    addi    x5, x4, g
+    li t0, g
```
jordancarlin: Use the LI macro.

tests/priv/pmp/pmp64/PMPSm/pmpsm_csr_walk-10.S
```diff
@@ -99,13 +99,7 @@ main:
...
 
     // --- pmpcfg even walking ones, bit 1 ---
-
-    .set pmpcfgi, CSR_PMPCFG0
-    .rept UDB_NUM_PMP_ENTRIES/8
-1:  li t2, (1 << 1)
-    RVTEST_SIGUPD_CSR_WRITE(pmpcfgi, t2, 1b, test_66_str)
-    .set pmpcfgi, pmpcfgi+2
-    .endr
+    // Skipped, R=0,W=1 is reserved
```
jordancarlin, line 102: Don't we still need to test the reserved behavior to make sure the WARL rules are obeyed?

tests/priv/pmp/pmp64/PMPSm/pmpsm_csr_walk-11.S
```diff
@@ -90,13 +90,7 @@ main:
...
 
     // --- pmpcfg even walking ones, bit 9 ---
-
-    .set pmpcfgi, CSR_PMPCFG0
-    .rept UDB_NUM_PMP_ENTRIES/8
-1:  li t2, (1 << 9)
-    RVTEST_SIGUPD_CSR_WRITE(pmpcfgi, t2, 1b, test_73_str)
-    .set pmpcfgi, pmpcfgi+2
-    .endr
+    // Skipped, R=0,W=1 is reserved
```
jordancarlin, line 93: See above comment.

tests/priv/pmp/pmp64/PMPSm/pmpsm_csr_walk-12.S
```diff
@@ -99,13 +99,7 @@ main:
...
 
     // --- pmpcfg even walking ones, bit 17 ---
-
-    .set pmpcfgi, CSR_PMPCFG0
-    .rept UDB_NUM_PMP_ENTRIES/8
-1:  li t2, (1 << 17)
-    RVTEST_SIGUPD_CSR_WRITE(pmpcfgi, t2, 1b, test_80_str)
-    .set pmpcfgi, pmpcfgi+2
-    .endr
+    // Skipped, R=0,W=1 is reserved
```
jordancarlin, line 102: See above comment.

tests/priv/pmp/pmp64/PMPSm/pmpsm_csr_walk-10.S
jordancarlin: There are a bunch of other instances of this. Not going to comment on all of them.

tests/priv/pmp/pmp64/PMPSm/pmpsm_na4_boundary-02.S
```diff
@@ -93,15 +93,16 @@ main:
     addi x4, x4, 4
     srl x4, x4, PMP_SHIFT
     csrw pmpaddr1, x4
-
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: Should be 2.

tests/priv/pmp/pmp64/PMPSm/pmpsm_na4_legal_lwxr.S
```diff
@@ -159,7 +159,8 @@ main:
     //Lock the Background region.
     LI(x4, PMP_REGION_GLOBAL)
     csrw pmpcfg2, x4
-
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: Should be 2.

tests/priv/pmp/pmp64/PMPSm/pmpsm_napot_legal_lxwr-01.S
```diff
@@ -62,35 +62,36 @@ RVTEST_BEGIN
...
-    addi a4, a4, 8                      // REGIONSTART + 4
+    addi a4, a4, 8                        // REGIONSTART + 4
     LA(x1, 3f)                            // Store the return Address in x1
     jalr ra, 0(a4)
     nop
 3:
     nop
 
-    addi a4, a4, (g-8)                  // REGIONSTART + g - 4
+    li t0, (g-8)
```
jordancarlin: Needs to use LI macro.

tests/priv/pmp/pmp64/PMPSm/pmpsm_priority.S
```diff
@@ -149,7 +151,11 @@ main:
 // Test Case: 2 : Accessing at end of region with XR Permissions
 
     VERIFICATION_RWX   (TEST_FOR_EXECUTION + 2*g - 4), test_2
-
+// For larger UDB_PMP_GRANUMARITY, it is raising
+// relocation truncation error to fit: R_RISCV_JAL
+// Regions get so large, trap handler, RVMODEL macros,
+// and failure macros gets far apart and start raising errors
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: Can we use jalr?

jordancarlin (dismissed):
Finished a more extensive review. Here is another round of comments.

config/whisper/whisper-rv64-max/whisper.json
UmerShahidengr: I am not setting up `UDB_PMP_GRANULARITY`, instead I am changing `PMP_GRANULARITY` parameter in the .yaml file, which is causing the error

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
UmerShahidengr: Done

config/whisper/whisper-rv64-max/whisper.json
jordancarlin: `UDB_PMP_GRANULARITY` is the macro generated from `PMP_GRANULARITY` in the YAML file. All of the UDB parameters have UDB_ prefixed to them when generating the header file to avoid namespace issues. They are the same thing.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
UmerShahidengr: Done

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
UmerShahidengr: It wont work on such DUTs, we have no control over such DUTs thus cant run any PMP test on it

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_zero.S
UmerShahidengr: Old comment, removed it

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_na4_all.S
UmerShahidengr: I am still confused, is `UDB_PMP_GRANULARITY = PMP_GRANULARITY - 2`?
I have assumed `UDB_PMP_GRANULARITY = PMP_GRANULARITY` everywhere

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S
UmerShahidengr: Not possible, regions are going so far apart, that failure codes cant find trap handler macros, trap handler cant find RVMODEL macros, this will be a huge effort to somehow run such tests on higher grain value. Cant run beyond this grain value as region size is increasing within this test

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
UmerShahidengr: It was a unified test but it was going beyond the code region and pmp regions were going on to the next aligned boundary so coverage was dropping, we split one test into multiple files, thus comment is refering to the old test. But it can be eliminated

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
UmerShahidengr: All sims are returning some different legal values, some are ignoring the instruction, some are returning zero, since it is a reserved behavior so it is out of scope, hence removed

tests/priv/pmp/pmp32/PMPSm/pmpsm_napot_legal_lwxr.S
UmerShahidengr: It is getting used by trap handler on fetch access fault. Trap handler check against this magic value and bring PC back to main

tests/priv/pmp/pmp32/PMPSm/pmpsm_priority.S
UmerShahidengr: Good catch, forgot to uncomment it, was done during debugging. Thanks

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S
jordancarlin: I don't think this is an acceptable solution long term. We need to be testing PMP just as extensively no matter what the granularity is (especially because most high performance cores will have PMP granularities of 12). If this is necessary for now to get the tests running, that is fine, but let's open an issue to figure out a more complete solution later. @davidharrishmc any thoughts?

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
jordancarlin: While there are many correct answers, I think we still need to check that it does not return the reserved value with R=0,W=1.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
jordancarlin: @UmerShahidengr Can you open an issue to track this? We need to support cores that have read-only 0 PMP regions in the upper PMP registers (there are at least a few actual cores that I know of that do this).

tests/priv/pmp/pmp32/PMPSm/pmpsm_napot_legal_lwxr.S
jordancarlin: Can you add a comment about that? The magic value is not obvious if you don't know this.

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S
UmerShahidengr: I agree, lets open the issue for future, and let this go. For larger grain size, I think we have to update the tests from scratch as it is using .rept / addi / LI pairs whose ranges will be exhausted on grain index > 10

tests/priv/pmp/pmp64/PMPSm/pmpsm_napot_legal_lxwr-01.S
UmerShahidengr: Done everywhere

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
UmerShahidengr: Sure, will do that

jordancarlin:
@UmerShahidengr I’d like to take another review before we merge. It’s no longer going to make a difference for the EU summit (the demo was earlier today).

config/cores/cve2/ci.yaml
```diff
@@ -5,6 +5,7 @@
 ci_enabled: true
 # Sm excluded: Insufficient WARL configuration options.
 # Ssstrict excluded: blocked by https://github.com/riscv/riscv-arch-test/issues/1601
+# PMP not supported, but still getting selected. Excluded it for the time being
```
jordancarlin: Outdated comment.

config/cores/cve4/ci.yaml
```diff
@@ -3,7 +3,9 @@
 # CI configuration for the cve4 family: CV32E40P and CV32E40X
 
 ci_enabled: true
-exclude_extensions: "Sm,SmF,SsstrictSm" # SmF: cv32e40p-rv32imcf DUT bug - https://github.com/openhwgroup/cv32e40p/issues/1060; SsstrictSm not supported
+exclude_extensions: "Sm,SmF,SsstrictSm"
+# SmF: cv32e40p-rv32imcf DUT bug - https://github.com/openhwgroup/cv32e40p/issues/1060; SsstrictSm not supported
+# PMP not supported, but still getting selected. Excluded it for the time being
```
jordancarlin: Outdated comment.

tests/priv/pmp/pmp32/PMPF/pmpf_cfg_wr.S
```diff
@@ -22,7 +22,7 @@
 # REQUIRED_EXTENSIONS:  ['F','Sm']
 # params:
 #   MXLEN: 32
-#   NUM_PMP_ENTRIES: '>0'
+#   UDB_NUM_PMP_ENTRIES: '>0'
```
jordancarlin: The YAML header parameter constraints should not have the UDB_ prefix.
```suggestion
#   NUM_PMP_ENTRIES: '>0'
```

tests/priv/pmp/pmp32/PMPF/pmpf_cfg_wr.S
```diff
@@ -141,8 +143,10 @@ main:
 
     LA(x5, REGIONSTART)
     srl x5, x5, PMP_SHIFT
-    andi x5, x5, PMP_MASK
-    ori x5, x5, PMP_REGION_SIZE
+    LI(x6, PMP_SHIFT)
```
jordancarlin: Same question. This is now using PMP_SHIFT twice and not using PMP_MASK. There are a bunch of instances of this throughout the file.

tests/priv/pmp/pmp32/PMPS/pmps_na4_legal_lxwr.S
```diff
@@ -43,6 +43,8 @@ RVTEST_BEGIN
 
 .macro VERIFICATION_RWX ADDRESS TEST_CASE
 
+    fence.i
```
jordancarlin: Global comment: why are all of the new `fence.i` instructions throughout these tests needed? Are they using self-modifying code? Right now these tests don't require Zifencei, so this won't work at the moment.

tests/priv/pmp/pmp32/PMPF/pmpf_cfg_wr.S
```diff
@@ -128,8 +128,10 @@ main:
...
+    and x5, x5, x6
+    LI(x6, PMP_REGION_SIZE)
+    or x5, x5, x6
     csrw pmpaddr5, x5
```
jordancarlin, line 135: Should this be PMP_MASK? That would make it consistent with what we had before.

config/whisper/whisper-rv32-max/whisper-rv32-max.yaml
```diff
@@ -279,7 +279,7 @@ params:
   MTVEC_BASE_ALIGNMENT_VECTORED: 4
   MTVEC_ILLEGAL_WRITE_BEHAVIOR: retain
   NUM_PMP_ENTRIES: 64
-  PMP_GRANULARITY: 12
+  PMP_GRANULARITY: 16
```
jordancarlin: I'm confused by the Whisper PMP configs:
- The sail config file says `G` is 2
- The UDB config file says `PMP_GRANULARITY` (which should be G+2) is 16
- The Whisper config file says the grain (not sure what the units are) is 0x10 (and was changed from the 0x1000 that we had before)

Can we keep the Whisper config file unchanged since that is the grain that was given to us by Tenstorrent and match it in the UDB and Sail configs? Right now the different config files do not seem to match up.

tests/priv/pmp/pmp32/PMPS/pmps_na4_legal_lxwr.S
```diff
@@ -144,7 +146,8 @@ main:
     //Lock the Background region.
     LI(x4, PMP_REGION_GLOBAL)
     csrw pmpcfg3, x4
-
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 4
```
jordancarlin: This still doesn't seem right. I think it should be `#if UDB_PMP_GRANULARITY == 2` because the parameter is "log2 of the smallest supported PMP region". NA4 works for regions that are 4 bytes, and log2 of 4 is 2. Global comment that applies many places.

jordancarlin (changes requested):

tests/priv/pmp/pmp32/PMPS/pmps_na4_legal_lxwr.S
UmerShahidengr: The verification macro performs a load/store operation on a target address and then jumps to that address to verify execute permissions. When a region has write (W) permission, the macro updates the value at that address — however, this updated value is held only in the **data cache (D-cache)**. When execution jumps to that address, the **instruction cache (I-cache)** and main memory still hold the stale, pre-update copy.

This D-cache/I-cache incoherency causes a mismatch between CVW and the SAIL reference model, leading to test failures on CVW. A `fence.i` instruction is therefore inserted to flush the cache and ensure it is synchronized with the most recently written data before execution resumes at that address.

tests/priv/pmp/pmp32/PMPS/pmps_na4_legal_lxwr.S
```diff
@@ -144,7 +146,8 @@ main:
     //Lock the Background region.
     LI(x4, PMP_REGION_GLOBAL)
     csrw pmpcfg3, x4
-
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 2
```
jordancarlin: For tests where we want the entire test to be skipped based on the granularity (like this one) let's add the constraint to the test YAML header instead of ifdefing the entire body of the test out. The following should work:
```yaml
##### START_TEST_CONFIG #####
# REQUIRED_EXTENSIONS:  ['S']
# params:
#   MXLEN: 32
#   NUM_PMP_ENTRIES: '>0'
#   PMP_GRANULARITY: 2
# MARCH: rv32i_zicsr_zifencei
##### END_TEST_CONFIG #####
```

This only applies to tests where we want to skip the entire test. If we just want to skip part of the test then the existing `#if` should be kept instead.

tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S
jordancarlin: @UmerShahidengr looks like an issue still needs to be opened for this. For example, some cores may implement 16 regions but make the top 8 read-only 0, so there are functionally only 8 regions. We need to be able to support this.

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_zero.S
```diff
@@ -58,7 +58,7 @@ RVTEST_BEGIN
     RVTEST_SIGUPD(x2, x5, x4, a4,\TEST_CASE\()_2,test_2_str)               // Signature update
 
     LI(x4, 0xACCE)
-    li a5, 0
+    LI(a5, RVMODEL_ACCESS_FAULT_ADDRESS)
```
jordancarlin: I think this should stay as 0 for this test. Isn't the point of it to check that region 0 extends from 0 to pmpaddr0? Probably worth adding a comment explaining why we use 0 instead of RVMODEL_ACCESS_FAULT_ADDRESS.

jordancarlin: Applies to all three instances in this test and the rv64 equivalent test.

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S
```diff
@@ -85,6 +85,11 @@ main:
     //Lock the Background region.
     LI(x4, (PMP_REGION_GLOBAL))
     csrw pmpcfg3, x4
+// For larger UDB_PMP_GRANUMARITY, it is raising
+// relocation truncation error to fit: R_RISCV_JAL
+// Regions get so large, trap handler, RVMODEL macros,
+// and failure macros gets far apart and start raising errors
+#if UDB_PMP_GRANULARITY == 2
```
jordancarlin: @UmerShahidengr Please open an issue for this before we merge this PR so it doesn't get forgotten. We need to find a solution eventually so that all cores are running this test no matter their granularity. Otherwise that is a significant coverage hole.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
jordancarlin: This was marked as resolved, but I still see the old comment. If this test skips the odd registers (which are not aliases like the comment says; they control unique regions), then it should mention which test does cover them. If there is no test that currently covers the odd regions, then please open an issue so that we remember to add a test because that is a coverage hole.

tests/priv/pmp/pmp32/PMPSm/pmpsm_na4_legal_lwxr.S
```diff
@@ -155,6 +155,8 @@ main:
     LI(x4, PMP_REGION_GLOBAL)
     csrw pmpcfg3, x4
 
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 2
```
jordancarlin: This one can use the test YAML header instead of the `#if`.

tests/priv/pmp/pmp32/PMPSm/pmpsm_napot_legal_lwxr.S
jordancarlin: This was marked resolved but there is still no comment here explaining this.

tests/priv/pmp/pmp32/PMPSm/pmpsm_napot_legal_lwxr.S
```diff
@@ -345,7 +369,7 @@ main:
 RVTEST_CODE_END
 
 RVTEST_DATA_BEGIN
-.align 4
+.align 10
```
jordancarlin, line 372: Why is this .align 10 needed before the test strings? This wastes a lot of space in the ELF.

tests/priv/pmp/pmp32/PMPZca/pmpzca_misaligned_na4.S
```diff
@@ -88,6 +88,8 @@ main:
     LI(x4, PMP_REGION_GLOBAL)
     csrw pmpcfg3, x4
 
+// For larger UDB_PMP_GRANUMARITY, NA4 is not supported
+#if UDB_PMP_GRANULARITY == 2
```
jordancarlin: All of these Zca NA4 tests should use the test yaml header instead of the `#if`.

tests/priv/pmp/pmp64/PMPF/pmpf_cfg_wr.S
```diff
@@ -22,7 +22,7 @@
 # REQUIRED_EXTENSIONS:  ['F','Sm']
 # params:
 #   MXLEN: 64
-#   NUM_PMP_ENTRIES: '>0'
+#   UDB_NUM_PMP_ENTRIES: '>0'
```
jordancarlin: The test YAML header should not have the `UDB_*` prefix.
```suggestion
#   NUM_PMP_ENTRIES: '>0'
```

tests/priv/pmp/pmp64/PMPS/pmps_na4_legal_lxwr.S
```diff
@@ -43,6 +43,8 @@ RVTEST_BEGIN                                // Encoded Instruction of nop (64 bi
 
 .macro VERIFICATION_RWX ADDRESS TEST_CASE
 
+    fence.i
```
jordancarlin: Should be using RVMODEL_FENCEI macro.

tests/priv/pmp/pmp64/PMPSm/pmpsm_cfg_A_tor_bot.S
```diff
@@ -101,6 +101,8 @@ RVTEST_BEGIN
     nop
     RVTEST_SIGUPD(x2, x5, x4, a4, \TEST_CASE\()_8, test_8_str)
 
+   fence.i
```
jordancarlin: Use RVMODEL macro.

tests/priv/pmp/pmp64/PMPSm/pmpsm_cfg_A_tor_zero.S
```diff
@@ -58,7 +58,7 @@ RVTEST_BEGIN
     RVTEST_SIGUPD(x2, x5, x4, a4,\TEST_CASE\()_2,test_2_str)               // Signature update
 
     LI(x4, 0xACCE)
-    li a5, 0
+    LI(a5, RVMODEL_ACCESS_FAULT_ADDRESS)
```
jordancarlin: Same comment as the other test. This should still be 0.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
jordancarlin: This still needs to be addressed (or open an issue to address it in a follow up PR). We should test that writing the reserved WARL value returns some legal value. It can be any legal value but it can't return the reserved value that we wrote.

jordancarlin (changes requested):

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_zero.S
UmerShahidengr: This will cause a bit of panic in trap handler in future, trap handler has an existing patch that does not do epc+4 if trap at epc = RVMODEL_ACCESS_FAULT_ADDRESS, for most simulators, RVMODEL_ACCESS_FAULT_ADDRESS = 0, and that patch works by default for this test, but for QEMU, RVMODEL_ACCESS_FAULT_ADDRESS ! = 0 and that patch does not work and gets stuck into an infinite loop while running on QEMU when we test on zero.

UmerShahidengr: Keeping it to zero for simulators where RVMODEL_ACCESS_FAULT_ADDRESS != 0 will require additional patch in trap handler, should we go on that path or just update the test plan that we are testing on either zero or broadly on RVMODEL_ACCESS_FAULT_ADDRESS

jordancarlin: I think testing 0 is important for PMP and is distinct from RVMODEL_ACCESS_FAULT_ADDRESS, so it is probably worth updating the trap handler if it is not too invasive. Let's change this back to 0 in this PR and then if you want to do the trap handler updates in a future PR that works (exclude QEMU from CI for now). Or add the trap handler changes to this PR if they are easy.

tests/priv/pmp/pmp32/PMPS/pmps_na4_legal_lxwr.S
UmerShahidengr: Thanks for the suggestion. Fixed it for all NA4 tests

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_A_tor_zero.S
UmerShahidengr: Fixed it, and updated trap handler. Good discussion. Thank you

tests/priv/pmp/pmp32/PMPS/pmps_na4_legal_lxwr.S
jordancarlin: Looks like you removed the #if from some of the PMPZca NA4 tests but didn't add the YAML constraint.

tests/priv/pmp/pmp32/PMPSm/pmpsm_csr_walk-5.S
UmerShahidengr: It is fixed now, it was updated to all other walk-* tests, but it left there unnoticed.

tests/priv/pmp/pmp32/PMPS/pmps_na4_legal_lxwr.S
UmerShahidengr: That yaml constraint was already there in those tests

tests/priv/pmp/pmp32/PMPSm/pmpsm_napot_legal_lwxr.S
UmerShahidengr: @hamza-1821 did it for coverage checking, he was trying to keep these labels on fixed addresses regardless of the code size, so that he keep that address fixed for hitting his bins

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S
```diff
@@ -173,17 +173,20 @@ main:
     VERIFICATION_RWX    TEST_FOR_EXECUTION_12, test_13
     VERIFICATION_RWX    TEST_FOR_EXECUTION_13, test_14
     VERIFICATION_RWX    TEST_FOR_EXECUTION_14, test_15
-
+//#endif
```
jordancarlin: Looks like a remnant from an earlier change?
```suggestion
```

tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S
UmerShahidengr: This test has worked, so I have removed the comments, previously I was putting UDB_PMP_GRANULARITY very very large because of my lack of understanding.

jordancarlin: There are a few other tests that still have this comment and check. Can it be removed there as well?

UmerShahidengr: For some, it has been removed, for rest of the tests (3 in total), I am masking the tests partially on large grain values, I am opening issues against those tests

jordancarlin: Sounds good.

jordancarlin (approved):
LGTM!

PR #1647, opened 2026-06-04
Apply verbose setting to prepare_dut_outputs

tynan-jdwk:
Using the `VERBOSE` flag can help debug errors during UDB config validation, but it currently doesn't get passed to the `prepare_dut_outputs` function. This step wasn't included when the `VERBOSE` flag was added in #1220.


jordancarlin (approved):

jordancarlin:
Good catch. Thanks for putting this together!

PR #1653, opened 2026-06-05
CP Custom Generation for Vector Crypto

fourth-bit:
This PR includes tests from NIST references for SHA-256, SHA-512, and GCM. As these algorithms don't use a byte substitution, the edge values were very sparse for each of them. These add another set of values to test each algorithm to make the testing more robust. 

There is no coverage because we are just using a set of test vectors.


generators/testgen/scripts/custom/data/gcm_test_vectors_wide.csv
```diff
@@ -0,0 +1,19 @@
+Test Case,K,P,IV,H,Y0,"E(K,Y0)",len(A)||len(C),"GHASH(H,A,C)",C,T,Y1,"E(K,Y1)",X1,Y2,"E(K,Y2)",Y3,"E(K,Y3)",Y4,"E(K,Y4)",X2,X3,X4,A,X5,X6,N1,len({})||len(IV),N2,N3,N4
```
davidharrishmc, line 4: Could you give a reference here to the standard (as well as the mention in the python), and a detailed description of exactly what you are testing with this with reference to certain portions of the example in the standard?

generators/testgen/scripts/custom/cp_custom_nist_sha.py
```diff
@@ -0,0 +1,116 @@
+# SPDX-License-Identifier: Apache-2.0
```
davidharrishmc, line 1: Every place you are using values from a standard, please reference the standard, explain exactly what part of the standard example you are testing, and give enough description for somebody to understand it.  The information should also be printed into the .S file, not just the generator or csv, because many users will only see the .S.

davidharrishmc (changes requested):

davidharrishmc:
This is great!  Will you also be able to add the AES tests from ACT3?  Fine to have them do exactly what was in ACT3, just adjusted to compile in the ACT4 flow.

fourth-bit:
The ACT3 tests for AES are the next thing on my plate.

generators/testgen/scripts/custom/cp_custom_nist_sha.py
fourth-bit: Added comments to the csvs, python, and placed them into the assembly file

fourth-bit:
Added the ACT3 tests to the zvkned

davidharrishmc (approved):

davidharrishmc (approved):

PR #1655, opened 2026-06-05
Add UDB FeatureExtractor M-extension probe in C

Jayant2908:
This PR adds an initial UDB Feature Extractor under tests-dev/priv/UDBFeatureExtractor.

It implements:

- A C-based Feature Extractor built around FeatureExtractorC.S, crt.S, feature_extractor.c, and syscalls.c.
- Detection of supported ISA extensions by executing probe instructions and determining support based on whether they trap (assuming Ssstrict).
- Initial support for detecting the I, E, and M extensions.
- A minimal bare-metal C runtime, local trap handler, lightweight printf implementation, and YAML output generation for implemented_extensions.
- Support for both RV32 and RV64 targets through shared code paths.
- A target-independent Makefile that can be run against different simulator/configuration directories.

Tested with:

- make -f tests-dev/priv/UDBFeatureExtractor/Makefile clean
- make -f tests-dev/priv/UDBFeatureExtractor/Makefile run
- make -f tests-dev/priv/UDBFeatureExtractor/Makefile run-no-M

This serves as the initial framework for automatically extracting ISA extensions and will be extended to support additional extensions and UDB parameters in follow-up work.

Screenshots:-
Screenshot 2026-06-05 011706
Screenshot 2026-06-05 011654


Jayant2908:
Hi @davidharrishmc and @jordancarlin , I have completed the C version of the UDB feature extractor as the next step in this PR. It now builds and runs on Sail with both M enabled and M disabled.
The C version uses `crt.S` for bare-metal startup, `syscalls.c` for minimal `printf` support through `RVMODEL_IO_WRITE_STR`, inline assembly for the encoded `DIV` probe, and a local trap handler/counter to detect whether the probe traps.
Could you please review this when you get a chance? If this approach looks okay, I’ll continue by adding more extension probes and parameter check.

tests/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,64 @@
...
+{
+    return read_trap_count_asm();
+}
+
+/*
+ * Probe M by executing:
+ *     div a2, a0, a1
+ *
+ * Raw encoding:
+ *     div a2, a0, a1 = 0x02b54633
```
davidharrishmc: Why is this raw encoding needed?

tests/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,64 @@
...
+static int probe_m_extension(void)
+{
+    uint32_t before = read_trap_count();
+
+    __asm__ volatile (
+        "li      a0, 10\n"
+        "li      a1, 2\n"
+        ".option push\n"
+        ".option norvc\n"
+        ".word   0x02b54633\n"   // div a2, a0, a1
```
davidharrishmc: Why not just do div a2, a0, a1 instead of .word?

tests/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,64 @@
...
+        ".option pop\n"
+        :
+        :
+        : "memory", "a0", "a1", "a2", "t0", "t1"
```
davidharrishmc: I don't remember the syntax here.  Could you add some comments about what all this does, including why memory and t0 and t1 are needed?

tests/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,64 @@
...
+    );
+
+    uint32_t after = read_trap_count();
+
+    return after != before;
```
davidharrishmc: How about changing the polarity to return true if m the extension is supported.  Then you don't need the ! in !div_trapped.

tests/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,64 @@
...
+}
+
+int main(void)
+{
+    int div_trapped = probe_m_extension();
```
davidharrishmc: Let's give names that will generalize well to other extensions and keep this simple.  How about changing this line to something like 

bool m_supported = check_m_supported()

tests/priv/UDBFeatureExtractor/FeatureExtractor.S
```diff
@@ -0,0 +1,75 @@
...
+#define SIGUPD_COUNT 4
+
+#include "riscv_arch_test.h"
+
+RVTEST_BEGIN
+
+main:
+    // Mtrap_sig stores the current trap-signature write pointer.
+    // If DIV traps, the framework trap handler advances this pointer.
+    LA      s0, Mtrap_sig
```
davidharrishmc: Works for now.  Eventually this may need to generalize for S and VS handlers too, and you may want to keep track of how many of each type of trap has occurred, or what the type of the last trap was.  No need to change until you need that.

tests/priv/UDBFeatureExtractor/FeatureExtractor.S
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: Apache-2.0
```
davidharrishmc: The c version seems to do everything the assembly version does, so is there a reason to keep this one?

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,114 @@
...
+
+
+// Called from feature_extractor.c.
+// Returns current trap count in a0.
+.globl read_trap_count_asm
+.type read_trap_count_asm, @function
+read_trap_count_asm:
+    la      t0, c_trap_count
+#if __riscv_xlen == 64
+    lwu     a0, 0(t0)
```
davidharrishmc: I think it is a safe assumption that the feature extractor takes < 2^31 traps, so you could drop the lwu and have a single code path.  However, I like that you are thinking about supporting both RV32 and RV64 in the shared extractor.

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,114 @@
...
+    lw      a0, 0(t0)
+#endif
+    ret
+.size read_trap_count_asm, . - read_trap_count_asm
+
+// Local trap handler for feature-probe instructions.
+// Handles only expected illegal-instruction traps from 32-bit probe opcodes.
+.align 4
+c_feature_trap_handler:
+    // Only illegal instruction is expected for the current DIV probe.
```
davidharrishmc: Generalize to not be restricted to DIV

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,114 @@
...
+    csrr    t0, mcause
+    li      t1, 2
+    bne     t0, t1, unexpected_trap
+
+    // Count the trap.
+    LA      t0, c_trap_count
+#if __riscv_xlen == 64
+    lwu     t1, 0(t0)
```
davidharrishmc: same comment

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,114 @@
...
+#else
+    lw      t1, 0(t0)
+#endif
+    addi    t1, t1, 1
+    sw      t1, 0(t0)
+
+    // Skip the faulting 4-byte probe instruction.
+    csrr    t0, mepc
+    addi    t0, t0, 4
```
davidharrishmc: Eventually this will need to support compressed instructions as well.  See SssstrictSm for an example of a trap handler that does that.

tests/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Raw encoding:
+ *     div a2, a0, a1 = 0x02b54633
+ *
+ * Returns:
+ *     1 if DIV trapped
+ *     0 if DIV executed successfully
+ */
+ 
+static int probe_m_extension(void)
```
davidharrishmc: Hopefully many or all of your extension probes will be of this form, testing an instruction or CSR access and checking if it faults.  It may be useful to write a check_extension generic function that takes a string with the instruction to test (e.g. "div a2, a0, a1") so then this could be recast as:

bool check_m_supported() {
  return check_extension_supported("div a2, a0, a1");
}

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,114 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 RISC-V International
+
+// FeatureExtractorC.S
+// ACT wrapper for the C version.
+// Sets up local trap counting, enters crt.S, and provides IO/trap helpers.
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Zicsr, Zifencei, Sm]
+# MARCH: rv${XLEN}i_zicsr_zifencei
```
davidharrishmc: could add m and all of the other RVA23 extensions to MARCH so it can compile all instructions without needing .word.

davidharrishmc (changes requested):

davidharrishmc:
This looks very good.  I gave some nuanced comments, but fundamentally this program seems to do the right stuff in the right way.

Let's get this merged.  Then the next steps are to add support for extensions and then for parameters.

For extensions, our priority order is:

* Extensions in RVI20
* Extensions in RVA22
* Extensions in RVB23
* Extensions in RVA23
* Other ratified extensions (can want until after parameters)

I think the idiom in this code can be used to check most extensions by executing a single instruction or CSR access and seeing if it traps.

For parameters, our goal is to capture all UDB parameters.  See https://github.com/riscv/riscv-unified-db/tree/main/spec/std/isa/param with a full list, one per .yaml file.  Many will be easy and some will be difficult to check.

tests/priv/UDBFeatureExtractor/feature_extractor.c
Jayant2908: I originally used the raw encoding so the extractor could be built without requiring M in the global MARCH, ensuring that the only M instruction was the intentional probe.

I’ll switch this to the `div a2, a0, a1` form using a local `.option arch, +m` block. This lets the assembler encode the instruction while still keeping the global MARCH minimal, so the compiler does not emit M instructions elsewhere outside the probe.

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
Jayant2908: Makes sense. I’ll simplify the trap-count loads to lw since the extractor should never approach that many traps.

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
Jayant2908: Agreed. I’ll add a TODO noting that compressed probes will need 2-byte skip support, and I’ll look at SssstrictSm before adding Zca/Zc* probes.

tests/priv/UDBFeatureExtractor/feature_extractor.c
Jayant2908: Sure will add in the latest commit with all the fixes.

tests/priv/UDBFeatureExtractor/feature_extractor.c
Jayant2908: Good point. I agree this could be generalized as more probes are added.
For now I kept the M check explicit to keep the first C version simple and easy to review. When I add the next set of extension probes, I’ll refactor this into a common helper pattern such as `check_extension_supported(...)` / a macro-based wrapper, so each extension-specific function can stay small, e.g. `check_m_supported()` just calling the common probe logic.

tests/priv/UDBFeatureExtractor/FeatureExtractorC.S
Jayant2908: I kept global MARCH minimal to avoid the compiler emitting optional-extension instructions outside the intended probe. For now, I switched this probe to the readable `div a2, a0, a1` form using a local `.option arch, +m` block. As we add more RVA23 probes, I’ll revisit whether it is cleaner to include more extensions in MARCH or keep enabling them locally around each probe.

Jayant2908:
> This looks very good. I gave some nuanced comments, but fundamentally this program seems to do the right stuff in the right way.
> 
> Let's get this merged. Then the next steps are to add support for extensions and then for parameters.
> 
> For extensions, our priority order is:
> 
> * Extensions in RVI20
> * Extensions in RVA22
> * Extensions in RVB23
> * Extensions in RVA23
> * Other ratified extensions (can want until after parameters)
> 
> I think the idiom in this code can be used to check most extensions by executing a single instruction or CSR access and seeing if it traps.
> 
> For parameters, our goal is to capture all UDB parameters. See https://github.com/riscv/riscv-unified-db/tree/main/spec/std/isa/param with a full list, one per .yaml file. Many will be easy and some will be difficult to check.

Thanks David, this is very helpful. I’ve addressed the review comments. Could you please take a look again!?
After this is merged, I’ll start adding extension probes in the priority order you listed. For parameters, I’ll start going through the UDB parameter YAML files and separate them into directly detectable, probe-based, and difficult/manual categories.
Thanks again for the guidance.

Jayant2908:
@davidharrishmc Just following up on this. I had addressed the review comments. Could you please take another look when you get a chance? If this looks good, I’ll use this as the base pattern for adding the remaining extension probes and then move on to the UDB parameter checks.

davidharrishmc (approved):

davidharrishmc:
Not passing recommit or Sail right now.

davidharrishmc:
Stalled pending resolving the copilot feedback and passing CI.

Jayant2908:
> Stalled pending resolving the copilot feedback and passing CI.

Hello @davidharrishmc , Copilot feedbacks are nonblockers and I will fix those. 
I got the C version  fully working standalone with a custom Makefile — it correctly detects M via the DIV probe and prints the YAML output.
The main blocker is that when I tried hooking the C version into the top-level make (the act framework), I hit a structural issue: act discovers tests via rglob("*.S") and builds each .S file as a single, self-contained unit against the shared link.ld — it has no support for multi-file C builds or custom linker scripts/symbols (__global_pointer$, __bss_start/end, _stack_top, main, etc., which the C runtime needs).
So crt.S/FeatureExtractorC.S either get picked up individually by act and fail to link (missing those symbols + no main), or I have to rename them away from .S, which then breaks the custom build too.

I don't see an existing mechanism in act to exclude files/dirs or support multi-file tests. Two options I see: either act needs a small extension for multi-file/custom-linker tests, or we keep the C version as a separate, manually-built artifact outside act's tests/ discovery for now.

Could you please provide any guidance in this!?

davidharrishmc:
We will want to draw on the ACT4 suite infrastructure, especially the config files and rvmodel macros.  These might change over time, so there are benefits to being part of the act framework.  If you can make a small extension for multifile while still taking advantage of the information in the config directory link.ld, that would be great.

Jayant2908:
Thanks @davidharrishmc , Sounds good. Putting this on draft for now — going to push a new PR to extend act to support multi-file C tests (new EXTRA_SOURCES key in the test header, plus minor link.ld additions for __global_pointer$/.bss/.stack) instead of using a separate custom Makefile, so we stay in sync with the shared config infrastructure. Will ping once it builds cleanly via make.

jordancarlin:
@Jayant2908 I hope you haven't put too much work into the updated build system yet, but I think we are going to need to go in a different direction. The existing ACT build system requires a config file as an input, so it does not make sense to add the feature extractor to it because that means the config generator would require a completed config.

For now, let's keep this as a separate test in `tests-dev/feature-extractor` (or similar) with a standalone Makefile. Once it is more complete, we can decide how to integrate it into the larger ACT flow.

Jayant2908:
> Jayant Malvi (@Jayant2908) I hope you haven't put too much work into the updated build system yet, but I think we are going to need to go in a different direction. The existing ACT build system requires a config file as an input, so it does not make sense to add the feature extractor to it because that means the config generator would require a completed config.
> 
> For now, let's keep this as a separate test in `tests-dev/feature-extractor` (or similar) with a standalone Makefile. Once it is more complete, we can decide how to integrate it into the larger ACT flow.

Yeah that makes sense — will move it to tests-dev/feature-extractor with the standalone Makefile for now in this PR.

Jayant2908:
@davidharrishmc Fixed small bugs as well as moved the UDB feature extractor into tests-dev/priv folder. Both sail and pre-commit should work fine. Please take a look when you get time. Thank you!

tests-dev/priv/UDBFeatureExtractor/crt.S
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 RISC-V International
+//
+// crt.S  –  Minimal bare-metal C runtime for UDBFeatureExtractor (C version).
+
+##### START_TEST_CONFIG #####
+# REQUIRED_EXTENSIONS: [I, Zicsr, Zifencei, Sm]
+# MARCH: rv${XLEN}i_zicsr_zifencei
+##### END_TEST_CONFIG #####
```
jordancarlin: Why does this include the test YAML header? `crt.S` will never run on its own.

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,120 @@
...
+
+CC      := riscv64-unknown-elf-gcc
+OBJDUMP := riscv64-unknown-elf-objdump
+SIM     := sail_riscv_sim
+
+ARCH := rv64i_zicsr_zifencei
+ABI  := lp64
+
+# riscv_arch_test.h lives in tests/env; rvtest_config.h lives in the per-target config dir (e.g. config/sail/sail-RVA23S64).
+# Both must be on the include path.
```
jordancarlin: `rvtest_config.h` is generated from the UDB config. The point of the Feature Extractor is to generate the UDB config, so you can't use `rvtest_config.h` as part of that.

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,120 @@
...
+    -DTEST_FLEN=32 \
+    -DSIGNATURE \
+    -Wl,--no-warn-rwx-segments \
+    -Wl,-e,rvtest_entry_point \
+
+SAIL_FLAGS := $(COMMON_FLAGS) \
+	-I$(SAIL_CONFIG) \
+	-I$(CVW_WORK_CONFIG) \
+	-I$(CVW_CONFIG) \
+	-I$(ARCH_TEST_INC)
```
jordancarlin: Why are there CVW variables in the sail flags?

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,120 @@
...
+
+# C-version extras:
+#   -std=gnu11   required for register-asm pin syntax (__asm__("a0"))
+#   -ffreestanding  no hosted-environment assumptions
+#   -O1 -g       mild optimisation + debug info
+C_EXTRA := -ffreestanding -O1 -g -std=gnu11 \
+           -fno-builtin -fno-builtin-printf -fno-builtin-puts
+
+YAML_C_OUT   := $(BUILD_DIR)/output-c.yaml
+YAML_C_NOM   := $(BUILD_DIR)/output-c-no-M.yaml
```
jordancarlin: Can we pick a default output filename that more closely aligns with what a user would actually want? Maybe `<config_name>.yaml` or `extracted_config.yaml` or similar.

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,120 @@
...
+    $(TEST_SRC)/feature_extractor.c \
+    $(TEST_SRC)/syscalls.c
+
+.PHONY: all sail-c sail-c-no-M cvw dump-c clean help
+all: sail-c
+
+$(BUILD_DIR):
+	mkdir -p $@
+
+# ── Linker-script generation───────────────────────────────────────
```
jordancarlin: Generating a linker script using Makefile syntax seems prone to errors. Can you just include a premade linker script for now?

tests-dev/priv/UDBFeatureExtractor/Makefile
jordancarlin, line 1: Can you generalize this Makefile so there is a single build target that uses variables defined at the top instead of lots of different rules for CVW and Sail?

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,120 @@
...
+
+# ────────────────────────────────────────────────────────────────────────────
+clean:
+	rm -rf $(BUILD_DIR)
+
+help:
+	@echo "make -f tests-dev/priv/UDBFeatureExtractor/Makefile <target>"
+	@echo ""
+	@echo "C version: sail-c  sail-c-no-M  dump-c"
+	@echo "shared:    cvw  clean"
```
jordancarlin: Not sure what shared means here.

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,143 @@
...
+// Called from syscalls.c.
+// a0 contains a null-terminated string pointer.
+.globl arch_write_str_asm
+.type arch_write_str_asm, @function
+arch_write_str_asm:
+    addi    sp, sp, -32
+    SREG    ra, 0(sp)
+    SREG    t0, 8(sp)
+    SREG    t1, 16(sp)
+    SREG    t2, 24(sp)
```
jordancarlin: Is saving t0-t2 needed? The temporary registers are not guaranteed to be saved across function calls in the standard ABI.

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
jordancarlin, line 1: It might make the overall test easier to understand if you move some of these functions into C using inline assembly.

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,143 @@
...
+
+    // Count the trap.
+    LA      t0, c_trap_count
+#if __riscv_xlen == 64
+    lw     t1, 0(t0)
+#else
+    lw      t1, 0(t0)
+#endif
+    addi    t1, t1, 1
+    sw      t1, 0(t0)
```
jordancarlin: Why is a count of the traps needed? Can we just simplify this to just indicate if we saw a trap or not?

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,82 @@
...
+{
+    bool m_supported = check_m_supported();
+
+    if (read_unexpected_trap_asm() != 0) {
+        printf("error: unexpected trap occurred\n");
+        return 1;
+    }
+
+    printf("implemented_extensions:\n");
+    printf("  - { name: I, version: '= 2.1' }\n");
```
jordancarlin: I isn't always supported. This should test an instruction accessing one of the upper 16 registers to check for I/E.

jordancarlin (changes requested):

tests-dev/priv/UDBFeatureExtractor/Makefile
Jayant2908: Removed — replaced the help text with clearly labeled targets (run, run-no-M, dump, clean). "shared" was a leftover label from when the Makefile had separate sail/CVW targets and doesn't apply anymore.

tests-dev/priv/UDBFeatureExtractor/Makefile
Jayant2908: You're right — I've addressed this by providing a minimal stub rvtest_config.h in the test directory itself (tests-dev/priv/UDBFeatureExtractor/rvtest_config.h). This stub only defines the bare minimum needed for the ACT env headers to assemble (UDB_MXLEN and UDB_TRAP_ON_UNIMPLEMENTED_INSTRUCTION) — it doesn't use any real UDB config values. The -I$(TEST_SRC) include path ensures this stub is picked up before any generated config. The feature extractor itself probes and detects all features at runtime rather than relying on any config defines.

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,119 @@
...
+
+REPO_ROOT := $(shell git rev-parse --show-toplevel 2>/dev/null || pwd)
+TEST_SRC  := $(REPO_ROOT)/tests-dev/priv/UDBFeatureExtractor
+BUILD_DIR := $(REPO_ROOT)/work/UDBFeatureExtractor
+
+# ── Per-target configuration (override on command line) ──────────────────────
+CONFIG_DIR ?= $(REPO_ROOT)/config/sail/sail-RVA23S64
+SIM        ?= sail_riscv_sim
+XLEN       ?= 64
+ABI        ?= lp64
```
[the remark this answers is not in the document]
Jayant2908: The I/E probe is intentionally best-effort — we build with rv64i/rv32i because the Feature Extractor itself must run (it can't be built targeting E if we don't yet know whether the core is E). The probe uses .word encoding if needed to ensure x16 is emitted regardless. On a true E-only core, the binary may trap earlier, but that itself confirms E — the unexpected trap handler will catch it. This is a known limitation and will be documented.

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,108 @@
...
+
+c_done:
+    RVTEST_CODE_END
+
+
+RVTEST_DATA_BEGIN
+
+.align 3
+saved_mtvec:
+    .dword 0
```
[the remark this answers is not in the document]
Jayant2908: saved_mtvec uses .dword because this tool currently only targets RV64 (XLEN=64). If RV32 support is added later this will be revisited. Not changing for now

tests-dev/priv/UDBFeatureExtractor/syscalls.c
```diff
@@ -0,0 +1,94 @@
...
+    while (*fmt != '\0') {
+        if (*fmt != '%') {
+            const char *span_start = fmt;
+            while (*fmt != '\0' && *fmt != '%')
+                fmt++;
+            count += (int)(fmt - span_start);
+            /* Write span using put_char to avoid VLA stack allocation. */
+            const char *p = span_start;
+            while (p < fmt)
+                put_char(*p++);
```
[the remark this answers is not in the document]
Jayant2908: The output is short YAML strings (under 100 chars total). Per-character overhead is negligible for this use case. Keeping simple code over premature optimization.

davidharrishmc:
Let’s make the tool general enough for both RV32 and RV64.  Use the LREG/SREG macros, which expand to word or double depending on the architecture.  There should be relatively little other work to support both.

davidharrishmc:
Why not compile for E?  Then it will run on both.

allenjbaum:
An RVE core is not guaranteed to trap if 16 is encoded; the encoding is
RSVD, which means ANYTHING might occur: alias to X0, halt and catch fire,
etc.
You cannot test for an E version using this approach.

jordancarlin:
> An RVE core is not guaranteed to trap if 16 is encoded; the encoding is RSVD, which means ANYTHING might occur: alias to X0, halt and catch fire, etc. You cannot test for an E version using this approach.

The FeatureExtractor is only designed to work on a system that implements Ssstrict, in which case it is guaranteed to trap. Trying to automatically extract supported extensions on a system without Ssstrict seems like an incredibly difficult (if not impossible) problem.

Jayant2908:
@davidharrishmc @jordancarlin Sorry for the delay in updating this PR—I was busy settling into a new place and juggling a few other commitments. I'm fully back now and hoping to get this wrapped up as soon as possible so I can move forward with the remaining tasks.

I've addressed the latest review comments and made a few additional improvements:

- Added support for both RV32 and RV64 in the FeatureExtractor.
- Kept E extension detection based on Jordan's clarification that the FeatureExtractor assumes Ssstrict, where probing reserved encodings is guaranteed to trap.
- Cleaned up a few portability and documentation issues along the way.

Could you please take another look when you have a chance? If everything looks good, I'd like to use this as the baseline and continue adding support for the remaining extensions and UDB parameters. Thanks!

davidharrishmc:
Glad you are settled and getting going on the feature extractor again.  I’m back from travels as well and will review this shortly.

We also are gravitating toward simplifying the parameter set as much as possible.  For example, we will not test the behavior of a WARL field when a reserved value is written to it, and will not test the behavior of WPRI fields.

This reduces the problem to identifying which named extensions are supported and which parameters are supported.  For the near term, let’s focus on mandatory and optional extensions in RVA23, and parameters from https://github.com/riscv/riscv-unified-db/tree/main/spec/std/isa/param that apply to RVA23 extensions.

Our next release goal is July 24.  How far do you think you’ll be able to get by that point?

Jayant2908:
@davidharrishmc Good to know you're back as well!

The simplified parameter scope definitely helps. For July 24, I’ll prioritize the mandatory and optional RVA23 extension probes first, then move to the applicable RVA23-related parameters from the UDB list.
Realistically, I think I can get a solid portion of the RVA23 extension checks done by then, along with some of the straightforward applicable parameters if time permits. The stretch goal would be covering most of the RVA23 extension set and starting a clean parameter-classification list.

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,127 @@
...
+
+    return !read_trap_flag();
+}
+
+int main(void)
+{
+    bool i_supported = check_i_supported();
+    bool m_supported = check_m_supported();
+
+    if (read_unexpected_trap() != 0) {
```
davidharrishmc: No longer needed because it can never return here.

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,110 @@
...
+    addi    t0, t0, 4
+    csrw    mepc, t0
+
+    mret
+
+unexpected_trap:
+    LA      t0, c_unexpected_trap
+    li      t1, 1
+    sw      t1, 0(t0)
+    j       rvmodel_halt_fail
```
davidharrishmc, line 82: It would be nice to print a failure message such as "Unexpected trap.  mcause = xxx.  mepc = xxx.  mtval = xxx"  However, that feature can be deferred for a later version.

davidharrishmc (changes requested):

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
Jayant2908: Good point. Since unexpected traps now fail through rvmodel_halt_fail, this check is no longer reachable. I've removed it.

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
Jayant2908: Agreed. That would definitely make debugging easier. I'll defer that improvement to a follow-up so this PR stays focused on the probing framework.

davidharrishmc (approved):
My comments have been fully addressed.  Still pending a final review from Jordan.

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
```diff
@@ -0,0 +1,110 @@
...
+// a0 contains a null-terminated string pointer.
+.globl arch_write_str_asm
+.type arch_write_str_asm, @function
+arch_write_str_asm:
+    addi    sp, sp, -16
+    SREG    ra, 0(sp)
+    RVMODEL_IO_WRITE_STR(t0, t1, t2, a0)
+    LREG    ra, 0(sp)
+    addi    sp, sp, 16
+    ret
```
jordancarlin, line 53: If this only writes to t0-t3 and a0, why does it need to save/restore from the stack?

tests-dev/priv/UDBFeatureExtractor/rvtest_config.h
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 RISC-V International
+//
+// Minimal stub rvtest_config.h for UDBFeatureExtractor.
+// The feature extractor runs before the real UDB config is generated.
+// Only the defines required by the ACT env headers are provided here.
```
jordancarlin, line 6: Rather than needing hacks like this stub, I wonder if you are better off just not including `riscv_arch_test.h` in the UDBFeatureExtractor. There is no need to set up the standard trap handler since you will overwrite it and most of the boot code is not needed since this will run in M mode anyway. I think you can get away with just `#include rvmodel_macros.h` which will simplify this whole flow. That also lets you pull a lot of the macros out of `FeatureExtractorC.S`.

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,115 @@
...
+ *     true  if I is supported
+ *     false if only E is supported
+ */
+static bool check_i_supported(void)
+{
+    reset_trap_flag();
+
+    __asm__ volatile (
+        ".option push\n"
+        ".option norvc\n"
```
jordancarlin, line 45: Is `.option norvc` needed? C is not in the `march` string, so the compiler will not use any compressed instructions anyway.

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,125 @@
...
+ifeq ($(XLEN),32)
+ABI        ?= ilp32
+else
+ABI        ?= lp64
+endif
+
+CC      := riscv64-unknown-elf-gcc
+OBJDUMP := riscv64-unknown-elf-objdump
+
+ARCH := rv$(XLEN)i_zicsr_zifencei
```
jordancarlin, line 33: Should we compile with `rv$(XLEN)e...` so that it only uses x0-x15 except when we are explicitly testing for the presence of the I extension? Otherwise an E core will trap all over the place and not be able to use this.

tests-dev/priv/UDBFeatureExtractor/Makefile
```diff
@@ -0,0 +1,125 @@
...
+    -Wl,--no-warn-rwx-segments \
+    -Wl,-e,rvtest_entry_point
+
+CFLAGS := $(COMMON_FLAGS) \
+	-I$(TEST_SRC) \
+    -I$(CONFIG_DIR) \
+    -I$(ARCH_TEST_INC)
+
+# C-version extras:
+#   -std=gnu11      required for inline asm syntax
```
jordancarlin, line 53: Pretty sure asm syntax is supported in gnu99.

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,115 @@
...
+        "li      a1, 2\n"
+        ".option push\n"
+        ".option arch, +m\n"
+        ".option norvc\n"
+        "div     a2, a0, a1\n"
+        ".option pop\n"
+        :
+        :
+        : "memory", "a0", "a1", "a2", "t0", "t1"
+    );
```
jordancarlin, line 92: If we're using C anyway, it might make more sense to do something like this:
```c
long dividend = 10;
long divisor = 2;
long result;

__asm__ volatile (
    ".option push\n"
    ".option arch, +m\n"
    "div %0, %1, %2\n"
    ".option pop\n"
    : "=r"(result)
    : "r"(dividend), "r"(divisor)
);
```

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,115 @@
...
+
+    if (i_supported) {
+        printf("  - { name: I, version: '= 2.1' }\n");
+    } else {
+        printf("  - { name: E, version: '= 2.0' }\n");
+    }
+
+    if (m_supported) {
+        printf("  - { name: M, version: '= 2.0' }\n");
+    }
```
jordancarlin, line 112: Thinking about scaling this to many extensions, it might make sense to have the `check_{i/m}_supported` functions print the relevant lines of the YAML before returning. Otherwise, we're going to end up with a giant list of if statements here.

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,115 @@
...
+{
+    bool i_supported = check_i_supported();
+    bool m_supported = check_m_supported();
+
+    printf("implemented_extensions:\n");
+
+    if (i_supported) {
+        printf("  - { name: I, version: '= 2.1' }\n");
+    } else {
+        printf("  - { name: E, version: '= 2.0' }\n");
```
jordancarlin, line 107: It might be worth a helper function that takes an extension name and version and prints the relevant line of YAML so all of the YAML syntax doesn't need to be repeated for each function. Something like the following:
```c
void print_extension(const char *name, const char *version) {
    printf("  - { name: %s, version: '= %s' }\n", name, version);
}
```

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
```diff
@@ -0,0 +1,115 @@
...
+        ".option norvc\n"
+        "addi x16, x16, 0\n"
+        ".option pop\n"
+        :
+        :
+        : "memory", "a6", "t0", "t1"
+    );
+
+    return !read_trap_flag();
+}
```
jordancarlin, line 54: Eventually we will probably want a more generic version of this (`check_extension_supported`) so that the setup doesn't need to be repeated for each extension, but that can come later.

tests-dev/priv/UDBFeatureExtractor/rvtest_config.h
jordancarlin: This would also help simplify the Makefile a lot.

jordancarlin (changes requested):

tests-dev/priv/UDBFeatureExtractor/FeatureExtractorC.S
Jayant2908: Good point. RVMODEL_IO_WRITE_STR doesn't modify ra, so the save/restore was unnecessary. I've removed it.

tests-dev/priv/UDBFeatureExtractor/feature_extractor.c
Jayant2908: Yes Since C isn't enabled in the build -march, norvc is redundant. I've removed it.

tests-dev/priv/UDBFeatureExtractor/rvtest_config.h
Jayant2908: That makes sense. The original implementation used riscv_arch_test.h to integrate with the existing ACT flow, but as the feature extractor has evolved it now relies on only a small subset of that functionality. I’ll refactor this PR to include only `rvmodel_macros.h`, keep the local M-mode startup/trap handling, and remove the ACT test-header dependency. That should also simplify the Makefile and make this a cleaner baseline for the remaining extension probes.

jordancarlin:
I see some of my comments were marked as resolved, but I don't see any changes pushed. Did you forget to push?

jordancarlin:
@Jayant2908 it's been about a month since there has been any update on this PR. Are you still working on it? If so, can you push an update addressing the open comments from my last review?

PR #1670, opened 2026-06-08
Update Sail Configs to 0.12

fourth-bit:
Current Failures with Known Root Causes:
- Non-segmented indexed vector stores were broken: This is a sail bug, and a PR has been opened with a fix
- ExceptionsVx: The length suite sig update macro is incorrectly emitted, leading to traps at different mepcs due to the differences in sig update macros
- ExceptionsVls: Testgen emitted an illegal instruction that is currently not caught by sail


config/cores/cve2/cv32e20/sail.json
```diff
@@ -185,21 +239,33 @@
...
     },
     "clint": {
+      "supported": true,
       // This must be in a suitable memory region (see `memory.regions`).
       "base": 33554432,
       "size": 786432
     },
     // Simple MMIO device to generate interrupts for testing purposes.
     "simple_interrupt_generator": {
+      "supported": false,
```
karabambus: All cve2/4 test bench's have this peripheral it should be enabled.

pmundkur:
Some of the Ssstrict failures I noticed with 0.12 sail-riscv and riscv-arch-test as of 2026-05-29 were due to these bug fixes:
https://github.com/riscv/sail-riscv/pull/1712  (add.uw)
https://github.com/riscv/sail-riscv/pull/1690 (vbrev8 and similar)
https://github.com/riscv/sail-riscv/pull/1698 (vwadd.vv and similar)
The tests might have been fixed since, I haven't checked.

config/cores/cve2/cv32e20/sail.json
fourth-bit: Thanks for the feedback.

fourth-bit:
Looks like CI is passing now, so it is ready for a review.

The only test that had to be excluded was S in spike. Part of the S tests is walking a one across all csrs, this includes the PMM bits of senvcfg. The clear test for bit 33 results in the PMM bits taking on the value 0b01, which is reserved. Spike and sail mismatch here: PMM = 0b00 or PMM = 0b11, and this mismatch is legal, so we are excluding it from CI for now.

With the changes to wfi in sail, the InterruptsS and InterruptsU tests are able to pass whisper, so they are back in CI.

davidharrishmc:
Can you find a good solution to S?  Perhaps we need to drop walking 1s or exclude that PMM case?  It would be good to keep Spike in CI.  Adjust the test plan as needed to match.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -516,7 +516,8 @@ def _generate_scsr_tests(test_data: TestData) -> list[str]:
     )
 
     for csr in csrs:
-        lines.extend(csr_walk_test(test_data, csr, covergroup, coverpoint))
+        unspecified_clears = [33] if csr[0] == "senvcfg" else []
+        lines.extend(csr_walk_test(test_data, csr, covergroup, coverpoint, unspecified_clears=unspecified_clears))
```
jordancarlin: I think this is a bit more subtle. We shouldn't fully skip this case. We still want to test writing 0b01 to the PMM field and then we should be checking that we read back a legal value (anything other than 0b01 in this case).

fourth-bit: I reverted these changes, and confirmed spike passes with your PR

jordancarlin:
@karabambus Can you review the changes to the CVE2/4 configs and open another PR if any tweaks are needed?

jordancarlin (approved):

Issue #1675, opened 2026-06-09
Possible vector test missing features


davidharrishmc:
@fourth-bit I asked claude for a sanity check of whether all vector features are tested.  It made these suggestions.  Take with a grain of salt, but perhaps some are useful.
  
 5. Segment load/store field interleaving is unchecked — HIGH  vlseg2..8 / vlsseg / vluxseg / vloxseg de-interleave NF fields into NF consecutive register groups. The tests check only field 0 (register vd) — even the _LEN path checks all elements of vd but never vd+1.  The entire de-interleave purpose is 0% result-verified. Stores have a memory-readback path (good) but readback runs at vl=1, so stride/index/segment placement past element 0 is unconfirmed.

  4. Fault-only-first vl truncation is never exercised — HIGH
  vleNff faults are constructed at element 0 (address 0), which raises a normal exception with no truncation. The defining behavior — fault at element ≥1 truncates vl and keeps preceding elements — is never
  triggered, and the post-load vl CSR is never checked. Currently indistinguishable from a plain vle.

  5. Unordered FP reductions may over-constrain (false-failure risk) — HIGH, verify
  vfredusum/vfwredusum reduce in implementation-defined order. The multi-element cr_vl_lmul cases reference-check element 0 of the result against Spike's specific reduction tree. A conformant DUT with a
  different valid tree would produce different rounding and fail. vl=1 cases are safe (order-independent). Worth confirming the multi-element operands are constructed to be rounding-order-invariant; if not,
  these cases are stricter than the spec.

  By-design simplifications that still leave blind spots (mostly "known" per the CTP docs)

  6. Permute/mask edge cases checked at element 0 only — MEDIUM. vslideup/down, vslide1*, vrgather[ei16], vcompress, viota.m, vid.v, vmsbf/vmsif/vmsof, mask-logicals: the relocation/sequence/boundary is
  verified only in the ~50 _LEN cases per file; the hundreds of per-operand/edge cases check element 0 — which for vid (always 0), viota (trivial), and slides is the least informative position. A wrong-position
  bug that leaves element 0 correct is caught only if it also perturbs a random _LEN case.

  7. Whole-register loads vl{2,4,8}re checked at element 0 only — MEDIUM. Inconsistent: the register moves vmv{1,2,4,8}r.v correctly use _LEN at matching LMUL and fully verify the group; the loads don't.

[...]

PR #1682, opened 2026-06-11
Vls Improvements from #1675

fourth-bit:
This addresses two areas for improvement identified in #1675
1. Whole register load/stores now always use the length suite sig update macro to check the whole register, even for base suite tests
2. Segmented load/store operations check all nf register groups


davidharrishmc (approved):
LGTM

PR #1697, opened 2026-06-12
ZawrsSU and ZawrsSm Coverpoints 

Teemooooooooo:
ZawrsSU and ZawrsSm Coverpoints written


coverpoints/unpriv/ZawrsSU_coverage.svh
```diff
@@ -0,0 +1,126 @@
...
+    }
+    mstatus_mie_one: coverpoint ins.prev.csr[CSR_MSTATUS][3] {
+        bins one = {1};
+    }
+
+    mip_mtip_one: coverpoint ins.current.csr[CSR_MIP][7] {
+        bins one = {1};
+    }
+
+    mstatus_sie: coverpoint ins.prev.csr[CSR_MSTATUS][1] {
```
fourth-bit: You have these ifdefed below in the crosses, but these are going to collect coverage in all cases, so they should get an ifdef too. Same comment for the hstatus coverpoint.

coverpoints/priv/ZawrsSm_coverage.svh
```diff
@@ -0,0 +1,86 @@
...
+        wildcard bins zeros = {16'b????0?0?0?0?0?0?}; // zero in all 6 interrupt enable bits
+    }
+    mie_mtie_one: coverpoint ins.current.csr[CSR_MIE][7] {
+        bins one = {1}
+    }
+
+
+
+    // main coverpoints
+    cp_wrs_sto_timeout:     cross priv_mode_m, wrs_sto, mstatus_tw, mstatus_mie_zero, mie_zeros, lr_w;
```
davidharrishmc, line 84: Putting the lr_w in coverpoint requires that the was be immediately proceeded by the lr.  Fine to leave it if that work, but if you need to drop the cross to make the test case easier to write, that's also ok.

coverpoints/priv/ZawrsSU_coverage.svh
```diff
@@ -0,0 +1,129 @@
...
+    cp_wrs_nto_timeout:     cross priv_mode_s_u, mstatus_tw_one, mstatus_mie_zero,
+                        `ifdef S_SUPPORTED
+                            mstatus_sie_zero,
+                        `endif
+                        mie_zeros, wrs_nto, lr_w;
+
+    // if H supported
+    `ifdef H_SUPPORTED
+        cp_wrs_nto_timeout_h:   cross priv_mode_vs_vu, mstatus_tw, mstatus_mie_zero,
+                        `ifdef S_SUPPORTED
```
davidharrishmc, line 142: S is always supported if H is supported so you can drop this ifdef.

coverpoints/priv/ZawrsSm_coverage.svh
davidharrishmc, line 84: These have a _m suffix in the testplan.  Suggest dropping m from testplan to match coverpoint.

davidharrishmc (changes requested):
These coverpoints look great, and are easy to follow from the test plans.  Some minor comments and then this is good to go.

coverpoints/priv/ZawrsSU_coverage.svh
```diff
@@ -0,0 +1,129 @@
...
+        }
+    `endif
+
+
+
+
+
+
+    // main coverpoints
+    cp_wrs_sto_timeout:     cross priv_mode_s_u, wrs_sto, mstatus_tw, mstatus_mie_zero,
```
jordancarlin: priv_mode_s_u will never get coverage for S when S is not supported.

coverpoints/priv/ZawrsSU_coverage.svh
```diff
@@ -0,0 +1,129 @@
...
+            // autofill 0/1
+        }
+        mstatus_sie_zero: coverpoint ins.prev.csr[CSR_MSTATUS][1] {
+            bins zero = {0};
+        }
+    `endif
+    mie_zeros: coverpoint ins.current.csr[CSR_MIE][15:0] {
+        wildcard bins zeros = {16'b????0?0?0?0?0?0?}; // zero in all 6 interrupt enable bits
+    }
+    mie_mtie_one: coverpoint ins.current.csr[CSR_MIE][7] {
```
jordancarlin: Doesn’t have to happen in this PR since usage is still inconsistent right now, but we are trying to move towards the get_csr_val function instead of directly accessing the CSR data structure.

jordancarlin (dismissed):

coverpoints/priv/ZawrsSm_coverage.svh
Teemooooooooo: I was planning to execute the wrs immediately after a lr_w so it should be fine

Teemooooooooo: I added in the _m suffix in the test plan so that it does not have the same exact name as the coverpoints in ZawrsSU. Should i add the _m suffix to the coverpoints for ZawrsSm or should I just drop it if it does not matter for the ZawrsSm and ZawrsSU coverpoints to look the same

jordancarlin: They are in different covergroups already, so it is fine for them to be the same.

coverpoints/priv/ZawrsSm_coverage.svh
```diff
@@ -0,0 +1,97 @@
...
+        bins zeros = {0}; // zero in all 6 interrupt enable bits
+    }
+    mie_mtie_one: coverpoint (get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mie", "mtie")) {
+        bins one = {1}
+    }
+
+    mip_ones: coverpoint ({get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "mtip"),
+                           get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "meip"),
+                           get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "msip"),}){
+        bins ones = {1};
```
davidharrishmc: I think you meant 3'b111

coverpoints/priv/ZawrsSU_coverage.svh
```diff
@@ -0,0 +1,162 @@
...
+
+    mip_ones: coverpoint ({get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "mtip"),
+                           get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "meip"),
+                           get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "msip"),
+                           `ifdef S_SUPPORTED
+                           get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "stip"),
+                           get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "seip"),
+                           get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "mip", "ssip")
+                           `endif}){
+        bins ones = {1};
```
davidharrishmc: {`1}.  This is a weird one with 3 or 6 bits.  You could also make the coverpoint always 6 bits and make the bin an `ifdef of either 111000 or 111111.

davidharrishmc (changes requested):

davidharrishmc (approved):

davidharrishmc (approved):

davidharrishmc (approved):

PR #1699, opened 2026-06-12
ExceptionsVx and ExceptionsVls Passing on all sims except Imperas

fourth-bit:
This PR includes fixes to testgen when relevant, and lots of fixes to the config files for the sims. The indexed load/stores stress the memory map configurations, and pointed out a lot of mismatches between the sim setup and the sail setup. The most significant difference was between qemu's memory map and sail's default memory map. The sail config now uses the same setup (roughly because sail does not accept memory regions with borders that are not page aligned), which was taken from [qemu/hw/riscv/virt.c:83](https://github.com/qemu/qemu/blob/193963218accbcf65f28ab1efa773ac6d1cc634c/hw/riscv/virt.c#L83). 

All non-imperas tests pass on the version of sail built from master, not Sail 0.12. Notably, we require the fix to non-segmented indexed loads that came after Sail 0.12 for all simulators. Imperas fails ExceptionsVx on this version of sail because of a difference in how the legal vstart values are calculated for scalar move instructions. Both implementations are legal, but sail doesn't have the ability to match imperas's behavior.


config/imperas/imperas-rv32gcv/imperas.ic
```diff
@@ -83,4 +83,4 @@
 --callcommand iss/cpu0/setPMA -lo 0x0002000000 -hi 0x000200FFFF -attributes " rw-aA- 1248 " # CLINT
 --callcommand iss/cpu0/setPMA -lo 0x000C000000 -hi 0x000FFFFFFF -attributes " rw-aA- --4- " # PLIC
 --callcommand iss/cpu0/setPMA -lo 0x0010010000 -hi 0x0010010007 -attributes " rw-aA- 1248 " # UART0
---callcommand iss/cpu0/setPMA -lo 0x0080000000 -hi 0x008FFFFFFF -attributes " rwx--- 1248 " # UNCORE_RAM
+--callcommand iss/cpu0/setPMA -lo 0x0080000000 -hi 0x00FFFFFFFF -attributes " rwx--- 1248 " # UNCORE_RAM
```
davidharrishmc: Why is this necessary?  Do tests need to consume this much memory?

config/qemu/qemu-rv32-max/sail.json
```diff
@@ -279,16 +342,257 @@
...
         },
-        "size": {
-          "len": 64,
-          "value": "0x08000000"
+        "include_in_device_tree": true
+      },
+
+      // ── PCIe PIO (I/O port space) ─────────────────────────────────────────────
+      {
+        "base": { "len": 64, "value": "0x03000000" },
```
davidharrishmc: Why do we need all of these peripheral memory definitions?  It is harmless to have them, but I hope we aren't accessing them for vector tests.

config/whisper/whisper-rv32-max/sail.json
```diff
@@ -392,7 +392,7 @@
       // The maximum index EEW for indexed vector addressing mode, as
       // a power of 2.  This must be at least 3 but must not exceed
       // the supported ELEN.
-      "max_index_eew_exp": 5,
+      "max_index_eew_exp": 6,
```
davidharrishmc, line 395: Are there any configs now where this is still 5?  Should we be trying to test 5 somewhere, or save that for later parameter testing?

config/whisper/whisper-rv64-max/whisper.json
```diff
@@ -21,8 +21,20 @@
...
         // Make address range 0 to 1023 non readable to support ACCESS_FAULT_ADDR in rvtest_config.h
         "attribs": [],
-        "high": 1023,
+        "high": "0x1FFFFFF",
         "low": 0
+      },
+      {
+        // Space between MMIO and RAM Should Also be Blocked
+        "attribs": [],
+        "high": "0x7FFFFFFF",
```
davidharrishmc: Why do we need all this? Hope tests are never accessing this address space.

davidharrishmc (commented):
Looks good, but some questions about memory map and config files.

config/imperas/imperas-rv32gcv/imperas.ic
fourth-bit: Its necessary to match the sail config, and is just the default value from sail. Without this, there are different load access faults because each simulator has read/write access to a different space. The indexed load/store tests in ExceptionsVls write to random points in memory, so this cleans up a mismatch between the simulators.

config/qemu/qemu-rv32-max/sail.json
fourth-bit: Same comment as above. These are all regions that qemu has read/write access to, and the indexed load/stores ended up writing to these regions when random inputs were given to it.

config/whisper/whisper-rv32-max/sail.json
fourth-bit: We use 5 in the spike and sail configs for rv32.

config/whisper/whisper-rv64-max/whisper.json
fourth-bit: This is meant to mark that part of the address space as inaccessible. Some of the indexed loads were touching it.

config/imperas/imperas-rv32gcv/imperas.ic
jordancarlin: We should be modifying the Sail memory map to match the DUT, not changing the DUT to match Sail.

More importantly though, we can’t access random addresses in memory. Some of those addresses will contain MMIO and may do unexpected things when we read/write to them. The tests should only be accessing main memory. Let’s discuss this further in Slack.

davidharrishmc: Agreed that we can't hit random points in memory.  Got to constraint it to safe places in main memory.  Perhaps just in a narrow region of main memory, such as scratch, or else require a bigger block and document what is required.

fourth-bit: I changed the indexed loads to only access reasonable values. Do we still want to have these changes to memory maps, or should I roll them back / spin them out into another PR?

davidharrishmc:
My feeling is that these detailed Sail memory maps add complexity without value, so I'm inclined to drop them.  However, @jordancarlin do you see any reason to keep them?

jordancarlin:
> My feeling is that these detailed Sail memory maps add complexity without value, so I'm inclined to drop them. However, Jordan Carlin (@jordancarlin) do you see any reason to keep them?

Agreed that the detailed memory map is probably not helpful. Maybe we should create a single IO region in the sail memory map to roughly approximate it?

davidharrishmc:
Could  do that but it is a bug if the tests ever access it memory so maybe
better to get a fault.

davidharrishmc (approved):

PR #1705, opened 2026-06-14
mhpmevent*h CSRs only exist when Sscofpmf supported

jordancarlin:


davidharrishmc (approved):

davidharrishmc:
Technically accessing them is reserved not illegal instruction, I believe. Consider updating comment.

PR #1714, opened 2026-06-15
Add CLEAN_INTERMEDIATES option for disk space savings

### LINKED ISSUE #1701: Add option to remove intermediate build files to save disk space
**Is your feature request related to a problem? Please describe.**

I was tinkering with caching in CI and discovered a few ways to reduce the amount of data being cached. `FAST` environment variable is mentioned in the readme and decreases disk usage substantially (and likely time needed to generate ELFs), however `build` directory is still present and is quite large too.

By only selectively caching `elfs` the total size of compressed cache for two DUTs (with vector tests) went down from over ~860M (IIRC) to 170M, which is a lot.

**Describe the solution you'd like**

I'm wondering if it would be possible to not write those `.sig` and `.results` files either, which would save disk write resource and possibly some work too.

I don't know where they are in the build pipeline though, it is possible that they are input to the next stage and need to be written for that reason, but if not then it'd be great to be able to skip them too.

**Describe alternatives you've considered**

I simply skip extra files explicitly so they don't end up in the cache, but they are still being produced and written to disk, which is ~2G per DUT for me, can hardly be considered negligible.

**Additional context**

jordancarlin:
This initial implementation forces a full rebuild whenever `make` is invoked if `CLEAN_INTERMEDIATES` is set.

Resolves #1701


davidharrishmc (approved):
Does this delete the entire build directory after all of build completes, or suite by suite. Suite by suite would reduce total disk need.

jordancarlin:
> Does this delete the entire build directory after all of build completes, or suite by suite. Suite by suite would reduce total disk need.

For now, it is at the end after the entire build is done. I am working on further changes to the build system to use hashes instead of mtime and will modify this to delete suites as they complete as part of that. With the current built system, that is not easy to do.

PR #1728, opened 2026-06-16
Update RISCV_imported_decode_pkg

### LINKED ISSUE #1726: Fcov build failure in Sm test suite due to missing `CSR_MEDELEGH` definition in auto-generated decode package
## Description

While working on counter wrap-around tests in the Sm test suite #1602 , I encountered an Fcov build failure during coverage generation.

The build fails because `CSR_MEDELEGH` is referenced in `Sm_coverage.svh`, but the CSR definition is missing from the auto-generated decode package (`RISCV_imported_decode_pkg.svh`).

The relevant error is:

```text
** at .../coverpoints/priv/Sm_coverage.svh(252): (vlog-2730) Undefined variable: 'CSR_MEDELEGH'.
```

This causes the coverage package compilation to fail and results in a cascade of subsequent errors.

## Reproduction

```bash
make coverage EXTENSIONS="Sm"
```

## Investigation

During debugging, I observed the following:

* `CSR_MEDELEGH` is referenced by the Sm coverage model.
* `CSR_MEDELEGH` is not present in `framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh`.
* Related CSR definitions are present in the same package.
* Manually adding the missing CSR definition allows the coverage build to complete successfully.

However, `RISCV_imported_decode_pkg.svh` is an auto-generated file and should not be modified manually.

I also checked the `riscv-opcodes` repository and found that `MEDELEGH` is already present in the upstream CSR definitions. Regenerating the corresponding SystemVerilog output includes the missing CSR definition.

Updating the generated decode package using the latest generated output from `riscv-opcodes` is not a straightforward change. The regenerated file introduces a large diff:

```diff
1 file changed, 618 insertions(+), 161 deletions(-)
```

Testing the newly generated file breaks downstream files and creates a large cascade of errors. This shows that a clean update requires extensive changes across the entire framework.

## Possible Fix 

As a short-term fix, the missing CSR definition could potentially be added to `RISCV_decode_pkg.svh`, allowing the coverage model to build successfully without requiring a full regeneration of the imported decode package.

For the longer term, it may be worth considering a more formal synchronization process between the Architecture Test framework and `riscv-opcodes`, so that updates to CSR and instruction definitions can be incorporated in a controlled manner and generated packages remain aligned with upstream changes.

jordancarlin:
Fixes #1726


framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh
```diff
@@ -765,6 +765,34 @@
   localparam [31:0] VL8RE8_V           = 32'b111000101000?????000?????0000111;
   localparam [31:0] VLE1_V             = 32'b000000101011?????000?????0000111;
   localparam [31:0] VLM_V              = 32'b000000101011?????000?????0000111;
+
+  // The following vector load instructions are currently commented out because they
+  // conflict with the more specific versionds defined in RISCV_decode_pkg.svh.
```
davidharrishmc: versions

framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh
```diff
@@ -90,7 +90,6 @@
   localparam [31:0] BSET               = 32'b0010100??????????001?????0110011;
   localparam [31:0] BSETI              = 32'b001010???????????001?????0010011;
   localparam [31:0] BSETI_RV32         = 32'b0010100??????????001?????0010011;
-  localparam [31:0] C_ILLEGAL          = 32'b????????????????0000000000000000;
```
davidharrishmc, line 93: Why did this need to be deleted?

davidharrishmc (changes requested):

framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh
amanmurad-10xe, line 1: Is this file manually editable?

While going through the Developer Guide, I found the following lines:

```
The encodings themselves come from the auto-generated [RISCV_imported_decode_pkg.svh](https://github.com/riscv/riscv-arch-test/blob/act4/framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh) header. This header is generated using [riscv-opcodes](https://github.com/riscv/riscv-opcodes) and should not be manually modified.
```

framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh
jordancarlin: I regenerated the file from riscv-opcodes and it is no longer included in the output of the generator. We don't actually use it anywhere as far as I can tell, so I saw no reason to manually keep it.

framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh
jordancarlin: I regenerated it from riscv-opcodes, but some changes to how riscv-opcodes handles vector loads would have broken our vector tests, so manual edits were needed. I eventually want to switch to generating it from riscv-unified-db, which generates the vector instructions in a manner more aligned with our vector tests, so it did not seem worth updating all of the tests to work with this version since when we switch to the new generator we would need to undo the changes to the tests anyway.

davidharrishmc (approved):

PR #1729, opened 2026-06-16
Pmp fence issue

UmerShahidengr:
Added RVMODEL_FENCEI in the Verification macros (where required), and moved Verification macros to rvtest_pmp_macros.h as centrailised location.


tests/env/rvtest_pmp_macros.h
```diff
@@ -0,0 +1,934 @@
...
+// normally defined), so define it here too. The #ifndef guard makes the later
+// definition in rvtest_trap_handler.h a no-op; ZIFENCEI_SUPPORTED comes from the
+// already-included derived_config.h. Keep this in sync with rvtest_trap_handler.h.
+#ifndef   RVMODEL_FENCEI
+  #ifndef ZIFENCEI_SUPPORTED
+       #define RVMODEL_FENCEI nop                // no Zifencei: assume coherent I-cache
+  #else
+       #define RVMODEL_FENCEI fence.i            // Zifencei available: use fence.i
+  #endif
+#endif
```
jordancarlin: This seems like a maintenance burden. I don't think there is any reason `RVMODEL_FENCEI` has to be in `rvtest_trap_handler.h`. How about just moving it to `utils.h` or similar?

tests/env/rvtest_pmp_macros.h
```diff
@@ -0,0 +1,934 @@
...
+//   ADDRESS   - region label to probe
+//   TEST_CASE - prefix for the local result labels (TEST_CASE_1 .. _3)
+.macro PMP_VERIFICATION_RWX ADDRESS, TEST_CASE
+    // Execution Access Check
+    LA (a4, \ADDRESS)
+    LI(x4, 0xACCE)                        // Store a value which is to be checked in trap handler
+    LA(x1, 1f)                            // Store the return Address in x1
+    RVMODEL_FENCEI                              // sync I-cache: a prior store may have updated this executable region
+    \TEST_CASE\()_1:
+    jalr ra, 0(a4)
```
jordancarlin: This is confusing to follow because you are switching between using x1 and ra which are the same register. Use a consistent name for it.

tests/env/rvtest_pmp_macros.h
```diff
@@ -0,0 +1,934 @@
...
+    // Load Access Check
+    \TEST_CASE\()_3:
+    lw a4, 0(a5)                                             // Word load test
+    nop
+    RVTEST_SIGUPD(x2, x5, x4, a4, \TEST_CASE\()_3, test_3_str)
+.endm
+
+// PMP_VERIFICATION_RWX_LEGAL: full boundary R/W/X check (the *_legal_lxwr/lwxr family).
+// Probes execute, then store, then load at five offsets relative to the region:
+// start, start-4, start+4, start+g-4, start+g (g = PMP granule). Execute is done first
```
jordancarlin: What is lowercase g? There has been so much confusion with the different terms for PMP granularity. Please be more specific which one you are referring to.

tests/env/rvtest_pmp_macros.h
```diff
@@ -0,0 +1,934 @@
...
+// for all five offsets (results recorded as TEST_CASE_11..15), then the five stores
+// (_1.._5) and five loads (_6.._10). RVMODEL_FENCEI at the top syncs the I-cache so a
+// prior invocation's store can't leave a stale instruction. Needs `g` defined and test_1_str..test_15_str.
```
jordancarlin: This says it needs `g` defined but it looks like it actually uses `UDB_PMP_GRANULARITY`.

jordancarlin (changes requested):
I haven't reviewed closely, but here are a few comments. Some of the comments in the `rvtest_pmp_macros.h` file also seem to explain the history of the macros instead of what they do so could use a bit of cleaning up.

tests/env/rvtest_pmp_macros.h
davidharrishmc: RVMODEL seems like a strange naming.  Is there anything that would be used other than fence.i?

davidharrishmc:
Claude review:

PR #1729 Review — "Pmp fence issue"

Overview

The PR centralizes ~90 per-test copies of VERIFICATION_RWX-family GAS macros into a single rvtest_pmp_macros.h, and ensures fence.i (via a new RVMODEL_FENCEI macro) is emitted before jumps into PMP-protected execute regions. Net effect: −2,333 lines, which is genuinely good. The motivation (I-cache coherency before jumping into a region that a prior store may have updated) is correct. But there are several correctness problems in the centralized macros.

---
Critical: Dead Code in Every Execute-Check Block

Every execution-check across all macros (PMP_VERIFICATION_RWX, _RWX_LEGAL, _RWX_ALL, _RWX_NAPOT, _NA4_RV32, _RWX_NAPOT_SM_RV64) follows this pattern:

LA(x1, 1f)               // sets ra = address of label 1:
RVMODEL_FENCEI
TEST_CASE_N:
jalr ra, 0(a4)           // OVERWRITES ra with PC+4 (the nop below)
nop                      // ← ra now points here, not to 1:
1:
nop
RVTEST_SIGUPD(...)

jalr ra, 0(a4) writes PC+4 (address of the first nop) into ra before jumping, soad. The trap handler — which appears to return to ra after a PMPinstruction-access fault — returns to the nop after the jalr, not to 1:. Since both paths (fault or successful execution + callee ret) land at the nop anyway, things work today, but:

- Every LA(x1, Nf) line is useless as written — it tells the assembler to load an address that is never used.
- This is exactly the register-naming inconsistency Jordan Carlin flagged, but t destination register of jalr should be x0 (preserving ra for the trap handler to use) or the LA lines should be removed entirely.

The original, correct pattern was almost certainly:
LA(ra, 1f)       // set skip target for callee + trap handler
jalr x0, 0(a4)  // jump without modifying ra

Every occurrence of jalr ra, 0(a4) paired with a preceding LA(x1, ...) should be audited against the base-branch per-test macros to confirm whether this was pre-existing or introduced by
this consolidation.

---
Critical: PMP_VERIFICATION_RWX_NAPOT_SM_RV64 — Execution Results Never Recorded

This macro performs five jalr jumps (at REGIONSTART, ±4, +g−4, +g) but has no RVTEST_SIGUPD after any of them:

LI(x4, 0xACCE)
LA(a4, ADDRESS)
LA(x1, 1f)
RVMODEL_FENCEI
jalr ra, 0(a4)     // no SIGUPD here
nop
1: nop
addi a4, a4, -4
LA(x1, 2f)
jalr ra, 0(a4)     // no SIGUPD here either
...
// All store/load checks DO have SIGUPD

Compare PMP_VERIFICATION_RWX_NAPOT, which records each of the 5 execute probes aSM_RV64 variant the execution results are silently discarded. This macro doesn'tverify execute access at all from a signature standpoint — exactly the bug the PR claims to fix.

---
Design: RVMODEL_FENCEI Defined Twice with "Keep in Sync" Comment

The new rvtest_pmp_macros.h defines RVMODEL_FENCEI with the comment:

// Keep this in sync with rvtest_trap_handler.h.

This is the maintenance problem: two definitions, no enforcement. Jordan Carlin'the single canonical definition to utils.h (already included by both files). The#ifndef guard means the second definition is a no-op, but the "sync" comment signals that someone could legitimately change one without updating the other.

---
Design: Wrong Naming Prefix for RVMODEL_FENCEI

RVMODEL_* macros are defined by the model under test (the RTL or simulator beingm-specific addresses, memory maps, and capabilities. fence.i is not something the model chooses — it's a test-framework decision about whether to emit a synchronization instruction. This should be RVTEST_FENCEI (defined by the test framework, like RVTEST_SIGUPD) or at minimum RVARCH_FENCEI. David Harris's question about "is there anything other th the answer is yes, nop, but the choice is driven by Zifencei support — atest-framework concern, not a model customization point.

---
Minor: Missing LI(x4, 0xACCE) Before Later jalr Calls in PMP_VERIFICATION_RWX_NA

PMP_VERIFICATION_RWX_NAPOT reloads LI(x4, 0xACCE) (the trap handler sentinel) beobes. The SM_RV64 variant only loads it once, before the first. If the traphandler clears or modifies x4 as part of fault handling (which would be the natural implementation), probes 2–5 will have a stale sentinel and may not be correctly identified as PMP execute
tests by the handler.

---
Minor: Granularity Documentation

As Jordan Carlin noted, "lowercase g" appears in comments without definition. The macro header says it needs g defined, but the actual assembler code uses UDB_PMP_GRANULARITY. The comment
should say:

▎ UDB_PMP_GRANULARITY — the PMP granule size in bits (granule = 1 << UDB_PMP_GRA

---
Minor: History Comments in rvtest_pmp_macros.h

Several comments describe why macros were moved here rather than what the macro does. As Jordan Carlin noted, these should describe behavior, not migration history.

---
Positive

- Net −2,333 lines from the consolidation is a clear improvement.
- The RVMODEL_FENCEI conditional (ZIFENCEI_SUPPORTED ? fence.i : nop) is correctly gated.
- Fence placement immediately before jalr (after the store that may have written
- Store and load checks are uniformly covered with RVTEST_SIGUPD.
- The #include "rvtest_pmp_macros.h" in riscv_arch_test.h makes the macros availon.

---
Summary

The PR needs two correctness fixes before merge: the missing RVTEST_SIGUPD for execution checks in PMP_VERIFICATION_RWX_NAPOT_SM_RV64, and the jalr ra / LA(x1, ...) dead-code audit to confirm whether execution-success and fault paths both land correctly. The RVMODing issues are design debts that should at minimum be resolved per JordanCarlin's utils.h suggestion.

tests/env/rvtest_pmp_macros.h
UmerShahidengr: No, either fence.i or nop will be used, and this macro has been in trap handler file since ages, but never used properly. We can move this one in utils.h, it wont make any difference

tests/env/rvtest_pmp_macros.h
UmerShahidengr: Sure, will do that

tests/env/rvtest_pmp_macros.h
UmerShahidengr: this lower case 'g' is a variable getting used in some tests where we are creating various overlapped regions of multiple of PMP_GRAIN value and checking priority and effect of matching, this variable will be defined in test file and will keep getting updated as per test logic

tests/env/rvtest_pmp_macros.h
UmerShahidengr: Yes, above comment will answer this  comment too

jordancarlin:
@UmerShahidengr any updates on this one?

UmerShahidengr:
@jordancarlin I was occupied in other things for the last few weeks, I will update it this week. Thanks for the ping

UmerShahidengr:
@jordancarlin this one is ready for your attention

tests/env/rvtest_trap_handler.h
```diff
@@ -197,28 +197,8 @@
...
-
-#ifndef   RVMODEL_FENCEI
-  #ifndef ZIFENCEI_SUPPORTED
-       #define RVMODEL_FENCEI nop                // no Zifencei: assume coherent I-cache
-  #else
-       #define RVMODEL_FENCEI fence.i            // Zifencei available: use fence.i
-  #endif
-#endif
+// RVTEST_FENCEI (used below to sync the instruction stream after writing code to
+// memory) is defined in utils.h, which is included before this header.
```
jordancarlin: No need for this comment. There are many other macros that are defined in other files that are used here. This comment will just get stale as the other files continue to change.

tests/env/utils.h
```diff
@@ -43,6 +43,22 @@
...
+// sync routine. Must stay a single instruction (or a JAL) so code size is constant.
+// Defined here so it is available to every consumer (rvtest_pmp_macros.h,
+// rvtest_trap_handler.h, ...); ZIFENCEI_SUPPORTED comes from derived_config.h.
+#ifndef   RVTEST_FENCEI
+  #ifndef ZIFENCEI_SUPPORTED
+       #define RVTEST_FENCEI nop                // no Zifencei: assume coherent I-cache
+  #else
+       #define RVTEST_FENCEI fence.i            // Zifencei available: use fence.i
+  #endif
+#endif
```
jordancarlin: We don't want users defining RVTEST macros. They should only define RVMODEL macros. I think we want something like the following:

```suggestion
#ifdef   RVMODEL_FENCEI
  #define RVTEST_FENCEI RVMODEL_FENCEI
#else
  #ifdef ZIFENCEI_SUPPORTED
    #define RVTEST_FENCEI fence.i
  #else
    #define RVTEST_FENCEI nop
  #endif
#endif
```

tests/env/utils.h
```diff
@@ -43,6 +43,22 @@
...
 #endif
 
+// RVTEST_FENCEI: instruction that synchronizes the instruction stream after code
+// has been written to memory (relocating the trampoline, writing a dynamic
+// instruction to scratch, or storing into an executable PMP region before jumping
+// into it). It is fence.i when the DUT supports Zifencei, otherwise nop (coherent
+// I-cache assumed). A DUT needing a custom mechanism may predefine it as a JAL to a
+// sync routine. Must stay a single instruction (or a JAL) so code size is constant.
+// Defined here so it is available to every consumer (rvtest_pmp_macros.h,
+// rvtest_trap_handler.h, ...); ZIFENCEI_SUPPORTED comes from derived_config.h.
```
jordancarlin: This is not correct. `ZIFENCEI_SUPPORTED` is defined in the UDB generated header (`rvtest_config.h`). I would just drop this part of the comment since it is already documented where all of the extension related #defines come from.

tests/env/rvtest_pmp_macros.h
```diff
@@ -0,0 +1,964 @@
...
+// Jumps (c.jalr) to ADDRESS and records whether execution was permitted.
+// No store occurs, so no RVTEST_FENCEI is required.
+//   ADDRESS   - region label to execute from
+//   TEST_CASE - prefix for the local result label (TEST_CASE_1)
+.macro PMP_VERIFICATION_X_C ADDRESS, TEST_CASE
+    \TEST_CASE\()_1:
+    LA(x15, \ADDRESS)               // Address to be verified
+    c.jalr x15
+    nop
+    RVTEST_SIGUPD(x2, x5, x4, a4, \TEST_CASE\()_1, test_1_str)
```
jordancarlin, line 34: Can these use the DEFAULT_{SIG/LINK/DATA}_REG macros instead of explicitly calling the macros? That way it is easier to change the registers later if needed.

tests/env/rvtest_pmp_macros.h
```diff
@@ -0,0 +1,964 @@
...
+    lw a4, 0(a5)                                             // Word load test
+    nop
+    RVTEST_SIGUPD(x2, x5, x4, a4, \TEST_CASE\()_3, test_3_str)
+.endm
+
+// PMP_VERIFICATION_RWX_LEGAL: full boundary R/W/X check (the *_legal_lxwr/lwxr family).
+// Probes execute, then store, then load at five offsets relative to the region:
+// start, start-4, start+4, start+g-4, start+g, where g = (1<<UDB_PMP_GRANULARITY) is the
+// PMP granule size in bytes. Execute is done first for all five offsets (results recorded
+// as TEST_CASE_11..15), then the five stores (_1.._5) and five loads (_6.._10).
```
jordancarlin: Why are the test cases not executed in order? Can we renumber them to be in order to make debugging easier?

jordancarlin (changes requested):
A few more small things from me and several Copilot comments, and then this looks good to go.

tests/priv/pmp/pmp64/PMPZca/pmpzcb_legal_lwxr.S
```diff
@@ -143,7 +95,7 @@ main:
     LI(x4, PMPREGION_LXWR_1000)
     csrw pmpcfg0, x4
 
-    VERIFICATION_RWX    TEST_FOR_EXECUTION, test_1
+    PMP_VERIFICATION_X_ZCB    TEST_FOR_EXECUTION, test_1
```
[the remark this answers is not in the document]
UmerShahidengr: Not required

tests/env/rvtest_pmp_macros.h
UmerShahidengr: We tried it, but it did not work earlier so adopted this approach for PMP and all Sv tests. It is consistent with all Sv tests so my suggestion is to let it go

tests/env/rvtest_pmp_macros.h
UmerShahidengr: Done

jordancarlin (approved):

PR #1737, opened 2026-06-17
Bring Vf, Vls, and ExceptionsVf to Full Coverage

fourth-bit:
There seems to have been a small error in the sail config for rv64-max that set Vlen incorrectly, which broke coverage in many places.

wvv_all didn't make sense to be in the custom coverpoints for vfwcvt.*.*.* instructions as it is the wrong type. I'm not sure why it was there in the first place, but the coverage it was asking for (vs1 values) is impossible when there isn't a vs1 in the instruction.

A fix to the way indexed data overlaps were being handled: now the index eew is taken from the width field of the instruction, and the data_overlap function does not throw an error anymore.

Finally, sew is being set correctly in the ExceptionsVf tests. It's getting full coverage, but not yet running on every sim.


coverpoints/unpriv/Vf16_coverage.svh
```diff
@@ -21560,139 +21072,17 @@ covergroup Vf16_vfwcvt_rtz_xu_f_v_cg with function sample(ins_t ins);
...
-        bins overlapping = {1'b0};
-    }
-
-    vs1_mod4_2: coverpoint ins.current.insn[16:15] {
-        bins mod4_2 = {2'b10};
-    }
-
-    vs1_mod8_4: coverpoint ins.current.insn[17:15] {
-        bins mod8_4 = {3'b100};
+    `ifdef D_SUPPORTED
```
davidharrishmc, line 21075: Why is this for just D supported, and why is the coverpoint vdoun, but only v checked?  Confirm this is intentional.

fourth-bit: I am not entirely certain about the reason for this coverpoint's existence, but this looks like the intention. The other flags (doun) are supposed to be checked by other coverpoints that specifically use them, and this is meant to be a coverpoint for fflags being able to be set (as if any flags can be raised, NV can be raised). Above this, there is a `ifndef COVER_VFCUSTOM64`, so this line is only hit in the Vf64 case. I'm still not sure that a D_SUPPORTED makes sense if we are running coverage of Vf64, because running coverage on Vf64 implies support for D. This is the version of the coverpoint from `cp_custom_vfp_flags_set.sv`, and this coverpoint was in the file before this PR.

davidharrishmc (approved):

PR #1738, opened 2026-06-17
Fixed ExceptionsF

### LINKED ISSUE #1734: ExceptionsF: RVTEST_SIGUPD_CSR_READ causes a second trap after CSRRS/CSRRC on FS=Off implementations
**Description:**

When testing FP CSR coverpoints under mstatus.FS=Off conditions, the generated test sequences follow this pattern for CSRRS and CSRRC coverpoints:

Image

Expanding RVTEST_SIGUPD_CSR_READ, the actual emitted instructions are:

Image

**Problem:**

CSRRC rd, csr, rs1 with rd≠x0 is an atomic read-modify-write. The instruction itself delivers the pre-clear value of fflags into x12 as part of its execution. There is no need to issue a second csrrs x12, fflags, x0 to obtain the result, the result is already in x12.
However, RVTEST_SIGUPD_CSR_READ unconditionally emits that second CSR read. On hardware where mstatus.FS=Off, every access to an FP CSR is illegal, so this second read is not a redundancy ; it is a second trap site.

When mstatus.FS=Off, the trap sequence is as follows:
First trap:    from csrrc x12, fflags, x12 (the actual coverpoint):
Traps as Illegal Instruction (scause=2) -> Trap handler records this into the signature slot at 0(x2) -> Handler advances x2 += 8 and returns to mepc+4

Second trap: from csrrs x12, fflags, x0 (inside RVTEST_SIGUPD_CSR_READ):
mstatus.FS is still Off (nothing restored it between the two instructions) -> This instruction also traps as Illegal Instruction (scause=2) -> Trap handler fires again, now trying to write to 0(x2) but x2 was already advanced by the first trap.

Depending on where x2 now points, this store either writes into a signature slot that belongs to the next coverpoint, hits a misaligned or unmapped address, generating a store/AMO misaligned exception (scause=6)
That new scause=6 overwrites the scause=2 that was correctly recorded by the first trap.
After this second trap, x2 has been advanced twice, it is now 8 bytes ahead of where the reference model (Sail) expects it to be.

**The Core Question**

Is RVTEST_SIGUPD_CSR_READ being intentionally emitted after CSRRS/CSRRC (with rd≠x0) to capture the post-modification CSR state as a separate check, with the expectation that the reference model also accounts for two signature slots? Or is this an unintentional macro misuse where the generator should have simply logged rd directly without issuing a second CSR access?

If it is intentional, then the trap handler needs to be aware that the macro can generate two traps from a single coverpoint and must not advance x2 on the second one otherwise the pointer walks off the allocated region on any hardware where the CSR is inaccessible.

If it is unintentional, the fix is straightforward: for CSRRS/CSRRC coverpoints where rd≠x0, replace RVTEST_SIGUPD_CSR_READ with a macro that records the value already in rd directly, without issuing another CSR read instruction. The coverpoint instruction is its own read.

Teemooooooooo:
Fixes #1734

Modified the test generator for cp_mstatus_fs_csr_write to use write_sigupd instead of gen_csr_write_sigupd so that the test does not double trap when FS = 00.


nazar-pc:
Just FYI: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

This helps closing linked issues automatically on PR merge.

jordancarlin (changes requested):
I'm not convinced that these changes are correct. We want to make sure the value in the CSR was not changed by the instruction when it traps. Checking the value from before the access is not meaningful. I think we might need to do the access, set `mstatus.FS` to 0b11, call `gen_csr_read_sigupd`, and then restore `mstatus.FS` to whatever it is supposed to be for that test.

jordancarlin:
The other option would be to not do a signature at all. The trap still gets recorded in the signature, and maybe we just assume that if it trapped, then the value was not erroneously updated. The approach described in the previous comment would be more robust, though.

generators/testgen/src/testgen/priv/extensions/ExceptionsF.py
```diff
@@ -354,22 +359,40 @@ def _generate_mstatus_fs_illegal_instr_tests(test_data: TestData) -> list[str]:
...
+            )
+            if csr_instr == "CSRR":
+                lines.append(f"{csr_instr}(x{check_reg}, {csr})")
+            else:
+                lines.append(f"{csr_instr}({csr}, x{check_reg})")
+            lines.extend(
+                [
+                    f"csrs mstatus, x{clear_mask_reg}",
+                    gen_csr_read_sigupd(check_reg, (csr, None), test_data),
+                    f"csrc mstatus, x{clear_mask_reg}",
```
jordancarlin: We clear mstatus.FS at the beginning of the loop, so I don't think we need to do it at the end of the loop too.

generators/testgen/src/testgen/priv/extensions/ExceptionsF.py
```diff
@@ -354,22 +359,40 @@ def _generate_mstatus_fs_illegal_instr_tests(test_data: TestData) -> list[str]:
...
+                ]
+            )
+            if csr_instr == "CSRR":
+                lines.append(f"{csr_instr}(x{check_reg}, {csr})")
+            else:
+                lines.append(f"{csr_instr}({csr}, x{check_reg})")
+            lines.extend(
+                [
+                    f"csrs mstatus, x{clear_mask_reg}",
+                    gen_csr_read_sigupd(check_reg, (csr, None), test_data),
```
jordancarlin: It's worth a comment explaining why we are enabling mstatus.FS here.

generators/testgen/src/testgen/priv/extensions/ExceptionsF.py
```diff
@@ -354,22 +359,40 @@ def _generate_mstatus_fs_illegal_instr_tests(test_data: TestData) -> list[str]:
...
+def _generate_mstatus_fs_csr_access_tests(test_data: TestData) -> list[str]:
+    covergroup, coverpoint = "ExceptionsF_cg", "cp_mstatus_fs_csr_access"
+    clear_mask_reg, check_reg = test_data.int_regs.get_registers(2)
 
     lines = [
         comment_banner(
             coverpoint,
             "Test that access to floating point CSRs trap when mstatus.fs is set to 0 (Off)",
         ),
         f"LI(x{clear_mask_reg}, 0x6000) # MSTATUS_FS mask",
```
jordancarlin: Might be worth renaming this register to `mstatus_fs_mask_reg` or similar since it is now used for both clears and sets.

jordancarlin (changes requested):
A few more minor comments, but this looks pretty good now.

jordancarlin (approved):

PR #1745, opened 2026-06-18
ZawrsSm, ZawrsS, and ZawrsU testgen at 100% coverage on everything other than H

Teemooooooooo:
ZawrsSm, ZawrsS and ZawrsU test generators
Does not have coverage for H since sail does not support hypervisor yet


jordancarlin:
@Teemooooooooo I haven't looked through all of the code yet, but is a custom trap handler really needed for this test? `mtvec`/`stvec` are WARL, so implementations have all sorts of different restrictions on the legal values. The standard trap handler loader already deals with those, but we would need to do the same here if we want to load a different trap handler. It would be a lot simpler if we are able to stick with the standard trap handler.

davidharrishmc:
Reviewed everything except Python code.  Will review python after possibly removing custom trap handler.

Teemooooooooo:
I think we would still need a custom trap handler because we need a mechanism to indicate a trap has been taken. 

The loop is need so that the resume/trap behavior of wrs.nto /wrs.sto can still be tested even if the implementation has it terminating early. (Same thing will need to be implemented in wfi tests)

When I was using the custom trap handler, this is what was used to indicate a trap has happened:
- initialize r_cause =0
- loop until trap, the trap handler records xcause value in r_cause
- return to the program: if r_cause != 0, a trap happened
- move on

I do not think the same thing could be done with the normal trap handler as the xcause value will only be saved in the csr, which will not be accessible from lower priv modes once we return from the trap handler. Even if we use sbi calls to access xcause, ecall would change the xcause value.

jordancarlin:
I think we had discussed the trap handler storing the number of traps that have been taken to a memory location. If you read that number before and after, you can tell whether a trap was taken. Would that be sufficient?

Teemooooooooo:
ohh that should be sufficient if that memory location accessible by all priv modes. I will look into rvtest_trap_handler.h and try to find that address :D

jordancarlin:
> ohh that should be sufficient if that memory location accessible by all priv modes. I will look into rvtest_trap_handler.h and try to find that address :D

To clarify, we discussed adding that, but I don't know if it has actually been done yet. @UmerShahidengr would know.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+) -> list[str]:
+    """WRS ops timeout behavior"""
+    lines = []
+    op = "WRS.STO" if coverpoint == "cp_wrs_sto_timeout" else "WRS.NTO"
+
+    tw_list = [1] if coverpoint == "cp_wrs_nto_timeout" else [0, 1]
+
+    if coverpoint == "cp_wrs_nto_timeout_h":
+        lines.append("#ifdef H_SUPPORTED")
+    for priv in priv_list:
```
davidharrishmc, line 364: Does priv_list ever have more than one element?  If not, drop iterating over it.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+        for tw_val in tw_list:
+            lines.extend(
+                [
+                    "###### Setup (M Mode) ######",
+                    "# Disable all interrupts in mie",
+                    "CSRW mie, zero",
+                    "# mstatus.MIE = 0 and MPIE = 0",
+                    f"LI(x{r_temp}, 0x88)",
```
davidharrishmc: Change to 8A and then you clear SIE for free. Harmless to clear it even if there is no S mode.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                    f"CSRC(mstatus, x{r_temp})",
+                    "# Write TW bit",
+                    f"LI(x{r_temp}, 0x200000)",
+                    f"{'CSRS' if tw_val else 'CSRC'}(mstatus, x{r_temp})",
+                ]
+            )
+            if priv != "M":
```
davidharrishmc: Can be deleted by using 8A above.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                        "# Clear mstatus.SIE",
+                        "csrci mstatus, 0x2",
+                        "#endif",
+                    ]
+                )
+            if coverpoint == "cp_wrs_nto_timeout_h":
+                lines.extend(
+                    [
+                        "# Set VTM",
+                        f"LI(x{r_temp}, 0x100000)",
```
davidharrishmc: Should this be VTW in bit 21 (0x200000)? There is no VTM.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                        f"CSRS(mstatus, x{r_temp})",
```
davidharrishmc: VTW is in hstatus, not mstatus.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                        "CSRW hedeleg, zero",
+                    ]
+                )
+            if op == "WRS.NTO":  # These are the cases that are expected to trap
+                lines.extend(
+                    [
+                        "# Set up trap handler - will trap to either S or M mode based on what is supported",
+                        f"# Install local trap handler in xtvec (direct mode); save old xtvec in x{r_scratch}",
+                        f"LA(x{r_temp}, xt_trap_handler)",
+                        f"csrrw x{r_scratch}, xtvec, x{r_temp}",
```
davidharrishmc: how does this work?  There is no such thing as xtvec.  You want to do stvec or mtvec (mtvec if not S_SUPPORTED).  I don't know how this compiled.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                    "1:",
+                    f"{op}",
+                ]
+            )
+            if op == "WRS.NTO":
+                lines.extend(
+                    [
+                        # This part commented out because the parameter is not defined yet
+                        # "#ifndef UDB_ZAWRS_NTO_IS_NOP",
+                        "# r_cause is 0, no trap happened",
```
davidharrishmc: What do you mean by r_cause is 0?  Is it necessarily 0?  Or do you mean if r_cause = 0, WRS.NTO terminated prematurely and should be repeated until timeout occurs.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+            )
+            if priv != "M":
+                lines.extend(
+                    [
+                        "RVTEST_GOTO_MMODE",
+                    ]
+                )
+
+    if coverpoint == "cp_wrs_nto_timeout_h":
+        lines.append("#endif")
```
davidharrishmc: #endif // H_SUPPORTED

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+def _wrs_no_res_helper(test_data: TestData, priv: str, covergroup: str) -> list[str]:
+    """Helper function for generating WRS instruction no reservation tests"""
+
+    r_scratch, r_temp, r_temp2 = test_data.int_regs.get_registers(3)
+    ######################################
+    coverpoint = "cp_wrs_no_res"
+    ######################################
+    lines = []
+
+    tw_list = [0] if priv != "M" else [0, 1]
```
davidharrishmc, line 303: The test plan specifies TW=1.  The plan and tests should be consistent.  Why do the tests exercise TW=1 for S and U but not M?  If there is a good reason, let's update the plan to match and explain the reason.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+
+    for tw_val in tw_list:
+        for wrs_ops in ["WRS.STO", "WRS.NTO"]:
+            lines.extend(
+                [
+                    "#### Setup (M mode) ####",
+                    "# Disable all interrupts in mie",
+                    "CSRW mie, zero",
+                    "# mstatus.MPIE = 0",
+                    f"LI(x{r_temp}, 0x80)",
```
davidharrishmc: See timeout comments about using 0x8A to clear MPIE, MIE, SIE.  

The test plan says to clear mstatus.MIE, but this test only touches MPIE.  Any reason the test can't be consistent with the plan?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+    r_scratch, r_temp, r_temp2 = test_data.int_regs.get_registers(3)
+    ######################################
+    coverpoint = "cp_wrs_no_res"
+    ######################################
+    lines = []
+
+    tw_list = [0] if priv != "M" else [0, 1]
+
+    for tw_val in tw_list:
+        for wrs_ops in ["WRS.STO", "WRS.NTO"]:
```
davidharrishmc: Shouldn't this be wrs_op, not ops?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+    priv: str,
+    covergroup: str,
+    r_cause: int,
+    r_scratch: int,
+    r_temp: int,
+    r_timecmp: int,
+    r_temp2: int,
+) -> list[str]:
+    """wrs resume when interrupt"""
+    # TODO: FIX THIS PART LATER
```
davidharrishmc: Resolve TODOs before merge.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+    ######################################
+
+    r_time, r_temp3 = test_data.int_regs.get_registers(2)
+
+    lines = []
+    if priv != "M":
+        sie_list = [0, 1]
+        tw_list = [0]
+    else:  # if test is for M mode
+        sie_list = [0]  # to ensure that the python for loop does not repeat for sie values
```
davidharrishmc: Comments are more useful if they focus on the underlying reason this matters for the test, not what Python does.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                            f"# mstatus.MPIE and mstatus.MIE = {mie_val}",
+                            f"LI(x{r_temp}, 0x88)",
+                            f"{'CSRS' if mie_val else 'CSRC'}(mstatus, x{r_temp})",
+                            "# Write mstatus.TW",
+                            f"LI(x{r_temp}, 0x200000)",
+                            f"{'CSRS' if tw_val else 'CSRC'}(mstatus, x{r_temp})",
+                            "",
+                            f"# Install local trap handler in xtvec (direct mode); save old xtvec in x{r_scratch}",
+                            f"LA(x{r_temp}, xt_int_handler)",
+                            f"csrrw x{r_scratch}, xtvec, x{r_temp}",
```
davidharrishmc: Same comment as in timeout about xtvec

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                    )
+
+                    if priv != "M":
+                        lines.extend(
+                            [
+                                "#ifdef S_SUPPORTED",
+                                f"# sstatus.SIE = {sie_val}",
+                                f"{'csrsi' if sie_val else 'csrci'} sstatus, 2",
+                                "#endif",
+                                "#if defined(S_SUPPORTED) && defined(SSTC_SUPPORTED)",
```
davidharrishmc: if you move #ifdef SSTC_SUPPORTED inside #ifdef S_SUPPORED, then you don't have to have the double #if defined.

I thought rvtest_config.h uses UDB_XXX_DEFINED instead of XXX_DEFINED.  Have you checked whether this is the right parameter and whether the #ifdef ever executes in real code?  If you haven't watched the test run in sim and compared to source, that would be prudent.

I am out of copilot review credits for the month.  Have you asked Claude for a review and fixed anything legit that it flags?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                                "# Enable Sstc (menvcfg.STCE) so stimecmp drives sip.STIP",
+                                *set_menvcfg_stce(r_temp, True),
```
davidharrishmc: Why not do the menvcfg.STCE and mideleg steps at the start of the program just once, instead of every time through the loop?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                                "# Set stimer interrupt soon",
+                                *set_stimer_int_soon_sstc(r_time, r_temp, r_temp2, r_temp3, r_cause),
+                                "#else",
+                                "# Set mie.MTIE",
+                                f"LI(x{r_temp}, 0x80)",
+                                f"CSRS(mie, x{r_temp})",
+                                "# Set mtimer interrupt soon",
+                                *set_mtimer_int_soon(r_time, r_timecmp, r_temp, r_temp2, r_temp3, r_cause),
+                                "#endif",
+                                f"RVTEST_GOTO_LOWER_MODE {priv}mode",
```
davidharrishmc, line 107: Why do we only go to lower priv mode if SSTC?  Should this be moved after the part of setting up the interrupt soon?

davidharrishmc: How do we know that we aren't getting the interrupt before we get to the lower priv mode?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                                f"CSRS(mie, x{r_temp})",
+                                "# Set mtimer interrupt soon",
+                                *set_mtimer_int_soon(r_time, r_timecmp, r_temp, r_temp2, r_temp3, r_cause),
+                            ]
+                        )
+
+                    lines.extend(
+                        [
+                            "",
+                            f"LI(x{r_cause}, 0)                 # nonzero means the trap was taken",
```
davidharrishmc: Seems prudent to put this before setting up the interrupt to fire, in case it fires earlier than expected.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                            ),
+                            "1:",
+                            f"{op}",
+                            f"bnez x{r_cause}, 2f              # MIE=1: handler recorded mcause -> done",
+                        ]
+                    )
+                    if (priv == "M") & (mie_val == 0):
+                        lines.extend(
+                            [
+                                "# Only moves on if MIE = 0 and mip.MTIP = 1, expect no interrupt should be taken",
```
davidharrishmc: Don't we already know mstatus.MIE=0?  Does it need to be said here.

Safer to talk about mstatus.MIE than MIE because there is also a mie register that is confusing.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                                f"csrr x{r_temp}, mip",
+                                f"andi x{r_temp}, x{r_temp}, 0x80  # Extract mip.MTIP",
+                                f"bnez x{r_temp}, 2f              # Interrupt pending -> done",
+                            ]
+                        )
+                    if (priv == "S") & (sie_val == 0):
+                        lines.extend(
+                            [
+                                "#if defined(S_SUPPORTED) && defined(SSTC_SUPPORTED)",
```
davidharrishmc: Don't need S_SUPPORTED because we already know priv == S just to get here.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+def _zawrs_define_helper(priv: str) -> list[str]:
+    """Define the CSRs based on priv mode"""
+    lines = []
+    if priv != "M":
+        lines.extend(
+            [
+                "#if defined(S_SUPPORTED) && defined(SSTC_SUPPORTED)",
+                "#define xt_int_handler st_int_handler",  # test using STIP
+                "#define xcause scause",
+                "#define xtvec stvec",
```
davidharrishmc: confusing to have macros with lower case names.  I was confused about xtvec several places below.  Use upper case macro names to help the reader realize these are macros.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+        [
+            "# Record xcause, restore xtvec",
+            f"csrr x{r_cause}, xcause",
+            f"csrw xtvec, x{r_scratch}",
+        ]
+    )
+    if is_int:
+        lines.extend(
+            [
+                "#if defined(S_SUPPORTED) && defined(SSTC_SUPPORTED)",
```
davidharrishmc: SSTC_SUPPORTED implies S_SUPPORTED so no need to check both.

davidharrishmc: Global change.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                "#define xt_int_handler st_int_handler",  # test using STIP
+                "#define xcause scause",
+                "#define xtvec stvec",
+                "#define xret sret",
+                "#define xt_trap_handler st_trap_handler",
+                "#define xepc sepc",
+                "#else",
+                "#define xt_int_handler mt_int_handler",
```
davidharrishmc: This is repeated below. With more careful condition checks, I think you only need to state it once.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                            f"lr.w x{r_temp2}, (x{r_temp})",
+                            test_data.add_testcase(
+                                f"tw_{tw_val}_mie_{mie_val}_sie_{sie_val}_{op}", coverpoint, covergroup
+                            ),
+                            "1:",
+                            f"{op}",
+                            f"bnez x{r_cause}, 2f              # MIE=1: handler recorded mcause -> done",
+                        ]
+                    )
+                    if (priv == "M") & (mie_val == 0):
```
davidharrishmc, line 142: I think this is right, but it took a long time to understand.  Better explanation of the big picture would be useful.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+    wrs_list = ["WRS.STO", "WRS.NTO"] if priv != "M" else ["WRS.STO"]
+
+    for op in wrs_list:
+        lines.extend(
+            [
+                "###### Setup (M Mode) ######",
+                "# Disable all interrupts in mie",
+                "CSRW mie, zero",
+                "# mstatus.MIE = 1 and MPIE = 1",
+                f"LI(x{r_temp}, 0x88)",
```
davidharrishmc: Why not set SIE here too?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                f"CSRS(mstatus, x{r_temp})",
+            ]
+        )
+        if op == "WRS.NTO":
+            lines.extend(
+                [
+                    # "#ifndef UDB_ZAWRS_NTO_IS_NOP",
```
davidharrishmc: Sounds like a good time to make a PR to UDB with this new parameter.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -0,0 +1,563 @@
...
+                test_data.add_testcase(f"{op}", coverpoint, covergroup),
+                "1:",
+                f"{op}",
+            ]
+        )
+        if op == "WRS.NTO":
+            lines.extend(
+                [
+                    # This part commented out because the parameter is not defined yet
+                    # "#ifndef UDB_ZAWRS_NTO_IS_NOP",
```
davidharrishmc: Can this be uncommented?  It won't be defined yet anyway.

davidharrishmc (changes requested):
Please ask Claude for a review as well.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
Teemooooooooo: it does have more than one element for the hypervisor case (VS and VU)

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
Teemooooooooo: I had another helper function that defines xtvec to be either mtvec or stvec based on whether SSTC and S are supported or not (and whether the test runs in M mode)

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
Teemooooooooo: The testplan specifies TW = {0/1} only for M mode, the reason for this is listed below:

For the coverpoint cp_wrs_no_res, the expected behavior of TW to not stall the hart's execution as there is no reservation. 
The spec states:
"When the TW (Timeout Wait) bit in mstatus is set and WRS.NTO is executed in any privilege mode other than M mode, and it does not complete within an implementation-specific bounded time limit, the WRS.NTO instruction will cause an illegal-instruction exception."
- A DUT (For example, Spike) can have WRS.NTO trigger illegal instruction immediately instead of having it retire successfully (which is what SAIL does when there is no reservation).

Therefore, the behavior of TW = 1 can be implementation specific on whether it causes illegal instruction or not. It has already been removed from the testplan

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
Teemooooooooo: We are going to a lower priv mode for any test that runs in non-M mode. 

I was trying to ensure that the interrupt does not happen before we get to the goto_lower_priv mode by setting up  timer interrupt soon right before using the GOTO_LOWER_MODE macro. I thought the delay in the helper function should be enough to ensure that the interrupt does not accidentally trigger before the intended area.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
Teemooooooooo: I am not sure how much of a risk it would be for the interrupt to fire earlier than expected. If that is a large risk adjustments should be made to the interrupts.py helper function because this will also affect the interrupts test. 

The reason of why I set up r_cause after setting up the interrupts is because set_mtimer_int_soon requires 6 registers and r_scratch was already used to save the trap handler address and we were hitting the limit of 7 registers (which will no longer be a problem if we switch to using the framework trap handler).

Also, if the interrupt triggers before r_cause is initialized, the test would not be able to test the resume behavior of the DUT even if the test pass. 

One way that I was thinking could ensure that the resume behavior of the WRS instruction is tested is to include another cross with mip.xTIP in the coverpoint. The issue with this is that we would not be able to keep the cross with lr instruction to ensure the reservation has been set up and that the interrupt might land on the branch instruction instead of actually on the WRS instruction.

How important is it to ensure that interrupt resume behavior of WRS instruction is tested?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
Teemooooooooo: This comment was added in specifying the condition that interrupt is not expected fire, so that it shows up in the objdump in case it is needed for debugging purposes.

davidharrishmc (approved):

Issue #1633, opened 2026-06-03
Ssstrict tests should not include cases that depend on system state


davidharrishmc:
Presently, SsstrictSm,SsstrictS,SsstrictU and probably SsstrictV test encodings that are reserved for certain system state, especially LMUL or SEW.  Per Andrew Waterman, the behavior of such instructions is UNSPECIFIED and should not be tested.  Ssstrict only applies to instructions whose reserved status solely depends on the opcode or CSR number.

Overlap constraints that apply to only some LMUL are not covered by Ssstrict, but constraints that apply independently of LMUL presumably should be Ssstrict.

Recommend commenting out these coverpoints and tests.

Examples (not all confirmed to be in the test suite) include:

```
The use of vtype encodings with LMUL < SEWMIN/ELEN is reserved, but implementations can set vill if
they do not support these configurations.

When LMUL=2, the vector register group contains vector register v n and vector register v n+1, providing
twice the vector length in bits. Instructions specifying an LMUL=2 vector register group with an odd-
numbered vector register are reserved.

When LMUL=4, the vector register group contains four vector registers, and instructions specifying an
LMUL=4 vector register group using vector register numbers that are not multiples of four are reserved.

When LMUL=8, the vector register group contains eight vector registers, and instructions specifying an
LMUL=8 vector register group using register numbers that are not multiples of eight are reserved.

The use of vstart values greater than the largest element index for the current vtype setting is reserved.

Vector operands or results may occupy one or more vector registers depending on EMUL, but are always
specified using the lowest-numbered vector register in the group. Using other than the lowest-numbered
vector register to specify a vector register group is a reserved encoding.

The destination vector register group for a masked vector instruction cannot overlap the source mask
register (v0), unless the destination vector register is being written with a mask value (e.g., compares) or the
scalar result of a reduction. These instruction encodings are reserved.

When rs1=x0 and rd=x0, the instructions operate as if the current vector length in vl is used as the AVL,
and the resulting value is written to vl, but not to a destination register. This form can only be used when
VLMAX and hence vl is not actually changed by the new SEW/LMUL ratio. Use of the instructions with a
new SEW/LMUL ratio that would result in a change of VLMAX is reserved. Use of the instructions is also
reserved if vill was 1 beforehand. Implementations may set vill in either case.

If the implementation does not support the EEW of the offset elements, the instruction is reserved.

Implementations must provide vector loads and stores with EEWs corresponding to all supported SEW
[...]

PR #1751, opened 2026-06-18
SsstrictV 100% Coverage, Passing Imperas and Whisper

fourth-bit:
This PR massively pares down the set of tests that SsstrictV runs, as is suggested by #1633. Several duplicated tests were removed, and tests for unaligned whole register loads and illegal nf values were added. This required me to run the `build_disasm_fallback.py` script which seems like it hadn't had results checked into the repo for a while, so many lines were added to the `RISCV_build_disasm_fallback.svh` file. 

Currently, with a modified version of sail, both whisper and imperas pass the test suite. Spike and Qemu do not pass, but I don't think that they claim to support Ssstrict.


jordancarlin:
@fourth-bit is this ready to review or should we wait for the next sail-riscv release?

fourth-bit:
This is ready for review. Passing Imperas and Whisper is contingent on the new sail release, but they were failing SsstrictV before PR anyways. Coverage should work without the new sail release.

coverpoints/priv/RISCV_coverage_vect_instr.svh
```diff
@@ -138,30 +138,167 @@
...
     v_IVV_f6 : coverpoint ins.current.insn[31:26]
         iff (ins.current.insn[6:0] == 7'b1010111 & ins.current.insn[14:12] == 3'b000) {
+        // Narrowing Instructions and Widening Instructions are reserved at the max SEW only, so
+        // SsstrictV tests should not run them.
+        ignore_bins vnsrl        = { 6'b101100 };
+        ignore_bins vnsra        = { 6'b101101 };
+        ignore_bins vnclipu      = { 6'b101110 };
+        ignore_bins vnclip       = { 6'b101111 };
+        ignore_bins vwredsumu    = { 6'b110000 };
+        ignore_bins vwredsum     = { 6'b110001 };
```
jordancarlin: Should this be wrapped in an `` `ifdef ZVE32X_SUPPORTED `` so that they are still tested in Ssstrict for implementations that don't support vector?

coverpoints/priv/RISCV_coverage_vect_instr.svh
```diff
@@ -138,30 +138,167 @@
...
+        ignore_bins vwaddu_w    = { 6'b110100 };
+        ignore_bins vwadd_w     = { 6'b110101 };
+        ignore_bins vwsubu_w    = { 6'b110110 };
+        ignore_bins vwsub_w     = { 6'b110111 };
+        ignore_bins vwmulu      = { 6'b111000 };
+        ignore_bins vwmulsu     = { 6'b111010 };
+        ignore_bins vwmul       = { 6'b111011 };
+        ignore_bins vwmaccu     = { 6'b111100 };
+        ignore_bins vwmacc      = { 6'b111101 };
+        ignore_bins vwmaccsu    = { 6'b111111 };
```
jordancarlin: Should this be wrapped in an `` `ifdef ZVE32X_SUPPORTED `` so that they are still tested in Ssstrict for implementations that don't support vector?

coverpoints/priv/RISCV_coverage_vect_instr.svh
```diff
@@ -138,30 +138,167 @@
...
     }
     // OPIVI: funct3 = 011
     v_IVI_f6 : coverpoint ins.current.insn[31:26]
         iff (ins.current.insn[6:0] == 7'b1010111 & ins.current.insn[14:12] == 3'b011) {
+        // Narrowing Instructions and Widening Instructions are reserved at the max SEW only, so
+        // SsstrictV tests should not run them.
+        ignore_bins vnsrl   = { 6'b101100 };
+        ignore_bins vnsra   = { 6'b101101 };
+        ignore_bins vnclipu = { 6'b101110 };
+        ignore_bins vnclip  = { 6'b101111 };
```
jordancarlin: Should this be wrapped in an `` `ifdef ZVE32X_SUPPORTED `` so that they are still tested in Ssstrict for implementations that don't support vector?

coverpoints/priv/RISCV_coverage_vect_instr.svh
```diff
@@ -138,30 +138,167 @@
...
     }
     // OPIVX: funct3 = 100
     v_IVX_f6 : coverpoint ins.current.insn[31:26]
         iff (ins.current.insn[6:0] == 7'b1010111 & ins.current.insn[14:12] == 3'b100) {
+        // Narrowing Instructions and Widening Instructions are reserved at the max SEW only, so
+        // SsstrictV tests should not run them.
+        ignore_bins vnsrl   = { 6'b101100 };
+        ignore_bins vnsra   = { 6'b101101 };
+        ignore_bins vnclipu = { 6'b101110 };
+        ignore_bins vnclip  = { 6'b101111 };
```
jordancarlin: Should this be wrapped in an `` `ifdef ZVE32X_SUPPORTED `` so that they are still tested in Ssstrict for implementations that don't support vector?

coverpoints/priv/RISCV_coverage_vect_instr.svh
```diff
@@ -138,30 +138,167 @@
...
+        ignore_bins vwadd_w     = { 6'b110101 };
+        ignore_bins vwsubu_w    = { 6'b110110 };
+        ignore_bins vwsub_w     = { 6'b110111 };
+        ignore_bins vwmulu      = { 6'b111000 };
+        ignore_bins vwmulsu     = { 6'b111010 };
+        ignore_bins vwmul       = { 6'b111011 };
+        ignore_bins vwmaccu     = { 6'b111100 };
+        ignore_bins vwmacc      = { 6'b111101 };
+        ignore_bins vwmaccus    = { 6'b111110 };
+        ignore_bins vwmaccsu    = { 6'b111111 };
```
jordancarlin: Should this be wrapped in an `` `ifdef ZVE32X_SUPPORTED `` so that they are still tested in Ssstrict for implementations that don't support vector?

jordancarlin: Add comment for why these are ignored similar to what you did for all the other exclusions.

framework/src/act/fcov/coverage/RISCV_disasm_fallback.svh
jordancarlin, line 1: We should eventually make sure this file is regenerated as part of the standard build/CI flow when covergroupgen is called (seems like it needs to be manually regenerated right now). It would be even better to eliminate the need for this file somehow since right now it takes a very different approach than the rest of the coverage model. I am especially concerned because all illegal instructions are sent through this even for non-vector tests.

generators/coverage/src/covergroupgen/templates/vector/priv/cp_exceptionsv_vd_vs1_overlap.sv
```diff
@@ -3,6 +3,7 @@
 // //////////////////////////////////////////////////////////////////////////////////////////////////////////
 
     // ExceptionsV: vd overlaps vs1 source register group
-    cp_exceptionsv_vd_vs1_overlap : cross std_trap_vec, vtype_all_lmulge1, vd_eq_vs1;
+    // cp_exceptionsv_vd_vs1_overlap : cross std_trap_vec, vtype_all_lmulge1, vd_eq_vs1;
+    cp_exceptionsv_vd_vs1_overlap : cross std_trap_vec, vd_eq_vs1;
```
jordancarlin, line 9: Why is the old version retained as a comment? Global question for many of these templates.

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_masking_vs1_eq_v0.sv
```diff
@@ -0,0 +1,9 @@
+// //////////////////////////////////////////////////////////////////////////////////////////////////////////
+// cp_ssstrictv_masking_vs1_eq_v0
+// //////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+    // Masked instruction with vs1=v0 at LMUL=1: vs1 group overlaps mask v0, reserved
+    vs1_is_v0_meqv0 : coverpoint ins.current.insn[19:15] { bins v0 = { 5'b00000 }; }
```
jordancarlin, line 6: Can this use `ins.current.vs1` instead of the explicit `insn[19:15]`?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_masking_vs2_eq_v0.sv
```diff
@@ -0,0 +1,9 @@
+// //////////////////////////////////////////////////////////////////////////////////////////////////////////
+// cp_ssstrictv_masking_vs2_eq_v0
+// //////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+    // Masked instruction with vs2=v0 at LMUL=1: vs2 group overlaps mask v0, reserved
+    vs2_is_v0_meqv0 : coverpoint ins.current.insn[24:20] { bins v0 = { 5'b00000 }; }
```
jordancarlin, line 6: Can this use `ins.current.vs2` instead of the explicit `insn[24:20]`?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vcompress_vd_v0_overlap.sv
```diff
@@ -4,12 +4,14 @@
...
-    cp_ssstrictv_vcompress_vd_v0_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vd_v0;
+    // cp_ssstrictv_vcompress_vd_v0_overlap_lmul1: cross std_trap_vec, vtype_lmul_1, vd_v0;
 
-    cp_ssstrictv_vcompress_vd_v0_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vd_v0;
+    // cp_ssstrictv_vcompress_vd_v0_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vd_v0;
 
-    cp_ssstrictv_vcompress_vd_v0_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vd_v0;
+    // cp_ssstrictv_vcompress_vd_v0_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vd_v0;
+
+    // cp_ssstrictv_vcompress_vd_v0_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vd_v0;
```
jordancarlin, line 17: Why are these commented out?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vcompress_vd_vs2_overlap.sv
```diff
@@ -4,12 +4,14 @@
...
-    cp_ssstrictv_vcompress_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
+    // cp_ssstrictv_vcompress_vd_vs2_overlap_lmul1: cross std_trap_vec, vtype_lmul_1, vd_eq_vs2;
 
-    cp_ssstrictv_vcompress_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+    // cp_ssstrictv_vcompress_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
 
-    cp_ssstrictv_vcompress_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
+    // cp_ssstrictv_vcompress_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+
+    // cp_ssstrictv_vcompress_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
```
jordancarlin, line 17: Why are these all commented out?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vrgather_vd_vs1_eq.sv
```diff
@@ -3,6 +3,7 @@
 // //////////////////////////////////////////////////////////////////////////////////////////////////////////
 
     // vrgather with vd==vs1 reserved (vd source overlap)
-    cp_ssstrictv_vrgather_vd_vs1_eq : cross std_trap_vec, vtype_all_lmulge1, vd_eq_vs1;
+    // cp_ssstrictv_vrgather_vd_vs1_eq : cross std_trap_vec, vtype_all_lmulge1, vd_eq_vs1;
+    cp_ssstrictv_vrgather_vd_vs1_eq : cross std_trap_vec, vd_eq_vs1;
```
jordancarlin, line 9: Can we delete the old version or is there a reason to keep it as a comment?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vrgather_vd_vs1_overlap.sv
```diff
@@ -4,12 +4,14 @@
...
-    cp_ssstrictv_vrgather_vd_vs1_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs1_vd_overlap_lmul1;
+    // cp_ssstrictv_vrgather_vd_vs1_overlap_lmul1: cross std_trap_vec, vtype_lmul_1, vd_eq_vs1;
 
-    cp_ssstrictv_vrgather_vd_vs1_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs1_vd_overlap_lmul2;
+    // cp_ssstrictv_vrgather_vd_vs1_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs1_vd_overlap_lmul1;
 
-    cp_ssstrictv_vrgather_vd_vs1_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs1_vd_overlap_lmul4;
+    // cp_ssstrictv_vrgather_vd_vs1_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs1_vd_overlap_lmul2;
+
+    // cp_ssstrictv_vrgather_vd_vs1_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs1_vd_overlap_lmul4;
```
jordancarlin, line 17: Why are these all commented out?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vrgather_vd_vs2_eq.sv
```diff
@@ -3,6 +3,7 @@
 // //////////////////////////////////////////////////////////////////////////////////////////////////////////
 
     // vrgather with vd==vs2 reserved (vd source overlap)
-    cp_ssstrictv_vrgather_vd_vs2_eq : cross std_trap_vec, vtype_all_lmulge1, vd_eq_vs2;
+    // cp_ssstrictv_vrgather_vd_vs2_eq : cross std_trap_vec, vtype_all_lmulge1, vd_eq_vs2;
+    cp_ssstrictv_vrgather_vd_vs2_eq : cross std_trap_vec, vd_eq_vs2;
```
jordancarlin, line 9: Can we delete the old version or is there a reason to keep it as a comment?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vrgather_vd_vs2_overlap.sv
```diff
@@ -4,12 +4,14 @@
...
-    cp_ssstrictv_vrgather_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
+    // cp_ssstrictv_vrgather_vd_vs2_overlap_lmul1: cross std_trap_vec, vtype_lmul_1, vd_eq_vs2;
 
-    cp_ssstrictv_vrgather_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+    // cp_ssstrictv_vrgather_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
 
-    cp_ssstrictv_vrgather_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
+    // cp_ssstrictv_vrgather_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+
+    // cp_ssstrictv_vrgather_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
```
jordancarlin, line 17: Why are these all commented out?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vslide1up_vd_vs2_overlap.sv
```diff
@@ -4,12 +4,14 @@
...
-    cp_ssstrictv_vslide1up_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
+    // cp_ssstrictv_vslide1up_vd_vs2_overlap_lmul1: cross std_trap_vec, vtype_lmul_1, vd_eq_vs2;
 
-    cp_ssstrictv_vslide1up_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+    // cp_ssstrictv_vslide1up_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
 
-    cp_ssstrictv_vslide1up_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
+    // cp_ssstrictv_vslide1up_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+
+    // cp_ssstrictv_vslide1up_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
```
jordancarlin, line 17: Why are these all commented out?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_vslideup_vd_vs2_overlap.sv
```diff
@@ -4,12 +4,14 @@
...
-    cp_ssstrictv_vslideup_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
+    // cp_ssstrictv_vslideup_vd_vs2_overlap_lmul1: cross std_trap_vec, vtype_lmul_1, vd_eq_vs2;
 
-    cp_ssstrictv_vslideup_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+    // cp_ssstrictv_vslideup_vd_vs2_overlap_lmul2: cross std_trap_vec, vtype_lmul_2, vs2_vd_overlap_lmul1;
 
-    cp_ssstrictv_vslideup_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
+    // cp_ssstrictv_vslideup_vd_vs2_overlap_lmul4: cross std_trap_vec, vtype_lmul_4, vs2_vd_overlap_lmul2;
+
+    // cp_ssstrictv_vslideup_vd_vs2_overlap_lmul8: cross std_trap_vec, vtype_lmul_8, vs2_vd_overlap_lmul4;
```
jordancarlin, line 17: Why are these all commented out?

jordancarlin (dismissed):
Not focusing too much on the Python since that is all going to be refactored as part of the port to standard testgen. Here are comments on the rest of this.

generators/coverage/src/covergroupgen/templates/vector/priv/cp_exceptionsv_vd_vs1_overlap.sv
fourth-bit: The idea was to leave everything in a state that could easily repurposed for a stricter version of Ssstrict (e.g. testing traps that are contingent on the value in lmul) instead of just deleting the work currently in the repo. Its also accessible through the commit history if I delete it, so I can go ahead and delete it if that's better.

jordancarlin:
@fourth-bit would be good to get this merged soon. Can you address all of the comments from me and Copilot?

generators/coverage/src/covergroupgen/templates/vector/priv/cp_ssstrictv_masking_vs1_eq_v0.sv
fourth-bit: It doesn't work and I'm not entirely sure why something like 
```systemverilog
cp_vd : coverpoint ins.get_vr_reg(ins.current.vd)  iff (ins.trap == 0 ) 
```
works while it doesn't work here. I suspect it has something to do with how SsstrictV is sampled or with the iff guard. The problem is that initially `ins.current.v__` is an empty string, so `ins.get_vr_reg` causes a fatal error in coverage. Not sure what the right fix is.

jordancarlin: Most likely the issue is that these are not decoded by the disassembler.svh file, so there is no string to parse. I think that is what the fallback decoder was trying to solve, but that file is kind of hacky and probably needs to be reworked. I wonder if we should just be more relaxed in the decoder and accept the reserved variants when decoding instructions normally?

fourth-bit: That seems rational, but it seems like the list of operations used in `dissassemble.svh` is autogenerated by [riscv/riscv-opcodes](https://github.com/riscv/riscv-opcodes). It doesn't seem maintainable to edit that file whenever we need a new version of the opcodes. Maybe build_disasm_fallback could generate constants like in `RISCV_imported_decode_pkg.svh` instead so that the disassembler could use them.

jordancarlin: Agreed. Let's leave it as is for now and open an issue to revisit this (and the fallback decoder in general).

jordancarlin (commented):
**AI Review**

 Blocking issues

 1. Vx8 vrgatherei16.vv coverage treats vs1 as SEW=8, but its index EEW is always 16
     - testplans/Vx.csv:209 now enables EFFEW8 for vrgatherei16.vv.
     - Generated coverpoints/unpriv/Vx8_coverage.svh:28115-28164 uses 8-bit bins for vs1 index corner cases, and :28268-28282 includes LMUL=8.
     - For vrgatherei16.vv at SEW=8, vs1 has EEW=16 and EMUL=2*LMUL; LMUL=8 implies EMUL=16 and is reserved. Also the index edge bins should be 16-bit (0xffff, 0x0002), not 8-bit (0xff, 0x02).
     - Suggest using a special coverpoint/testplan entry for vrgatherei16.vv that caps LMUL at 4 for SEW=8 and sizes vs1 edge/corner checks by EEW=16.

 2. cp_ssstrictv_masking_vs2_eq_v0 is applied to mask-source instructions where vs2=v0 is legal
     - The template says vs2=v0 with vm=0 is reserved because it overlaps mask v0 (cp_ssstrictv_masking_vs2_eq_v0.sv:5-7), and the generator blindly emits it (cp_ssstrictv_masking_vd_eq_v0.py:54-60).
     - That is true for normal vector sources because the same register is read as EEW=SEW and EEW=1, but not for mask-source instructions where vs2 itself has EEW=1.
     - testplans/priv/SsstrictV.csv currently marks this CP for vcpop.m, vfirst.m, vmsbf.m, vmsif.m, vmsof.m, and viota.m. Those should not be Ssstrict reserved cases solely because vs2=v0 under masking.

 3. cp_ssstrictv_ls_eew_lt_sewmin has coverpoint bins the generator never emits
     - Template cp_ssstrictv_ls_eew_lt_sewmin.sv:24-43 creates standard, vl=0, and vstart>=vl crosses.
     - Generator cp_ssstrictv_ls_family.py:265-278 only emits the standard vl=1, vstart=0 case.
     - On any SEWMIN>8 configuration, the *_vl0 and *_vstart_ge_vl crosses look unhit/uncoverable unless the generator adds those cases or the template removes those extra crosses.

 Nits / maintainability

 - Some cross names still include _lmul1 even after the LMUL coverpoint was removed (e.g. cp_ssstrictv_vcompress_vd_vs2_overlap_lmul1), which is now misleading.
 - The masking_vs{1,2}_eq_v0 comments should cite the actual source-EEW conflict rule and note the mask-source exception.

generators/testgen/scripts/priv/_ssstrictv_helpers.py
```diff
@@ -209,8 +185,59 @@ def make_dest_zero_overrides(instruction: str) -> dict:
...
+    register = vec_data['vs2']['reg']
+
+    vtypeReg = common.pickPrivScratch(scalar_data, (scratch,))
+    vlmaxReg = common.pickPrivScratch(scalar_data, (scratch, vtypeReg))
+    avlReg = scratch
+
+    if   sew == 8  : sew_aligned = -1#"0x1F"
+    elif sew == 16 : sew_aligned = -2#"0x1E"
+    elif sew == 32 : sew_aligned = -4#"0x1C"
+    elif sew == 64 : sew_aligned = -8#"0x18"
```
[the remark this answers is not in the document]
fourth-bit: This causes compile errors as the immediate is supposed to be signed

fourth-bit:
Removed the vrgatherei16 changes as that is too big of an unrelated change for this PR. I removed the legal overlaps with v0, I corrected cp_ssstrictv_ls_eew_lt_sewmin, and fixed both of the nits.

jordancarlin:
I'm going to hold off on merging this until https://github.com/riscv/riscv-isa-manual/issues/3258#issuecomment-5040391787 is resolved.

jordancarlin:
Looks like we might need to keep most of these after all: https://github.com/riscv/riscv-isa-manual/issues/3258#issuecomment-5052593171

fourth-bit:
What does that imply Ssstrict covers? That case would have been covered by this set of tests because it is an instruction encoding that is reserved in all csr states. Does this mean I should also consider SsstrictV to be tests of all reserved vector instruction encodings, dependent on CSRs?

davidharrishmc (approved):

PR #1755, opened 2026-06-19
Integrate Vx into Main Testgen

fourth-bit:
This PR generates the Vx test suite in the main ACT4 testgen module. It revamps the previous instruction type system to include more granular instruction types that provide information about overlap constraints in ACT4, e.g. a mask producing vector - vector operation is now specified as:
```python
mvv_config = InstructionTypeConfig(
    required_params={"vd", "vs1", "vs2"},
    vector_data=VectorTypeConfig(
        mask_regs={"vd"},
        masked_constraints={("vs1", "v0"), ("vs2", "v0")},
    ),
)
```
These new systems allow for a more explicit specification of the constraints on vector operations. These changes to instruction types did not introduce any differences in the generated coverage.

The randomization logic, now present in `formatters/vector_params.py`, is roughly the same as what was present before in `vector_testgen_common.py`. I made some attempts to simplify the logic, and reduce the monolithic nature of the function calls. In the future, this logic should be more tightly tied to the `VectorRegisterFile`, and more simply allow for generating random register assignments without needing overlap checks after the randomization. 

The only meaningful test change that is made in this PR is changing the signature to take the mask register as a macro argument. This impacts tests of `cp_custom_vmask_write_v0_masked` where v0 is written to during the instruction. This allows the mask value to be accurately checked for those instructions.


jordancarlin:
> It feels like the major effort here is going to have to be figuring out how to put all of the information that is stored in these lists into classes. We may need more targeted instruction types in the csv files, probably something more akin to what sail uses to decode instructions, rather than the general instruction types that we currently have in the csv (for example VVVM includes narrowing, widening, and normal instructions). This would help create a better system to store the information about all of these instructions.

We had to do something similar for the integer instructions and defined many very similar types, so that seems like a reasonable approach for vector.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,183 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
     """
     test_data.int_regs.return_registers(params.used_int_regs)
     test_data.float_regs.return_registers(params.used_float_regs)
+    test_data.vec_regs.return_registers(params.used_vector_regs)
```
jordancarlin: Be consistent with the naming. Either use vec_regs or vector_regs for everything.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,183 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    # Preloads Require Special Handling for V
+    if lmul is not None and vl is not None:
+        if vl == "vlmax":
+            lines.append(f"vsetvli x{temp_reg}, x0, e{sew}, m{_lmul_flag(lmul)}, tu, mu")
+        else:
+            lines.append(f"vsetvli x{temp_reg}, {vl}, e{sew}, m{_lmul_flag(lmul)}, tu, mu")
+
+    lines.extend(
+        [
+            f"la x{temp_reg}, {val_pointer}",
```
jordancarlin: Make sure to use the LA macro everywhere.

generators/testgen/src/testgen/io/testplans.py
```diff
@@ -24,6 +24,45 @@ def get_extensions(testplan_dir: Path) -> list[str]:
     return extensions
 
 
+def get_vector_extensions(testplan_dir: Path, *, priv: bool) -> list[str]:
+    if priv:
+        testplan_dir = testplan_dir / "priv"
+    testplans = []
+    for file in testplan_dir.glob("*"):
+        if file.suffix == ".csv":
+            arch = file.name[: file.name.rfind(".csv")]
```
jordancarlin:
```suggestion
            arch = file.name.stem
```

generators/testgen/src/testgen/io/testplans.py
```diff
@@ -24,6 +24,45 @@ def get_extensions(testplan_dir: Path) -> list[str]:
     return extensions
 
 
+def get_vector_extensions(testplan_dir: Path, *, priv: bool) -> list[str]:
+    if priv:
+        testplan_dir = testplan_dir / "priv"
+    testplans = []
+    for file in testplan_dir.glob("*"):
```
jordancarlin:
```suggestion
    for file in testplan_dir.glob("*.csv"):
```

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,183 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    lmulflag = str(lmul)
+
+    if lmul == 0.5:
+        lmulflag = "f2"
+    elif lmul == 0.25:
+        lmulflag = "f4"
+    elif lmul == 0.125:
+        lmulflag = "f8"
+
+    return lmulflag
```
jordancarlin: Could this function simplify to something like the following:

```suggestion
def _lmul_flag(lmul: float) -> str:
    if lmul < 1:
        return f"f{1/lmul}"
    else:
        return str(lmul)
```

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,183 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+) -> list[str]:
+    lines = []
+
+    assert (lmul is None) == (vl is None), "When Setting Preload Information, both lmul and vl must be set together"
+    # Preloads Require Special Handling for V
+    if lmul is not None and vl is not None:
+        if vl == "vlmax":
+            lines.append(f"vsetvli x{temp_reg}, x0, e{sew}, m{_lmul_flag(lmul)}, tu, mu")
+        else:
+            lines.append(f"vsetvli x{temp_reg}, {vl}, e{sew}, m{_lmul_flag(lmul)}, tu, mu")
```
jordancarlin: Might be simpler like this so you don’t need to repeat the rest of the line in both the if and else blocks.
```suggestion
        lines.append(f"vsetvli x{temp_reg}, {'x0' if vl == 'vlmax' else 'vl'}, e{sew}, m{_lmul_flag(lmul)}, tu, mu")
```

generators/testgen/src/testgen/generate/unpriv.py
```diff
@@ -83,6 +84,81 @@ def generate_unpriv_extension_tests(
...
+        xlen: Target XLEN (32 or 64)
+        E_ext: Whether to generate RV32E tests
+        testsuite: Testsuite to generate tests for (e.g., 'I', 'M', 'ZcbM', 'MisalignD')
+        testplan_dir: Directory containing testplan CSV files
+        output_test_dir: Directory to output generated tests
+    """
+    # Read testplan for this testsuite
+    match = re.search(r"([^0-9]*)\d*$", testsuite)
+    if match is None:
+        return  # This branch shouldn't be taken ever because we are matching with *s
```
jordancarlin: If it should never be taken, then it is better for it to throw an error instead of silently failing.

generators/testgen/src/testgen/io/testplans.py
```diff
@@ -24,6 +24,45 @@ def get_extensions(testplan_dir: Path) -> list[str]:
...
+                is_vector = arch.startswith(("ExceptionsV", "SsstrictV", "MisalignedV", "V", "Zv"))
+            else:
+                is_vector = arch.startswith(("V", "Zv"))
+            if is_vector:
+                if "Vx" in arch and not arch.startswith("Exceptions") and not arch.startswith("Ssstrict"):
+                    for effew in ["8", "16", "32", "64"]:
+                        testplans.append("Vx" + effew)
+                elif arch == "Vls":
+                    for effew in ["8", "16", "32", "64"]:
+                        testplans.append("Vls" + effew)
```
jordancarlin: Combine and use `arch` instead of hardcoding `"Vx"`.

generators/testgen/src/testgen/io/testplans.py
```diff
@@ -24,6 +24,45 @@ def get_extensions(testplan_dir: Path) -> list[str]:
...
+                elif arch == "ExceptionsVf":
+                    # Mirror unpriv Vf: expand into per-SEW pseudo-extensions so
+                    # each generated test runs vector-FP at a non-reserved SEW
+                    # (SEW=8 is reserved for FP). The driver filters instructions
+                    # by EFFEW{N} and emits ExceptionsVf{N}_rv{xlen}.S.
+                    for effew in ["16", "32", "64"]:
+                        testplans.append("ExceptionsVf" + effew)
+                elif arch in ["Zvbb", "Zvkb"]:
+                    for effew in ["8", "16", "32", "64"]:
+                        testplans.append(arch + effew)
```
jordancarlin: This can get merged in above too.

generators/testgen/src/testgen/io/testplans.py
```diff
@@ -24,6 +24,45 @@ def get_extensions(testplan_dir: Path) -> list[str]:
...
+                elif arch == "Vf":
+                    for effew in ["16", "32", "64"]:
+                        testplans.append("Vf" + effew)
+                elif arch == "ExceptionsVf":
+                    # Mirror unpriv Vf: expand into per-SEW pseudo-extensions so
+                    # each generated test runs vector-FP at a non-reserved SEW
+                    # (SEW=8 is reserved for FP). The driver filters instructions
+                    # by EFFEW{N} and emits ExceptionsVf{N}_rv{xlen}.S.
+                    for effew in ["16", "32", "64"]:
+                        testplans.append("ExceptionsVf" + effew)
```
jordancarlin: These ones can also be combined.

generators/testgen/src/testgen/cli.py
```diff
@@ -50,6 +50,16 @@ class UnprivTask:
...
 
+@dataclass
+class UnprivVectorTask:
+    """Task for generating unprivileged tests."""
+
+    xlen: int
+    testsuite: str
+    testplan_dir: Path
+    output_test_dir: Path
+
```
jordancarlin, line 52: Is a new `UnprivVectorTask` really needed? It seems like the builder this dispatches to (`generate_unpriv_vector_tests`) is very similar to the standard unpriv one with just a few extra function calls that could be guarded by `if is_vector`.

jordancarlin (changes requested):
I haven't reviewed most of the code in detail yet, but here are a few comments to help steer this in the right direction as you continue working on it. Looks like a great start!

generators/testgen/src/testgen/asm/helpers.py
fourth-bit: The second one should have been `vsetivli` so I think they need to stay distinct

generators/testgen/src/testgen/asm/helpers.py
fourth-bit: Should I also be using LI in place of li everywhere?

jordancarlin: Yes. That ensures consistent results across compilers.

fourth-bit:
This PR is now ready for a full review. Currently, it supports Vx with 100% coverage and is passing on all sims. Here is an assembly file that it generated: [Vx32-vadd.vv-00.S.gz](https://github.com/user-attachments/files/29524287/Vx32-vadd.vv-00.S.gz)

The generated assembly for when the vtype needs to change during the setup is very clunky still, like in this file: [Vx32-vwadd.vv-00.S.gz](https://github.com/user-attachments/files/29524318/Vx32-vwadd.vv-00.S.gz).

tests/env/signature.h
jordancarlin, line 1: Why are these changes to the signature logic needed? Why weren't these issues found with the old testgen if the new tests are supposed to be the same?

generators/testgen/src/testgen/constants.py
```diff
@@ -131,3 +131,10 @@ def get_flen_for_extension(extension: str) -> int:
         "cp_fclass",
     }
 )
+
+# =============================================================================
+# Vector Configuration
+# =============================================================================
+MIN_SEW_MIN = 8
+ELEN = 64
```
jordancarlin: Why is ELEN a constant? Isn't ELEN a DUT parameter?

generators/testgen/src/testgen/io/testplans.py
```diff
@@ -18,12 +18,44 @@ def get_extensions(testplan_dir: Path) -> list[str]:
     extensions: list[str] = []
     for testplan in testplan_dir.glob("*.csv"):
         extension = testplan.stem
-        # TODO: Remove once vector testgen is merged
         if ("V" not in extension and "Zv" not in extension) or extension == "ZfaZvfh":
             extensions.append(extension)
```
jordancarlin: What about something like the following:

```suggestion
        if extension.startswith(("V", "Zv")):
            extensions.extend(expand_vector_extensions(extension))
        else:
            extensions.append(extension)
```

Where `expand_vector_extensions` is something along these lines:
```py
def expand_vector_extensions(extension: str) -> list[str]:
    if extension in ["Vx", "Vls", "Zvbb", "Zvkb"]:
          return [extension + effew for effete in ("8", "16", "32", "64")]
    elif extension == "Vf":
          return [extension + effew for effete in ("16", "32", "64")]
    elif extension == "Zvknhb":
          return [extension + effew for effete in ("32", "64")]
    else:
         return [extension]
```

Then you can drop `get_vector_extensions` here and simplify `cli.py` to be completely agnostic to vector.

generators/testgen/src/testgen/io/templates.py
```diff
@@ -79,7 +85,12 @@ def insert_footer_template(test_data_section: str, test_string_section: str) ->
 
 
 def canonicalize_extensions(
-    testsuite: str, xlen: int, E_ext: bool, required_extensions: list[str] | None = None
+    testsuite: str,
+    xlen: int,
+    E_ext: bool,
+    required_extensions: list[str] | None = None,
+    sew: int | None = None,
+    instruction: str | None = None,
```
jordancarlin: Be consistent with the naming. This switches between `instr_name` and `instruction` in the different functions in this file but is really the same thing.

generators/testgen/src/testgen/cli.py
```diff
@@ -124,7 +130,11 @@ def generate_all_tests(
             for testsuite in sorted(unpriv_ext_list):
                 if E_ext and testsuite not in E_EXTENSION_TESTS:
                     continue
-                tasks.append(UnprivTask(xlen, E_ext, testsuite, testplan_dir, output_test_dir))
+                tasks.append(UnprivTask(xlen, E_ext, testsuite, testplan_dir, output_test_dir, False))
+
+    for xlen in [32, 64]:
```
jordancarlin: See my comment in `testplans.py`, but I think you can drop all of this vector logic.

generators/testgen/src/testgen/io/templates.py
```diff
@@ -108,14 +121,93 @@ def canonicalize_extensions(
...
+
+        # Get the most minimal V subextension for Vx, Vls, and Vf (e.g. Zve32f for Vf32)
+        maybe_mapped = get_vector_base_extension(testsuite, instruction, xlen, sew)
+        if maybe_mapped is not None:
+            ext_components.extend(maybe_mapped)
+            no_sew_suffix = re.sub(r"\d+$", "", testsuite)
+            if no_sew_suffix in ext_components:
+                ext_components.remove(no_sew_suffix)
+
+    if any(ext in ext_components for ext in ["V", "Zvfh"]) or testsuite.startswith(("V", "Zv")):
```
jordancarlin: I think this is functionally the same:

```suggestion
    if any(ext.startswith(("V", "Zv")) for ext in ext_components):
```

generators/testgen/src/testgen/io/templates.py
```diff
@@ -108,14 +121,93 @@ def canonicalize_extensions(
...
+        assert sew is not None, "SEW must be set for all unpriv vector tests"
+        assert instruction is not None, "Passing an instruction is required for all vector extensions"
+
+        # Get the most minimal V subextension for Vx, Vls, and Vf (e.g. Zve32f for Vf32)
+        maybe_mapped = get_vector_base_extension(testsuite, instruction, xlen, sew)
+        if maybe_mapped is not None:
+            ext_components.extend(maybe_mapped)
+            no_sew_suffix = re.sub(r"\d+$", "", testsuite)
+            if no_sew_suffix in ext_components:
+                ext_components.remove(no_sew_suffix)
```
jordancarlin, line 137: A comment explaining this part would be helpful.

generators/testgen/src/testgen/io/templates.py
```diff
@@ -108,14 +121,93 @@ def canonicalize_extensions(
...
         ext_components.append("M")  # Add M if V is present (required for gcc 15)
 
     ext_components = list(dict.fromkeys(ext_components))  # Remove duplicates while preserving order
 
     return ext_components, params
 
 
+def get_vector_base_extension(testsuite: str, instruction: str, xlen: int, sew: int) -> list[str] | None:
+    """
+    Helper function to derive the smallest possible V extension containing an instruction. This is necessary because
```
jordancarlin:
```suggestion
    Helper function to derive the smallest possible vector extension containing an instruction. This is necessary because
```

generators/testgen/src/testgen/io/templates.py
```diff
@@ -108,14 +121,93 @@ def canonicalize_extensions(
...
+        # those with EEW=64 floating-point operands.
+        # The first part of this requirement is handled by placing those operands into Vf.
+        if (
+            zve_ext in mapped
+            and instruction in ["vfmv.f.s", "vfmv.s.f", "vfslide1up.vf", "vfslide1down.vf"]
+            and sew == 64
+        ):
+            mapped.remove(zve_ext)
+
+    if "Zve32x" in mapped and instruction and instruction.startswith(("vw", "vn")) and sew == 32:
```
jordancarlin:
```suggestion
    if "Zve32x" in mapped and instruction.startswith(("vw", "vn")) and sew == 32:
```

generators/testgen/src/testgen/io/templates.py
```diff
@@ -108,14 +121,93 @@ def canonicalize_extensions(
...
+        mapped.remove("Zve32x")
+        mapped.append("Zve64x")
```
jordancarlin: This should have a comment with the rule from the spec like all of the other blocks in this function.

generators/testgen/src/testgen/generate/unpriv.py
```diff
@@ -66,29 +79,68 @@ def generate_unpriv_extension_tests(
...
+
+    for pattern, sew in [
+        (r"Zvfbfmin$", 16),
+        (r"Zvfhmin$", 16),
+        (r"Zvfbfwma$", 16),
+        (r"Zvk(g|nha|ned|sed|sh)$", 32),  # codespell:ignore ned
+    ]:
+        match = re.search(pattern, testsuite)
+        if match:
+            return sew
```
jordancarlin: Could this be simplified to just the second loop of matches (Zvfbfmin, etc.) and then if none of those match just extract a number from the end of the string? That way all of the extensions don't need to be listed and this is easier to maintain.

generators/testgen/src/testgen/generate/unpriv.py
```diff
@@ -100,14 +152,15 @@ def _generate_unpriv_tests_for_instruction(
...
     """
     test_data = TestData(test_config, instr_name)
     all_test_chunks: list[TestChunk] = []
 
     # Iterate through each coverpoint and generate test chunks
     for coverpoint in coverpoints:
-        # Skip cp_asm_count if mixed with other coverpoints
-        if coverpoint == "cp_asm_count" and len(coverpoints) > 1:
+        # Skip cp_asm_count and std_vec if mixed with other coverpoints
+        if (coverpoint == "cp_asm_count" or coverpoint == "std_vec") and len(coverpoints) > 1:
```
jordancarlin:
```suggestion
        if coverpoint in ["cp_asm_count", "std_vec"] and len(coverpoints) > 1:
```

generators/testgen/src/testgen/generate/unpriv.py
```diff
@@ -116,7 +169,20 @@ def _generate_unpriv_tests_for_instruction(
...
+        if is_vector:
+            assert test_config.sew is not None, "SEW must be set for vector tests"
+            sew = test_config.sew
+            vdsew = sew
+            info = extract_instruction_info(instr_name, instr_type)
+            if "vd" in info.widened_regs:
+                vdsew *= 2
+            elif info.load_store_eew == 64:
+                vdsew = 64
+            extra_defines = ["#define RVTEST_VECTOR", f"#define RVTEST_SEW {sew}", f"#define VDSEW {vdsew}"]
```
jordancarlin: I know we are working on dropping `RVTEST_VECTOR` (it may already not be needed). This might be a good time to finish that work.

generators/coverage/src/covergroupgen/generate.py
```diff
@@ -158,6 +158,38 @@ def _sew_variants_for(arch: str) -> list[str] | None:
     "VSXM": (True, False, False),
     "VSXVM": (True, False, True),
     "VSXXM": (True, False, False),
+    # Vector integer types from Vx.csv
+    "WVV": (True, True, True),
+    "WVX": (True, False, True),
+    "WVV_ACC": (True, True, True),
```
jordancarlin, line 164: It feels like there must be a better way for this to work than listing out every vector type in covergroupgen. Fine for now since that is how the vector implementation already works, but please put this on your list to look into later.

generators/coverage/src/covergroupgen/templates/vector/cr_vl_lmul_nomask_sew8.sv
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    cp_csr_vl_edges : coverpoint vl_check(ins.hart, ins.issue)  iff (ins.trap == 0 )  {
+        // Edges values of VL (vector length)
+        bins one        = {vl_one       };
+        bins vlmax      = {vl_vlmax     };
+        bins legal      = {vl_legal     };
+    }
+
+    cr_vl_lmul_sew8 : cross cp_csr_vtype_lmul_all_sew8_lmul_le_8, cp_csr_vl_edges  iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vsew") == 0)  {
```
jordancarlin: You probably don't need the `iff (ins.trap == 0 & get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "vtype", "vsew") == 0)` in both the cross and one of the constituent coverpoints.

jordancarlin (changes requested):
Still working my way through this (there is a lot to review!), but here is the next round of feedback. I will try to review `vector_params` on the my next pass.

tests/env/signature.h
fourth-bit: The main change to the signature is adding the ability to check against the mask register even when it was overwritten. I'm not sure why this didn't cause issues before, but it was a logical error that came up when the register file was stopping me from doubly allocating v0 (because vd was already there).

generators/testgen/src/testgen/generate/unpriv.py
fourth-bit: Its only needed in the failure code. It looks like they should just be replaced with `ZVL32B_SUPPORTED`

jordancarlin: Can you open an issue to do this so we don't forget about it?

generators/coverage/src/covergroupgen/generate.py
jordancarlin: Maybe open an issue so this doesn't get forgotten?

jordancarlin:
I will review this again soon. Looks like there are also merge conflicts that need to be resolved.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
     """
     test_data.int_regs.return_registers(params.used_int_regs)
     test_data.float_regs.return_registers(params.used_float_regs)
+    test_data.vec_regs.deallocate_operands()
+    assert len(test_data.vec_regs.reg_list) == 32, (
+        f"Not all vector registers returned: {len(test_data.vec_regs.reg_list)} remaining, they are {test_data.vec_regs.reg_list}"
```
jordancarlin, line 149: Why is the assert needed for vector registers but not the other register files? Shouldn't this already be covered when the register file is destroyed?

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+
+    return lines
+
+
+def load_vxrm(vxrm: str) -> list[str]:
+    """
+    Generates the instructions to load an individual vxrm.
+    """
+    vxrm_set_map = {"rod": "0x6", "rdn": "0x4", "rne": "0x2", "rnu": "0x0"}
+    return [f"# Clear vxrm and vcsr, then set vxrm = {vxrm}", "csrci vcsr, 0x6", f"csrsi vcsr, {vxrm_set_map[vxrm]}"]
```
jordancarlin: Put each assembly line on its own line. Ruff will do that for you if you put a comma after the last element.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+) -> list[str]:
+    """
+    Load a vector register.
+
+    Args:
+        register: The number register to load
+        val_pointer: A string that points to a label where the load data exists
+        params: The InstructionParams generated for this instruction. Gives data about default sew and lmul
+
+    Optional Args:
```
jordancarlin: Missing sew_override in docstring.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    assert test_data.test_chunk is not None, "No active test chunk — call begin_test_chunk() first"
+
+    sig_reg = test_data.int_regs.sig_reg
+    link_reg = test_data.int_regs.link_reg
+    temp_reg = test_data.int_regs.temp_reg
+    label = test_data.current_testcase_label
+    vdsew = params.sew * (2 if widen_vd else 1)
+
+    egs = params.egs if params.egs is not None else 1
+    worst_bytes = vdsew * egs // 8
```
jordancarlin: Maybe `max_bytes` would be more obvious?

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    sig_stride = max(test_data.xlen, test_data.flen) // 8 if test_data.flen > 0 else test_data.xlen // 8
```
jordancarlin: Is the `if` needed? Won't `test_data.xlen` always be greater than `test_data.flen` is 0?

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    offset_bytes = (worst_bytes + 4 + 7) & ~7
```
jordancarlin: Add a comment explaining the magic numbers. Why `+4+7 & ~7`?

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    test_data.test_chunk.sigupd_count += max(1, (offset_bytes + sig_stride - 1) // sig_stride)
+
+    vtmp, mtmp = test_data.vec_regs.get_registers(2, lmul=1, exclude_regs=[0])
+
+    lines = [
+        f"vsetivli x0, 1, e{vdsew}, m1, tu, mu",
+        f"# set SEW={vdsew}, LMUL=1, VL=1 before signature check",
```
jordancarlin: Put the comment on the same line as the relevant code or on the line before the relevant code, not on the line after the code. Global change.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+        "# RVTEST_SIGUPD_V(_CMP, _SIG_PTR, _LINK_REG, _TEMP_REG, _VTMP, _MTMP, _SEW, _VREG, _INST_PTR, _STR_PTR)",
```
jordancarlin: Why is there a commented out sigupd macro? Drop this.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    ]
+
+    if mask_producing:
+        lines.extend(
+            [
+                f"RVTEST_SIGUPD_V(vmxor.mm, x{sig_reg}, x{link_reg}, x{temp_reg}, v{vtmp}, v{mtmp}, {vdsew}, v{params.vd}, {label}, {label}_str)",
+                f"# Check if v{params.vd} contains the expected result. x{sig_reg} is the signature ptr, x{link_reg} is the link ptr, x{temp_reg} is a temp reg.",
```
jordancarlin: Extend this description to include the vector specific parts and the other registers.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+        segments: The number of segments used in the instruction. When there are multiple, the sigupd
+            macro may have to be emitted multiple times. (TODO: make this an optional arg)
+        lmul: The lmul that this sigupd should check. Note that this can be different from the test lmul
+
+    Optional Args:
+        widen_vd: When active, SIGUPD will run with 2*SEW as the element width
+        vcompress: Active when the instruction under test is vcompress because vcompress needs special handling
+        scalar_dest: Active when vd only contains one element and needs different handling.
+        mask_reg: Set to the value where the mask register lies (0 by default). Used when the mask is overwritten
+            by the instruction under test.
```
jordancarlin: Missing `mask_producing`. Global comment: double check all of your docstrings. Several have incorrect or missing arguments.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    """
+    assert params.sew is not None
+    assert test_data.test_chunk is not None, "No active test chunk — call begin_test_chunk() first"
+
+    vdsew = params.sew * (2 if widen_vd else 1)
+
+    emul_for_bytes = int(lmul) if lmul is not None and lmul >= 1 else 1
```
jordancarlin: The function call says it requires lmul (type is float, not float | None), so the `is not None` can be dropped.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    worst_bytes = VLEN_MAX * emul_for_bytes // 8
+    sig_stride = max(test_data.xlen, test_data.flen, vdsew) // 8 if test_data.flen > 0 else test_data.xlen // 8
```
jordancarlin: This line doesn't make much sense to me. Why would we want to ignore vdsew if flen is 0?

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+                f"RVTEST_SIGUPD_V(vmsne.vv, x{sig_reg}, x{link_reg}, x{temp_reg}, v{vtmp}, v{mtmp}, {vdsew}, v{params.vd}, {label}, {label}_str)",
+                f"# Check if v{params.vd} contains the expected result. x{sig_reg} is the signature ptr, x{link_reg} is the link ptr, x{temp_reg} is a temp reg.",
+            ]
+        )
+
+    test_data.vec_regs.return_registers([vtmp, mtmp])
+    return lines
+
+
+def write_sigupd_v_len(
```
jordancarlin: There is a lot of duplication between the write_sigupd_v_* functions. I wonder if some of it can be factored out into a helper function so it doesn't get out of sync.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    """
+    Variant of SIGUPD for mask producing operations that nops in self-check and puts an additional value
+    in the signature in non-self checking mode. This ensures that we can test both the behavior at vlmax
+    and at the test vl, as a valid implementation can do either.
+    """
+    assert params.sew is not None, "SEW must be set for vector tests"
+    assert params.lmul is not None, "LMUL must be set for vector tests"
+    assert test_data.test_chunk is not None, "No active test chunk — call begin_test_chunk() first"
+
+    # Unfortunately, I think this is still required, as the same macro as length suite is used to advance sigupd
```
jordancarlin: Confusing comment.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    emul_for_bytes = int(params.lmul) if params.lmul is not None and params.lmul >= 1 else 1
+    worst_bytes = VLEN_MAX * emul_for_bytes // 8
+    sig_stride = max(test_data.xlen, test_data.flen, params.sew) // 8 if test_data.flen > 0 else test_data.xlen // 8
```
jordancarlin: Why ignore sew if flen is 0?

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    assert params.lmul is not None, "lmul must be set for vector instructions"
+    lmul_flag = "m" + _lmul_flag(params.lmul)
+
+    mask_flags = ""
+    if params.ta is not None:
+        mask_flags += ", ta" if params.ta else ", tu"
+    if params.ma is not None:
+        mask_flags += ", ma" if params.ma else ", mu"
+    else:
+        mask_flags = ", ta, ma"
```
jordancarlin: Would it be simpler to express this as the following? This is confusing logic right now. I think this ends up being the same.
```py
if param.ta is False:
    mask_flags += ", tu"
else:
    mask_flags += ", ta"
if param.ma is False:
    mask_flags += ", mu"
else:
    mask_flags += ", ma"
```

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+    else:
+        mask_flags = ", tu, mu"
+
+    flags = lmul_flag + mask_flags
+    vl_register_or_imm: str | int = 0
+
+    if params.vl == "random":
+        temp_reg, vlmax_reg = test_data.int_regs.get_registers(2, exclude_regs=[0])
+
+        randomVl = random.getrandbits(32)
```
jordancarlin: Try to use the random functions defined in `testgen/data/random.py` to ensure consistency with the random seed. You can add additional functions to that file if needed.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+        lines.append(f"vsetvli x0, x{temp_reg}, e{params.sew}, {flags}")
+
+        test_data.int_regs.return_registers([vlmax_reg])
+        vl_register_or_imm = f"x{temp_reg}"
+    elif params.vl == "vlmax":
+        lines.extend(["# Load Vl=VLMAX", f"vsetvli x{params.temp_reg}, x0, e{params.sew}, {flags}"])
+        vl_register_or_imm = "x0"
+    else:
+        vl = params.vl if params.vl is not None else 1
+        assert isinstance(vl, int), "vl must be an integer if it is not 'vlmax' or 'random'"
```
jordancarlin: How about making the type of vl `Literal["vlmax", "random"] | int` so that this is guaranteed and the assert is not needed?

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+                f"LI (x{params.temp_reg}, {vl})",
+                f"vsetvli x0, x{params.temp_reg}, e{params.sew}, {flags}",
+            ]
+        )
+
+        vl_register_or_imm = vl
+
+    if params.vstart is not None:
+        lines.extend(
+            ["# Load Desired Vstart", f"LI (x{params.temp_reg}, {params.vstart})", f"csrw vstart, x{params.temp_reg}"]
```
jordancarlin: vstart should be lowercase. Put the different assembly lines on different lines.

generators/testgen/src/testgen/asm/helpers.py
jordancarlin, line 1: There are enough functions here that it probably makes sense to break this file apart at this point. Maybe a `vector_helper.py` file or something similar.

generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -144,3 +145,433 @@ def return_test_regs(test_data: TestData, params: InstructionParams) -> None:
...
+
+
+def prep_mask_v(
+    mask_val: str, test_data: TestData, params: InstructionParams, *, clobber_vd: bool = False, vd_v0: bool = False
+) -> list[str]:
+    """
+    Prepares the mask value for an instruction under test.
+
+    Args:
+        mask_val: Either a label that has the value for the mask, or one of ones, zeros, vlmaxm1_ones, or vlmax2dp1_ones.
```
jordancarlin: You could make this an enum or string (where the enum choices are the possible non-label options) for better type safety.

generators/testgen/src/testgen/asm/sections.py
```diff
@@ -47,3 +47,34 @@ def generate_test_string_section(data_strings: list[str]) -> str:
...
     return "\n".join(data_strings)
+
+
+def generate_vector_data_section(vector_data_labels: list[tuple[str, list[int], int]]) -> str:
+    """
+    Generate the .data section containing all vector data, aligned to eew
+
+    Args:
+        vector_data_labels: List of triples (label, data, eew) containing the vector data
+        xlen: Target XLEN (32 or 64)
```
jordancarlin: Doesn't look like xlen is an argument?

generators/testgen/src/testgen/asm/sections.py
```diff
@@ -47,3 +47,34 @@ def generate_test_string_section(data_strings: list[str]) -> str:
         Assembly code for the .data section
     """
     return "\n".join(data_strings)
+
+
+def generate_vector_data_section(vector_data_labels: list[tuple[str, list[int], int]]) -> str:
```
jordancarlin, line 52: `vector_data_labels` might make more sense as a dict with the key being the label and the value as the data/eew pair.

generators/testgen/src/testgen/coverpoints/vector/cmp_overlaps.py
```diff
@@ -0,0 +1,162 @@
...
+            test_data.vec_regs.deallocate_operands()
+            continue
+
+        desc = f"cmp_{v1}_{v2} (Test {v1} = {v2} = v{v})"
+        bin_name = f"cp_{v1}_{v2}_b{v}"
+
+        tc = format_single_testcase(instr_name, instr_type, test_data, params, desc, bin_name, coverpoint)
+
+        test_chunks.append(tc)
+        return_test_regs(test_data, params)
```
jordancarlin, line 123: Seems like this can fail silently right now. What if none of the possibilities for v work? Right now it would end up returning an empty list. Presumably that should be an error instead. Also, it is not obvious why this whole looping approach is needed, so please add a comment explaining it.

generators/testgen/src/testgen/coverpoints/vector/cp_custom.py
jordancarlin, line 1: Would it be simpler to have a function like `expand_coverpoints` as part of the initial loading of the coverpoints that has a dictionary mapping from individual coverpoints to lists of coverpoints and replaces any entries that match with the list from the dict?

generators/testgen/src/testgen/coverpoints/vector/vector_helpers.py
```diff
@@ -0,0 +1,364 @@
...
+    "vs_corner_f_onePulp",
+    "vs_corner_f_largestsubnorm",
+    "vs_corner_f_negSubnormLeadingOne",
+    "vs_corner_f_min_subnorm",
+    "vs_corner_f_canonicalQNaN",
+    "vs_corner_f_negNoncanonicalQNaN",
+    "vs_corner_f_sNaN_payload1",
+]
+
+fedges = {
```
jordancarlin: Seems like a lot of this would fit better in `data/edges.py`.

generators/testgen/src/testgen/coverpoints/vector/vector_helpers.py
```diff
@@ -0,0 +1,364 @@
+##################################
+# vector_helpers.py
+#
+# rwolk@hmc.edu June 2026
+# SPDX-License-Identifier: Apache-2.0
+##################################
+
+import dataclasses
```
jordancarlin: `from dataclasses import dataclass` is the standard for this.

generators/testgen/src/testgen/coverpoints/vector/vector_helpers.py
```diff
@@ -0,0 +1,364 @@
...
+                if "random" in corner2:
+                    continue
+                if random_val == _corner_value(corner2, eew):
+                    conflict = True
+                    break
+        return random_val
+    raise ValueError(f"Unknown corner: {corner}")
+
+
+@dataclasses.dataclass
```
jordancarlin:
```suggestion
@dataclass
```

generators/testgen/src/testgen/data/state.py
```diff
@@ -218,3 +229,32 @@ def destroy(self) -> None:
...
+        """
+        Registers a label and sew pair to be emitted in the data section. Used in vector tests to not be wasteful
+        by incrementing the data pointer in every test.
+
+        Exactly one of two optional arguments must be passed:
+            elements: List of SEW-wide elements to be emitted in that order in the data section
+            random_elements: Integer number of random elements to be emitted in the data section
+        """
+        assert (elements is None) ^ (random_elements is None), (
+            "Exactly One of Bytes and Random Bytes Must Be Set for register_vector_data"
```
jordancarlin: The error message names don't align with the argument names.

generators/testgen/src/testgen/formatters/registry.py
```diff
@@ -64,6 +88,8 @@ class InstructionTypeConfig:
     imm_signed: bool = True
     imm_nonzero: bool = False
     pair_regs: set[str] | None = None  # Registers that use register pairs (e.g., {"rd", "rs2"})
+    instruction_class: str = ""
```
jordancarlin: Probably makes more sense for instruction_class to be an Enum or Literal so that we have a defined list of valid classes.

generators/testgen/src/testgen/formatters/vector_params.py
```diff
@@ -0,0 +1,499 @@
...
+                test_data.config.sew,
+                random_elements=VLEN_MAX // test_data.config.sew,
+            )
+        elif new_params.rs1val is None:
+            new_params.rs1val = random_int(test_data.config.xlen)
+    if "rd" in registers:
+        new_params.rd = randomize_register("rd", test_data, instr_type_config, lmul, info, new_params.rd)
+
+    if "fs1" in registers:
+        new_params.fs1 = randomize_register("fs1", test_data, instr_type_config, lmul, info, new_params.fs2)
```
jordancarlin:
```suggestion
        new_params.fs1 = randomize_register("fs1", test_data, instr_type_config, lmul, info, new_params.fs1)
```

jordancarlin (changes requested):

generators/testgen/src/testgen/formatters/vector_params.py
jordancarlin: **Copy-paste bug: `fs1` is validated against `fs2`'s preset.**

Every other register passes its own preset (`new_params.vd` for `vd`, etc.), but this passes `new_params.fs2` as the preset for `fs1`. Since `randomize_register` returns an `f`-register preset verbatim when non-`None`, an instruction with both `fs1` and `fs2` set would clobber `fs1` with `fs2`'s value.

```suggestion
        new_params.fs1 = randomize_register("fs1", test_data, instr_type_config, lmul, info, new_params.fs1)
```

Latent today (no in-scope vector type uses `fs1`), but a clear defect in a shared helper.

---
*🤖 Generated by Claude Code (Claude Opus 4.8). Automated review — please verify before acting.*

generators/testgen/src/testgen/formatters/vector_params.py
```diff
@@ -0,0 +1,499 @@
...
+    no_overlap: set[tuple[str, str]] = set()
+
+    # Normal constrains
+    assert instr_type_config.vector_data is not None, "vector_data must be provided for vector instruction types"
+    no_overlap |= instr_type_config.vector_data.overlap_constraints
+
+    # Specific Masked Constraints
+    if masked:
+        # Either use the overrides, or assume that there is no overlap with v0 in the masked case
+        if instr_type_config.vector_data.masked_constraints is not None:
```
jordancarlin, line 198: **The masked-v0 fallback default is unreachable dead code.**

`VectorTypeConfig.masked_constraints` is declared `field(default_factory=set)`, so it is **never `None`**. This `if … is not None` is therefore always true, and the `elif` fallback below — "unless otherwise specified, do not overlap v0 when masked" — can never run.

Safe today only because all masked callers pass a full `additional_no_overlap` set. But any new masked coverpoint that omits it will silently generate **illegal masked instructions whose dest/source group overlaps the mask v0**. Consider defaulting `masked_constraints` to `None` in the dataclass (which also lets the several `is not None` guards on these always-set fields do real work).

---
*🤖 Generated by Claude Code (Claude Opus 4.8). Automated review — please verify before acting.*

generators/testgen/src/testgen/formatters/registry.py
jordancarlin: **`instruction_class` is never assigned anywhere, so every conditional that reads it is dead.**

This field defaults to `""` and is never set in any `InstructionTypeConfig`/`VectorTypeConfig` instantiation. As a result, all three consumers in `vector_params.py` are unreachable:
- the index-segment special-case (`"index" in instruction_class`)
- the load/store `rs1val_pointer` handling (`instruction_class in ["load", "store"]`)
- the segment-reset branch (`"index" not in instruction_class`)

Either wire it up in the type configs or drop it before it misleads a future reader into thinking those paths are active.

---
*🤖 Generated by Claude Code (Claude Opus 4.8). Automated review — please verify before acting.*

generators/testgen/src/testgen/formatters/vector_params.py
```diff
@@ -0,0 +1,499 @@
...
+
+    # We can check that the register that can't overlap is even assigned now that the suffixes have been removed
+    if params_dict[register] is None:
+        return []
+
+    smallest_emul = lmul
+    if info.vext_multiplier is not None:
+        smallest_emul = min(smallest_emul, lmul * info.vext_multiplier)
+    if info.load_store_eew is not None:
+        smallest_emul = min(smallest_emul, lmul * info.load_store_eew)
```
jordancarlin: **`smallest_emul` mixes EMUL with raw EEW bit-widths — these two `min()` calls are dead.**

`info.load_store_eew` and `info.index_eew` are raw bit-widths (8/16/32/64), so `lmul * eew` is always ≥ `lmul` and can never be the minimum. The value you want is the size *multiplier* (`eew / sew`), i.e. what `info.get_size_multiplier(...)` returns — matching how `emul` is derived two lines below.

Consequence: for indexed/segment load-stores the `_top`/`_bottom` partial-overlap region is miscomputed, permitting some illegal dest/source overlaps. Currently inert because `expand_vector_extension` only emits `Vx` suites, but it will bite the moment indexed suites are enabled.

---
*🤖 Generated by Claude Code (Claude Opus 4.8). Automated review — please verify before acting.*

generators/testgen/src/testgen/formatters/vector_params.py
```diff
@@ -0,0 +1,499 @@
...
+    if "vs1" in registers:
+        new_params.vs1 = randomize_register("vs1", test_data, instr_type_config, lmul, info, new_params.vs1)
+    if "vd" in registers:
+        new_params.vd = randomize_register("vd", test_data, instr_type_config, lmul, info, new_params.vd)
+
+    if "rs2" in registers:
+        new_params.rs2 = randomize_register("rs2", test_data, instr_type_config, lmul, info, new_params.rs2)
+        if new_params.rs2val is not None:
+            if info.load_store_eew is not None:
+                new_params.rs2val = random.randint(-2, 2 + 1) * int(info.load_store_eew / 8)
```
jordancarlin: **`random.randint` is inclusive on both ends, so `(-2, 2 + 1)` yields −2..3, not the symmetric −2..2 the `2 + 1` seems to intend.**

If exclusive-upper was intended (as with `range`/`random_range`), drop the `+ 1`. Separately, this uses the module-global `random` rather than the project's seeded `random_int`/`random_range` helpers used elsewhere, which hurts reproducibility of generated tests. Latent (load-store scope).

---
*🤖 Generated by Claude Code (Claude Opus 4.8). Automated review — please verify before acting.*

generators/testgen/src/testgen/data/registers.py
```diff
@@ -372,3 +372,164 @@ def consume_registers(self, regs: list[int]) -> None:
...
+        Optional Args:
+            suppress_overlap: If the allocation would result in double allocation of a register, this normally throws an
+                error, but when this argument is set no errors are thrown on overlaps.
+        """
+        for reg in range(register, register + width):
+            if reg in self.reg_list:
+                self.consume_registers([reg])
+            elif not suppress_overlap or reg >= self.reg_count:
+                raise ValueError(f"Unable to Allocate Registers for Operand {name}, v{register} with width {width}")
+        self.operand_map[name] = (register, width)
```
jordancarlin, line 482: **`allocate_operand` can leak registers if the loop raises partway through.**

Registers are consumed one at a time in the loop above, but `operand_map[name]` is only written *after* the loop completes. If the loop raises mid-way (e.g. a group member `>= self.reg_count`, or a non-`reg_list` reg with `suppress_overlap=False`), the already-consumed registers are removed from `reg_list` with **no `operand_map` entry**, so `deallocate_operand(s)` can never return them. The leak later surfaces as the `len(reg_list) == 32` assertion in `return_test_regs`. Consider consuming into a local list and rolling back before re-raising.

---
*🤖 Generated by Claude Code (Claude Opus 4.8). Automated review — please verify before acting.*

generators/testgen/src/testgen/data/state.py
```diff
@@ -218,3 +229,32 @@ def destroy(self) -> None:
         """Clean up resources used by TestData."""
         self._int_regs.destroy()
         self._float_regs.destroy()
```
jordancarlin, line 231: **`destroy()` never calls `self._vec_regs.destroy()`.**

Only the int and float register files are destroyed here, so the inherited `RegisterFile.destroy` leak check (`len(reg_list) != reg_count`) that guards those files is silently bypassed for vectors — an intra-file vector-register leak won't be caught at teardown. Add `self._vec_regs.destroy()` alongside the other two.

---
*🤖 Generated by Claude Code (Claude Opus 4.8). Automated review — please verify before acting.*

generators/testgen/src/testgen/asm/helpers.py
fourth-bit: The other register files holds a temporary register whereas the vector register file does not. I found that placing an assert here gives a more localized error than running out of registers in the another coverpoint or having an error at the end of test generation.

generators/testgen/src/testgen/asm/helpers.py
fourth-bit: The only duplication is three lines for the number of bytes consumed for sigupd given some `max_bytes` which differs between the two. Unless I'm missing something

jordancarlin: Seems fine to leave as is then. Too much code to look through so I'm sure some of my comments are wrong.

generators/testgen/src/testgen/asm/sections.py
fourth-bit: I believe that the storage of the data within TestData is a dictionary. The data here has to be a list so that the right labels end up in the right files with test chunk splits. I believe that I chose to do it this way to avoid passing TestData everywhere, but perhaps that wasn't the right choice

jordancarlin: I wasn't thinking about splits. That seems like a reasonable justification for keeping it as a list.

generators/testgen/src/testgen/formatters/registry.py
fourth-bit: It is in there for the future when Vls is added.

generators/testgen/src/testgen/formatters/vector_params.py
fourth-bit: This is the same range that the previous testgen gave. I will change it to random_range though

generators/testgen/src/testgen/formatters/vector_params.py
fourth-bit: The fix is elsewhere (masked_constraints now defaults to None)

fourth-bit:
Resolved most of the comments. I'm have a regression run that I'll look at the results of on Monday, and I still have to go back and look through the code globally for the globally applicable comments.

jordancarlin:
> Resolved most of the comments. I'm have a regression run that I'll look at the results of on Monday, and I still have to go back and look through the code globally for the globally applicable comments.

Sounds good. Let me know when it is ready to merge.

fourth-bit:
All the global comments have been addressed now, so it should be ready to merge.

jordancarlin:
A nice side effect of this change is that the Vx tests are now enabled by default for all users with vector in their UDB config and Vx is finally tested in CI.

jordancarlin (approved):
LGTM. Great work with this one!

PR #1765, opened 2026-06-24
Sm: Add counter wraparound coverpoints and tests for mcycle, minstret, and mtime

### LINKED ISSUE #1602: Counter wraparound tests needed
Need new tests in testplan, coverpoints, tests to check wraparound of counters (mcycle, minstret, mtime).  Set counter to highest value, wait for it to increment, check it is small again.

These probably belong in the Sm test suite.

amanmurad-10xe:
Added coverpoints and assembly tests to verify that the mcycle, minstret, and mtime counters wrap from all-ones to zero without saturating or raising an exception.

For RV64, the test writes maximum value to each counter, waits one increment, and checks the counter reads back a small value. 
For RV32, the high half is written before the low half to avoid a carry race condition. mcountinhibit is explicitly cleared before wraparound tests to ensure counters are running. The mtime wraparound uses a polling loop on the time CSR to guarantee the coverage tracker observes the wrapped value.

New normative rules mcycle_wraparound, minstret_wraparound, and mtime_wraparound are added in Sm.yaml linking these coverpoints to the specification.

Closes #1602


coverpoints/norm/Sm.yaml
```diff
@@ -1303,3 +1311,21 @@ normative_rule_definitions:
...
+      ["Sm_mcsr_cg/cp_mcycle_wraparound", "Sm_mcsr_cg/cp_mcycleh_wraparound"]
+
+  - name: minstret_wraparound
+    coverpoint:
+      [
+        "Sm_mcsr_cg/cp_minstret_wraparound",
+        "Sm_mcsr_cg/cp_minstreth_wraparound",
+      ]
+
+  - name: mtime_wraparound
```
davidharrishmc: According to https://github.com/riscv/riscv-isa-manual/blob/main/normative_rule_defs/machine.yaml, the rule is named mtime_wrap.

coverpoints/norm/Sm.yaml
```diff
@@ -1303,3 +1311,21 @@ normative_rule_definitions:
   # WARL field.
   - name: Zicsr_warl
     coverpoint: ["Sm_mcsr_cg/cp_mcsr_access"]
+
+  # mcycle and minstret are 64-bit and must wrap from all-ones to zero without
+  # saturating or raising an exception (implied by unrestricted modular arithmetic
+  # on WARL counter fields)
+  - name: mcycle_wraparound
```
davidharrishmc: mcycle_wraparound and minstret_wraparound rules do not exist in the isa manual.  Delete them.  Only reference rules that exist and are tagged.  I don't see good quotes in the ISA manual specifying wraparound behavior, except for mtime.

coverpoints/param/Sm.yaml
```diff
@@ -8,8 +8,12 @@ parameter_definitions:
     coverpoint: ["cp_mcsr_access"]
     effect: ["misa bin results depend on parameter"]
   - name: TIME_CSR_IMPLEMENTED
-    coverpoint: ["cp_mtime_write"]
-    effect: ["coverpoint only applies if parameter is True"]
+    coverpoint:
```
davidharrishmc: drop all changes to this file.  They are not exercising the unprivileged TIME csr.

coverpoints/param/Sm.yaml
```diff
@@ -34,3 +38,10 @@ parameter_definitions:
   - name: VENDOR_ID_OFFSET
     coverpoint: ["cp_mcsr_access"]
     effect: ["`mvendorid` value"]
+  - name: RVMODEL_MTIME_ADDRESS
```
davidharrishmc: Drop this parameter.  It is a RVMODEL DUT-specific macro, not a UDB parameter.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -465,6 +465,54 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
         `endif // RVMODEL_MSIP_ADDRESS
     `endif // S1P13P0_SUPPORTED
 
+    // Counter Wraparound Coverage
+
+    `ifdef UDB_MXLEN_32
+        cp_mcycle_wraparound : coverpoint get_csr_val_addr(ins.hart, ins.issue, `SAMPLE_AFTER, CSR_MCYCLE ,"mcycle", "") {
+            bins wrapped = ( [32'hFFFFFF00:32'hFFFFFFFF] => [0:32'hFF] );
```
davidharrishmc: I don't understand how this relates to wraparound.  Seems like the value before the clock should be all 1s and the value after the clock should be all 0s.  Why is there a range before, and FF after?  Same query applies to all of these coverpoints.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -465,6 +465,54 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        }
+        cp_mcycleh_wraparound : coverpoint get_csr_val_addr(ins.hart, ins.issue, `SAMPLE_AFTER, CSR_MCYCLEH ,"mcycleh", "") {
+            bins wrapped =( [32'hFFFFFFF0:32'hFFFFFFFF] => [0:32'h1] );
```
davidharrishmc: For RV32, I think we want a combination of {mcycle, mcycle} = all 1s -> all 0s.  
Are these coverpoints being hit?

coverpoints/priv/Sm_coverage.svh
```diff
@@ -465,6 +465,54 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        cp_minstreth_wraparound : coverpoint get_csr_val_addr(ins.hart, ins.issue, `SAMPLE_AFTER, CSR_MINSTRETH ,"minstreth", "") {
+            bins wrapped =( [32'hFFFFFFF0:32'hFFFFFFFF] => [0:32'h1] );
+        }
+    `else
+        cp_minstret_wraparound : coverpoint get_csr_val_addr(ins.hart, ins.issue, `SAMPLE_AFTER, CSR_MINSTRET ,"minstret", "") {
+            bins wrapped =( [64'hFFFFFFFFFFFFFF00:64'hFFFFFFFFFFFFFFFF] => [0:64'hFF] );
+        }
+    `endif
+
+    `ifdef UDB_TIME_CSR_IMPLEMENTED
```
davidharrishmc: This only controls access to the user-level TIME csr, not the existence of MTIME.  Delete the `ifdef.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -465,6 +465,54 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+    `ifndef UDB_MXLEN_32
```
davidharrishmc: The other blocks are `ifdef.  This is `ifndef.  Be consistent.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -964,6 +964,130 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            "#else // __riscv_xlen == 32",
+            f"LI(x{r_temp}, -1)                    # Load 0xFFFFFFFF",
+            test_data.add_testcase("mcycle_wrap_32", coverpoint, covergroup),
+            f"CSRW(mcycleh, x{r_temp})             # Set upper bits to max FIRST",
+            f"CSRW(mcycle, x{r_temp})              # Set lower bits to max LAST",
+            "nop                                   # instruction increment to cause wrap",
+            f"CSRR(x{r_val2}, mcycle)              # Read lower bits of wrapped value",
+            f"CSRR(x{r_val}, mcycleh)              # Check if upper bits wrapped to 0",
+            f"sub x{r_val}, x{r_val}, x{r_temp}    # Subtract max value to get wrapped value",
+            f"sltiu x{r_val}, x{r_val}, 2          # Signature = 1 if upper bits wrapped to 0 or 1",
```
davidharrishmc: Upper bits should never be 1.  That would take 4 billion cycles.
r_val2 doesn't appear in this logic but should.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -964,6 +964,130 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+    lines.extend(
+        [
+            "#if __riscv_xlen == 64",
+            f"LI(x{r_temp}, -1)                    # Load 0xFFFFFFFFFFFFFFFF",
+            test_data.add_testcase("minstret_wrap_64", coverpoint, covergroup),
+            f"CSRW(minstret, x{r_temp})            # Set minstret to max value",
+            "nop                                   # instruction increment to cause wrap",
+            f"CSRR(x{r_val}, minstret)             # Read wrapped value",
+            f"sub x{r_val}, x{r_val}, x{r_temp}    # Subtract max value to get wrapped value",
+            f"sltiu x{r_val}, x{r_val}, 0x100      # Signature = 1 if wrapped value < 256",
```
davidharrishmc: minstret is precisely known by the reference model because a deterministic number of instructions are executed after it was written (e.g. 2). Just sigupd it directly rather than using sltiu because the value should be exact.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -964,6 +964,130 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            f"LI(x{r_temp}, -1)                    # Load 0xFFFFFFFF",
+            test_data.add_testcase("minstret_wrap_32", coverpoint, covergroup),
+            f"CSRW(minstreth, x{r_temp})           # Set upper bits to max",
+            f"CSRW(minstret, x{r_temp})            # Set lower bits to max",
+            "nop                                   # instruction increment to cause wrap",
+            f"CSRR(x{r_val2}, minstret)            # Read lower bits of wrapped value",
+            f"CSRR(x{r_val}, minstreth)            # Check if upper bits wrapped to 0",
+            f"sub x{r_val}, x{r_val}, x{r_temp}    # Subtract max value to get wrapped value",
+            f"sltiu x{r_val}, x{r_val}, 3          # Signature = 1 if upper bits wrapped to 0 or 1",
+            write_sigupd(r_val, test_data),
```
davidharrishmc, line 1049: Just sigupd both r_val2 and r_val.

coverpoints/priv/Sm_coverage.svh
davidharrishmc: RVMODEL_TIME_ADDRESS controls whether the DUT has a mtime csr.  You use it properly in the test.  We've been leaving it out of the coverpoints because the reference model may have mtime even if the DUT does not.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -964,6 +964,130 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+    ######################################
+    coverpoint = "cp_mtime_wraparound"
+    ######################################
+    lines.append(
+        comment_banner(coverpoint, "Write max value to memory-mapped mtime and read wrapped value via time CSR")
+    )
+
+    # Generate unique local assembly labels using the current test count
+    tc_num = test_data.test_count + 1
+    loop_label_64 = f"mtime_wait_64_tc{tc_num}"
```
davidharrishmc: Use 1b rather than creating label names like this.  e.g.

1:
CSRR x1, time
li x2, 0xFF
bgeu x1, x2, 1b

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -964,6 +964,130 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            f"LI(x{r_val}, 1)                      # Write success signature",
+            write_sigupd(r_val, test_data),
+            "#else // __riscv_xlen == 32",
+            f"SREG x{r_val}, 4(x{r_temp})          # Write upper 32-bit max",
+            f"SREG x{r_val}, 0(x{r_temp})          # Write lower 32-bit max",
+            # Loop until timeh wraps to 0 or 1
+            f"{loop_label_32}:",
+            f"CSRR(x{r_val2}, time)                # Read shadow CSR time (lower 32 bits)",
+            f"CSRR(x{r_val}, timeh)                # Read shadow CSR timeh",
+            f"LI(x{r_temp}, 2)",
```
davidharrishmc: It's better not go to 1.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -964,6 +964,130 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            f"bgeu x{r_val}, x{r_temp}, {loop_label_32} # Wait for high bits to clear",
```
davidharrishmc: r_val2 is never checked.

.codespellrc
```diff
@@ -1,2 +1,2 @@
 [codespell]
-ignore-words-list = hart, vill, vor, FPR, fpr, mone, sie, fle, stip, sxl, VAs, ExceptionsS, rouge, Synopsys, ExtInt, USEED, useed
+ignore-words-list = hart, vill, vor, FPR, fpr, mone, sie, fle, stip, sxl, VAs, ExceptionsS, rouge, Synopsys, ExtInt, USEED, useed, caf, daa, ded, ede
```
davidharrishmc: Why are these new words needed?

davidharrishmc (changes requested):
There seem to be a number of misconceptions in this PR.  See detailed feedback.

Also, do not check in Sm-00.S.  We aren't checking privileged .S files in yet because they are in too much flux.

.codespellrc
amanmurad-10xe: These words were only added to avoid codespell pre-commit check failures caused by the Sm-00.S file. Since I have reverted that test file, these additions are no longer needed.

coverpoints/norm/Sm.yaml
amanmurad-10xe: Thanks for pointing this out. I hadn't checked the normative rule definitions in the referenced machine.yaml file and had added the mcycle and minstret wraparound rules based on the issue description. After reviewing the file, I confirmed that only mtime_wrap is defined. I've now removed the mcycle_wraparound and minstret_wraparound rules and kept only the existing mtime_wrap reference.

amanmurad-10xe:
@davidharrishmc - sent you a clarification email on this earlier, but posting a summary here as well in case it's easier to follow up on the PR directly.

I've pushed the changes addressing your review comments on the counter wraparound coverpoints/tests for `mcycle`, `minstret`, and `mtime`. Using the approach you suggested (read all-1s in one cycle, then all-0s the next), I'm now getting 100% coverage on both RV32 and RV64.

**Issue found while testing:** In Sail, `mcycle`/`mtime` appear to increment once every **two** retired instructions. This becomes a problem when the test also writes the counter to the signature region via `write_sigupd()`, a different value ends up in the signature than expected.

**Current test status:** the `mtime` wraparound tests currently **fail with a signature mismatch** whenever the counter is written to the signature region, because of this sampling shift. It's not a bug in the coverpoint logic itself — coverage is at 100% — it's specifically the signature comparison that breaks.

Trace evidence — same instruction sequence, but the `csrrs` reading `time` lands on a different tick:

<details>
<summary>Trace comparison (coverage run vs. build/signature run)</summary>

```
Coverage run:
[197037] sd x12, 0x0(x11)        clint mtime <- 0xFFFFFFFFFFFFFFFE → ticks to 0xFFFFFFFFFFFFFFFF
[197038] csrrs x12, time, x0     CSR time -> 0xFFFFFFFFFFFFFFFF
[197040] csrrs x12, time, x0     CSR time -> 0x0000000000000000  (wraparound)
                                  write_sigupd() -> 0x0000000000000000

Build/signature run:
[197042] sd x12, 0x0(x11)        clint mtime <- 0xFFFFFFFFFFFFFFFE (still 0xFFFFFFFFFFFFFFFE this cycle)
[197043] csrrs x12, time, x0     CSR time -> 0xFFFFFFFFFFFFFFFE
[197045] csrrs x12, time, x0     CSR time -> 0xFFFFFFFFFFFFFFFF
                                  write_sigupd() -> 0xFFFFFFFFFFFFFFFF
```

</details>

I'd expect the same to eventually apply to `mcycle` once the DUT implementation catches up, whereas `minstret` should stay deterministic since it only increments on retirement — not affected by this.

Since `mcycle`/`mtime` update timing may legitimately differ across DUTs (and vs. Sail), would it make sense to:
- keep `mcycle`/`mtime` for **coverage only**, and
- continue signature-checking `minstret`?

Let me know your thoughts.

jordancarlin:
We should check all of them with signatures. Otherwise, there is no reason to add the test because everything will pass. For `minstret`, we should check an exact match. For `mcycle` and `mtime` let's set it to all 100s, loop over 100 nops, and then check that the resulting counter is less than or equal to 200 (or something similar to that). That way we still test the wraparound behavior while allowing some flexibility with the exact timing of how often `mtime` and `minstret` increment.

coverpoints/norm/Sm.yaml
amanmurad-10xe: Fixed.

coverpoints/priv/Sm_coverage.svh
amanmurad-10xe: Fixed these to cover all 1s transition to all 0s properly.

coverpoints/priv/Sm_coverage.svh
amanmurad-10xe: These got covered too. But i have properly added the new coverpoint to capture this transition properly in RV32.

davidharrishmc:
Please post when this is fully working and ready for review.

amanmurad-10xe:
Tests are passing. Ready for review.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -469,6 +469,45 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+    `ifdef UDB_MXLEN_32
+        cp_mtime_wraparound : coverpoint {
+            get_csr_val_addr(ins.hart, ins.issue, `SAMPLE_BEFORE, CSR_TIME, "time", ""),
+            get_csr_val_addr(ins.hart, ins.issue, `SAMPLE_BEFORE, CSR_TIME, "timeh", "")
+        } {
+            bins wrapped = (64'hFFFFFFFF_FFFFFFFF => 64'h00000000_00000000);
+        }
+    `else
+        cp_mtime_wraparound : coverpoint get_csr_val_addr(ins.hart, ins.issue, `SAMPLE_BEFORE, CSR_TIME, "time", "") {
+            bins wrapped = (64'hFFFFFFFF_FFFFFFFF => 64'h00000000_00000000);
```
jordancarlin, line 508: I think we need to relax these coverpoints slightly. `mtime` might increment too fast for us to see if go from all ones to all zeros. We just need to see if go from all ones to all of the upper bits being zeros and the lower bits should be wildcards.

jordancarlin: The other option would be to modify the coverpoints to just check that we write all ones to the CSRs (instead of checking the actual CSR value). Then it is the test's responsibility to read back the new value and ensure it wrapped around correctly. That would be more in line with how most of the other coverpoints are written.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,126 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
 
     test_data.int_regs.return_registers([r1, r2])
 
+    # Counter Wraparound Verification
+    r_val, r_val2, r_temp, r_temp2 = test_data.int_regs.get_registers(4)
+
+    # Re-enable all counters before trying to wrap them!
+    lines.append(comment_banner("Enable Counters", "Ensure mcountinhibit is 0 so counters can run"))
```
jordancarlin, line 988: No need for a full comment banner for the little bit of setup happening here.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,126 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+    lines.extend(
+        [
+            f"LI(x{r_val}, 0)                      # Load 0 (enable all counters)",
+            f"CSRW(mcountinhibit, x{r_val})        # Clear inhibit register",
+            "nop\nnop\nnop",
```
jordancarlin, line 993: Why do we need three nops here?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,126 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
 
+    # Counter Wraparound Verification
+    r_val, r_val2, r_temp, r_temp2 = test_data.int_regs.get_registers(4)
+
+    # Re-enable all counters before trying to wrap them!
+    lines.append(comment_banner("Enable Counters", "Ensure mcountinhibit is 0 so counters can run"))
+    lines.extend(
+        [
+            f"LI(x{r_val}, 0)                      # Load 0 (enable all counters)",
+            f"CSRW(mcountinhibit, x{r_val})        # Clear inhibit register",
```
jordancarlin, line 992: The `CSRW` macros have been removed. Use regular `csrw` instructions. Global comment.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,126 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+    lines.append(comment_banner(coverpoint, "Write max value to mcycle and verify it wraps around cleanly"))
+
+    lines.extend(
+        [
+            "#if __riscv_xlen == 64",
+            f"LI(x{r_temp}, -1)                    # Load 0xFFFFFFFFFFFFFFFF",
+            test_data.add_testcase("mcycle_wrap_64", coverpoint, covergroup),
+            f"CSRW(mcycle, x{r_temp})              # Initialize mcycle to its maximum value",
+            f"CSRR(x{r_val}, mcycle)               # Read mcycle after wraparound",
+            f"sub x{r_val}, x{r_val}, x{r_temp}    # Compute the increment since the maximum value",
```
jordancarlin, line 1009: Why do we need to subtract the initial value off? Can't we just check the new value and make sure it is small? If it wrapped around, the old value should no longer be relevant.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,126 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            f"LA(x{r_temp}, RVMODEL_MTIME_ADDRESS) # base address of mtime",
+            "#if __riscv_xlen == 64",
+            test_data.add_testcase("mtime_wrap_64", coverpoint, covergroup),
+            f"LI(x{r_val}, -2)                     # 0xFFFFFFFFFFFFFFFE, two ticks before wrap",
+            f"SREG x{r_val}, 0(x{r_temp})          # arm mtime just before it wraps",
+            f"CSRR(x{r_temp2}, time)",
+            "1:",
+            f"CSRR(x{r_val}, time)                 # read time each loop pass",
+            f"LI(x{r_val2}, 0x10)                  # loop exit bound",
+            f"bgeu x{r_val}, x{r_val2}, 1b         # keep looping while time is still large",
```
jordancarlin, line 1074: Seems like this will get stuck in an infinite loop if mtime does not wraparound correctly. We want the test to fail in that case, not loop. How about just looping for 100 instructions or something and then checking that mtime is small after the loop?

jordancarlin (changes requested):

coverpoints/priv/Sm_coverage.svh
amanmurad-10xe: I'll implement the second option and push the changes soon. I had previously tried a similar approach to the first option, but it was suggested that I change it in an earlier review.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: I'll remove it.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Removed. I initially thought we needed to wait a few cycles for the counters to start incrementing, but it is not required.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Fixed.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Yes, we can. I'll update this to check only the new value after wraparound.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Sure, I'll update the test accordingly so that it fails if the wraparound does not happen correctly.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -469,6 +469,42 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+            bins minstreth = {CSR_MINSTRETH};
+        }
+    `endif
+
+    sw: coverpoint ins.current.insn {
+        wildcard bins sw = {SW};
+    }
+    sd: coverpoint ins.current.insn {
+        wildcard bins sd = {SD};
+    }
```
jordancarlin: This needs to be guarded on `UDB_MXLEN_64`.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -469,6 +469,42 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+
+    `ifdef UDB_MXLEN_64
+        cp_mtime_wraparound: cross priv_mode_m, sd, rs2_ones;
+    `else
+        cp_mtime_wraparound: cross priv_mode_m, sw, rs2_ones;
+    `endif
```
jordancarlin, line 495: Let's make these a little more specific and check that the address they are accessing is `RVMODEL_MTIME_ADDRESS`. Also, these coverpoints need to be guarded by `` `ifdef RVMODEL_MTIME_ADDRESS `` to support DUTs that do not have `mtime`.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -469,6 +469,42 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+    cp_mcycle_wraparound:   cross priv_mode_m, csrrw_allones, mcycle;
+    cp_minstret_wraparound: cross priv_mode_m, csrrw_allones, minstret;
+
+    `ifdef UDB_MXLEN_32
+        cp_mcycleh_wraparound:   cross priv_mode_m, csrrw_allones, mcycleh;
+        cp_minstreth_wraparound: cross priv_mode_m, csrrw_allones, minstreth;
+        cp_mtimeh_wraparound: cross priv_mode_m, sw, rs2_ones;
```
jordancarlin: Same comments as for cp_mtime_wraparound.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,120 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+    # Counter Wraparound Verification
+    r_val, r_val2, r_temp, r_temp2 = test_data.int_regs.get_registers(4)
+
+    # Re-enable all counters before trying to wrap them!
+    lines.extend(
+        [
+            f"LI(x{r_val}, 0)                      # Load 0 (enable all counters)",
+            f"csrw mcountinhibit, x{r_val}        # Clear inhibit register",
+        ]
+    )
```
jordancarlin: This can just become `lines.append("csrw mcountinhibit, x0       # Clear inhibit register")`.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,120 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+    ######################################
+    coverpoint = "cp_mcycle_wraparound"
+    ######################################
+    lines.append(comment_banner(coverpoint, "Write max value to mcycle and verify it wraps around cleanly"))
+
+    lines.extend(
+        [
+            "#if __riscv_xlen == 64",
+            f"LI(x{r_temp}, -1)                    # Load 0xFFFFFFFFFFFFFFFF",
+            test_data.add_testcase("mcycle_wrap_64", coverpoint, covergroup),
```
jordancarlin: Let's use `mcycle_wrap` as the name for this instead of using different rv64 and rv32 versions. Also, a lot of the code in the 32 and 64 paths are the same. It would be better to pull most of it outside the `#if` and only put the reads/writes of `mcycleh` in the `#if`.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,120 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            f"csrw mcycle, x{r_temp}              # Initialize mcycle to its maximum value",
+            "nop",
```
jordancarlin: We don't know how how many instructions the hart executes per cycle, so I think this needs a longer delay. Let's loop over the `nop` 100 times and then read `mcycle` and make sure it is small (less than 1000 or something like that).

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,120 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+    ######################################
+    coverpoint = "cp_minstret_wraparound"
+    ######################################
+    lines.append(comment_banner(coverpoint, "Write max value to minstret and verify it wraps around cleanly"))
+
+    lines.extend(
+        [
+            "#if __riscv_xlen == 64",
+            f"LI(x{r_temp}, -1)                    # Load 0xFFFFFFFFFFFFFFFF",
+            test_data.add_testcase("minstret_wrap_64", coverpoint, covergroup),
```
jordancarlin: Same comment about moving most of this out of the `#if`.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,120 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            f"csrw minstret, x{r_temp}            # Initialize minstret to its maximum value",
+            f"csrr x{r_val}, minstret             # Read minstret after wraparound",
```
jordancarlin: Do we need one `nop` in between the read and write here? The write sets it to all ones, and when the read happens, no other instruction has retired yet. I think it would still be all ones.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,120 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            f"LA(x{r_temp}, RVMODEL_MTIME_ADDRESS) # base address of mtime",
+            "#if __riscv_xlen == 64",
+            f"LI(x{r_val}, -1)                      # 0xFFFFFFFFFFFFFFFF",
+            test_data.add_testcase("mtime_wrap_64", coverpoint, covergroup),
+            f"SREG x{r_val}, 0(x{r_temp})           # write all ones to mtime",
+            f"LI(x{r_counter}, 100)                 # bounded wait: fixed instruction count, never hangs",
+            "1:",
+            f"addi x{r_counter}, x{r_counter}, -1   # decrement the counter",
+            f"bnez x{r_counter}, 1b                 # wait for mtime to tick over",
+            f"csrr x{r_val2}, time                   # read time after the bounded wait",
```
jordancarlin: Shouldn't we be reading `mtime` here instead of `time`? `time` does not exist on DUTs without Zicntr.

jordancarlin (changes requested):
Getting closer. Here is another round of feedback.

coverpoints/priv/Sm_coverage.svh
amanmurad-10xe: Done.

coverpoints/priv/Sm_coverage.svh
amanmurad-10xe: Fixed.

coverpoints/priv/Sm_coverage.svh
amanmurad-10xe: Done.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Fixed.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Address as suggested.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Sure. Added.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Yes, we need a `nop` between write and read so minstret increments and wraps around cleanly. Added.

generators/testgen/src/testgen/priv/extensions/Sm.py
amanmurad-10xe: Updated to read memory-mapped `mtime` instead of the `time` CSR.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -981,6 +981,103 @@ def _generate_mcsr_cntr_tests(test_data: TestData) -> list[str]:
...
+            "#endif",
+            test_data.add_testcase("mcycle_wrap", coverpoint, covergroup),
+            f"csrw mcycle, x{r_temp}             # Set mcycle to its maximum value",
+            f"LI(x{r_counter}, 100)                # Wait loop for counter ticks",
+            "1:",
+            "nop",
+            f"addi x{r_counter}, x{r_counter}, -1",
+            f"bnez x{r_counter}, 1b",
+            f"csrr x{r_val}, mcycle               # Read mcycle after the bounded wait",
+            f"sltiu x{r_val}, x{r_val}, 1000       # Pass if mcycle wrapped to a small value",
```
jordancarlin, line 1009: This could cause issues on DUTs if mcycle increments more than 1000 times in 100 instructions, but it seems fine to start with this.

jordancarlin (approved):
Thanks for sticking with this one! LGTM.

amanmurad-10xe:
@davidharrishmc If there are no more changes that need to be addressed, can we get this merged to avoid any conflicts later?

davidharrishmc (approved):
Looks good to me.  Thanks for the effort pushing this all through.

PR #1770, opened 2026-06-24
TrapHandler fix - CSR renaming for V mode

Teemooooooooo:
When accessing the vCSRs from VS mode, the vCSRs should be referred to as the corresponding Supervisor level CSRs instead of vCSRs. 
    - For example, when in VS mode, `csrr    T3, CSR_STVAL` reads from VSTVAL. 
    
The trap handler had the macros CSR_*** in V mode all be renamed to vCSRs, causing infinite loop in the trap handler as attemping to access the vCSRs when V = 1 cause virtual instruction exception. This PR ensures that the vCSRs are written as the corresponding S level CSRs when accessing them from V mode.


allenjbaum:
Not sure this is the best way to fix this. Inside the trap handler, it
knows which mode it is in.
The real problem is in the XCSR_RENAME_<mode> macro that defines which CSR
should be accessed for each mode.
I think the fix is to change then lines in XCSR_RENAME_V from
  .set CSR_XSTATUS, CSR_VSSTATUS
 to
  .set CSR_XSTATUS, CSR_SSTATUS
 and likewise for all the other CSRs that are renamed.

Tests still are not written at the time this code was, so there are other
misses here,
e.g. access to VS CSRs in all modes should map to VS CSRs, which will trap
if actually accessed in VS mode
(though accesses to lower priv mode CSR in the trap handler are probably
already explicit, i.e. they don't use XCSR)

Teemooooooooo:
Hi! I gated it at the specific sites where the VS CSRs were accessed from VS mode instead of just changing it at the CSR renaming macro because the renamed registers are also used in the setup code, which runs in M mode
For example, on line 859:
init_\__MODE__\()scratch:
        csrrw   T3, CSR_XSCRATCH, T1    // swap xscratch with save area ptr (will be used by handler)
        SREG    T3, xscr_save_off(T1)   // save old mscratch in xscratch_save
Even init_VSscratch would still run in M mode, if we just change the VS CSR renaming macro, it would break this part of the code. 

Another way to fix this is to change the VS CSR renaming macro and just gate them at the initiation code instead. 

Or we can also have two rename macros the the V mode, where one is the VS CSRs for setup in M mode and the other one is the S CSRs for everywhere else. 

Is there a more preferred fix?

jordancarlin:
Are the renamed CSR macros used more from VS mode or M mode? I think we should go with whatever approach requires the least special-casing.

tests/env/rvtest_trap_handler.h
```diff
@@ -368,19 +368,24 @@
...
-  .set CSR_XTVEC,   CSR_VSTVEC                  // vstvec — VS-mode trap vector
-  .set CSR_XSCRATCH,CSR_VSSCRATCH               // vsscratch — VS-mode scratch register
-  .set CSR_XEPC,    CSR_VSEPC                   // vsepc — VS-mode exception PC
-  .set CSR_XCAUSE,  CSR_VSCAUSE                 // vscause — VS-mode trap cause
+  .set CSR_XTVEC,   CSR_STVEC                   // stvec aliases to vstvec
+  .set CSR_XSCRATCH,CSR_SSCRATCH                // sscratch aliases to vsscratch
+  .set CSR_XEPC,    CSR_SEPC                    // sepc aliases to vsepc
+  .set CSR_XCAUSE,  CSR_SCAUSE                  // scause aliases to vscause
 #if (UDB_MXLEN==32)
   .set CSR_XEDELEGH, CSR_SEDELEGH               // sedelegh — upper half of sedeleg (RV32 only)
```
jordancarlin: If this is now identical to the S version, I think it would be simpler to drop all of this and just change the `XCSR_RENAME __MODE__` macro to point at the S CSRs when set to V mode.

allenjbaum:
Actually, I believe that is warranted (and I take responsibility for
getting it wrong to begin with.)

From the point of view of a trap handler in VS mode, it should not be
able to distinguish (legal) accesses to its CSR between VS and S.
so invoking the rename macro to use S instead of V should be the way to go.

In general, handlers that access CSRs at their priv level should use
renamed XCSRs.
Handlers that access CSRs at priv levels below where they (think they)
are should use the explicit name (which might trap, as in the example
above,  because they aren't actually at the priv level they think they
are).
And, of course, handlers that (try to) access CSRs at priv levels
above where they (think they) are should also use the explicit name,
and will always properly trap.

davidharrishmc:
Umer believes there is a simple way to fix rename macro that handles all of this.  He and Allen plan to review this week and push a fix if that works.

jordancarlin:
@UmerShahidengr are you still planning to push an alternate fix to this? If not, we should get this merged soon.

jordancarlin:
@UmerShahidengr, any update?

davidharrishmc:
PR planned today.

jordancarlin:
Closing as superseded by #2056.

PR #1780, opened 2026-06-28
Fix ExceptionsS abort when misaligned load/store raises AlignmentException

### LINKED ISSUE #1753: ExceptionsS test generation fails when misaligned load/store exceptions are enabled in sail.json
**Description**

While generating the ExceptionsS test with a platform configuration where misaligned scalar load/store accesses raise AlignmentException, the generated test fails during trap handling.

The platform configuration is changed from allowing misaligned accesses:

```
"misaligned": {
  "exceptions": {
    "load_store": {
      "None": null
    }
  }
}
```

to:

```
"misaligned": {
  "exceptions": {
    "load_store": {
      "Some": "AlignmentException"
    }
  }
}
```

After enabling AlignmentException, the generated ExceptionsS-00 test fails with:

```
RVCP: Failure: "The trap handler aborted the test before normal completion!"

RVCP: Unrecognized trap failure context.
RVCP: Instruction that trapped: 0x00d523a3
RVCP: MEPC:    0x0000000080015640
RVCP: MCAUSE:  0x0000000000000006
RVCP: MTVAL:   0x000000008001c007
```

The trapped instruction corresponds to the generated store-misaligned coverpoint:

```
ExceptionsS_cg_cp_store_address_misaligned_sw_off0:

    LI(x13, 0xDEADBEEF)
    LA(x14, scratch)
    addi x14, x14, 0
    sw x13, 0(x14)
```

With the new configuration, this instruction correctly generates a Store Address Misaligned exception:

```
mcause = 6
mtval  = 0x8001c007
```

However, instead of completing the expected trap signature update, the RVCP handler reports:

```
Unrecognized trap failure context
```

and aborts the test.

The same ExceptionsS test generation succeeds when misaligned accesses are configured as:

```
"load_store": {
    "None": null
}
```

This suggests that enabling AlignmentException changes the expected trap flow/signature handling for the misaligned load/store coverpoints, but the generated test or trap handler is not correctly accounting for this configuration.

The ExceptionsS test plan explicitly includes:

* cp_load_address_misaligned
* cp_store_address_misaligned
* cp_misaligned_priority

with the expectation that platforms may generate misaligned exceptions depending on implementation behavior.

**Expected behavior:**

When misaligned load/store exceptions are configured as AlignmentException, the generated ExceptionsS test should correctly recognize and validate:

* Load address misaligned (mcause=4)
* Store address misaligned (mcause=6)

without aborting due to an unrecognized trap context.

Vaibhav805:
Fixes #1753

## Root cause

Two issues combined to produce the reported failure.

**1. Spurious abort in `chk_\__MODE__\()trapsig_overrun`**

In `tests/env/rvtest_trap_handler.h`, this block double-adjusted `sp`
before reading the trap signature bounds:

\`\`\`asm
chk_\__MODE__\()trapsig_overrun:
        addi    sp, sp, -1*sv_area_sz              // temp adjust sp
        LREG    T4, sv_area_off+trapsig_ptr_off(sp)
        LREG    T2, sv_area_off+sig_bgn_off(sp)
        LREG    T1, sv_area_off+sig_seg_siz(sp)
        addi    sp, sp, 1*sv_area_sz               // undo adjustment
\`\`\`

`sv_area_off` already accounts for the per-mode save-area offset, so
this extra `addi` pair caused the three `LREG`s to read from the wrong
location — garbage values for the trap-signature pointer, begin
address, and size. The garbage values spuriously satisfied
`bgtu T4, T1, abort_test`, producing the `"Unrecognized trap failure
context"` / `"The trap handler aborted the test before normal
completion!"` messages from the issue.

This stayed latent under the default `load_store: None` config because
misaligned scalar loads/stores never trap in that case, so this check
is never reached for those coverpoints. Setting `load_store` to
`AlignmentException` is what first drives a trap through this exact
path for `cp_store_address_misaligned`, exposing the stale offset math.

**2. Trap signature region overflow**

Once (1) is fixed, the test runs much further (it no longer aborts on
the first misaligned trap) and produces more trap-signature entries
than the default `TRAP_SIGUPD_COUNT` (15000) allows for, overflowing
into adjacent HTIF memory and corrupting the syscall/exit write —
visible as a bogus `FAILURE: 0x06020331` with no diagnostic text.
Raised `TRAP_SIGUPD_COUNT` to 50000 for `ExceptionsS` specifically via
`extra_defines`.

## Fix

- `tests/env/rvtest_trap_handler.h`: removed the redundant `sp`
  adjustment in `chk_\__MODE__\()trapsig_overrun`.
- `generators/testgen/src/testgen/priv/extensions/ExceptionsS.py`:
  added `#define TRAP_SIGUPD_COUNT 50000` to `extra_defines`.

## Testing

Reproduced the original failure on `sail-RVA20S64` with
`misaligned.exceptions.load_store` set to
`{"Some": "AlignmentException"}` (matching the issue), confirmed the
fix resolves it:

\`\`\`
✓ Build complete: 5 succeeded
══════ sail-RVA20S64 ══════
  RESULT: All 1 tests passed.
\`\`\`

Also re-ran with `load_store` reverted to `None` to confirm no
regression on the existing passing path.

issue1753_1


.gitignore
```diff
@@ -45,3 +45,5 @@ docs/crd/build
 coverpoints/norm/norm-rules.json
 coverpoints/norm/yaml/new
 coverpoints/norm/yaml/chapters
+framework/src/act/data/vendor/
```
jordancarlin: Can you drop this from the project-specific gitignore? You can add it to your local `.git/info/exclude` file if needed.

tests/env/rvtest_trap_handler.h
```diff
@@ -1942,11 +1942,9 @@ skp_\__MODE__\()tval:
 1:
 // --- Check for trap signature overrun ---
 chk_\__MODE__\()trapsig_overrun:
-        addi    sp, sp, -1*sv_area_sz              // temp adjust sp
         LREG    T4, sv_area_off+trapsig_ptr_off(sp) // T4 = updated trap sig ptr
         LREG    T2, sv_area_off+sig_bgn_off(sp)    // T2 = sig begin
         LREG    T1, sv_area_off+sig_seg_siz(sp)    // T1 = sig size
-        addi    sp, sp, 1*sv_area_sz               // undo adjustment
```
jordancarlin, line 2019: This change seems to have broken the tests on CVW. @UmerShahidengr any ideas on what might be going wrong with the trap handler?

jordancarlin (changes requested):

tests/env/rvtest_trap_handler.h
Vaibhav805: @jordancarlin I traced the sp flow through this path. The temporary adjustment in trap_sig_sv is restored before execution reaches chk_trapsig_overrun, so sp appears to already point to the save area on entry. Based on that, the removed addi pair looks redundant, and I wouldn't expect it to change the effective address of the LREGs here.

Given that, I'm not convinced this hunk alone explains the CVW ExceptionsU U-00 (mcause=3/ebreak) failure. Could you confirm whether U-00 also fails on act4 without this PR? That would help determine whether this is an existing issue that's now being exposed or a regression introduced by this change.

jordancarlin: CI is passing on the act4 branch (and is required to pass for all PRs before merging), so the test is definitely passing before this change.

Vaibhav805: @jordancarlin The functional CI checks are passing now. The only remaining failure is the prek Markdown Link Check, which reports https://lists.riscv.org/g/sig-arch-test returning HTTP 503 from README.md.

jordancarlin: The prek failure seems like a transient error. Glad to hear that the rest of CI is passing. I'd like @UmerShahidengr to review the trap handler changes before merging.

UmerShahidengr (approved):
LGTM. Thanks for the fix

tests/env/rvtest_trap_handler.h
UmerShahidengr: Indeed it was a minor offset miscalculation bug which was getting shown while running trap handler in Smode without Hmode. Thanks for the fix

Vaibhav805:
@jordancarlin The trap handler is reviewed and pr is ready to merge when
you get a chance. Happy to make any additional changes if needed.

jordancarlin:
@Vaibhav805 looks like there are merge conflicts now. Once you resolve those, this look good to merge to me.

jordancarlin (approved):

Vaibhav805:
@jordancarlin Conflicts resolved — All things should be fine now. Ready for merge when you get a chance!

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -418,6 +418,12 @@ def _generate_xstatus_ie_tests(test_data: TestData, mode_tag: str, priv_mode: in
 @add_priv_test_generator(
     "ExceptionsS",
     required_extensions=["S"],
+
+    extra_defines=[
+        "#define SKIP_MEPC",
```
jordancarlin: SKIP_MEPC has been deprecated, please don't add it back.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
jordancarlin: This was marked as resolved, but I still see it. Remove the `#define SKIP_MEPC` line.

Vaibhav805: @jordancarlin  In my latest commit removed the line u can re trigger ci workflow for the final checks in prev check only prek issues were there which are now finally done.

jordancarlin (approved):

PR #1781, opened 2026-06-29
Replace all uses of .align with .balign and .p2align

### LINKED ISSUE #1704: Replace all uses of `.align` with `.balign` and `.p2align`
There have been at least a few instances recently where `.align` was used with a byte argument instead of a power-of-two argument. We should replace all uses of the somewhat ambiguous `.align` directive with the more explicit `.balign` and `.p2align` directives to try to avoid these subtle bugs.

pyarchana:
Fixes #1704

Replaced all uses of `.align` with the explicit `.p2align` or `.balign` directives:
- `.p2align N` where power-of-two alignment was intended (page tables, trap handlers, instruction alignment)
- `.balign N` where byte alignment was intended (tohost/fromhost with .dword)

Changes cover generator scripts, header files (.h), and checked-in generated .S files.


generators/testgen/scripts/vector_testgen_common.py
```diff
@@ -560,7 +560,7 @@ def getLengthSuiteTestCount():
 ]
 
 
-# ─── Segment length 2 ──────────────────────────────────────────────
+# â”€â”€â”€ Segment length 2 â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€â”€
```
jordancarlin: Seems like a bunch of characters in this PR got garbled. Please revert all of these changes that do not relate to the .align refactor.

jordancarlin (changes requested):
Looks like something got messed up in this PR. Please fix the garbled characters first and then I will review the rest of the PR.

generators/testgen/scripts/vector_testgen_common.py
pyarchana: sorry for the noise. the replacement script i used added a UTF-8 BOM to files and corrupted non-ASCII characters due to an encoding issue on windows. i have fixed it to do a clean byte-preserving substitution. pls take another look when you get a chance.

jordancarlin:
@areychana This PR was targeting main, but all PRs need to target the act4 branch. I updated the PR to target act4, but now there are many merge conflicts. Please update the PR to resolve the merge conflict.

jordancarlin (approved):

PR #1793, opened 2026-07-01
Fix cp_misaligned_priority_load to exclude non-misalign-capable ops

### LINKED ISSUE #1752: ExceptionsSm-00.S: cp_misaligned_priority_load uses access-fault base address
Test name :ExceptionsSm-00.S

Description
ACT generated `ExceptionsSm-00.S` uses `RVMODEL_ACCESS_FAULT_ADDRESS` as the base address for `cp_misaligned_priority_load`.
On VisionFive2 / SiFive U74 hardware, this makes the test depend on exception priority between an access fault and a misaligned access, instead of testing a clean misaligned load.
cp_misaligned_priority_load is intended to test misaligned exception priority. But the generated address is based on RVMODEL_ACCESS_FAULT_ADDRESS. With RVMODEL_ACCESS_FAULT_ADDRESS = 0x0, the off1 access targets 0x1.

That address is both  invalid/unmapped -> Load access fault, mcause 5 and misaligned for lh -> Load address misaligned, mcause 4
VF2/U74 reports the access fault first mcause = 5, mtval = 0x1.

Expected behavior

Test should set a valid address as RVMODEL_ACCESS_FAULT_ADDRESS which should be an accessible for hardware!  
That would make the effective address accessible but misaligned, avoiding an unrelated access fault.
The test should not use RVMODEL_ACCESS_FAULT_ADDRESS for this coverpoint unless the intended result is an access fault.
 
=======================================================================

Console snippet::

=======================================================================

RVCP-SUMMARY: TEST FAILED - Test File "ExceptionsSm-00.S"
RVCP: Test Info: "Mismatch in mcause value! Trap was being handled in M-Mode."
RVCP: Mismatching field: XCAUSE (trap signature word 1)
RVCP: Expected value: 0x0000000000000004
RVCP: Actual value:   0x0000000000000005
RVCP: Expected cause: Load address misaligned
RVCP: Actual cause:   Load access fault
RVCP: MEPC:    0x0000000080003e40
RVCP: MCAUSE:  0x0000000000000005
RVCP: MTVAL:   0x0000000000000001
RVCP: MSTATUS: 0x8000000a00007800

=====================================================================================

Instructions snippet

======================================================================================

/////////////////////////////////
// cp_misaligned_priority_load
//   Misaligned Priority Load
/////////////////////////////////
  LA(x13, RVMODEL_ACCESS_FAULT_ADDRESS)
    80003de0: 0000069b           addiw x13,x0,0
  addi x15, x13, 0
    80003e00: 00068793           addi x15,x13,0
ExceptionsSm_cg_cp_misaligned_priority_load_lh_off0_priority:
    80003e04: 00079603           lh x12,0(x15)
  addi x15, x13, 1
    80003e3c: 00168793           addi x15,x13,1

ExceptionsSm_cg_cp_misaligned_priority_load_lh_off1_priority:          
    80003e40: 00079603           lh x12,0(x15)                                          (faulting instructions)

=========================================================

Suggested Fix
Do not use RVMODEL_ACCESS_FAULT_ADDRESS for misaligned-priority load/store tests.
Use a valid mapped scratch/data address, then add the misalignment offset:

===============================================================
Environment::
Board: VisionFive2
Core: SiFive U74
Test: ExceptionsSm-00.S
Coverpoint: cp_misaligned_priority_load
Failing bin: lh_off1_priority
Reference model: Sail 0.12

Vaibhav805:
lb and lbu have byte alignment (1), so they can never produce a misaligned exception regardless of address offset. Including them in generate_misaligned_priority_load_tests meant testcases were generated for RVMODEL_ACCESS_FAULT_ADDRESS + offset where the only possible exception was an access fault, not a misaligned/access-fault priority test as intended.

Removed lb/lbu from load_ops_base in this function and added an alignment filter (_ALIGN dict) that skips any (op, offset) pair where the access is naturally aligned -- those cannot produce a misaligned exception either, so including them would have the same problem.

Verified on sail-rv64-max: ExceptionsSm RESULT: All 1 tests passed.

Fixes #1752

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
@@ -433,7 +433,9 @@ def generate_misaligned_priority_load_tests(
     addr_reg, temp_reg, check_reg = test_data.int_regs.get_registers(3)
 
     lines = ["#ifdef RVMODEL_ACCESS_FAULT_ADDRESS", comment_banner(coverpoint, "Misaligned Priority Load")]
-    load_ops_base = ["lh", "lhu", "lw", "lb", "lbu"]
+    # Only ops where misalignment is possible (lb/lbu excluded: alignment=1)
+    _ALIGN = {"lh": 2, "lhu": 2, "lw": 4, "lwu": 4, "ld": 8}
+    load_ops_base = ["lh", "lhu", "lw"]
     load_ops_64 = ["lwu", "ld"]
 
     lines.append(f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)")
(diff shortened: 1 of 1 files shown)


davidharrishmc:
This is hiding the problem rather than fixing it.  It should be possible to run byte accesses at any address offset and never see misaligned faults.  If Sail sometimes produces misaligned faults, the problem should be fixed in Sail rather than by removing tests.  Suggest closing this PR.

PR #1807, opened 2026-07-02
CI speedups

jordancarlin:
- Split configs that take a long time across multiple runners. The number of runners per simulator/config is specified in the appropriate `ci.yaml`. Splitting is done equally across the specified number of runners based on the size of the tests.
- More parallelism in CI
- More caching of UDB results, gem installation, and uv installation


jordancarlin:
Brings total CI time down from ~38 minutes to ~13 minutes.

davidharrishmc (approved):
Spiffy!

PR #1819, opened 2026-07-06
Add coverpoints for Sscofpmf

ayeshaanwaar05:
**Add functional coverage for Sscofpmf extension**
        Adds Sscofpmf_cg covergroup (Sscofpmf_coverage_init.svh) for the Sscofpmf (count overflow / mode-based filtering)   extension.
**Out of scope:** 
      VS/VU modes (no virtualization support yet); only one HPM counter tested per spec.


ayeshaanwaar05:
Ready for review

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+//
+// Copyright (C) 2024 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+////////////////////////////////////////////////////////////////////////////////////////////////
+
+`define COVER_SSCOFPMF
+`ifdef SSCOFPMF_SUPPORTED
+    `ifndef RVMODEL_MHPMEVENT
```
davidharrishmc: The rest of the RVMODEL checking is in tests/env/check_defines.h.  How about moving these checks there and also checking RVMODEL_MHPMEVENT_VAL and RVMODEL_MHPMEVENT_CODE?

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    `ifndef RVMODEL_MHPMCOUNTER
+        `error "RVMODEL_MHPMCOUNTER must be defined when SSCOFPMF_SUPPORTED"
+    `endif
+`endif
+
+covergroup Sscofpmf_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include "general/RISCV_coverage_standard_coverpoints.svh"
+
+    priv_mode_m: coverpoint ins.current.mode {
```
davidharrishmc: priv_mode_m/s/u are already defined in RISCV_coverage_standard_coverpoints and don't need to be redefined here.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins none = {6'b000000};
+            bins meip = {6'b100000};
+            bins mtip = {6'b010000};
+            bins msip = {6'b001000};
+            bins seip = {6'b000100};
+            bins stip = {6'b000010};
+            bins ssip = {6'b000001};
+    }
+
+    cp_minh_inhibits_mmode:  cross priv_mode_m, mhpmevent_minh, hpmcounter_nonzero, mhpmevent_of { ignore_bins of_one = binsof(mhpmevent_of.one); }
```
davidharrishmc: I think it is going to be too hard to identify which tests to run in the test/coverage files.  Let's split this into three coverage files SscofpmfSm, SscofpmfS, SscofpmfU that just check the corresponding privilege mode.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins U_mode = {2'b00};
+    }
+    priv_mode_m_maybes_u: coverpoint {ins.current.mode_virt, ins.prev.mode} {
+            bins M_mode = {3'b011};
+            bins U_mode = {3'b000};
+        `ifdef S_SUPPORTED
+                bins S_mode = {3'b001};
+        `endif
+    }
+    mhpmevent_of: coverpoint ins.current.csr[RVMODEL_MHPMEVENT][63] {
```
davidharrishmc: Put all these parts that are shared across coverage files in coverpoints/general/RISCV_coverage_sscofpmf_coverpoints.svh (as was done for pmm).

When developing tests, most of the test logic will be common and can go in testgen/priv/exgtensions/SscofpmfCommon.py the way that is done with Ssstrict.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins zero = {0};
```
davidharrishmc: It's not necessary to list zero and one bins - all the bins are created implicitly if { } is blank.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins yes = {1};
+            bins no  = {0};
+    }
+    mip_clear: coverpoint (ins.current.csr[CSR_MIP] == 0) {
+            bins yes = {1};
+    }
+    mie_clear: coverpoint (ins.current.csr[CSR_MIE] == 0) {
+            bins yes = {1};
+    }
+    mhpmcounter_all_ones: coverpoint (ins.current.csr[RVMODEL_MHPMCOUNTER] == {64{1'b1}}) {
```
davidharrishmc: Use '1 instead of {64{1'b1}}.  '1 fills all bits with 1 and works gracefully for both RV32 and 64.

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins yes = {1};
+    }
+    mhpmcounter_all_ones: coverpoint (ins.current.csr[RVMODEL_MHPMCOUNTER] == {64{1'b1}}) {
+            bins yes = {1};
+    }
+    mhpmcounter_all_zeros: coverpoint (ins.current.csr[RVMODEL_MHPMCOUNTER] == 64'h0) {
+            bins yes = {1};
+    }
+    mhpmcounter_extremes: coverpoint ins.current.csr[RVMODEL_MHPMCOUNTER] {
+            bins all_ones  = {'1};
```
davidharrishmc: Good style

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+                     ins.current.csr[CSR_MHPMEVENT17][63], ins.current.csr[CSR_MHPMEVENT16][63], \
+                     ins.current.csr[CSR_MHPMEVENT15][63], ins.current.csr[CSR_MHPMEVENT14][63], \
+                     ins.current.csr[CSR_MHPMEVENT13][63], ins.current.csr[CSR_MHPMEVENT12][63], \
+                     ins.current.csr[CSR_MHPMEVENT11][63], ins.current.csr[CSR_MHPMEVENT10][63], \
+                     ins.current.csr[CSR_MHPMEVENT9][63],  ins.current.csr[CSR_MHPMEVENT8][63], \
+                     ins.current.csr[CSR_MHPMEVENT7][63],  ins.current.csr[CSR_MHPMEVENT6][63], \
+                     ins.current.csr[CSR_MHPMEVENT5][63],  ins.current.csr[CSR_MHPMEVENT4][63], \
+                     ins.current.csr[CSR_MHPMEVENT3][63]}
+
+    mcounteren_all_ones: coverpoint (ins.current.csr[CSR_MCOUNTEREN] == '1) {
```
davidharrishmc: mcounteren is WARL and may have any bit read-only zero, so it may be impossible to hit this coverpoint.  Let's just write it to all 1s in the test (and it will take on whatever value it can), and drop this coverpoint from the cross-product because it is hard to test.  Alternatively, we could check that the previous instruction is a csr write of all 1s to mcounteren if you really want to be careful, but then you need to write 1s before each pattern.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins yes = {1};
+    }
+    of_pattern_class: coverpoint $countones(`OF_VEC) {
+            bins all_zeros   = {0};
+            bins walking_one = {1};
```
davidharrishmc: Does this check that all 1s are hit?  

See Sm_coverage.svh:
    walking_ones: coverpoint $clog2(ins.current.rs1_val) iff ($onehot(ins.current.rs1_val)) {
        bins b_1[] = { [0:`UDB_MXLEN-1] };
    }

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins match = {1};
+    }
+
+
+    of_write_pattern: coverpoint (`OF_VEC) {
+            bins all_ones     = {29'h1FFFFFFF};
+            bins checker_even = {29'b1_0101_0101_0101_0101_0101_0101_0101}; // even-indexed OF bits set
+            bins checker_odd  = {29'b0_1010_1010_1010_1010_1010_1010_1010}; // odd-indexed OF bits set
+    }
+    mcounteren_hpm_pattern: coverpoint (ins.current.csr[CSR_MCOUNTEREN][31:3]) {
```
davidharrishmc: Check the value written to mcounteren rather than the value in mcounteren because of WARL.

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins walking_one = {1};
+            bins all_ones    = {29};
+    }
+
+    scountovf_of_match: coverpoint (ins.current.csr[CSR_SCOUNTOVF][31:3] == `OF_VEC) {
+            bins match = {1};
+    }
+
+
+    of_write_pattern: coverpoint (`OF_VEC) {
```
davidharrishmc: We can only check what is written, not what is in mcounteren because of WARL.

coverpoints/priv/Sscofpmf_coverage.svh
davidharrishmc: Could use the same one hot notation for walking ones as above.

To reduce runtime, make sure the changes to mcounteren are in the inner loop in the test so mhpmcevent doesn't need to be rewritten each time mcounteren is tested.  It will take 32 register accesses to modify mhpmevent, and these will require T-SBI calls eventually, which is slow.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+                                29'h1000000, 29'h2000000, 29'h4000000, 29'h8000000,
+                                29'h10000000};
+    }
+    csrops: coverpoint ins.current.insn {
+            wildcard bins csrw = {CSRRW};
+            wildcard bins csrs = {CSRRS};
+            wildcard bins csrc = {CSRRC};
+    }
+
+    csr_write_value_pattern: coverpoint ins.current.rs1_val {
```
davidharrishmc: For this and crops, see csraccesses in Sm_coverage.svh/Sm_mcsr_cg.

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            wildcard bins csrc = {CSRRC};
+    }
+
+    csr_write_value_pattern: coverpoint ins.current.rs1_val {
+            bins all_ones  = {'1};
+            bins all_zeros = {'0};
+    }
+    hpm_csr_target: coverpoint ins.current.insn[31:20] {
+            bins scountovf   = {CSR_SCOUNTOVF};
+            bins mcounteren  = {CSR_MCOUNTEREN};
```
davidharrishmc: I don't see mcounteren in the testplan.  Why is it here.  I think it can be deleted here because it is tested in Sm, but otherwise the test plan would need to be updated.

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins mhpmevent[] = {CSR_MHPMEVENT3,  CSR_MHPMEVENT4,  CSR_MHPMEVENT5,
```
davidharrishmc: Testplan calls for mhpmeventh, not mhpmevent.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    mstatus_sie_set: coverpoint ins.current.csr[CSR_MSTATUS][1] {
+            bins one = {1};
+    }
+    sstatus_sie_set: coverpoint ins.current.csr[CSR_SSTATUS][1] {
+            bins one = {1};
+    }
+    sstatus_sie_clear: coverpoint ins.current.csr[CSR_SSTATUS][1] {
+            bins zero = {0};
+    }
+    mie_all_ones: coverpoint (ins.current.csr[CSR_MIE] == '1) {
```
davidharrishmc: We can write all 1s, but can't guarantee they will all be ones.  Easiest to leave out of the cross product.

coverpoints/priv/Sscofpmf_coverage.svh
davidharrishmc: If you just want to know mhpmevent.OF = 0, make a coverpoint with just that bin instead of using ignore_bins.

davidharrishmc: I don't see a coverpoint for MHPMEVENT[62:58].  It will need to be carefully written as a cross of five separate bits.  Most bits can only be 1 if the corresponding mode is supported, so the bits will need an `ifdef ?_SUPPORTED gating the 1 bin.

davidharrishmc: Actually just check the H/S/U bits, al of which are guaranteed to exist, so no `ifdef needed.  8 bins.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            bins msip = {6'b001000};
+            bins seip = {6'b000100};
+            bins stip = {6'b000010};
+            bins ssip = {6'b000001};
+    }
+
+    cp_minh_inhibits_mmode:  cross priv_mode_m, mhpmevent_minh, hpmcounter_nonzero, mhpmevent_of { ignore_bins of_one = binsof(mhpmevent_of.one); }
+    cp_sinh_inhibits_smode:  cross priv_mode_s, mhpmevent_sinh, hpmcounter_nonzero, mhpmevent_of { ignore_bins of_one = binsof(mhpmevent_of.one); }
+    cp_uinh_inhibits_umode:  cross priv_mode_u, mhpmevent_uinh, hpmcounter_nonzero, mhpmevent_of { ignore_bins of_one = binsof(mhpmevent_of.one); }
+    cp_of_set_on_overflow:   cross priv_mode_m_maybes_u, mip_clear, mie_clear, mhpmevent_of { ignore_bins of_zero = binsof(mhpmevent_of.zero); }
```
davidharrishmc: Why is this skipping the MHPMEVENT conditions?  Why is it checking mhpmevent_of = 1; the test plan says 0/1.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    cp_overflow_hw_only:     cross priv_mode_m_maybes_u, mip_clear, mie_clear, mhpmcounter_extremes, mhpmevent_of { ignore_bins of_one = binsof(mhpmevent_of.one); }
```
davidharrishmc: what about mhpmevent = 0? 
No need to check mhpmevent_of; the self-checking test checks that.  Coverpoints only need to check inputs, not expected outputs.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    cp_lcofip_hw_only:       cross priv_mode_s, mhpmevent_of_set_then_cleared, mip_lcofip_zero;
```
davidharrishmc: Don't need to check lcofip output

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    cp_scountovf_shadow:     cross priv_mode_s, mcounteren_all_ones, of_pattern_class, scountovf_of_match;
```
davidharrishmc: can't rely on mcounteren holding all ones or all of of being writable.  Your scountovf_of_match  is clever but comes useless when it isn't crossed with all 1s.

davidharrishmc: Most of these coverpoints will need trimming to reflect WARL registers that can't be guaranteed to hold a value.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    cp_scountovf_mcounteren: cross priv_mode_m_maybes_u, of_write_pattern, mcounteren_hpm_pattern {ignore_bins u_mode = binsof(priv_mode_m_maybes_u.U_mode);}
+    cp_sscofpmf_write:       cross priv_mode_m_maybes_u, csr_write_value_pattern, hpm_csr_target {
```
davidharrishmc: See Sm for how to do write, set, clear in one batch of coverpoints.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+            ignore_bins u_mode = binsof(priv_mode_m_maybes_u.U_mode);
+    }
+    cp_sscofpmf_setclear:    cross priv_mode_m_maybes_u, csrops, hpm_csr_target {
+            ignore_bins u_mode  = binsof(priv_mode_m_maybes_u.U_mode);
+            ignore_bins no_write = binsof(csrops.csrw);  // csrw handled by the cross above
+    }
+    cp_lcofi:                cross priv_mode_m_maybes_u, lcofi_ip, lcofi_ie, lcofi_mideleg, mstatus_mie_clear, mstatus_sie_set;
```
davidharrishmc: Testing of mideleg = 0 will need to be in a coverpoint that only is in SscofpmfSm, even though it runs from S and U mode, because S and U won't disable delegation.

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    cp_lcofi_sip_s:          cross priv_mode_s, sstatus_sie_set, lcofi_ie, lcofi_ip, lcofi_mideleg { ignore_bins mideleg_zero = binsof(lcofi_mideleg.zero); }
+    cp_lcofi_sip_u:          cross priv_mode_u, sstatus_sie_clear, lcofi_ie, lcofi_ip, lcofi_mideleg { ignore_bins mideleg_zero = binsof(lcofi_mideleg.zero); }
```
davidharrishmc: Add lcifi_mideleg_one coverpoint so you don't need ignore_bins.

davidharrishmc: Two places

coverpoints/priv/Sscofpmf_coverage.svh
```diff
@@ -0,0 +1,229 @@
...
+    cp_lcofip_priority:      cross priv_mode_m_maybes_u, mstatus_mie_set, sstatus_sie_set, mie_all_ones, lcofi_ip, mip_other_pending {
```
davidharrishmc: lcofi_ip_one

davidharrishmc: Add a csrw of all 1s to mie in cross product.

davidharrishmc (changes requested):
Good stuff.  I gave you bad advice to put all of these tests in one file.  It will be cleaner with a per-priv mode separation, like with many of our other test suites.

Be careful to cover as much of the input condition as possible, but don't need to have cover points for the output condition because self-checking catches that.  WARL fields are subtle and you can't guarantee which ones might have certain values, so we can only reliably check what we try to write, not what they actually contain.

coverpoints/priv/Sscofpmf_coverage.svh
ayeshaanwaar05: You meant M/S/U

coverpoints/priv/SscofpmfS_coverage.svh
ayeshaanwaar05: cp_scountovf_shadow says mcounteren = all 1's

davidharrishmc (commented):
A lot of this is repeated across files.  Can you create a RVTEST_coverage_sscofpmf file containing the common parts, similar to how RISCV_coverage_instr is used in SsstrictSm/S/U?

davidharrishmc:
Please click resolve on copilot feedback when you have addressed it.

config/imperas/imperas-rv32gck/rvmodel_macros.h
davidharrishmc, line 1: Something is funky that all of the lines are showing up as changed.

config/whisper/whisper-rv64-max/rvmodel_macros.h
```diff
@@ -213,4 +213,11 @@
   li _R2, WHISPER_SSIP_ADDRESS; \
   sw zero, 0(_R2);
 
+##### Sscofpmf #####
+
+#define RVMODEL_MHPMEVENT   mhpmevent3
+#define RVMODEL_MHPMCOUNTER mhpmcounter3
+#define RVMODEL_MHPMEVENT_VAL
```
davidharrishmc: Looks like this would not increment mhpmcounter3, so tests wouldn't pass?

coverpoints/priv/RVTEST_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,197 @@
...
+                     ins.current.csr[CSR_MHPMEVENTH15][31], ins.current.csr[CSR_MHPMEVENTH14][31], \
+                     ins.current.csr[CSR_MHPMEVENTH13][31], ins.current.csr[CSR_MHPMEVENTH12][31], \
+                     ins.current.csr[CSR_MHPMEVENTH11][31], ins.current.csr[CSR_MHPMEVENTH10][31], \
+                     ins.current.csr[CSR_MHPMEVENTH9][31],  ins.current.csr[CSR_MHPMEVENTH8][31], \
+                     ins.current.csr[CSR_MHPMEVENTH7][31],  ins.current.csr[CSR_MHPMEVENTH6][31], \
+                     ins.current.csr[CSR_MHPMEVENTH5][31],  ins.current.csr[CSR_MHPMEVENTH4][31], \
+                     ins.current.csr[CSR_MHPMEVENTH3][31]}
+    `endif
+    mcounteren_write_all_ones: coverpoint ins.current.insn {
+                wildcard bins write_ones = {CSRRW} iff (ins.current.insn[31:20] == CSR_MCOUNTEREN &&
```
davidharrishmc, line 91: extra tab

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    scountovf_of_match: coverpoint ((ins.current.csr[CSR_SCOUNTOVF][31:3] & ins.current.csr[CSR_MCOUNTEREN][31:3]) == (`OF_VEC & ins.current.csr[CSR_MCOUNTEREN][31:3])) {
+            bins match = {1};
+    }
+    sstatus_sie_set: coverpoint ins.current.csr[CSR_SSTATUS][1] {
+            bins one = {1};
+    }
+
+    cp_sinh_inhibits_smode:    cross priv_mode_s, mhpmevent_sinh, hpmcounter_nonzero, mhpmevent_of_zero ;
```
davidharrishmc, line 33: Where is mhpmevent_sinh defined?  (And similar bits)?

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,46 @@
...
+    cp_of_set_on_overflow:     cross priv_mode_s, mip_clear, mie_clear, mhpmevent_of;
```
davidharrishmc, line 34: Why isn't this able to include MHPMEVENT[62:58] = 11100?

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,46 @@
...
+    cp_overflow_hw_only:       cross priv_mode_s, mip_clear, mie_clear, mhpmcounter_extremes, mhpmevent_all_zero;
+    cp_lcofip_hw_only:         cross priv_mode_s, mhpmevent_of_set_then_cleared;
+    cp_scountovf_shadow:       cross priv_mode_s, mcounteren_write_all_ones, of_pattern_class, of_walking_one, scountovf_of_match;
```
davidharrishmc, line 37: of_pattern_class and of_walkign_one are mutually exclusive.  There needs to be a coverpoint that is the union of both, not the cross of these two coverpoints.

davidharrishmc: scountovf_of_match is an output, not an input.  Coverpoints only need to list inputs.  SIGUPD checks outputs match reference model.

coverpoints/priv/RVTEST_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,197 @@
...
+    `else
+        // On RV32, MINH/SINH/UINH/VSINH/VUINH live in mhpmevent*h[30:26] (address + 0x400)
+        mhpmevent_inhibits_all_set: coverpoint ins.current.insn {
+                wildcard bins write_pattern = {CSRRW} iff (ins.current.insn[31:20] == RVMODEL_MHPMEVENT + 12'h400 && ins.current.rs1_val[30:26] == 5'b11100);
+        }
+    `endif
+
+    hpm_csr_target: coverpoint ins.current.insn[31:20] {
+            bins scountovf   = {CSR_SCOUNTOVF};
+            bins mhpmevent[] = {CSR_MHPMEVENTH3,  CSR_MHPMEVENTH4,  CSR_MHPMEVENTH5,
```
davidharrishmc, line 150: The test plan only calls to read mhpmeventh if in rv32 and M-mode.  Right now  they are accessed from all modes and from RV64.  They are reserved if not implemented, so should not be tested in RV64.

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,46 @@
...
+    cp_scountovf_mcounteren:   cross priv_mode_s, of_write_pattern, mcounteren_write_pattern ;
+    cp_sscofpmf_access:        cross priv_mode_s, csr_access_pattern, hpm_csr_target ;
+    cp_lcofi_sip_s:            cross priv_mode_s, sstatus_sie_set, lcofi_ie, lcofi_ip, lcofi_mideleg_one ;
+    cp_lcofip_priority:        cross priv_mode_s, mstatus_mie_set, sstatus_sie_set, csrops,  lcofi_ip_one, mip_other_pending {ignore_bins not_csrw    = binsof(csrops.csrs) || binsof(csrops.csrc);}
```
davidharrishmc, line 41: What does crops have to do with this?

davidharrishmc: Why is there an ignore_bins?

coverpoints/priv/SscofpmfS_coverage.svh
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    cp_sinh_inhibits_smode:    cross priv_mode_s, mhpmevent_sinh, hpmcounter_nonzero, mhpmevent_of_zero ;
+    cp_of_set_on_overflow:     cross priv_mode_s, mip_clear, mie_clear, mhpmevent_of;
+    cp_overflow_hw_only:       cross priv_mode_s, mip_clear, mie_clear, mhpmcounter_extremes, mhpmevent_all_zero;
+    cp_lcofip_hw_only:         cross priv_mode_s, mhpmevent_of_set_then_cleared;
+    cp_scountovf_shadow:       cross priv_mode_s, mcounteren_write_all_ones, of_pattern_class, of_walking_one, scountovf_of_match;
+    cp_scountovf_mcounteren:   cross priv_mode_s, of_write_pattern, mcounteren_write_pattern ;
+    cp_sscofpmf_access:        cross priv_mode_s, csr_access_pattern, hpm_csr_target ;
+    cp_lcofi_sip_s:            cross priv_mode_s, sstatus_sie_set, lcofi_ie, lcofi_ip, lcofi_mideleg_one ;
```
davidharrishmc, line 40: Looks like cp_lcofi is missing.

davidharrishmc (changes requested):
Doesn't look like the coverpoints are implemented in accordance with the test plan.  I've flagged a number of things, but this needs a closer inspection.

config/imperas/imperas-rv32gck/rvmodel_macros.h
davidharrishmc, line 1: Why does this file still show up as entirely changed?  Same for other rvmodel_macros.h.

davidharrishmc: Why is this repeated?

davidharrishmc: There is no event val and code.  Put in a comment that Imperas doesn't have a way to trigger hpm events yet and so Sscofpmf can't be tested on this sim.  Same for other sims that lack the ability to trigger events.

coverpoints/priv/RISCV_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,217 @@
+///////////////////////////////////////////
```
davidharrishmc, line 1: This file should be named RISCV_coverage_sscofpmf to be similar to other shared files in the coverpoints directory.

tests/env/check_defines.h
```diff
@@ -5,7 +5,7 @@
 
 ########## test.S CHECKS ##########
 #ifndef TEST_FILE
-  #error "TEST_FILE not defined. It should be passed on the compiler command line."
+  #error "TEST_FILE not defined. It should be passed on the compiler command line."
```
davidharrishmc, line 8: what changed?

config/imperas/imperas-rv32gck/rvmodel_macros.h
davidharrishmc: Is it worth defining these at all if they have no value?

coverpoints/priv/RISCV_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,217 @@
...
+                wildcard bins write_pattern = {CSRRW} iff (ins.current.insn[31:20] == RVMODEL_MHPMEVENT + 12'h400 && ins.current.rs1_val[30:26] == 5'b11100);
+        }
+    `endif
+
+    `ifdef UDB_MXLEN_64
+        mhpmevent_of: coverpoint ins.current.csr[RVMODEL_MHPMEVENT][63] {
+                bins zero = {0};
+                bins one  = {1};
+        }
+        mhpmevent_of_zero: coverpoint ins.current.csr[RVMODEL_MHPMEVENT][63] {
```
davidharrishmc, line 35: Does a value for RVMODEL_MHPMEVENT get generated in work/<sim>/rvmodel_macros.svh from the .h file?  Does a script need to be updated to do this generation?  Are these coverpoints being hit?

coverpoints/priv/RVTEST_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,217 @@
...
+                bins zero = {0};
+                bins one  = {1};
+        }
+        mhpmevent_of_zero: coverpoint ins.current.csr[RVMODEL_MHPMEVENT][63] {
+                bins zero = {0};
+        }
+    `else
+        // On RV32, Sscofpmf bits (including OF) live in mhpmevent*h[31:28] (CSR address + 0x400)
+        mhpmevent_of: coverpoint ins.current.csr[RVMODEL_MHPMEVENT + 12'h400][31] {
+                bins zero = {0};
```
davidharrishmc: Don't need the zero and one.  See line 51 about bins automatically generated.

coverpoints/priv/RVTEST_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,217 @@
...
+                bins one  = {1};
+        }
+        mhpmevent_of_zero: coverpoint ins.current.csr[RVMODEL_MHPMEVENT + 12'h400][31] {
+                bins zero = {0};
+        }
+    `endif
+    sip_lcofi: coverpoint ins.current.csr[CSR_SIP][13] {}
+    sie_lcofi: coverpoint ins.current.csr[CSR_SIE][13] {}
+    hpmcounter_nonzero: coverpoint (ins.current.csr[RVMODEL_MHPMCOUNTER] != 0) {
+            bins yes = {1};
```
davidharrishmc: Don't need to list two bins - just {}

coverpoints/priv/RVTEST_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,217 @@
...
+            bins no  = {0};
+    }
+    mip_clear: coverpoint (ins.current.csr[CSR_MIP] == 0) {
+            bins yes = {1};
+    }
+    mie_clear: coverpoint (ins.current.csr[CSR_MIE] == 0) {
+            bins yes = {1};
+    }
+    mie_all_ones: coverpoint (ins.current.csr[CSR_MIE] == '1) {
+        bins yes = {1};
```
davidharrishmc: This bin can't be hit.  Many bits of MIE are read-only zero.  All we can do is attempt to write mie to all 1s.  Can't be sure all of the other mip bits are writable either.  Suggest dropping the coverpoints.

coverpoints/priv/RISCV_coverage_sscofpmf.svh
ayeshaanwaar05: Yes i manually added these macros in every config/rvmodel_macros.h. And no support on sail so idk if it gets hit or not only working on whisper and qemu

PR #1822, opened 2026-07-07
ExceptionsU using new boot code to boot to U-mode and use T-SBI calls

davidharrishmc:
Incorporated new trap handler.  Adjusted boot code and ExceptionsU. 
Fixed other tests that use ecall that broke T-SBI assumptions.
Clean up names and Sdtrig UDB in Sail Clang configs.


davidharrishmc:
Note: trap handler is saving and restoring 7 registers.  This interferes with registers that TSBI would like to access.  May have to return to the original design.

davidharrishmc:
Fails on cvw-rv32imc because of self-modifying code.  Otherwise working.

tests/env/rvtest_setup.h
```diff
@@ -122,7 +122,8 @@
   // Terminate test with a failure message
   abort_test:
     LI(     T5, 0xBAD0DEAD)
-    j       cleanup_epilogs
+//    j       cleanup_epilogs
+    j       rvtest_code_end // switch to M-mode and clean up and terminate
```
jordancarlin: Remove the commented out `j       cleanup_epilogs`. Both labels are at the same location, but `rvtest_code_end` is easier to understand.

tests/env/rvtest_setup.h
```diff
@@ -155,7 +156,10 @@
     #ifdef RVMODEL_IO_INIT
       RVMODEL_IO_INIT(T1, T2, T3)
     #endif
+
+    // boot to the lowest supported privilege mode unless a higher mode is specified by BOOT_TO_MMODE or BOOT_TO_SMODE
     // always boot to at least M-mode
+    // TODO: RVTEST_bOOT_TO_S/UMODE are temporarily gated, so we remain in M-mode unless BOOT_TO_S/UMODE parameters are set.  Remove this once all tests are ported to T-SBI.
```
jordancarlin:
```suggestion
    // TODO: RVTEST_BOOT_TO_S/UMODE are temporarily gated, so we remain in M-mode unless BOOT_TO_S/UMODE parameters are set.  Remove this once all tests are ported to T-SBI.
```

tests/env/rvtest_setup.h
```diff
@@ -232,41 +237,41 @@
...
+      RVMODEL_SET_SEXT_INT(a0, a1)
       ret
 
     rvtest_clr_sext_int:
-      RVMODEL_CLR_SEXT_INT(T2, T5)
-      LI(T3, 512)
-      csrc sip, T3 // clear sip.SEIP
+      RVMODEL_CLR_SEXT_INT(a0, a1)
+      LI(a1, 512)
+      csrc sip, a1 // clear sip.SEIP
```
jordancarlin: Use `a0` for consistency with `rvtest_clr_ssw_int`.

tests/env/rvtest_trap_handler.h
jordancarlin:
```suggestion
csrr    a0, CSR_XEPC                       // a0 = mepc = address of the ecall instruction
addi    T3, a0, 4                              // T3 = mepc + 4 (skip past ecall)
```

tests/env/rvtest_trap_handler.h
```diff
@@ -1649,6 +1634,114 @@ tsbi_\__MODE__\()handle_forwarded:
...
+        beq     T3, a0, found_instr                 // if it matches tsbi request, handle it
+        beqz    T3, tsbi_instr_not_found           // if we hit the end of the table (0 sentinel), not found
+        addi    T2, T2, 8                          // advance to next entry (4 bytes for instruction, 4 bytes for return)
+        j       tsbi_instr_table_search_loop          // repeat search
+found_instr:
+        mv      T4, x1                          // preserve caller's ra (T4's live value is already saved; restored by resto)
+        LA(     x1, tsbi_instr_rtn)             // table entries end in ret: link them to the epilogue below
+        jr      T2                                   // jump to the matching instruction in the table
+tsbi_instr_rtn:
+        mv      x1, T4                          // restore caller's ra
```
jordancarlin: ABI name makes intent clearer.

```suggestion
        mv      T4, ra                          // preserve caller's ra (T4's live value is already saved; restored by resto)
        LA(     ra, tsbi_instr_rtn)             // table entries end in ret: link them to the epilogue below
        jr      T2                                   // jump to the matching instruction in the table
tsbi_instr_rtn:
        mv      ra, T4                          // restore caller's ra
```

tests/env/rvtest_trap_handler.h
```diff
@@ -1649,6 +1634,114 @@ tsbi_\__MODE__\()handle_forwarded:
...
+        LA(a0, ascii_buffer)
+        call    rvmodel_io_write_str
+        LA(a0, newlinestr)
+        call    rvmodel_io_write_str
+        LA(a0, failstr)                         // RVCP-SUMMARY: TEST FAILED line so the harness classifies this run
+        call    rvmodel_io_write_str
+        call    rvmodel_halt_fail
+
+tsbi_instr_not_found_str:
+        .string "\nT-SBI ERROR: requested instruction not found in tsbi_instr_table: "
```
jordancarlin: Move this string into the `RVTEST_FAILURE_DATA` macro with the other strings so it is placed in a data region by the linker.

tests/env/rvtest_trap_handler.h
```diff
@@ -1649,6 +1634,114 @@ tsbi_\__MODE__\()handle_forwarded:
...
+.balign 4
+
+.macro TSBI_INSTR_TABLE csr_addr
```
jordancarlin: Probably worth naming this macro something that makes it clear it is CSR specific. Maybe TSBI_CSR_INSTR.

tests/env/rvtest_trap_handler.h
```diff
@@ -1649,6 +1634,114 @@ tsbi_\__MODE__\()handle_forwarded:
...
+// This table contains all instructions that the T-SBI knows how to run.
+// The dispatcher searches through the table for the matching instruction, runs it, and returns.
+// This avoids the need for self-modifying code.
+// Only CSRs that lower-privilege code might be interested in accessing are needed here.
+// Each instruction must be followed by a ret to return to the caller
+tsbi_instr_table:
+
+        TSBI_INSTR_TABLE(0x300) // mstatus
+        TSBI_INSTR_TABLE(0x302) // medeleg  // TODO: This might need to record the intended delegation state for delegating in software when bits are read-only zero
+        TSBI_INSTR_TABLE(0x303) // mideleg  // TODO: This might need to record the intended delegation state for delegating in software when bits are read-only zero
```
jordancarlin: When would these ever be accessed by the T-SBI? They should have everything possible delegated for S/U mode tests and will be directly accessed for M mode tests.

tests/env/rvtest_trap_handler.h
```diff
@@ -1649,6 +1634,114 @@ tsbi_\__MODE__\()handle_forwarded:
...
+        TSBI_INSTR_TABLE(0xB00) // mcycle
+        TSBI_INSTR_TABLE(0xB02) // minstret
+        TSBI_INSTR_TABLE(0x100) // sstatus
+        TSBI_INSTR_TABLE(0x104) // sie
+        //TSBI_INSTR_TABLE(0x105) // stvec
+        TSBI_INSTR_TABLE(0x106) // scounteren
+        TSBI_INSTR_TABLE(0x10A) // senvcfg
+        TSBI_INSTR_TABLE(0x144) // sip
+        TSBI_INSTR_TABLE(0x14D) // stimecmp
+        TSBI_INSTR_TABLE(0x180) // satp
```
jordancarlin: Shouldn't these be handled by an S-mode T-SBI dispatch? No need to include them in the M-mode handler because S-mode can directly access them.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -39,23 +40,29 @@ def _generate_mstatus_ie_tests(test_data: TestData) -> list[str]:
 
     lines = [
         comment_banner(coverpoint, "ecall from user mode with MIE=0 vs MIE=1"),
-        "RVTEST_GOTO_MMODE",
-        f"csrr x{save_reg}, mstatus",
-        f"LI(x{mask_reg}, 0x88)",
-        f"csrc mstatus, x{mask_reg}",
-        "RVTEST_GOTO_LOWER_MODE Umode",
+        # "RVTEST_GOTO_MMODE",
+        # f"csrr x{save_reg}, mstatus",
```
jordancarlin, line 44: Drop the comments with the old version. Many instances throughout this file.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -39,23 +40,29 @@ def _generate_mstatus_ie_tests(test_data: TestData) -> list[str]:
...
+        add_tsbi(f"csrr x{save_reg}, mstatus"),
+        f"LI(x{mask_reg}, 0x88) # mstatus.MIE=0, mstatus.MPIE=0",
+        # f"csrc mstatus, x{mask_reg}",
+        add_tsbi(f"csrc mstatus, x{mask_reg}"),
+        # RVTEST_GOTO_LOWER_MODE Umode",
         test_data.add_testcase("ecall_mie_0", coverpoint, covergroup),
+        "LI(a0, 0x00000073) # test ecall to execution environment that just returns",
         "ecall",
         "nop",
```
jordancarlin: Use `RVTEST_TSBI_ECALL_TEST`.

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
```diff
@@ -39,23 +40,29 @@ def _generate_mstatus_ie_tests(test_data: TestData) -> list[str]:
...
-        "RVTEST_GOTO_LOWER_MODE Umode",
+        # "RVTEST_GOTO_MMODE",
+        f"LI(x{mask_reg}, 0x80) # mstatus.MIE=0, mstatus.MPIE=1, then when returning from SBI call, mstatus.MIE becomes 1",
+        add_tsbi(f"csrs mstatus, x{mask_reg}"),
+        # f"csrrs x0, mstatus, x{mask_reg}",
+        # "RVTEST_GOTO_LOWER_MODE Umode",
         test_data.add_testcase("ecall_mie_1", coverpoint, covergroup),
+        "LI(a0, 0x00000073) # test ecall to execution environment that just returns",
         "ecall",
         "nop",
```
jordancarlin: Use `RVTEST_TSBI_ECALL_TEST`.

generators/testgen/src/testgen/asm/tsbi.py
```diff
@@ -0,0 +1,288 @@
...
+
+import re
+
+from testgen.constants import INDENT
+
+_CSR_INSTR_RE = re.compile(r"(?i)(csrr|csrw|csrs|csrc)\s+([^,\s]+)\s*,\s*([^,\s]+)")
+_MEM_INSTR_RE = re.compile(r"(?i)(lw|ld|sw|sd)\s+([^,\s]+)\s*,\s*([-+]?(?:0x[0-9a-f]+|\d+))\s*\(\s*([^\)\s]+)\s*\)")
+
+
+def add_tsbi(instr: str) -> str:
```
jordancarlin: Maybe `tsbi_call` (or something similar) would be a better name.

generators/testgen/src/testgen/asm/tsbi.py
```diff
@@ -0,0 +1,288 @@
...
+_CSR_ALIASES = {
+    "fflags": 0x001,
+    "frm": 0x002,
+    "fcsr": 0x003,
+    "cycle": 0xC00,
+    "time": 0xC01,
+    "instret": 0xC02,
+    "cycleh": 0xC80,
+    "timeh": 0xC81,
+    "instreth": 0xC82,
```
jordancarlin: These are all unprivileged CSRs accessible from any mode. No need to include them.

generators/testgen/src/testgen/asm/tsbi.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+def _normalize_instr(instr: str) -> str:
+    return instr.split("#", 1)[0].strip()
+
+
+def get_rd(instr: str) -> int:
+    normalized = _normalize_instr(instr)
+
+    if csr_match := _CSR_INSTR_RE.fullmatch(normalized):
```
jordancarlin: Rather than checking the instruction type in each of these helpers, it probably makes more sense to determine the instruction type in the main `add_tsbi` function and then call these helpers as appropriate based on that.

generators/testgen/src/testgen/asm/tsbi.py
```diff
@@ -0,0 +1,288 @@
...
+        (((imm12 >> 5) & 0x7F) << 25)
+        | ((rs2 & 0x1F) << 20)
+        | ((rs1 & 0x1F) << 15)
+        | ((funct3 & 0x7) << 12)
+        | ((imm12 & 0x1F) << 7)
+        | (opcode & 0x7F)
+    )
+
+
+def add_opcode(instr: str) -> str:
```
jordancarlin: A lot of this function does work that was already done in other functions. For example, `get_{rd,rs1,rs2}` were already called in `add_tsbi` a few lines before it calls `add_opcode`. The results of those other functions should be passed into this function rather than calling them twice.

jordancarlin (changes requested):

tests/env/rvtest_trap_handler.h
```diff
@@ -1649,6 +1634,114 @@ tsbi_\__MODE__\()handle_forwarded:
...
+        // loads and stores (these must not fault; the recursive trap handler may not save registers correctly)
+        lw a0, 0(a1)
+        ret
+        lw a0, 4(a1)
+        ret
```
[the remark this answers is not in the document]
davidharrishmc: Will do S-mode in the next PR

generators/testgen/src/testgen/priv/extensions/ExceptionsU.py
davidharrishmc: Leaving these comments as reference until other suites are ported.

generators/testgen/src/testgen/asm/tsbi.py
davidharrishmc: Made that change.  tsbi.py was made by AI and is working, and I'd rather move ahead than optimize it further.

generators/testgen/src/testgen/asm/tsbi.py
davidharrishmc: Fair, but runtime won't matter here and I don't want to spend time messing with AI-generated code too much.  It works and there are more important things to get to.

jordancarlin (approved):

tests/env/rvtest_trap_handler.h
```diff
@@ -1649,6 +1568,116 @@ tsbi_\__MODE__\()handle_forwarded:
...
+// tsbi_instr_table_dispatch
+tsbi_instr_table_dispatch:
+        LA(T2, tsbi_instr_table)                // initialize T2 = base address of instruction table
+tsbi_instr_table_search_loop:
+        #if (UDB_MXLEN==32)
+            lw      T3, 0(T2)                   // fetch current instruction (32-bit)
+        #else
+            lwu      T3, 0(T2)                  // fetch current instruction (64-bit); zero extend to 64 bits for comparison
+        #endif
+        lw      T3, 0(T2)                       // fetch current instruction
```
jordancarlin: Something seems to have gone wrong here. We are now loading the instruction twice.

jordancarlin (changes requested):

jordancarlin (approved):

PR #1825, opened 2026-07-08
Fix failedtest_saveresults_int secondary trap on invalid pointer

### LINKED ISSUE #1727: failedtest_saveresults_int triggers secondary trap and overwrites original exception cause
**Description:**

When failedtest_saveresults_int is entered due to a trap signature mismatch, the routine obtains an invalid pointer and dereferences it, causing a secondary trap that overwrites the original mcause value. This makes the true failure cause invisible in the exit report and produces a misleading diagnostic.

**Examples observed:**

* SiFive U74:
  invalid pointer = 0x8000c002
  secondary trap = Store/AMO address misaligned (mcause=6)

* Sail:
  invalid pointer = 0xBAAAAAADBAAAAD10
  secondary trap = Load access fault (mcause=5)
The secondary trap overwrites the original trap state before failure information is reported.

**Observed Failure on SiFive Hardware:**

Image

The original failure cause is mcause=2 (Illegal Instruction) at 0x80000c98. However the exit report shows mcause=6 (Store/AMO address misaligned) because the failure routine corrupted its working pointer to 0x8000c002 and a subsequent load or store via that pointer caused a secondary trap which overwrote mcause.
0x8000c002 is not in the signature region (sig_begin=0x80014dc0, sig_end=0x80033230) and not a valid aligned address (0x8000c002 % 8 = 2)

**Observed Failure on Sail (Forced Mismatch):**
To confirm the corruption is inside the failure routine itself and not hardware-specific, the same test was run on Sail with the expected signature deliberately set incorrectly to force entry into failedtest_saveresults_int. The same corruption pattern appeared with a different secondary trap:

Image

**Objdump of failedtest_saveresults_int at Crash Point:**

0000000080001c0c                      <failedtest_saveresults_int>:
  80001c0c: lhu  x6, -6(x5)
  80001c10: lhu  x7, -8(x5)
  80001c14: slli x6, x6, 0x10
  80001c18: or   x6, x6, x7
  80001c1c: srli x7, x6, 0xf
  80001c20: andi x7, x7, 31
  80001c24: srli x8, x6, 0x14
  80001c28: andi x8, x8, 31
  80001c2c: sw   x8, 260(x4)
  80001c30: slli x6, x8, 0x3
  80001c34: add  x6, x4, x6
  80001c38: ld   x6, 0(x6)                -> first indirect load via computed offset
  80001c3c: sd   x6, 272(x4)
  80001c40: lhu  x6, -10(x5)
  80001c44: lhu  x7, -12(x5)
  80001c48: slli x6, x6, 0x10
  80001c4c: or   x6, x6, x7
  80001c50: srai x7, x6, 0x14
  80001c54: srli x6, x6, 0xf
  80001c58: andi x6, x6, 31
  80001c5c: slli x6, x6, 0x3
  80001c60: add  x6, x4, x6
  80001c64: ld   x6, 0(x6)                 -> second indirect load via computed offset
  80001c68: add  x6, x6, x7
  80001c6c: ld   x6, 0(x6)                  -> Crash: x6 = 0xBAAAAAADBAAAAD10 (sentinel)
  80001c70: sd   x6, 280(x4)
  80001c74: jal  x0, 80002100             <failedtest_saveresults_common>

The routine computes a pointer by extracting fields from the signature word, using them as an index into a lookup table via x4, then dereferencing the result. When the signature word contains an unexpected value (due to mismatch in mstatus.FS bits), the extracted index lands on a sentinel entry (0xBAAAAAADBAAAAD10) in the lookup table. The routine then blindly dereferences this sentinel as a pointer, causing the secondary trap.

Vaibhav805:
Fixes #1727

## Root cause

`failedtest_saveresults_int` extracts register indices from the
instruction encoding around the failure point, uses them to index
into `DEFAULT_TEMP_REG`'s save area, then dereferences the result
as a pointer to load the expected signature value. When the
extracted index lands on a sentinel slot in the save area
(initialized with `0xBAAAAAADBAAAAD10` / `0xDEADBEEF`), the routine
blindly dereferences the sentinel as a pointer, causing a secondary
trap that overwrites the original `mcause` and produces misleading
diagnostics — the true failure cause becomes invisible in the exit
report.

Confirmed on both SiFive U74 hardware (secondary `mcause=6`,
Store/AMO address misaligned) and Sail with forced mismatch
(secondary `mcause=5`, Load access fault).

## Fix

Added null and alignment validation before the final `LREG`
dereference in `failedtest_saveresults_int`. Sentinel values are
always misaligned (`0xBAAAAAADBAAAAD10 % 8 != 0`), so an alignment
check catches all sentinel cases without needing to know the
signature region bounds. On invalid pointer, records `0` as the
expected value and continues to `failedtest_saveresults_common`,
preserving the original failure cause in `mcause`.

## Testing

Verified on `sail-rv64-max` with `ExceptionsS`:
```
✓ Build complete: 5 succeeded
RESULT: All 1 tests passed.
```


davidharrishmc:
Claude review:

Review: PR #1825 — Fix failedtest_saveresults_int secondary trap on invalid pointer

  Overview

  When a self-checking test fails, failedtest_saveresults_int (tests/env/rvtest_failure_code.h) reconstructs the ld/beq instructions preceding the failure point, extracts a register index and immediate, and dereferences saved_reg + imm to report the
  "expected value." When the extracted fields are garbage (issue #1727), that pointer is a save-area sentinel and the dereference raises a secondary trap that destroys the original mcause. The PR adds a null + alignment guard before the dereference and
  records 0 as the expected value when the guard fails. +10/−1, one file, targets act4.

  The direction is right — failure-reporting code must never trap — and the change is small, contained, and doesn't disturb the success path. But the guard's core premise doesn't hold, and I'd ask for one revision before merging.

  Main issue: the "sentinels are always misaligned" premise is false

  The PR's own quoted pointer disproves it: 0xBAAAAAADBAAAAD10 % 8 == 0 (any value ending in hex 0 or 8 is 8-byte aligned). That exact observed value passes both the beqz and the andi x8, x6, (REGWIDTH-1) check, gets dereferenced, and still raises the
  secondary trap this PR sets out to eliminate.

  The underlying reason: the guarded value is not the raw sentinel but sentinel + a garbage 12-bit immediate. The actual fills are .fill N, 4, 0xbaaaaaad and 0xdeadbeef (rvtest_failure_code.h:1936-1963, rvtest_setup.h:395), so the raw 64-bit patterns
  (…AAAD, …BEEF) are indeed misaligned — but adding an immediate with imm ≡ 3 (mod 8) (e.g. 0x…AAAD + 0x63 = 0x…AB10) re-aligns them. Roughly 1 in 8 garbage immediates slips through. The fix reduces the failure rate; it doesn't fix the bug.

  Suggested fix — validate the instruction, not the pointer. The root cause is trusting bytes at -12(DEFAULT_LINK_REG) that aren't actually a load. After reassembling x6, check opcode == 0b0000011 (LOAD) and funct3 matches the LREG width (011 ld / 010
  lw) before extracting rs1/imm; branch to the existing 1: fallback otherwise. That's ~4 instructions and deterministic. Alternatively (or additionally), bounds-check the pointer against [begin_signature, end_signature) — both symbols are linked into the
  same image. Either approach also catches aligned-but-wild pointers that the alignment check can never see (e.g. a stale-but-aligned register value).

  Secondary issues

  - Same unguarded pattern in two sibling paths. failedtest_saveresults_fflags (line 466) and failedtest_saveresults_fp (lines 504-506) do the identical reconstruct-and-dereference with no validation, so the same secondary-trap failure exists there. Fix
  them in this PR or note a follow-up issue — as-is, #1727 is only partially resolved even for the misaligned cases.
  - Recording 0 as "expected value" is ambiguous — 0 is a perfectly plausible expected value, so the report can't be distinguished from a real expected-0 mismatch. Consider storing the invalid pointer value itself (it makes the sentinel visible in the
  report, which is diagnostic gold) or a distinctive marker constant.
  - Comment needs correcting regardless of the approach taken: "sentinel values like 0xBAAAAAADBAAAAD10 or 0xDEADBEEF are misaligned" — the first value is aligned, and the real fill patterns are 0xbaaaaaad/0xdeadbeef words.

  What's correct and clean

  - x8 scratch use is safe: x8 (rs2 of the beq) is dead after line 418, and failedtest_saveresults_common overwrites x8 immediately (line 1102). Verified no live-range conflict.
  - Numeric local label 1: matches existing file style (the common handler already uses them) and can't collide.
  - REGWIDTH-1 mask is correct on both RV32 (4-byte LREG) and RV64.
  - No false rejections: legitimate signature pointers are always REGWIDTH-aligned by construction, so the guard can't suppress a real expected value.
  - Nice failure-mode analysis in the PR description (U74 hardware + forced Sail mismatch) — exactly the kind of evidence these fixes need.

  Testing

  Verified on sail-rv64-max ExceptionsS with a forced mismatch — good, but that exercised one lucky path. Worth forcing a mismatch where the garbage immediate is ≡ 3 mod 8 (aligned result) to demonstrate the residual hole; with the opcode-validation
  approach that case passes too.

  Verdict

  Request changes. The guard as written would have passed the very pointer value quoted in the PR description. Recommend validating the reconstructed instruction (opcode/funct3) or bounds-checking against the signature region, applying the same guard to
  the fflags/fp paths, and correcting the comment. The scaffolding (fallback label, store-0-and-continue flow, register choice) is all reusable as-is.

Vaibhav805:
@davidharrishmc @jordancarlin As per the previous review the pointer validation here was insufficient so in the latest commit i replaced the alignment check with opcode validation: reconstructed the
instruction at -12(DEFAULT_LINK_REG) and verify opcode[6:0] ==
0b0000011 (LOAD) before trusting its rs1/immediate fields and all the ci check passed successfully

jordancarlin:
@Vaibhav805 thanks for working on this. We've decided to go with a more holistic refactor of how the trap handler recovers the expected value in #1825, so I am going to close this. Thanks again for contributing!

PR #1827, opened 2026-07-08
Adding Ssnpm (Pointer masking ) test generator

Ammarahwakeel:
This draft PR introduces the initial implementation of the Ssnpm pointer-masking test generator.

It is being opened early to provide visibility into the overall structure and gather feedback.

Once Ssnpm reaches complete coverage, the common logic shared across the four pointer-masking test generators ( Ssnpm, Smmpm, SmnpmS, SmnpmU ) will be refactored into a shared module (ZpmCommon.py) in a follow-up PR.


Ammarahwakeel:
Prof. @davidharrishmc  and Sir @UmerShahidengr ,

 I'm currently debugging the Ssnpm test generator. Here's what I've narrowed down so far.

### Working case (Bare mode)

* In **Bare mode**, the generator builds successfully for **every upper 16-bit address pattern** and **every PMM value**.
* The generated tests run correctly and currently achieve around **15% coverage**.

image


### Issue when enabling virtual memory

As soon as I enable **Sv39**, the tests no longer complete—they appear to build/run indefinitely.

To isolate the problem, I tried the following:

1. **Upper pattern = `0x0000`**

   * All PMM values
   * All `satp.MODE` values (Sv39, Sv48, Sv57)

   This works correctly and produces coverage.

   
image


2. **PMM = `11` (PMLEN = 16)**

   * All `satp.MODE` values

   This also works correctly. Since PMLEN = 16 masks all of the upper pattern bits in this case, the tests complete successfully.

3. **Upper pattern = `0x0001` (bit 48 set)**

   This is where the problem begins.

   * With **PMM = `11` (PMLEN = 16)**, the tests still complete successfully.
   * However, with **PMM = `7` (PMLEN = 7)**, I get this error.
   
image

   

### My current understanding

My current hypothesis is that with **PMM = 7**, although pointer masking does work but bit 48 remains set, so for **Sv39** the requirement that bits **63:39** equal bit 38 is violated.

Based on my understanding, this should result in a **page-fault exception**. Instead, the test appears to get stuck in an infinite trap loop

At this point, I believe the issue is related to trap handler, but I'm not yet sure where the problem lies. It could be:

* an issue in the page table setup,
* a problem with how the trap handler is being used, or
* a bug in my generator.

P.S : for upper pattern 0x0001, PMM=10 and SATP.mode =sv39. This trap is correct to occur. In our testplan we have written that we will record any traps. Now I seem to not fully understand this part. As I donot think so this trap should throw an error  like that 

I'd really appreciate a review to determine whether my reasoning is correct or if I'm overlooking something.

Ammarahwakeel:
Also for now I am only testing  cp_pmlen_masking entry of testplan in this generator  and it doesnot yet include Vector instructions and Zicfiss instruction's test. I will push those soon.

jordancarlin:
Closing now that #2031 has been opened to replace this.

PR #1837, opened 2026-07-09
Trap handler updates for T-SBI / lower-privilege-mode execution

UmerShahidengr:
This PR fixes the register-clobber and mode-precondition bugs reported during the switchover.

**Changes**
* Trap handler no longer uses `a0/a1` (the T-SBI argument/result registers): `T5/T6` remapped from `x10/x11` to `x14/x15`. The dispatch now reads the caller's `a0/a1` directly from the live registers instead of the save area. This also fixes a latent bug where `resto_Xrtn` restored `x10` and thereby wiped T-SBI return values (e.g. `TSBI_ECALL_TEST` returned the opcode instead of xEPC).
* `a0` is now the single designated clobber register for all mode-switching macros: `RVTEST_GOTO_MMODE / RVTEST_GOTO_DELEGATED_MMODE / RVTEST_GOTO_SMODE` save `x3` in `a0` instead of silently clobbering `t0 (x5)`, matching the existing `RVTEST_TSBI_*` convention. testgen now reserves `x10/x11` for generated priv tests.
* `RVTEST_GOTO_LOWER_MODE` is now clobber-free: `T1/T2/T4` (and the pointer register `T3`) are saved to dedicated slots in the M-mode save area (`goto_lower_sv_off`, located via `mscratch`) and restored immediately after the `mret` in the target mode.
* `ra` is never left modified by the handler: the `CSR_ACCESS` scratch-execution `jalr ra` is wrapped with a save/restore (spare `trapreg_sv` slot 0) in both the M- and S-mode handlers, preserving the Sv/PMP fetch-fault contract (x4=0xACCE -> resume at ra).
* cleanup_epilogs now guarantees M-mode for every entry path: the `RVTEST_GOTO_MMODE` moved below the cleanup_epilogs label so `abort_test` and the default unexpected-interrupt handlers (which can run in S/U/VS mode) no longer reach the mscratch-reading epilogs in a lower mode.
* RV32 fix: `rvmodel_sv_off` was computed with 8-byte slots and pointed past the save area on RV32; now expressed in REGWIDTH (no change on RV64).
* `Sclr_Sext_int` uses a handler-context-safe inline `GOTO_MMODE` (via `T5`) so the interrupted test's `a0` is not clobbered.

### DIFF
--- a/framework/src/act/data/Gemfile.lock
+++ b/framework/src/act/data/Gemfile.lock
@@ -223,7 +223,7 @@ CHECKSUMS
   awesome_print (1.9.2) sha256=e99b32b704acff16d768b3468680793ced40bfdc4537eb07e06a4be11133786e
   base64 (0.3.0) sha256=27337aeabad6ffae05c265c450490628ef3ebd4b67be58257393227588f5a97b
   bigdecimal (4.1.2) sha256=53d217666027eab4280346fba98e7d5b66baaae1b9c3c1c0ffe89d48188a3fbd
-  bundler (4.0.15) sha256=a4ceb882fe94a0e0ac63cd0813932bbfd631a14e5ac0b7975189b19a4d28d9e7
+  bundler (4.0.13) sha256=19f08be7f27022cf0b89f27da0b044ae075e8270a9ef44ad248a932614e1ca3b
   commander (5.0.0) sha256=22ca42866ce5ed1a7f3e9a9c831a23d03b75b3743c7ad0302c3cbcd9eb5f257d
   concurrent-ruby (1.3.6) sha256=6b56837e1e7e5292f9864f34b69c5a2cbc75c0cf5338f1ce9903d10fa762d5ab
   connection_pool (3.0.2) sha256=33fff5ba71a12d2aa26cb72b1db8bba2a1a01823559fb01d29eb74c286e62e0a
@@ -304,4 +304,4 @@ RUBY VERSION
   ruby 3.4.9
 
 BUNDLED WITH
-  4.0.15
+  4.0.13
--- a/generators/testgen/src/testgen/generate/priv.py
+++ b/generators/testgen/src/testgen/generate/priv.py
@@ -54,9 +54,11 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     # Reserve registers for priv tests:
     #   - x0: avoid so desired values are actually loaded into registers
     #   - x1/ra: used as the return address for function calls
-    #   - x6, x7, x9: handler/boot temporaries (T1, T2, T4)
+    #   - x10, x11 (a0/a1): designated clobber registers of the mode-switching
+    #     macros. RVTEST_GOTO_MMODE/SMODE/DELEGATED_MMODE signal via a0 and leave
+    #     it as -1; RVTEST_TSBI_* pass arguments/results in a0/a1.
     #   - x16-x31: ensure the same test can be used for I or E bases
-    priv_exclude_regs = [0, 1, 6, 7, 9, *range(16, 32)]
+    priv_exclude_regs = [0, 1, 10, 11, *range(16, 32)]
     test_data.int_regs.consume_registers(priv_exclude_regs)
     seed(reproducible_hash(testsuite))
 
--- a/generators/testgen/src/testgen/generate/priv.py
+++ b/generators/testgen/src/testgen/generate/priv.py
@@ -55,12 +55,8 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     #   - x0: avoid so desired values are actually loaded into registers
     #   - x1/ra: used as the return address for function calls
     #   - x6, x7, x9: handler/boot temporaries (T1, T2, T4)
-    #   - x10, x11 (a0/a1): the designated clobber registers of every
-    #     mode-switching macro (RVTEST_GOTO_MMODE/SMODE/DELEGATED_MMODE and all
-    #     RVTEST_TSBI_* macros pass arguments / return results in a0/a1), so
-    #     tests must never hold live values there across a mode switch
     #   - x16-x31: ensure the same test can be used for I or E bases
-    priv_exclude_regs = [0, 1, 6, 7, 9, 10, 11, *range(16, 32)]
+    priv_exclude_regs = [0, 1, 6, 7, 9, *range(16, 32)]
     test_data.int_regs.consume_registers(priv_exclude_regs)
     seed(reproducible_hash(testsuite))
 
--- a/generators/testgen/src/testgen/generate/priv.py
+++ b/generators/testgen/src/testgen/generate/priv.py
@@ -54,9 +54,13 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     # Reserve registers for priv tests:
     #   - x0: avoid so desired values are actually loaded into registers
     #   - x1/ra: used as the return address for function calls
-    #   - x6, x7, x9: used by the RVTEST_GOTO_LOWER_MODE macro
+    #   - x6, x7, x9: handler/boot temporaries (T1, T2, T4)
+    #   - x10, x11 (a0/a1): the designated clobber registers of every
+    #     mode-switching macro (RVTEST_GOTO_MMODE/SMODE/DELEGATED_MMODE and all
+    #     RVTEST_TSBI_* macros pass arguments / return results in a0/a1), so
+    #     tests must never hold live values there across a mode switch
     #   - x16-x31: ensure the same test can be used for I or E bases
-    priv_exclude_regs = [0, 1, 6, 7, 9, *range(16, 32)]
+    priv_exclude_regs = [0, 1, 6, 7, 9, 10, 11, *range(16, 32)]
     test_data.int_regs.consume_registers(priv_exclude_regs)
     seed(reproducible_hash(testsuite))
 
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
@@ -139,7 +139,6 @@ def generate_instr_access_fault_tests(test_data: TestData, covergroup: str) -> l
         "#ifdef RVMODEL_ACCESS_FAULT_ADDRESS",
         comment_banner(coverpoint, "Instruction Access Fault"),
         f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
-        "LI(x4, 0xACCE)",  # trap handler checks x4 value and uses x1 (ra) as return address instead of mepc
         test_data.add_testcase("instr_access_fault", coverpoint, covergroup),
         f"jalr x1, 0(x{addr_reg})",
         "nop",
@@ -536,7 +535,6 @@ def generate_misaligned_priority_fetch_tests(
             "\n# misaligned fetch - existent address",
             f"LA(x{addr_reg}, {target_label})",
             f"addi x{addr_reg}, x{addr_reg}, 2",
-            "LI(x4, 0xACCE)",  # trap handler checks x4 value to use x1 (ra) as return address instead of mepc
             test_data.add_testcase(f"{name_prefix}misaligned_existent{name_suffix}", coverpoint, covergroup),
             f"jalr x1, 0(x{addr_reg})",
             "nop",
@@ -547,7 +545,6 @@ def generate_misaligned_priority_fetch_tests(
             "\n# misaligned fetch - non-existent (fault) address",
             f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
             f"addi x{addr_reg}, x{addr_reg}, 2",
-            "LI(x4, 0xACCE)",  # trap handler checks x4 value to use x1 (ra) as return address instead of mepc
             test_data.add_testcase(f"{name_prefix}misaligned_nonexistent{name_suffix}", coverpoint, covergroup),
             f"jalr x1, 0(x{addr_reg})",
             "nop",
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
@@ -114,22 +114,23 @@ def generate_instr_adr_misaligned_jalr_tests(test_data: TestData, covergroup: st
             lines.extend(
                 [
                     f"\n# rs1[1:0]={rs1_lsb:02b}, offset[1:0]={offset_lsb:02b}",
-                    ".p2align 2",                                        # align to 4-byte boundary
-                    f"auipc x{addr_reg}, 0",                          # PC+0  setup: addr_reg = PC
-                    f"addi x{addr_reg}, x{addr_reg}, {base_off}",     # PC+4  setup: addr_reg[1:0] = rs1_lsb
+                    ".p2align 2",  # align to 4-byte boundary
+                    f"auipc x{addr_reg}, 0",  # PC+0  setup: addr_reg = PC
+                    f"addi x{addr_reg}, x{addr_reg}, {base_off}",  # PC+4  setup: addr_reg[1:0] = rs1_lsb
                     test_data.add_testcase(
                         f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup
-                    ),                                                 # PC+8  (label)
-                    f"jalr x1, {jalr_off}(x{addr_reg})",              # PC+8  THE TEST
+                    ),  # PC+8  (label)
+                    f"jalr x1, {jalr_off}(x{addr_reg})",  # PC+8  THE TEST
                     "# branch by 6 lands in upper half of next instruction 0x0001 which is generated into a c.nop",
-                    "addi x0, x2, 0",                                 # PC+12 return for aligned jumps
-                    "nop",                                             # PC+16 safe landing pad
-                    "nop",                                             # PC+20 safe landing pad
+                    "addi x0, x2, 0",  # PC+12 return for aligned jumps
+                    "nop",  # PC+16 safe landing pad
+                    "nop",  # PC+20 safe landing pad
                 ]
             )
 
     test_data.int_regs.return_registers([addr_reg])
     return lines
+
 
 def generate_instr_access_fault_tests(test_data: TestData, covergroup: str) -> list[str]:
     coverpoint = "cp_instr_access_fault"
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
@@ -114,22 +114,22 @@ def generate_instr_adr_misaligned_jalr_tests(test_data: TestData, covergroup: st
             lines.extend(
                 [
                     f"\n# rs1[1:0]={rs1_lsb:02b}, offset[1:0]={offset_lsb:02b}",
-                    ".p2align 2",
-                    f"auipc x{addr_reg}, 0 # PC+0 addr_reg = PC",
-                    f"addi x{addr_reg}, x{addr_reg}, {base_off} # PC+4 addr_reg[1:0] = rs1_lsb",
-                    test_data.add_testcase(f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup),
-                    f"jalr x1, {jalr_off}(x{addr_reg}) # PC+8 jump target is PC + base_off + jalr_off (bit 0 cleared)",
-                    "# JALR target may land on the upper halfword (0x0001) of a padding ADDI, which decodes as a valid c.nop",
-                    "# With base_off/jalr_off in {6,7,8,9}, (base_off + jalr_off) spans 12..18 bytes",
-                    "# after clearing bit 0 the possible targets are 12/14/16/18, so use two padding instructions",
-                    "addi x0, x2, 0  # PC+12 padding (upper halfword is 0x0001)",
-                    "addi x0, x2, 0  # PC+16 padding (upper halfword is 0x0001)",
+                    ".p2align 2",                                        # align to 4-byte boundary
+                    f"auipc x{addr_reg}, 0",                          # PC+0  setup: addr_reg = PC
+                    f"addi x{addr_reg}, x{addr_reg}, {base_off}",     # PC+4  setup: addr_reg[1:0] = rs1_lsb
+                    test_data.add_testcase(
+                        f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup
+                    ),                                                 # PC+8  (label)
+                    f"jalr x1, {jalr_off}(x{addr_reg})",              # PC+8  THE TEST
+                    "# branch by 6 lands in upper half of next instruction 0x0001 which is generated into a c.nop",
+                    "addi x0, x2, 0",                                 # PC+12 return for aligned jumps
+                    "nop",                                             # PC+16 safe landing pad
+                    "nop",                                             # PC+20 safe landing pad
                 ]
             )
 
     test_data.int_regs.return_registers([addr_reg])
     return lines
-
 
 def generate_instr_access_fault_tests(test_data: TestData, covergroup: str) -> list[str]:
     coverpoint = "cp_instr_access_fault"
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
@@ -114,17 +114,16 @@ def generate_instr_adr_misaligned_jalr_tests(test_data: TestData, covergroup: st
             lines.extend(
                 [
                     f"\n# rs1[1:0]={rs1_lsb:02b}, offset[1:0]={offset_lsb:02b}",
-                    ".p2align 2",  # align to 4-byte boundary
-                    f"auipc x{addr_reg}, 0",  # PC+0  setup: addr_reg = PC
-                    f"addi x{addr_reg}, x{addr_reg}, {base_off}",  # PC+4  setup: addr_reg[1:0] = rs1_lsb
-                    test_data.add_testcase(
-                        f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup
-                    ),  # PC+8  (label)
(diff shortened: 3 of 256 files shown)


generators/testgen/src/testgen/generate/priv.py
```diff
@@ -54,9 +54,13 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
...
     #   - x1/ra: used as the return address for function calls
-    #   - x6, x7, x9: used by the RVTEST_GOTO_LOWER_MODE macro
+    #   - x6, x7, x9: handler/boot temporaries (T1, T2, T4)
+    #   - x10, x11 (a0/a1): the designated clobber registers of every
+    #     mode-switching macro (RVTEST_GOTO_MMODE/SMODE/DELEGATED_MMODE and all
+    #     RVTEST_TSBI_* macros pass arguments / return results in a0/a1), so
+    #     tests must never hold live values there across a mode switch
     #   - x16-x31: ensure the same test can be used for I or E bases
-    priv_exclude_regs = [0, 1, 6, 7, 9, *range(16, 32)]
+    priv_exclude_regs = [0, 1, 6, 7, 9, 10, 11, *range(16, 32)]
```
davidharrishmc: Tests are passing 0xACCE in x4 to indicate that they should return to ra from instruction fetch/page/guest page faults.  This means x4 also needs to be excluded.  It's also a nasty and poorly documented hacky interface to the trap handler.

I'm wondering if instead of using and excluding x4, can we always have fetch/page/guest page faults return to ra?  If they return to the caller, they will just trap again.  Is there any circumstance where this doesn't work?

tests/env/rvtest_trap_handler.h
```diff
@@ -87,6 +91,10 @@
 //    - a1 = optional argument (e.g., CSR rs1 value)
 //    - Result returned in a0
 //    - x3 must be non-zero (x3==0 is reserved for legacy GOTO_MMODE)
+//    - a0/a1 are the ONLY registers the handler may leave modified (a0 as the
```
davidharrishmc, line 93: We are going to need a2 for sw TSBI, but then that x12, which ties up the x12/13 pair, which seems to interfere with the framework (maybe for failure reporting).  Any ideas?

tests/env/rvtest_trap_handler.h
```diff
@@ -465,7 +482,10 @@
 //   - S-mode handler detects x3==0, sets SPP=1, bumps sepc+4, srets to S-mode
 //   - Only available when S_SUPPORTED is defined
 //
-// CLOBBERS: t0 (used to save/restore x3)
+// CLOBBERS: a0 (used to save/restore x3). a0 is the single designated clobber
+//   register for ALL mode-switching macros (matching the T-SBI convention,
+//   which also clobbers a0), so tests only ever have to keep one register
+//   free across a mode switch.
 // CONSTRAINT: x3 must not already be 0 when used for normal ecall handling
```
davidharrishmc: Why are we still using x3?  How about GOTO_M/S_OP just expecting the value in a0?

tests/env/rvtest_trap_handler.h
```diff
@@ -1454,9 +1510,9 @@ tsbi_\__MODE__\()csr_access:
         bnez    T3, \__MODE__\()trapsig_ptr_upd   // not a U-mode ecall -> normal trap sig recording
         beqz    x3, \__MODE__\()rtn2smode          // x3==0 -> legacy GOTO_SMODE -> rtn2smode handler
 
-        //--- T-SBI dispatch: recover caller's a0 and dispatch ---
+        //--- T-SBI dispatch: caller's a0 is live in its register ---
 tsbi_\__MODE__\()dispatch:
-        LREG    T3, trap_sv_off+5*REGWIDTH(sp)   // T3 = caller's original a0 (SBI operation code)
+        mv      T3, a0                            // T3 = caller's a0 (SBI operation code, live in register)
```
davidharrishmc: Why not just substitute a0 for T3 in the code below and not use T3 here?

tests/env/rvtest_trap_handler.h
```diff
@@ -2136,7 +2198,16 @@ excpt_\__MODE__\()hndlr_tbl:
...
                 .ifc \__MODE__ , S
-                        RVTEST_GOTO_MMODE
+                        // Handler-context GOTO_MMODE: RVTEST_GOTO_MMODE clobbers a0,
+                        // but here a0 belongs to the interrupted test and is NOT part
+                        // of the saved/restored register set. Use T5 to hold x3
+                        // instead — T5 is restored from the save area by resto_Srtn,
+                        // so its use here is invisible to the test.
+                        mv   T5, x3               // save x3 (T5 survives the nested trap: the
+                                                  // M-handler restores it from its save slot)
+                        li   x3, 0                // x3==0 signals legacy GOTO_MMODE
```
davidharrishmc: Is this the right time to switch from x3 to a0 as the register with the code for GOTO_MMODE?

generators/testgen/src/testgen/generate/priv.py
```diff
@@ -54,7 +54,7 @@ def generate_priv_test(testsuite: str, output_test_dir: Path) -> None:
     # Reserve registers for priv tests:
     #   - x0: avoid so desired values are actually loaded into registers
     #   - x1/ra: used as the return address for function calls
-    #   - x6, x7, x9: used by the RVTEST_GOTO_LOWER_MODE macro
+    #   - x6, x7, x9: handler/boot temporaries (T1, T2, T4)
     #   - x16-x31: ensure the same test can be used for I or E bases
     priv_exclude_regs = [0, 1, 6, 7, 9, *range(16, 32)]
```
davidharrishmc: I don't think this is right.  I thought we freed up 6/7/9 with your GOTO_MMODE change.

I think we need to exclude 10, 11 for a0/a1 to the T-SBI, and 12 a2 for sw call to T-SBI.

Right now the trap handler uses x4 to hold ACCE magic value.  I'm suggesting below to get rid of that.  If not, then x4 needs to be reserved too.

davidharrishmc (changes requested):

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: I am removing that part, working on removing SKIP_MEPC altogether, it is taking some time in debugging.

generators/testgen/src/testgen/generate/priv.py
UmerShahidengr: When I excluded x10 & x11, we ran out of registers to generate priv tests, python framework started giving errors, trying to free up x6, x7, and x9 and then I will include x10 & x11 into priv_exclude_regs list. That work is it progress

tests/env/rvtest_trap_handler.h
UmerShahidengr: a2 is free to use, I have switched T5/T6 to x14/x15, so both x12 & x13 are available. When we will add that SBI macro, we will update the comments

framework/src/act/data/Gemfile.lock
```diff
@@ -223,7 +223,7 @@ CHECKSUMS
   awesome_print (1.9.2) sha256=e99b32b704acff16d768b3468680793ced40bfdc4537eb07e06a4be11133786e
   base64 (0.3.0) sha256=27337aeabad6ffae05c265c450490628ef3ebd4b67be58257393227588f5a97b
   bigdecimal (4.1.2) sha256=53d217666027eab4280346fba98e7d5b66baaae1b9c3c1c0ffe89d48188a3fbd
-  bundler (4.0.15) sha256=a4ceb882fe94a0e0ac63cd0813932bbfd631a14e5ac0b7975189b19a4d28d9e7
+  bundler (4.0.13) sha256=19f08be7f27022cf0b89f27da0b044ae075e8270a9ef44ad248a932614e1ca3b
```
davidharrishmc, line 226: Is it necessary to downgrade bundler?

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -114,16 +114,17 @@ def generate_instr_adr_misaligned_jalr_tests(test_data: TestData, covergroup: st
...
-                    f"addi x{addr_reg}, x{addr_reg}, {base_off} # PC+4 addr_reg[1:0] = rs1_lsb",
-                    test_data.add_testcase(f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup),
-                    f"jalr x1, {jalr_off}(x{addr_reg}) # PC+8 jump target is PC + base_off + jalr_off (bit 0 cleared)",
-                    "# JALR target may land on the upper halfword (0x0001) of a padding ADDI, which decodes as a valid c.nop",
-                    "# With base_off/jalr_off in {6,7,8,9}, (base_off + jalr_off) spans 12..18 bytes",
-                    "# after clearing bit 0 the possible targets are 12/14/16/18, so use two padding instructions",
-                    "addi x0, x2, 0  # PC+12 padding (upper halfword is 0x0001)",
-                    "addi x0, x2, 0  # PC+16 padding (upper halfword is 0x0001)",
+                    ".p2align 2",  # align to 4-byte boundary
+                    f"auipc x{addr_reg}, 0",  # PC+0  setup: addr_reg = PC
```
davidharrishmc: Keep comments inside quotes so they appear in the assembly language.

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
```diff
@@ -114,16 +114,17 @@ def generate_instr_adr_misaligned_jalr_tests(test_data: TestData, covergroup: st
...
                     f"\n# rs1[1:0]={rs1_lsb:02b}, offset[1:0]={offset_lsb:02b}",
-                    ".p2align 2",
-                    f"auipc x{addr_reg}, 0 # PC+0 addr_reg = PC",
-                    f"addi x{addr_reg}, x{addr_reg}, {base_off} # PC+4 addr_reg[1:0] = rs1_lsb",
-                    test_data.add_testcase(f"jalr_rs1_{rs1_lsb}_off_{offset_lsb}", coverpoint, covergroup),
-                    f"jalr x1, {jalr_off}(x{addr_reg}) # PC+8 jump target is PC + base_off + jalr_off (bit 0 cleared)",
-                    "# JALR target may land on the upper halfword (0x0001) of a padding ADDI, which decodes as a valid c.nop",
-                    "# With base_off/jalr_off in {6,7,8,9}, (base_off + jalr_off) spans 12..18 bytes",
-                    "# after clearing bit 0 the possible targets are 12/14/16/18, so use two padding instructions",
-                    "addi x0, x2, 0  # PC+12 padding (upper halfword is 0x0001)",
```
davidharrishmc, line 125: This code had been working, after a recent fix.  Unclear if this might undo the fix.  Does it need more addi x0, x2, 0 so the upper halfword will be 0x0001 and not be an illegal compressed instruction?  Not sure why another word of distance is needed.

tests/priv/ExceptionsSv/sv32_exceptions_Umode.S
```diff
@@ -66,7 +66,6 @@ RVTEST_BEGIN
   nop
 
   // Test execute permission
-  LI(  x4, 0xACCE)            // Store a value which is to be checked in trap handler
```
davidharrishmc, line 69: Looks like sv{32/39}_exceptions_mprv_{S/U}_Mmode.S, sv32_exceptions_Smode.S,
many other sv tests
many pmp
still contain 0xACCE.
Suggest global search.

tests/env/rvtest_trap_handler.h
```diff
@@ -12,11 +12,14 @@
...
-//    1. REGISTER ALIASES        — T1..T6 mapped to x6..x11; DEFAULT_SIG/DATA/TEMP/LINK
+//    1. REGISTER ALIASES        — T1..T4 mapped to x6..x9, T5/T6 to x14/x15;
+//                                  DEFAULT_SIG/DATA/TEMP/LINK. a0/a1 (x10/x11) are
+//                                  deliberately NOT used as handler temporaries: they
+//                                  are the T-SBI argument/return registers.
 //    2. ARCHITECTURE CONSTANTS  — interrupt/exception cause counts, masks, mode encodings
 //    3. T-SBI DEFINITIONS       — Test Supervisor Binary Interface operation codes and macros
 //    4. CSR RENAME MACROS       — XCSR_RENAME parameterizes CSR names by privilege mode
-//    5. GOTO_MMODE / GOTO_SMODE — Legacy macros using x3==0 convention for mode switching
+//    5. GOTO_MMODE / GOTO_SMODE — Legacy macros, overwritten with the new T-SBI macros
```
davidharrishmc, line 22: Do we need an issue to find and remove these legacy macros?

tests/env/rvtest_trap_handler.h
```diff
@@ -182,6 +182,7 @@
 //   4. Otherwise       -> RESERVED        (return -1)
 //==============================================================================
 
+#define ALT_GOTO_MMODE      0x00000000           // a0 value: Used by RVTEST_GOTO_DELEGATED_MMODE
```
davidharrishmc, line 185: Please add more information about why this is needed.  Should it be part of the T-SBI documentation?  Or should there be an issue to replace this with TSBI_GOTO_MMODE?

tests/env/rvtest_trap_handler.h
davidharrishmc: As far as I can tell, Smstateen.py and Ssstateen.py are the only places that use GOTO_MMODE as opposed to RVTEST_GOTO_MMODE (and similar GOTO_SMODE).

tests/env/rvtest_trap_handler.h
davidharrishmc: I don't see ALT_GOTO_MMODE referenced in the PR or code anywhere else. Maybe it can be deleted?

tests/env/rvtest_trap_handler.h
```diff
@@ -440,87 +452,11 @@
...
-//==============================================================================
-// SECTION 11: T-SBI CONVENIENCE MACROS FOR TESTS
+// SECTION 10: T-SBI CONVENIENCE MACROS FOR TESTS
 //
 // These macros provide a clean test-facing interface for the a0-based T-SBI
-// calling convention. Unlike the legacy GOTO_MMODE (which uses x3==0), these
-// set a0 to the operation code and ecall. The handler distinguishes these from
-// legacy calls by checking x3 != 0.
+// calling convention. Unlike the legacy GOTO_MMODE, these
+// set a0 to the operation code and ecall.
```
davidharrishmc, line 459: I think we can drop mention of legacy GOTO_MMODE if it is is no longer in this file.

tests/env/rvtest_trap_handler.h
```diff
@@ -628,6 +564,81 @@
...
+.macro  RVTEST_GOTO_MMODE
+  .option push
+  .option norvc                                  // disable compressed instructions for consistent code size
+  li   a0, 0                                    // set a0=0 to signal GOTO_MMODE to handler
+  GOTO_M_OP                                      // ecall: traps to M-mode handler
+  li   a0, -1                                   // handler returns HERE (in M-mode): kill the a0==0 signal
+  .option pop
+.endm
+
+.macro  RVTEST_GOTO_DELEGATED_MMODE
```
davidharrishmc, line 621: I don't think there is a need for this anymore.  When ecall from U-mode is delegated to S-mode, then a T-SBI call to goto M-mode will first trap to S-mode.  The handler should detect it is a GOTO_MMODE code in a0 and then make another trap to go to M-mode.  I think the code will be less confusing if we can drop RVTEST_GOTO_DELEGATED_MMODE, and then replace GOTO_M_OP with ecall.  This could be a followup PR if we make this an issue.

tests/env/rvtest_trap_handler.h
```diff
@@ -628,6 +564,81 @@
...
+
+#ifndef CAUSE_SPCL_GO2MMODE_OP
+    #define ALT_GOTO_M_CAUSE CAUSE_ILLEGAL_INSTRUCTION  // alt cause: illegal instruction
+    #define ALT_GOTO_M_OP    .word 0                     // alt op: emit illegal instruction encoding
+#endif
+
+.macro  RVTEST_GOTO_MMODE
+  .option push
+  .option norvc                                  // disable compressed instructions for consistent code size
+  li   a0, 0                                    // set a0=0 to signal GOTO_MMODE to handler
```
davidharrishmc, line 615: I thought GOTO_MMODE is now 1.

tests/env/rvtest_trap_handler.h
```diff
@@ -628,6 +564,81 @@
...
+  GOTO_M_OP                                      // ecall: traps to M-mode handler
+  li   a0, -1                                   // handler returns HERE (in M-mode): kill the a0==0 signal
```
davidharrishmc, line 617: Why is this needed?

tests/env/rvtest_trap_handler.h
```diff
@@ -628,6 +564,81 @@
...
+  ALT_GOTO_M_OP                                  // .word 0: triggers illegal instruction (not delegated)
+  li   a0, -1                                   // handler returns HERE (in M-mode): kill the a0==0 signal
+  .option pop
+.endm
+
+.macro  RVTEST_GOTO_SMODE
+  .option push
+  .option norvc                                  // disable compressed instructions
+  #ifdef  S_SUPPORTED
+    li   a0, 0                                  // set a0=0 to signal GOTO_SMODE
```
davidharrishmc, line 634: I thought this is now 2.

tests/env/rvtest_trap_handler.h
```diff
@@ -628,6 +564,81 @@
...
+  li   a0, 0                                    // set a0=0 to signal GOTO_MMODE
+  ALT_GOTO_M_OP                                  // .word 0: triggers illegal instruction (not delegated)
+  li   a0, -1                                   // handler returns HERE (in M-mode): kill the a0==0 signal
+  .option pop
+.endm
+
+.macro  RVTEST_GOTO_SMODE
+  .option push
+  .option norvc                                  // disable compressed instructions
+  #ifdef  S_SUPPORTED
```
davidharrishmc, line 633: This should never be called if S is not supported.  Should the compiler generate an error with a #else if no S-mode?

tests/env/rvtest_trap_handler.h
```diff
@@ -628,6 +564,81 @@
 .endm
 #endif // H_SUPPORTED
 
+//==============================================================================
+// SECTION 11: LEGACY MODE-SWITCHING MACROS
```
davidharrishmc, line 568: Open an issue to deprecate these?

tests/env/rvtest_trap_handler.h
```diff
@@ -1152,30 +1196,29 @@ spcl_\__MODE__\()chk4ecall:                      // Step 2: Check if cause is an
...
+        // so a0/a1 were never overwritten).
         //
         // Available registers: T2, T3, T4 (T5 = xcause, T1/T6 saved)
-        // Caller's a0: at trap_sv_off+5*REGWIDTH(sp)
-        // Caller's a1: at trap_sv_off+6*REGWIDTH(sp)
         //==============================================================
 
 tsbi_\__MODE__\()dispatch:
-        LREG    T3, trap_sv_off+5*REGWIDTH(sp)   // T3 = caller's original a0 (SBI operation code)
+        mv      T3, a0                            // T3 = caller's a0 (SBI operation code, live in register)
```
davidharrishmc: What if we just replace T3 with a0 in the code below and drop the mv?  Does it still work?

tests/env/rvtest_trap_handler.h
```diff
@@ -1452,11 +1500,11 @@ tsbi_\__MODE__\()csr_access:
...
         addi    T3, T4, -CAUSE_USER_ECALL          // T3 = masked_cause - 8 (U-mode ecall = cause 8)
         bnez    T3, \__MODE__\()trapsig_ptr_upd   // not a U-mode ecall -> normal trap sig recording
-        beqz    x3, \__MODE__\()rtn2smode          // x3==0 -> legacy GOTO_SMODE -> rtn2smode handler
+        beqz    a0, \__MODE__\()rtn2smode          // a0==0 -> legacy GOTO_SMODE -> rtn2smode handler
 
-        //--- T-SBI dispatch: recover caller's a0 and dispatch ---
+        //--- T-SBI dispatch: caller's a0 is live in its register ---
 tsbi_\__MODE__\()dispatch:
-        LREG    T3, trap_sv_off+5*REGWIDTH(sp)   // T3 = caller's original a0 (SBI operation code)
+        mv      T3, a0                            // T3 = caller's a0 (SBI operation code, live in register)
```
davidharrishmc: Just use a0 in place of T3 below and drop the mv?

tests/env/rvtest_trap_handler.h
```diff
@@ -1885,12 +1944,43 @@ sv_\__MODE__\()epc:
...
+  .ifc \__MODE__ , M
+        // M-mode has no translation of its own. Use MPRV so the load is
+        // performed with MPP's privilege + translation (resolves a virtual
+        // xEPC). SUM lets an effective-S load reach U pages; MXR lets the
+        // load read execute-only (R=0,X=1) pages. Relies on mstatus.MPP
+        // still holding the trapped privilege (don't write MPP before here).
+        csrr    T6, CSR_MSTATUS                      // save full mstatus
+        li      T2, (1 << MPRV_LSB) | (1 << SUM_LSB) | (1 << MXR_LSB)
+        or      T2, T6, T2
+        csrw    CSR_MSTATUS, T2
```
davidharrishmc, line 1971: This could have been csrs CSR_MSTATUS, T2 and then drop the or.

tests/env/rvtest_trap_handler.h
```diff
@@ -1885,12 +1944,43 @@ sv_\__MODE__\()epc:
...
+        lhu     T2, 0(T3)                            // fetch via trapped context
+        csrw    CSR_MSTATUS, T6                      // restore mstatus verbatim
+  .else
+        // Below M: the load already runs in the trapped privilege with the
+        // active satp, so no MPRV. Still need SUM (read U pages) and MXR
+        // (read execute-only pages).
+        csrr    T6, CSR_SSTATUS                      // save full sstatus
+        li      T2, (1 << SUM_LSB) | (1 << MXR_LSB)
+        or      T2, T6, T2
+        csrw    CSR_SSTATUS, T2
```
davidharrishmc, line 1981: same csrs

davidharrishmc (changes requested):

tests/env/rvtest_trap_handler.h
UmerShahidengr: It can still work, but since T3 will revived at the end of trap handler, so I did not put major effort in using a0 and seeing if it will break anything. All logic is getting driven by T0-T5 registers so I kept that flow intact.

tests/env/rvtest_trap_handler.h
UmerShahidengr: Same comment, T3 will be revived, so I ran the logic with T* registers.

framework/src/act/data/Gemfile.lock
UmerShahidengr: It was accidental. Recovered now

tests/priv/ExceptionsSv/sv32_exceptions_Umode.S
UmerShahidengr: Did global search and removed from everywhere

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
UmerShahidengr: Updated

generators/testgen/src/testgen/priv/extensions/ExceptionsCommon.py
UmerShahidengr: I compiled the exact instruction sequence and disassembled it: with this test's march (rv32i_zicsr_zifencei, no C extension), nop assembles to 0x00000013 — the RISC-V spec's reserved all-zero illegal-instruction encoding. The original code deliberately used addi x0, x2, 0 instead (0x00010013, whose upper halfword 0x0001 is a valid c.nop), because these tests can legitimately land mid-instruction on the padding when a DUT does support C. The last changes replaced the second padding instruction with plain nops, silently reintroducing an illegal-instruction fault into a test that's supposed to cleanly validate JALR misalignment only. I have now restored the two-addi padding and, per other comment, I moved the trailing comments back inside the emitted assembly strings.

tests/env/rvtest_trap_handler.h
UmerShahidengr: Yes, I will add this issue

tests/env/rvtest_trap_handler.h
UmerShahidengr: Sv tests are using these macros to switch back to Mmode, in case if ecall is delegated, this will require some more work to remove it. I have kept it here to make current Sv tests work

tests/env/rvtest_trap_handler.h
UmerShahidengr: TSBI_GOTO_MMODE is 1, this is the older one which is getting used in multiple tests along with the GOTO_DELEGATED_MMODE, these macros are still here in this file

tests/env/rvtest_trap_handler.h
UmerShahidengr: Killing this signal was needed as trap handler was getting confused with a0==0 on the next trap. It is just an additional safety net

tests/env/rvtest_trap_handler.h
UmerShahidengr: Yes, you are right, this will require updates in Sv tests to make them fully compatible with new TSBI macros, once they are done, I will remove these macros

tests/env/rvtest_trap_handler.h
UmerShahidengr: No need to generate an error, this macro will do nothing, so it will silently fall through

tests/env/rvtest_trap_handler.h
UmerShahidengr: This is 2 in latest TSBI macros, once again it is old implementation. I will park an issue against it

tests/env/rvtest_trap_handler.h
UmerShahidengr: Agree. Will do it

davidharrishmc (approved):

framework/src/act/data/Gemfile.lock
davidharrishmc: I will fix.

PR #1842, opened 2026-07-10
Add pre-commit hook to enforce LI/LA macros

### LINKED ISSUE #1810: Add pre-commit check to enforce the use of LI and LA macros
Similar to what was done in #1804 for `.align`, we should add a pre-commit hook that prohibits the use of `li`/`la` pseudoinstructions and suggest using the `LI`/`LA` macros instead.

pyarchana:
Closes #1810

this PR adds a pre-commit hook that enforces the use of `LI()` and `LA()` macros instead of raw `li` and `la` pseudoinstructions in assembly test files under `tests/`. the hook is scoped to `tests/` only, since `tests-dev/` uses a different framework that doesn't have these macros available.

along with the hook, all existing test files and generator scripts have been updated to use `LI()`/`LA()` so the hook passes on the current codebase.


.github/scripts/check-li-la-pseudoinstructions.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+pattern='(^\s*(li|la)\s)|(;\s*(li|la)\s)'
+
+if [ -t 1 ] && [ -z "${NO_COLOR:-}" ]; then
+  red='\033[31m'; bold='\033[1m'; reset='\033[0m'
+else
+  red=''; bold=''; reset=''
+fi
```
jordancarlin, line 15: Split each of these into separate lines to make them easier to read.

.github/scripts/check-li-la-pseudoinstructions.sh
```diff
@@ -0,0 +1,26 @@
...
+
+found=0
+for f in "$@"; do
+  matches=$(grep -nE "$pattern" "$f" || true)
+  if [ -n "$matches" ]; then
+    found=1
+    while IFS= read -r line; do echo "$f:$line"; done <<< "$matches"
```
jordancarlin:
```suggestion
    while IFS= read -r line; do
      echo "$f:$line"
    done <<< "$matches"
```

.pre-commit-config.yaml
```diff
@@ -103,3 +103,13 @@ repos:
...
+
+  # Forbid raw li/la pseudoinstructions in assembly test files;
+  # require the LI()/LA() macros from tests/env/utils.h instead.
+  - repo: local
+    hooks:
+      - id: forbid-li-la-pseudoinstructions
+        name: forbid li/la pseudoinstructions (use LI/LA macros)
+        language: script
+        entry: .github/scripts/check-li-la-pseudoinstructions.sh
+        files: ^tests/.*\.(S|s)$
```
jordancarlin: We should run this on `*.py` files as well so that it gets flagged in the generators. `*.h` files should be checked too.

.github/scripts/check-li-la-pseudoinstructions.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
jordancarlin, line 1: Missing license header. Add `# SPDX-License-Identifier: Apache-2.0`.

.github/scripts/check-li-la-pseudoinstructions.sh
```diff
@@ -0,0 +1,26 @@
...
+set -euo pipefail
+
+pattern='(^\s*(li|la)\s)|(;\s*(li|la)\s)'
```
jordancarlin: This pattern is too strict. The pseudoinstruction can be capitalized or can come after a label, and both of those would be missed. I think this needs to be generalized to catch any instance of li or la (case-insensitive) that is not followed by a parenthesis and is preceded by either a new line, a space, a colon, a semicolon, or a quotation mark.

jordancarlin (changes requested):

.pre-commit-config.yaml
pyarchana: okay

.pre-commit-config.yaml
```diff
@@ -103,3 +103,14 @@ repos:
...
+
+  # Forbid raw li/la pseudoinstructions in assembly test files;
+  # require the LI()/LA() macros from tests/env/utils.h instead.
+  - repo: local
+    hooks:
+      - id: forbid-li-la-pseudoinstructions
+        name: forbid li/la pseudoinstructions (use LI/LA macros)
+        language: script
+        entry: .github/scripts/check-li-la-pseudoinstructions.sh
+        files: \.(S|s|py)$
```
jordancarlin: This still needs to add `*.h` header files.

jordancarlin (approved):

jordancarlin (changes requested):
Hit the wrong button. See my previous comment.

.pre-commit-config.yaml
pyarchana: my bad. will fix this soon.

.pre-commit-config.yaml
```diff
@@ -103,3 +103,14 @@ repos:
...
+  # Forbid raw li/la pseudoinstructions in assembly test files;
+  # require the LI()/LA() macros from tests/env/utils.h instead.
+  - repo: local
+    hooks:
+      - id: forbid-li-la-pseudoinstructions
+        name: forbid li/la pseudoinstructions (use LI/LA macros)
+        language: script
+        entry: .github/scripts/check-li-la-pseudoinstructions.sh
+        files: \.(S|s|py|h)$
+        exclude: ^(tests-dev/|tests/env/|config/)
```
jordancarlin: Do not exclude tests-env. That is where all of the `*.h` header files are and are the main reason for including `*.h` files in the first place.

jordancarlin (changes requested):

tests/env/sail_macros.h
```diff
@@ -107,55 +107,55 @@
...
+  LI(_R1, (1 << 31) | (1 << 1);               \)
+  LI(_R2, SAIL_SIG_ADDRESS;    \)
   sw _R1, 0(_R2)            ; /* Set SSW interrupt */ \
 
 #undef RVMODEL_CLR_SSW_INT
 #define RVMODEL_CLR_SSW_INT(_R1, _R2)        \
-  li _R1, (1 << 1);               \
-  li _R2, SAIL_SIG_ADDRESS;    \
+  LI(_R1, (1 << 1);               \)
+  LI(_R2, SAIL_SIG_ADDRESS;    \)
```
jordancarlin: The closing parentheses all throughout this file seem wrong. They should be before the semicolon, not after the backslash.

tests/env/rvtest_pmp_macros.h
jordancarlin, line 1: Why are there so many unrelated changes in this file? Please only change li/la to LI/LA.

jordancarlin (changes requested):
Looks like there are merge conflicts that need to be resolved.

PR #1843, opened 2026-07-10
T-SBI Documentation Updates

davidharrishmc:
Fixed GOTO_MODE codes to match implementation
Add load and store functions to T-TSBI definition


UmerShahidengr (approved):
LGTM. Lets lock this documentation, and adjust trap handler to make everything consistent.

PR #1844, opened 2026-07-10
split InterruptsS into Sm and S

Teemooooooooo:
Splitting InterruptsS into InterruptsS and InterruptsSSm based on whether the test needs to run in M mode.

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/InterruptsS.py
+++ b/generators/testgen/src/testgen/priv/extensions/InterruptsS.py
@@ -22,7 +22,7 @@ from testgen.data.test_chunk import TestChunk
 from testgen.priv.registry import add_priv_test_generator
 
 
-def _generate_trigger_sti_tests(test_data: TestData) -> list[str]:
+def _generate_trigger_sti_tests(test_data: TestData) -> list[TestChunk]:
     """Generate STIP trigger tests.
 
     With mstatus.MIE = 0, mstatus.SIE = {0/1}, mideleg = {0s/STI+SEI+SSI},
@@ -34,6 +34,7 @@ def _generate_trigger_sti_tests(test_data: TestData) -> list[str]:
 
     r_mtime, r_temp, r_temp2, r_stimecmp, r_scratch, r_stce = test_data.int_regs.get_registers(6)
 
+    tc = test_data.begin_test_chunk("S")
     lines = [
         comment_banner(
             "cp_trigger_sti",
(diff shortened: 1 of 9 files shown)


generators/testgen/src/testgen/priv/extensions/InterruptsS.py
jordancarlin: Should this be updated to no longer require Sm?

jordancarlin: Also, while we are here, why does it require I? Shouldn't it work with E too?

generators/testgen/src/testgen/priv/extensions/InterruptsSSm.py
```diff
@@ -0,0 +1,1606 @@
...
+                if is_timer:
+                    lines.extend(clr_mtimer_int(r_temp, r_stimecmp))
+                else:
+                    lines.append(int_clr)
+
+    test_data.int_regs.return_registers([r_mtime, r_temp, r_temp2, r_stimecmp, r_scratch, r_stce])
+    return lines
+
+
+@add_priv_test_generator("InterruptsSSm", required_extensions=["Sm", "S", "I", "Zicsr"])
```
jordancarlin: Same comment, no need for I.

generators/testgen/src/testgen/priv/extensions/InterruptsSSm.py
```diff
@@ -0,0 +1,1606 @@
+##################################
+# priv/extensions/interruptsSSm.py
```
jordancarlin:
```suggestion
# priv/extensions/InterruptsSSm.py
```

generators/testgen/src/testgen/priv/extensions/InterruptsSSm.py
```diff
@@ -0,0 +1,1606 @@
...
+    """Generate interrupt tests in M-mode.
+
+    Cross: MIE={0,1} × mideleg={0,0x222} × 6 mip walking × 6 mie walking
+    Total: 2 × 2 × 6 × 6 = 144 bins
+
+    Routes to:
+    - cp_interrupts_m: Non-delegated (mideleg=0) OR M-interrupts (always M-mode)
+    - cp_interrupts_m_deleg: Delegated S-interrupts (mideleg=0x222 + SSIP/STIP/SEIP)
+    """
+    covergroup = "InterruptsS_S_cg"
```
jordancarlin: Global change in this file.
```suggestion
    covergroup = "InterruptsSSm_cg"
```

jordancarlin (changes requested):
A few minor comments, but looks pretty good.

jordancarlin:
Also, the CTP needs to be updated. At a minimum the `docs/ctp/src/priv.adoc` description of InterruptsS is now outdated.

jordancarlin (approved):

PR #1863, opened 2026-07-14
Improvements on Sm and S

davidharrishmc:


generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -713,7 +713,17 @@ def make_s(test_data: TestData) -> list[TestChunk]:
...
-    tc.code.extend(_generate_scsr_tests(test_data))
+    test_chunks.append(test_data.end_test_chunk())
 
+    tc = test_data.begin_test_chunk("scsr")
+    tc.code.extend(
+        [
+            "",
+            "",
+            "RVTEST_GOTO_MMODE  # Get back to machine mode to prepare to go to supervisor mode",
+            "RVTEST_GOTO_LOWER_MODE Smode  # Run remaining tests in supervisor mode",
```
jordancarlin: No need for RVTEST_GOTO_MMODE. A named test_chunk starts a new file, which begins in M mode.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -322,11 +322,10 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
     csrs = [
         # TODO: sail does not yet support sstatus.S/M/UBE; mask it until available to avoid mismatches.  Delete mask when Sail has endian support.
         ("mstatus", 0xFFFFFFCFFFFFFFBF),
-        ("medeleg", 0xFFFFFF),  # mask off custom bits and reserved bits
+        # ("medeleg", 0xDBBFE),  # mask off custom bits and reserved bits and bits that should be ROZ (Double trap [16], ecall from M [11]; instr misaligned [0] depends on MISALIGNED_LDST so don't check it
```
jordancarlin: Why do we need to drop `medeleg`? Especially now that Sail lets you configure the writable bits?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -387,6 +386,8 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
     lines.extend(csr_access_test(test_data, ("mseccfg", None), covergroup, coverpoint))
     lines.append("#endif")
 
+    return lines
```
jordancarlin: Why was this added? I think this means all of the other tests generated in this function (read-only CSRs, cp_mcsrwalk, cp_csr_insufficient_priv, etc.) are no longer tested. Did you check coverage on this PR?

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -322,11 +322,13 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
     csrs = [
         # TODO: sail does not yet support sstatus.S/M/UBE; mask it until available to avoid mismatches.  Delete mask when Sail has endian support.
         ("mstatus", 0xFFFFFFCFFFFFFFBF),
-        ("medeleg", 0xFFFFFF),  # mask off custom bits and reserved bits
+        (
+            "medeleg",
+            0xDBBFE,
+        ),  # mask off custom bits and reserved bits and bits that should be ROZ (Double trap [16], ecall from M [11]; instr misaligned [0] depends on MISALIGNED_LDST so don't check it
```
jordancarlin: Still not clear why we are masking off bits that should be read-only zero. Don't we need to test that those bits are read-only zero? Especially 16 and 11.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -335,35 +337,35 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
-        ("mhpmevent23", None),
-        ("mhpmevent24", None),
-        ("mhpmevent25", None),
-        ("mhpmevent26", None),
-        ("mhpmevent27", None),
-        ("mhpmevent28", None),
-        ("mhpmevent29", None),
-        ("mhpmevent30", None),
-        ("mhpmevent31", None),
+        ("mhpmevent3", 0),  # mask all bits because they are WARL can can all be ROZ
```
jordancarlin: Global comment
```suggestion
        ("mhpmevent3", 0),  # mask all bits because they are WARL and can all be ROZ
```

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -322,11 +322,13 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
     csrs = [
         # TODO: sail does not yet support sstatus.S/M/UBE; mask it until available to avoid mismatches.  Delete mask when Sail has endian support.
         ("mstatus", 0xFFFFFFCFFFFFFFBF),
-        ("medeleg", 0xFFFFFF),  # mask off custom bits and reserved bits
+        (
+            "medeleg",
+            0xDBBFE,
+        ),  # mask off custom bits and reserved bits; instr misaligned [0] depends on MISALIGNED_LDST so don't check it
```
jordancarlin: Doesn't instruction misaligned depend on the C extension, not MISALIGNED_LDST?

jordancarlin: That bit is also configurable in the sail model now, so we could check it.

jordancarlin (changes requested):
A few last minor nits.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: Fixed language.  Leaving it unchecked.

jordancarlin (approved):

PR #1867, opened 2026-07-14
Remove Bad Dead Link Check from CI

fourth-bit:


jordancarlin (changes requested):
I believe you also need to update `pre-commit-config.yaml` to pass this file with the `-c` flag to the check.

jordancarlin (approved):

PR #1868, opened 2026-07-14
Norm rule mapping for Zawrs

Teemooooooooo:
Norm rule mapping for Zawrs

also added in spec description of the normative rules for ExceptionsZalrsc and ExceptionsZaamo


davidharrishmc (approved):
LGTM

coverpoints/norm/Zawrs.yaml
jordancarlin, line 1: If we deleted this file and added the priv mode specific versions, the CTP probably needs to be updated.

PR #1901, opened 2026-07-17
RVA23 CRD with tightly limited M-mode requirements

davidharrishmc:


jordancarlin:
There are a lot of non CRD changes in here. Please pull those out of this PR and then I will review.

docs/crd/src/rva23_crd.adoc
```diff
@@ -105,12 +99,11 @@ RVA23 certification requires that the SBI provide methods of emulating every wri
...
+| `mseccfg` | PMLEN must be able to hold at least PMLEN=0 and 7.  SSEED and USEED must be controllable if Zkr is supported; otherwise they are read-only zero.  MLPE, RLB, MMWP, MML are OUT-OF-SCOPE because they do not relate to RVA23 features.
+| `mstateen` | SE0, ENVCFG, CONTEXT may be controllable to allow testing of the behavior of lower privilege modes when certain features are disabled. Other fields are OUT-OF-SCOPE because they do not relate to RVA23 features.
+| `medeleg` | Each exception cause except `ecall` from S-mode is delegated to S-mode if the corresponding bit can be written to 1.  Otherwise, a user-supplied M-mode trap handler intercepts invisible traps and logs the rest or relays them to S-mode.
+| `mideleg` | SEI, STI, SSI, and LCOFI are delegated to S-mode if the corresponding bit can be written to 1.  Otherwise, a user-supplied M-mode trap handler relays these to S-mode. VSEI, VSTI, VSSI, and SGEI are read-only 1 because hypervisor is a mandatory part of the profile.
```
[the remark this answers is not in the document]
davidharrishmc: RO1 per priv spec 22.4.2.

docs/crd/src/rva23_crd.adoc
```diff
@@ -105,12 +99,11 @@ RVA23 certification requires that the SBI provide methods of emulating every wri
...
-| `mstateen` | SE0, ENVCFG, CONTEXT must be controllable to allow testing of the behavior of lower privilege modes when certain features are disabled. Note that some bits may be read-only zero.  Other fields are OUT-OF-SCOPE because they do not relate to RVA23 features.
-| `medeleg` | Each exception cause is tested both with (if possible) and without delegation.
-| `mideleg` | Each interrupt cause is tested both with (if possible) and without delegation except that MTI, MEI, MSI are not delegated.
+| `menvcfg` | STCE, PBMTE, ADUE, PMM, CBZE, CBCFE, CBIE, SSE, LPE may be controllable to allow testing of the behavior of lower privilege modes with these features enabled and disabled. Note that ADUE may be read-only zero because Svadue is optional, and that PMM may only be able to hold PMLEN=0 and 7 per the RVA23 profile. CBIE must be able to hold 00 and at least one of 01 and/or 11. CDE, DTE are OUT-OF-SCOPE because they do not relate to RVA23 features. FIOM is OUT-OF-SCOPE because I/O memory is not required by the RVA23 profile. All writable fields must be controllable to allow testing of the behavior of lower privilege modes when certain features are disabled.
```
jordancarlin:
```suggestion
| `menvcfg` | STCE, PBMTE, ADUE, PMM, CBZE, CBCFE, CBIE, SSE, LPE may be controllable to allow testing of the behavior of lower privilege modes with these features enabled and disabled. Note that ADUE may be read-only zero because Svadu is optional, and that PMM may only be able to hold PMLEN=0 and 7 per the RVA23 profile. CBIE must be able to hold 00 and at least one of 01 and/or 11. CDE, DTE are OUT-OF-SCOPE because they do not relate to RVA23 features. FIOM is OUT-OF-SCOPE because I/O memory is not required by the RVA23 profile. All writable fields must be controllable to allow testing of the behavior of lower privilege modes when certain features are disabled.
```

docs/crd/src/rva23_crd.adoc
```diff
@@ -105,12 +99,11 @@ RVA23 certification requires that the SBI provide methods of emulating every wri
...
+| `mseccfg` | PMLEN must be able to hold at least PMLEN=0 and 7.  SSEED and USEED must be controllable if Zkr is supported; otherwise they are read-only zero.  MLPE, RLB, MMWP, MML are OUT-OF-SCOPE because they do not relate to RVA23 features.
```
jordancarlin: `mseccfg.PMM` controls pointer-masking in M-mode (Smmpm), which is OUT-OF-SCOPE. Drop PMLEN from this entry.

docs/crd/src/rva23_crd.adoc
```diff
@@ -143,24 +136,27 @@ The following M-mode CSRs/fields are OUT-OF-SCOPE and not tested:
 | `mtval` | Does not affect the behavior of lower privilege modes.
 | `mtinst` | Does not affect the behavior of lower privilege modes.
 | `mtval2` | Does not affect the behavior of lower privilege modes.
+| `mtimecmp` | MTI is not tested.
```
jordancarlin, line 139: How do we test STI with Sstc disabled?

jordancarlin (changes requested):

docs/crd/src/rva23_crd.adoc
davidharrishmc: By writing mip.STIP.

jordancarlin: Oops. Duh.

docs/crd/src/rva23_crd.adoc
```diff
@@ -105,12 +99,11 @@ RVA23 certification requires that the SBI provide methods of emulating every wri
...
-| `medeleg` | Each exception cause is tested both with (if possible) and without delegation.
-| `mideleg` | Each interrupt cause is tested both with (if possible) and without delegation except that MTI, MEI, MSI are not delegated.
+| `menvcfg` | STCE, PBMTE, ADUE, PMM, CBZE, CBCFE, CBIE, SSE, LPE may be controllable to allow testing of the behavior of lower privilege modes with these features enabled and disabled. Note that ADUE may be read-only zero because Svadu is optional, and that PMM may only be able to hold PMLEN=0 and 7 per the RVA23 profile. CBIE must be able to hold 00 and at least one of 01 and/or 11. CDE, DTE are OUT-OF-SCOPE because they do not relate to RVA23 features. FIOM is OUT-OF-SCOPE because I/O memory is not required by the RVA23 profile. All writable fields must be controllable to allow testing of the behavior of lower privilege modes when certain features are disabled.
+| `mseccfg` | SSEED and USEED must be controllable if Zkr is supported; otherwise they are read-only zero.  PMLEN, MLPE, RLB, MMWP, MML are OUT-OF-SCOPE because they do not relate to RVA23 features.  Note `mseccfg.PMLEN` affects Smmpm, which is not part of RVA23.
```
jordancarlin:
```suggestion
| `mseccfg` | SSEED and USEED must be controllable if Zkr is supported; otherwise they are read-only zero.  PMM, MLPE, RLB, MMWP, MML are OUT-OF-SCOPE because they do not relate to RVA23 features.  Note `mseccfg.PMM` affects Smmpm, which is not part of RVA23.
```

jordancarlin (changes requested):

jordancarlin (approved):

docs/crd/src/rva23_crd.adoc
```diff
@@ -47,55 +50,47 @@ NOTE: While it is permitted for an implementation to also support RV32I instruct
...
 
-As defined in the Unprivileged ISA, the SEE, which will typically include M-mode hardware, must be able to provide certain capabilities to the underlying lower privileged implementation. For example, an SBI can be used to request that some or all traps be delegated to be directly handled in S-mode. Which traps can be delegated is up to the SEE.
+RVA23 requires that the execution environment boot to a deterministic state that allows software to run in S-mode, and to handle traps to the execution environment that implements a test SBI (described below). If a custom M-mode execution environment does not directly implement a feature of standard M-mode in hardware, it must be able to emulate that feature in a way that is transparent to software running in lower privilege modes.  For example, if it does not support delegating interrupts to S-mode in hardware, it must accomplish the same purpose in software by trapping to the execution environment, which in turn delivers the interrupt to S-mode.
```
kdockser, line 69: The RVA23 _certification tests_ require boot to a a deterministic state. Not RVA23 itself.

docs/crd/src/rva23_crd.adoc
```diff
@@ -47,55 +50,47 @@ NOTE: While it is permitted for an implementation to also support RV32I instruct
...
-
-RVA23 requires that the execution environment boot to a deterministic state that allows software to run in S-mode, and to handle traps to the execution environment that implements a test SBI (described below). If a custom M-mode execution environment does not directly implement a feature of standard M-mode in hardware, it must be able to emulate that feature in a way that is transparent to software running in lower privilege modes.  For example, if it does not support delegating interrupts to S-mode in hardware, it must accomplish the same purpose in software by trapping to the execution environment, which delivers the interrupt to S-mode.
+Nevertheless, RVA23S64 runs within a supervisor execution environment.  For example, an `ECALL` from supervisor mode causes a requested trap to this execution environment. Similarly, the profile indicates that Sspm requires an execution environment providing a means to select `PMLEN`.
```
kdockser, line 67: Remove "Nevertheless" as it is unnecessary and can cause confusion,

docs/crd/src/rva23_crd.adoc
```diff
@@ -47,55 +50,47 @@ NOTE: While it is permitted for an implementation to also support RV32I instruct
...
 
 === Custom M-mode
 
 The RVA23S64 profile purposefully excludes Machine mode (M-mode).  It does not require a _Standard M-mode_ consistent with the ratified M-mode specification.
-Therefore, certification does not run M-mode code directly.
 
-Custom M-mode is a supervisor execution environment that deviates from the M-mode specification.  For example, it might have a bug that affects M-mode behavior but is invisible from S-mode.  The firmware could work around this bug by not running M-mode instructions affected by the bug.  Alternatively, Custom M-mode might have some security features or proprietary CSRs that break compatibility with ratified M-mode.
+Custom M-mode is a supervisor execution environment that deviates from the M-mode specification.  For example, it might have a bug that affects M-mode behavior but is invisible from S-mode.  The firmware could work around this bug by not running M-mode instructions affected by the bug.  Alternatively, Custom M-mode might have some security features or custom CSRs that break compatibility with ratified M-mode.
```
kdockser, line 59: M-mode is not a supervisor execution environment.
The talk of custom M-mode is likely to cause more confusion than clarity.
M-Mode 1.13 is a base specification that allows for custom extensions without being considered a "custom M-mode". Many (perhaps most) implementations add custom features yet are compliant with 1.13. Even CHERI (which is soon to be standard) is not a custom M-mode.
The SEE is used to abstract the implemented M-mode and implementation specific details so that OSs can work with all M-mode implementations.

docs/crd/src/rva23_crd.adoc
```diff
@@ -47,55 +50,47 @@ NOTE: While it is permitted for an implementation to also support RV32I instruct
 
 == Discussion
 
-The RVA23S64 profile implies requirements that might not be immediately
-obvious.  This section explains them in detail.
+The primary goal of RVA23 is to provide binary compatibility for S- and U-mode operating system and application code.  It does not impose requirements on M-mode firmware.  Certification deliberately avoids imposing requirements on M-mode behavior except where called out in the profile or where M-mode controls behaviors of lower privilege modes.
```
kdockser, line 53: RVA23 does impose requirements on firmware but we don't need to get into that level of detail. What it doesn't do is impose _M-level ISA requirements_.

docs/crd/src/rva23_crd.adoc
```diff
@@ -22,13 +22,16 @@ History of documentation changes that eventually lead to releases.
...
+| 0.2
+a| * Do not test traps into M-mode except ecall from S-mode
 
 |===
 
-// include::glossary.adoc[]
-
 == Introduction
 
+This document describes the requirements for a processor to be eligible for a RVA23 certificate.
```
kdockser, line 33: This describes the _test_ requirements for a certificate. They are not the only requirements.
We need to be a bit more specific or readers will misinterpret these to be the _only_ requirements.

docs/crd/src/rva23_crd.adoc
```diff
@@ -47,55 +50,47 @@ NOTE: While it is permitted for an implementation to also support RV32I instruct
...
 
-NOTE: There is no requirement that any traps can be delegated. But if they can be, the SEE must provide a means to request that they are and a means to determine which requested traps have been delegated.
+The execution environment must provide a Test Supervisor Binary Interface (T-SBI) that allows S-mode software to make requests of the execution environment.  These requests are made by placing commands in registers and performing an `ECALL`.  The certification suite provides a reference T-SBI implementation that runs on standard M-mode, but custom M-mode that differs significantly from standard M-mode might have to provide its own T-SBI implementation.
 
-The execution environment must provide a Test Supervisor Binary Interface (T-SBI) that allows S-mode software to make requests of the execution environment.  These requests are made by placing commands in registers and performing an `ECALL`.
+The T-SBI also encompasses boot code and an M-mode trap handler.  Both might need to be modified by the user to accommodate custom M-mode behavior.
```
kdockser, line 73: Both might need to be modified by the user to accommodate implementation specific behavior.

docs/crd/src/rva23_crd.adoc
```diff
@@ -47,55 +50,47 @@ NOTE: While it is permitted for an implementation to also support RV32I instruct
...
-. Check if the trap is a test of undelegated traps to M-mode, and if so, write/compare a trap signature of the standard values expected in `mepc`, `mcause`, `mtval` and `mstatus.MPV/MPP/MPIE/MPELP`.
-. Set up `sstatus`, `sepc`, `scause`, `stval` with values coming from the machine-mode trap CSRs, and return to the S-mode trap handler to effectively delegate to S-mode.
+. For traps that are not delegated in hardware but should still go to S-mode, set up `sstatus`, `sepc`, `scause`, `stval` with values coming from the machine-mode trap CSRs, and return to the S-mode trap handler to effectively delegate to S-mode. [*** This should be optional if the user prefers to handle them in M-mode; the traps must still be logged in a way compatible with standard M-mode of the reference model]
```
kdockser, line 89: What is meant by "traps that are not delegated in hardware"? Are we talking about traps that go to M mode? Those are out of scope for RVA23. However, they are required for a RISC-V implementation.

docs/crd/src/rva23_crd.adoc
```diff
@@ -47,55 +50,47 @@ NOTE: While it is permitted for an implementation to also support RV32I instruct
...
 
 === SBI Control of M-mode CSRs
 
-RVA23 certification requires that the SBI provide methods of emulating every writable standard M-mode CSR field that can affect the behavior of lower privilege modes.  With a standard M-mode, the SBI may just write bitfields in the M-mode registers.  With custom M-mode, the SBI must make it appear to S-mode software that the field holds the value written by the SBI, even though the implementation might be different. These CSRs and memory-mapped I/O registers are:
+RVA23 certification requires that the T-SBI provide methods of emulating every writable standard M-mode CSR field that can affect the behavior of lower privilege modes.  With a standard M-mode, the T-SBI may just write bitfields in the M-mode registers.  With custom M-mode, the T-SBI must make it appear to S-mode software that the field holds the value written by the T-SBI, even though the implementation might be different. These CSRs and memory-mapped I/O registers are:
```
kdockser, line 93: They don't emulate. In the rare case that someone doesn't implement the 1.13 CSRs, they need to be able to provide the same information through the use of the CSRs and M-mode code. Emulate implies more than that.

docs/crd/src/rva23_crd.adoc
```diff
@@ -105,12 +100,11 @@ RVA23 certification requires that the SBI provide methods of emulating every wri
...
+| `mseccfg` | PMLEN must be able to hold at least PMLEN=0 and 7.  SSEED and USEED must be controllable if Zkr is supported; otherwise they are read-only zero.  MLPE, RLB, MMWP, MML are OUT-OF-SCOPE because they do not relate to RVA23 features.
+| `mstateen` | SE0, ENVCFG, CONTEXT must be controllable to allow testing of the behavior of lower privilege modes when certain features are disabled. Other fields are OUT-OF-SCOPE because they do not relate to RVA23 features.
+| `medeleg` | Each exception cause except `ecall` from S-mode is delegated to S-mode if possible.  Otherwise, a user-supplied M-mode trap handler intercepts invisible traps and logs the rest or relays them to S-mode.
+| `mideleg` | SEI, STI, SSI, and LCOFI are delegated to S-mode if the corresponding bit can be written to 1.  Otherwise, a user-supplied M-mode trap handler relays these to S-mode. VSEI, VSTI, VSSI, and SGEI are read-only 1 because hypervisor is a mandatory part of the profile.
```
kdockser, line 107: For the purposes of RVA23 testing, there are no M-mode trap handers other than for ECALL. All other M-mode traps are fatal or transparent (change no additional state).

docs/crd/src/rva23_crd.adoc
```diff
@@ -143,24 +137,79 @@ The following M-mode CSRs/fields are OUT-OF-SCOPE and not tested:
...
 | `mtval2` | Does not affect the behavior of lower privilege modes.
+| `mtimecmp` | MTI is not tested.
+| `msip` | MSI is not tested.
 | PMP CSRs | PMP restrictions on access are indistinguishable below M-mode from a PMA that imposes similar restrictions.
 |===
 
 === Interrupts
 
-All interrupts taken from S and U-mode are IN-SCOPE for testing.  MTI, MEI, and MSI are only tested without delegation to S-mode, while the other interrupts are tested with and without delegation, if `mideleg` supports this.
+All supervisor interrupts (SEI, STI, SSI, and LCOFIP) are IN-SCOPE for testing. Hypervisor interrupts VSEI, VSTI, VSSI, and SGEI are also IN-SCOPE. Machine interrupts are OUT-OF-SCOPE.
```
kdockser, line 147: Out of scope of RVA23.
Sometime we use the term OUT-OF-SCOPE for what we don't do in certification tests but is in scope for RVA23.

docs/crd/src/rva23_crd.adoc
```diff
@@ -143,24 +137,79 @@ The following M-mode CSRs/fields are OUT-OF-SCOPE and not tested:
...
+Machine interrupts (MEI, MTI, MSI) are outside the RVA23 profile and are OUT-OF-SCOPE for certification.
 
 === User Requirements
 
 The user must provide:
 
 * DUT configuration information described in the CTP, including UDB description, linker file describing memory map, sail.json configured to match DUT, RVMODEL macros specific to DUT.
 * Custom firmware, if any, to handle invisible traps.
-* Custom boot code, if standard M-mode boot code does not run on the DUT or if additional configuration is necessary.
+* Custom boot code and/or trap handler, if standard M-mode boot code does not run on the DUT or if additional configuration is necessary.
```
kdockser, line 159: Boot code tends to be implementation specific. What is meant by standard M-mode boot code?

docs/ctp/src/strategy.adoc
```diff
@@ -62,7 +62,7 @@ Some RISC-V behaviors are UNSPECIFIED, which means an implementation can do anyt
 Unimplemented opcodes (except in the custom space) and instructions accessing non-existent CSRs are reserved.  The behavior of reserved instructions is UNSPECIFIED, so they should not be tested by most testplans.  However, when the Ssstrict extension is enabled, reserved instructions raise an illegal instruction exception.  This is tested by the <<SsstrictS, SsstrictS>> test plan.
 
 [NOTE]
-Ssstrict states that the illegal instruction is contained in the supervisor-mode trap handler.  Certification tests do not exercise the execution environment, so the illegal instruction traps to machine mode.
+If the instruction is reserved only in certain architectural states, it is not required to raise an illegal instruction and Ssstrict does not test it.  For example, vector register groups that are not a multiple of LMUL are reserved, but LMUL is architectural state not encoded in the instruction, so instructions with groups not aligned to LMUL are reserved but untested even by Ssstrict.
```
kdockser, line 65: CSRs _are_ considered as part of the instruction encoding. This is critical to the proper interpretation of RISC-V.

The issue [[Clarification] Definition of 'reserved opcodes' in Ssstrict #3258](https://github.com/riscv/riscv-isa-manual/issues/3258) was not related to CSRs as the EEWs in the specified instruction differed independent of vtype.

docs/ctp/src/strategy.adoc
```diff
@@ -83,13 +83,9 @@ TODO: Generate this table from norm rules via YAML when Ssstrict is tagged.
...
 
-*WPRI* fields are not tested. *WLRL* fields are tested with all legal values but no illegal values.   *WARL* fields are tested by trying all possible values of the field.  CSR tests (<<Zicsr>>) for *WARL* fields generally involve setting all bits to 1, setting all to 0, and writing walking 1s.   Therefore, they inherently exercise all possible values of 1 and 2-bit *WARL* fields.  Longer fields can be tested exhaustively if the number of possibilities is small (such as `satp.MODE`), or just with walking 1s if the number of possibilities is too large (such as `misa.EXTENSIONS`).
+*WPRI* fields are not tested. *WLRL* fields are tested with all legal values but no illegal values. Similarly, *WARL* fields are tested with all legal values but no illegal values. Reserved values are never tested in CSRs because the behavior is unpredictable.
```
kdockser, line 86: What don't we test WPRI? Their behavior is well defined.

docs/crd/src/rva23_crd.adoc
```diff
@@ -22,13 +22,16 @@ History of documentation changes that eventually lead to releases.
 | April 29, 2026
 | 0.1
 a| * Initial draft
+| July 27, 2026
+| 0.2
+a| * Do not test traps into M-mode except ecall from S-mode
```
kdockser, line 27: Do not execute traps to M-mode except for the ECALL.

kdockser (commented):
Here are comments on some of the areas in this document that needs cleanup. It looks like there is still a goodly amount of work to be done

docs/ctp/src/strategy.adoc
jordancarlin: https://github.com/riscv/riscv-isa-manual/issues/3105 seems to contradict this.

kdockser: Which _this_ is contradicted

#3104 does _not_ contradict "The issue https://github.com/riscv/riscv-isa-manual/issues/3258 was not related to CSRs as the EEWs in the specified instruction differed independent of vtype."

Furthermore, Andrew's comment specifically referred to _reserved_ CSR values. Was he asked about a specific example with a reserved vl/vstart? In any case, there are plenty of cases in the ISA that explicitly state that an encoding is reserved based on, for example, vstart.

docs/crd/src/rva23_crd.adoc
davidharrishmc: In initial drafts I had used the term nonconforming M-mode to describe an M-mode that was incapable of running standard M-mode code consistent with Chapter 3 of the Priv ISA spec.  You had asked to change it to custom M-mode.  What do you want to call it next?

docs/crd/src/rva23_crd.adoc
davidharrishmc: Line 57 defines standard M mode as consistent with the ratified M-mode specification.  Standard M-mode boot code is the reference boot code supplied in rvtest_setup.h that executes properly on standard M-mode.  It only has the parts to initialize standard M-mode and test state, not the implementation-specific parts such as initializing memory controllers or PLLs ("additional configuration necessary").  Would you like some language to clarify this?

docs/ctp/src/strategy.adoc
davidharrishmc: Ssstrict states: "No non-conforming extensions are present. Attempts to execute unimplemented opcodes
or access unimplemented CSRs in the standard or reserved encoding spaces raises an illegal
instruction exception that results in a contained trap to the supervisor-mode trap handler."

The instruction encoding is a 32-bit number.  For csrrw/s/c type instructions, the 12-bit CSR specifier is part of the instruction encoding.

Andrew's explanation, which is consistent with the plain and ordinary meaning of instruction encoding, is that the contents of a CSR are not part of the instruction encoding.  For example, vgmul.vv has a certain encoding (e.g. A208A77).  The instruction is reserved if SEW is any value other than 32.  SEW is in bits 5:3 of vtype.  A decoder cannot look at the 0xA208A77 by itself and know whether that instruction is reserved or not.

Hence, in a RVA23 processor supporting the Zvkng and Ssstrict options, A208A77 is not covered by Ssstrict because it is not part of a nonconforming extension and is not an unimplemented opcode.  The instruction happens to be reserved when SEW=8, and its behavior in that case is unspecified.  It is legal to throw an illegal instruction trap, but is not required to trap.  Therefore, the Ssstrict tests don't check it across different SEW.

We had written those tests earlier, and took them out per Andrew's guidance.

I suggest we leave the language as it is for now, and revise it if TSC responds in writing to interpret Ssstrict differently than in https://github.com/riscv/riscv-isa-manual/issues/3258.

docs/ctp/src/strategy.adoc
davidharrishmc: According to the spec: "Some whole read/write fields are reserved for future use. Software should ignore the values read from these
fields, and should preserve the values held in these fields when writing values to other fields of the same
register. For forward compatibility, implementations that do not furnish these fields must make them
read-only zero.  These fields are labeled WPRI in the register descriptions."

henvcfg[60] is WPRI.  How do we test this bit?  

The spec says software shouldn't change the value, and should ignore the value read, so there are no wrong answers.  If an implementation does not furnish this field, it should be read-only zero.  There are no ratified extensions that presently furnish henvcfg[60] so we could check that it is 0.  But if an extension is ratified in the future that defines henvcfg[60] and a RVA23-compliant core chose to implement that extension, henvcfg would not necessarily be 0 and the test that checks for 0 would fail.

What do you want a WPRI test for henvcfg[60] to do?

docs/crd/src/rva23_crd.adoc
davidharrishmc: I think all we can say is "do not test."  An implementation might have some bit of mideleg tied to 0 so it necessarily executes the trap to M-mode when we take that trap, but we will only check that the T-SBI forwards the trap to S-mode and S-mode CSRs get the expected values.

docs/crd/src/rva23_crd.adoc
davidharrishmc: Modifying the reference boot code and trap handler is costly and we would prefer not to force users to do this. There are already hooks to run implementation-specific boot code and interrupt-clearing code.  If a user implements standard M-mode (i.e. compatible with chapter 3 of the priv spec), why would they need to modify the boot code and trap handler, and are there any other hooks we should provide to avoid this need?

docs/ctp/src/strategy.adoc
kdockser: Hi David,

The example given with _Zvkng_ is reserved no matter what value is in SEW. Hard stop.

Andrew's answer contradicts Greg's earlier answer ([Ssstrict: Do "unimplemented opcodes" included reserved encodings #197](https://github.com/riscv/riscv-profiles/issues/197)) to which Andrew and the rest if of the ARC previously agreed. Many implementations have been built based on the interpretation (which had been the general understanding prior to my seeking that the text be clarified).

docs/crd/src/rva23_crd.adoc
davidharrishmc: Consider a machine that patches an FDIV bug by taking an invisible trap to M-mode.  It must hardwire medeleg[2] = 0 so that the firmware receives FDIV instructions, inspects them, and fixes the problem invisibly.  Other illegal instructions should still go to the illegal instruction handler, which we expect is in S-mode.  Therefore, the T-SBI receives the undelegated illegal instruction trap and sends it on to S-mode as if it had been delegated.  The tests had expected it was delegated, and attempted to write medeleg[2] = 1.  They don't know FDIV is fixed by an invisible trap.  

This is the architecture that Allen has discussed for the past several months. It is also what the OS expects. If you don't agree with architecture, please describe what you expect should happen.

docs/crd/src/rva23_crd.adoc
davidharrishmc: What word would you like?

Standard M-mode defines mstatus.TSR as mandatory.  The tests check the behavior of SRET with TSR = 0 and 1.  If a custom M-mode does not implement mstatus.TSR, this paragraph requires a custom T-SBI that intercepts SRET and causes it to throw an illegal instruction trap (normally delegated to S-mode) when the T-SBI thinks it had set mstatus.TSR = 1.

If we don't do this, how does the test know what to expect SRET to do?

docs/crd/src/rva23_crd.adoc
davidharrishmc: See above about what Allen has been telling us for the past few months.

Alibaba does not delegate ebreak to S-mode.  https://github.com/riscv/riscv-arch-test/issues/1892  That is legal within the spec.  

When U-mode executes ebreak, S-mode expects to see a breakpoint trap.  If the M-mode trap is fatal, the test crashes.  If the M-mode trap is silent, S-mode thinks breakpoint did not trap.

In general, "All M-mode traps are fatal or transparent" would mean that every instruction that might trap would need to be tested to either trap properly or crash or do nothing.  Seems like a major hole in the testing, as compared to Allen's approach of relaying the trap to S-mode.

kdockser: If an instruction traps to M-mode, the test fails. That is a good way to tell if an instruction incorrectly trapped.

Having an M-mode handler in RVA23 to redirect traps to S-mode is out of scope; it is testing M-mode.

docs/ctp/src/strategy.adoc
davidharrishmc: I fear we are talking about two different things and not understanding each other.

When Zvkng and Ssstrict are implemented and SEW = 32, the behavior of vgmul.vv is precisely defined to put a certain value in vd.  It is not reserved and is not unspecified and will not throw an illegal instruction trap.

If SEW=8, Andrew says the behavior is reserved but not covered by Ssstrict, so it is optional whether it traps.
It's fine if an implementation is built on an interpretation that vgmul.vv traps in this case, but the certification test can't demand that it trap.

I don't know what you mean by "reserved no matter what."

Can you be more specific about what you think the tests should cover and expect with respect to vgmul.vv when SEW != 32 and Ssstrict is implemented?

docs/crd/src/rva23_crd.adoc
davidharrishmc: So I think you are saying the ebreak test on Alibaba hardware will fail.  But Alibaba has a legal implementation of medeleg under RVA23.  Why do they need to fail?

Why is redirecting undelegatable traps to S-mode testing M-mode any more than a T-SBI setting menvcfg.STCE testing M-mode?

docs/crd/src/rva23_crd.adoc
allenjbaum: Exactly; it may always trap to Mmode. All we can do is expect that traps to Mmode are invisible unless explicitly undelegated, and we won't test anything undelegated that isn't part of TSBI

docs/crd/src/rva23_crd.adoc
allenjbaum: I don't agree that emulate means more than that. We even talk about emulating a divide instruction in Mmode if the M-extension is not implemented; emulating a standard (==m1p13) CSR is just another example of emulation

allenjbaum:
What Davi said. To pile up on that:
We only test that traps that S-mode *expects* to handle are sent to Smode
with all the CSRs populated *as if *they had been delegated in HW to Smode.
I.e. anything that instead traps to M-mode is *invisibly* emulated.  So, we
are not testing traps into M-mode, we are testing traps that aren't going
to Mode (because that is what Smode expects).
Ecalls from Smode, on the other hand, are *not expected* to be handled by
Smode, so those are tested by observing that whatever the Ecall was
requesting have taken effect.

kdockser:
Allen, what do you mean by "anything that instead traps to M-mode is *invisibly* emulated." If you are referring to them being redirected to S-mode by an M-Mode trap handler, they are no longer invisible. This _is_ testing M-mode.

allenjbaum:
As I've said earlier, Mmode must do what Smode expects it to do.
A hart that has an incomplete/missing/buggy implementation of (F)DIV must
have Mmode emulate it because Smode *expects* it to execute correctly.
We won't provide code to do that, so that will be custom firmware, but not
terribly complicated.
That must be an invisible trap, else the test should fail (though we won't
test that because CSC tests would have to check that very register is
unchanged for all tested instructions)
That custom/modified firmware may have many instances where it knows it
must perform special handling, and our reference firmware will have many
instances as well (e.g. ECALL).

You are correct that if it traps to Mmode, and neither the custom nor
standard special handling applies, then the tests fails,
either because the HW is buggy, or the custom firmware is incomplete/buggy.

allenjbaum:
Huh, it thinks Allen Chang sent this. I'm the other Allen that salso
pells "Allen" correctly.

And, no, if Smode has no way of noticing that there was trap  at all,
and sent to Mmode in order to update state as if it had been handled
in HW, then it is invisible to Smode.
The only exceptions are ECALLs that Smode expects to be handled by the
SEE (which I think includes RVMODEL macros)

kdockser:
`And, no, if Smode has no way of noticing that there was trap  at all,`
I think that this is actually a _yes_ not a _no_ as we agree on this.

allenjbaum:
I was responding to
    "If an instruction traps to M-mode, the test fails. That is a good way
to tell if an instruction incorrectly trapped."
and disagreeing,
Trapping to Mmode is allowed, and as long as Smode expects to handle that
trap cause, and Mmode delegates in in SW, the test succeeds. That's an
invisible delegation
Mmode will have a list of traps that should be delegated, which is the
values set in medeleg (or an emulated version) which is set via a TSBI call.

If a trap occurs that Smode was not expecting (e.g. [F]DIV fixup) and
Mmode  expects to handle it, it will emulate and return to Smode (as
opposed to delegate to Smode) That's an invisible emulation.
If a trap occurs that Smode was not expecting (e.g. [F]DIV fixup) and
Mmode  doesn't have a handler for it, only then it is a test failure.

docs/ctp/src/strategy.adoc
davidharrishmc: Note deleted per Ssstrict discussion linked above.

docs/ctp/src/strategy.adoc
davidharrishmc: clarified the reason

docs/ctp/src/strategy.adoc
kdockser: My bad, I am trying to do too much at once and while I was still talking about the original example (_vnsrl.wv v16, v0, v0_) in [[Clarification] Definition of 'reserved opcodes' in Ssstrict #3258](https://github.com/riscv/riscv-isa-manual/issues/3258) I missed that you had moved on to a different different example for a different (yet related) issue.

The original _intent_ was for Ssstrict to cover this case, but then it was reworded and no one seems to know what it really means now. It is my contention that Zvkng and Ssstrict and SEW=8 would raise an illegal instruction exception. Actually, it was my _intention_ that that this be the case. I originally defined the behavior as SEW <>32 would take an illegal instruction exception. I later changed it to _reserved_ after a discussion with Andrew. My intention was that we define an extension or make it a requirement that reserved encodings would take illegal instruction exceptions. This was to get an expected and reasonable behavior in these cases for a variety of reasons, including compatibility testing.

At this point, we still need a clarification from the ARC as they have provided spot rulings rather than a complete and unambiguous rule.

davidharrishmc:
Thank you.  Will you run that one down with ARC?

This is not a part of the CRD so it doesn’t need to block the CRD decision, but I’d like to wrap it up soon anyway.

kdockser:
I tired to prompt them in the Clarification issue, but I will also try a more direct approach.

allenjbaum:
I must quibble
with  "*WARL* fields are tested by trying *all* possible values of the
field." and
with "*WLRL* fields are tested with *all* legal values but no illegal
values."
We can't do that, and the wording that follows these illustrates that.
"all" needs to be replaced with some other wording.

davidharrishmc:
@allenjbaum cleaned up WARL and WLRL "all" language

PR #1909, opened 2026-07-18
Reserve a2 in priv tests

### LINKED ISSUE #1858: Need to reserve a2 for TSBI
**Describe the bug**
generators/testgen/src/testgen/generate/priv.py should reserve register 12 (a2) as an argument for SBI store requests.  

However, adding it and invoking "make spike" gives:

```
ValueError: Not enough registers available to select from. Requested 1, but only 0 available.
```

jordancarlin:
Update tests so that a2 can be reserved for the T-SBI:
- Change available link/temp regs to avoid a2
- Update register allocation in ExceptionsZaamo, InterruptsSm, InterruptsU, and Ssstateen to use fewer registers to reduce register pressure.
- Regenerate tests

Fixes #1858


davidharrishmc (approved):
LGTM

PR #1919, opened 2026-07-19
Work-in-progress notes on UDB parameters needed by the CTP

davidharrishmc:
Splits the `docs/ctp/src/params.adoc` changes out of #1901 into their own PR, so the RVA23 CRD discussion there can proceed separately.

These are work-in-progress notes enumerating the UDB parameters the CTP will need: parameters affecting coverpoints, boot code / T-SBI, and expected results in the reference model. More will be added as the tests mature and the parameters working group identifies them.


jordancarlin:
@davidharrishmc what is the plan with this PR?

davidharrishmc:
Just to frame some thoughts when parameters arrive.  Not high importance.

Issue #1866, opened 2026-07-14
trap handler rtn_fm_mmode trashes T2 (x7)


davidharrishmc:
in rvtest_trap_handler.h, T2 is not restored.  Thus T2 must be reserved.

```
rtn_fm_mmode:
        add     T2, T4, T2                             // T2 = M-mode code_begin + relative offset = return addr

        LREG    T1, trap_sv_off+1*REGWIDTH(sp)        // restore T1
        LREG    T3, trap_sv_off+3*REGWIDTH(sp)        // restore T3
        LREG    T4, trap_sv_off+4*REGWIDTH(sp)        // restore T4
        LREG    T5, trap_sv_off+5*REGWIDTH(sp)        // restore T5/a0
        LREG    T6, trap_sv_off+6*REGWIDTH(sp)        // restore T6/a1
        LREG    sp, trap_sv_off+7*REGWIDTH(sp)        // restore original sp
        jr      4(T2)                                  // jump to ecall+4 in M-mode address space

.endif  // end of M-mode rtn2mmode
```

davidharrishmc:
This was silently breaking tests, such as ZicntrS getting partial coverage.  Fixed by excluding x7 in priv.py priv_exclude_regs.

davidharrishmc:
Fixed with #1861

PR #1920, opened 2026-07-19
Enable Smdbltrp and Ssdbltrp in Imperas and Whisper and QEMU configs and fixed double trap error

davidharrishmc:
Enables the Smdbltrp and Ssdbltrp (double trap) extensions in the Imperas rv32gck/rv32gcv/rv64-max configs and the Whisper rv32-max/rv64-max ISA strings, and QEMU.

Fixes trap handler to clear MDT before returning without meet.

Better error reporting instead of hanging.
Disable SmV because of Whisper bug.


run_tests.py
```diff
@@ -25,6 +25,14 @@
 _DEBUG_PLACEHOLDER_RE = re.compile(r"\{debug:([^}]*)\}")
 _TRACEFILE_PLACEHOLDER = "__TRACEFILE__"
 _SUMMARYFILE_PLACEHOLDER = "__SUMMARYFILE__"
+_TRACE_PC_RE = re.compile(rb"core\s+\d+: (?:\d+ )?0x([0-9a-f]+)")
+# Simulator-reported failure reasons worth surfacing verbatim in FAIL messages, e.g.
+# whisper's "Error: Failed stop: Hart 0: Core entered critical-error state ..." or
+# qemu's "qemu: fatal: M-mode double trap"
+_ERROR_LINE_RE = re.compile(r"Error:|Failed stop:|critical.error|fatal", re.IGNORECASE)
```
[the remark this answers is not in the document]
davidharrishmc: The regex `.` wildcard did match the underscore variant (`critical_error`), and the "Monitoring net" exclusion is a plain substring check on the line, so no diagnostics were being missed — verified against captured whisper (`critical-error`) and Imperas (`net_cpu0_critical_error`) logs. Agreed the explicit form is clearer though: changed to `critical[-_ ]error` in 642b7bda.

run_tests.py
```diff
@@ -287,6 +336,10 @@ def main() -> int:
             rel_log = str(elf_path.relative_to(elf_dir).with_suffix(".log"))
             entries.append((rel_log, rvcp_summary))
             print(f"{rel_log}  {rvcp_summary}", file=f, flush=True)
+        # Let workers exit normally so their buffered stdout (per-test FAIL details) is
+        # flushed before the context manager's terminate() would kill them.
+        pool.close()
+        pool.join()
```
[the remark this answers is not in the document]
davidharrishmc: Claude says Copilot is full of it.

config/whisper/ci.yaml
```diff
@@ -14,8 +14,13 @@
 # SSTATUS shadow view whose write mask excludes MPRV, so the clear is silently dropped
 # The MRET path writes MSTATUS directly and is unaffected. The S suite's
 # cp_sret_m mprv=1 bins catch this: sail clears MPRV, whisper keeps it -> selfcheck mismatch.
+# SmV excluded due to Whisper bug (to be reported upstream): with Smdbltrp enabled, any write
+# to misa (even one that changes nothing) spuriously sets mstatus.MDT back to 1, because
+# Hart::processExtensions runs on misa writes and CsRegs::enableSmdbltrp re-applies the MDT
+# reset value on every call. The SmV cp_misa_v_clear_set coverpoint writes misa, so the next
+# intentional illegal-instruction trap hits with MDT=1 and whisper enters critical-error state.
```
jordancarlin: Trim this down. Link to Whisper issue and remove discussion of Whisper internals.

tests/env/rvtest_trap_handler.h
```diff
@@ -2316,13 +2316,24 @@ excpt_\__MODE__\()hndlr_tbl:
...
+  #endif // SMDBLTRP_SUPPORTED
+
         LREG    T1, trap_sv_off+1*REGWIDTH(sp)        // restore T1
         LREG    T3, trap_sv_off+3*REGWIDTH(sp)        // restore T3
         LREG    T4, trap_sv_off+4*REGWIDTH(sp)        // restore T4
         LREG    T5, trap_sv_off+5*REGWIDTH(sp)        // restore T5/a0
         LREG    T6, trap_sv_off+6*REGWIDTH(sp)        // restore T6/a1
         LREG    sp, trap_sv_off+7*REGWIDTH(sp)        // restore original sp
-        jr      4(T2)                                  // jump to ecall+4 in M-mode address space
+        jr      4(T2)                                 // jump to ecall+4 in M-mode address space
```
jordancarlin, line 2336: Seems like a better fix would be to change this to return using an `mret`. That would automatically clear `mstatus.MDT` and would avoid #1866.

run_tests.py
```diff
@@ -70,6 +78,31 @@ def dim(text: str) -> str:
...
+        with trace_file.open("rb") as f:
+            f.seek(0, os.SEEK_END)
+            f.seek(max(0, f.tell() - 4096))
+            tail = f.read()
+    except OSError:
+        return False
+    # The final line may be truncated mid-write when the process is killed, so
+    # accept a match in any of the last few PCs rather than only the very last.
+    pcs = _TRACE_PC_RE.findall(tail)
+    return any(int(pc, 16) in _DEBUG_ROM_RANGE for pc in pcs[-5:])
```
jordancarlin: I'd rather avoid adding simulator specific logic to this script (especially because the debug ROM address can be moved with CLI flags to Spike).

run_tests.py
```diff
@@ -25,6 +25,14 @@
 _DEBUG_PLACEHOLDER_RE = re.compile(r"\{debug:([^}]*)\}")
 _TRACEFILE_PLACEHOLDER = "__TRACEFILE__"
 _SUMMARYFILE_PLACEHOLDER = "__SUMMARYFILE__"
+_TRACE_PC_RE = re.compile(rb"core\s+\d+: (?:\d+ )?0x([0-9a-f]+)")
+# Simulator-reported failure reasons worth surfacing verbatim in FAIL messages, e.g.
+# whisper's "Error: Failed stop: Hart 0: Core entered critical-error state ..." or
+# qemu's "qemu: fatal: M-mode double trap"
+_ERROR_LINE_RE = re.compile(r"Error:|Failed stop:|critical[-_ ]error|fatal", re.IGNORECASE)
```
jordancarlin, line 31: Can this be more generic? Something like `re.compile(r"error|fail|fatal")`.

jordancarlin (changes requested):

tests/env/rvtest_trap_handler.h
davidharrishmc: Moving to T-SBI will move to mret.  This is legacy code and I'd rather not spend more time on it.  Not certain mret does the right thing for this one anyway.

run_tests.py
davidharrishmc: I hear you, but don't have a better solution.  It was annoying for Spike to just hang on double traps.  I think terminating with an error message is much better.  Any thoughts of a way to do this without touching this file?

run_tests.py
davidharrishmc: Too generic.  Want to catch double trap messages without catching other RVCP errors.

run_tests.py
jordancarlin: This is not preventing the hang. This just adds additional error output to the console which should already be visible in the log. The `--instructions` flag that you added to `run_cmd.txt` is what is stopping the hang, and I agree that we should keep that.

run_tests.py
```diff
@@ -25,6 +25,14 @@
...
+# Spike's debug module ROM occupies 0x800-0xfff; a hart that enters the critical-error
+# state (e.g. a double trap with Smdbltrp) is parked in its wfi loop forever.
+_DEBUG_ROM_RANGE = range(0x800, 0x1000)
```
jordancarlin: `_TRACE_PC_RE` and `_DEBUG_ROM_RANGE` are no longer used.

jordancarlin (changes requested):

jordancarlin (approved):

PR #1929, opened 2026-07-20
Adjusted Interrupts tests based on testplan

Teemooooooooo:
Adjusted Interrupts tests based on testplan
- InterruptsSSm excluded from whisper CI since whisper supports M mode interrupt delegation and SAIL currently only support S mode interrupt delegation
- adjusted cp_interrupt_m to attempt to delegate all interrupts instead of just S mode interrupts. 
- added a test to attempt writing to mip.STIP with SSTC supported (normative rule mapping of this is done in [PR1906](https://github.com/riscv/riscv-arch-test/pull/1906))
- test U mode wfi with TW = {0/1} instead of just testing TW = 1
- slight modification on  _generate_wfi_timeout_s_tests removing unnecessary code


coverpoints/priv/InterruptsSSm_coverage.svh
```diff
@@ -89,6 +89,11 @@ covergroup InterruptsSSm_cg with function sample(ins_t ins);
     mip_stip_one: coverpoint ins.current.csr[CSR_MIP][5] {
         bins one = {1};
     }
+    `ifdef SSTC_SUPPORTED
+        prev_mip_stip_zero: coverpoint ins.prev.csr[CSR_MIP][5] {
```
davidharrishmc, line 93: Why does this need to be gated by SSTC?  Can't we always test writing to mip.STIP?

coverpoints/priv/InterruptsSSm_coverage.svh
```diff
@@ -227,6 +232,9 @@ covergroup InterruptsSSm_cg with function sample(ins_t ins);
     cp_sei5:                    cross priv_mode_m, mideleg_zeros, mstatus_mie_zero, prev_mip_seip_one, mip_seip_one, csrrc, write_mip_seip;
     cp_sei6_7:                  cross priv_mode_m, mideleg_zeros, mstatus_mie_zero, prev_mip_seip_one, mip_seip;
     cp_global_ie:               cross priv_mode_m, mstatus_mie, mstatus_sie, mip_walking_m, mip_mie_eq;
+    `ifdef SSTC_SUPPORTED
+        cp_stip_write_stimecmp: cross priv_mode_m, prev_mip_stip_zero, csrrw;
```
davidharrishmc: Can this also have a constraint on rs1 to check we are actually writing the to STIP?

generators/testgen/src/testgen/priv/extensions/InterruptsS.py
```diff
@@ -1774,13 +1774,13 @@ def _generate_wfi_s_tests(test_data: TestData) -> list[str]:
 
 
 def _generate_wfi_timeout_s_tests(test_data: TestData) -> list[str]:
-    """Generate S-mode and U-mode WFI timeout tests.
+    """Generate S-mode WFI timeout tests.
 
     Test WFI timeout with TW=1, MTIMECMP=max.
     Cross: MIE={0,1} × SIE={0,1} × mideleg={0,0x222} × MTIE={0,1} × mode={S,U}
```
davidharrishmc: Lists U mode.

generators/testgen/src/testgen/priv/extensions/InterruptsS.py
```diff
@@ -2531,12 +2528,12 @@ def _generate_wfi_timeout_u_tests(test_data: TestData) -> list[str]:
     Test WFI timeout from U-mode with TW=1.
```
davidharrishmc: now says 0/1 below

generators/testgen/src/testgen/priv/extensions/InterruptsSSm.py
```diff
@@ -1556,6 +1555,44 @@ def _generate_global_ie_tests(test_data: TestData) -> list[str]:
...
+            "# set menvcfg.STCE",
+            *set_menvcfg_stce(r_temp, True),
+            "# clear STIP through STIMECMP",
+            *set_stimecmp_max(r_temp),
+            "",
+            test_data.add_testcase("STIP", coverpoint, covergroup),
+            "# attempt to write mip.STIP",
+            f"LI(x{r_temp}, 0x20)",
+            f"CSRS(mip, x{r_temp})",
+            write_sigupd(r_temp, test_data),
```
davidharrishmc, line 1593: mie = 0 to prevent interrupt
Read back value.
Also test case of writing 0 when SSTC makes it 1.

davidharrishmc (changes requested):

coverpoints/priv/InterruptsSSm_coverage.svh
Teemooooooooo: this is gated by SSTC because some of the conditions we check (prev_mip_stip_zero) is set up by writing to STIMECMP in the test

davidharrishmc (approved):

PR #1935, opened 2026-07-21
Select Zicsr test CSR via RVTEST_TEST_CSR with privileged-mode fallback

### LINKED ISSUE #1572: Zicsr tests fail to compile for U-mode cores without F, V, or Zicntr

karabambus:
Fixes #1572. Implements the chain @jordancarlin proposed there, reusing the reviewed pieces of #1573 (utils.h placement, ZVE32X condition, RVTEST_READ_ONLY_TEST_CSR),  rebased on current act4.

  ## Problem
The 12 generated Zicsr tests pick their test CSR from an inline preprocessor chain repeated at every CSR access site. The chain has no branch for cores with U-mode but no F, V, or Zicntr (cv32e40s), so every test fails to compile with `#error no CSR known for testing`.

  ## Change
**tests/env/utils.h**: the CSR selection now lives in one place, gated on ZICSR_SUPPORTED. In order of preference: fflags (F), vxsat (Zve32x), mepc (cores without U-mode), instret (Zicntr, read-only), sepc (S, test boots to S-mode), mepc (STANDARD_SM_SUPPORTED, test boots to M-mode), otherwise `#error`.

The boot-mode pinning and the `#error` only apply to files that define RVTEST_ZICSR, which testgen emits in the Zicsr tests. then the boot refactor starts running unpriv tests in the lowest privilege mode, the Zicsr tests keep the mode their CSR needs and no other suite is affected. No RVMODEL macro additions, no new YAML header keys.

**csr_type.py / csri_type.py**: the formatters previously printed the whole selection chain at every access site, about 550 times per file:
```
#if defined(F_SUPPORTED)
csrrw x5, fflags, x12
#elif defined(V_SUPPORTED)
csrrw x5, vxsat, x12
#elif !defined(U_SUPPORTED)
csrrw x5, mepc, x12
li x5, 0 ...
#else
#endif
```

Now they print:
```

#ifdef RVTEST_READ_ONLY_TEST_CSR
li x5, 0 ...
#else
csrrw x5, RVTEST_TEST_CSR, x12
#endif
```

utils.h defines RVTEST_READ_ONLY_TEST_CSR only when it selects instret, so the existing read-only handling (skip writes, check the counter incremented) is preserved unchanged. The rd/rs1 zero-register special cases are untouched. After preprocessing the emitted instructions are identical to the old chain for every previously working config. The regenerated tests shrink from about 29.5k to 10.8k lines.

**tests/env/riscv_arch_test.h**: utils.h is now included after rvmodel_macros.h so the selection can see STANDARD_SM_SUPPORTED. Side effect: the RVMODEL_FENCEI hook, previously included before any DUT could define it, now works.

## Validation
- Validated on a cv32e40s U-mode config: full 940-test build plus Verilator RTL run,  all 6 Zicsr tests pass via the STANDARD_SM fallback.

## Boot refactor notes (
1. U-mode access to the selected CSR needs enablement in the U-boot path: mcounteren.IR for instret (cvw-rv32imc is the only instret config) and mstatus.FS  for fflags.
2. The disabled BOOT-define emission in templates.py keys BOOT_TO_MMODE on Sm in REQUIRED_EXTENSIONS, which also matches the U-mode priv suites (ExceptionsU etc.); those must not be pinned to M-mode when it is re-enabled.

  ## Open questions
 1. RVMODEL_TEST_CSR: dropped per #1573 review, re-added in the #1572 snippet. Currently out; two lines to re-add if wanted.
  2. Zicntr ordering: kept before the privileged fallbacks per the original rationale (Zicntr CSRs can exist without M-mode). Cores with both Zicntr and an escalation path (today only cvw-rv32imc) get the read-only half-test where a writable CSR is available. No behavior change in this PR either way; happy to reorder if preferred.

### DIFF
--- a/generators/testgen/src/testgen/formatters/types/csr_type.py
+++ b/generators/testgen/src/testgen/formatters/types/csr_type.py
@@ -15,31 +15,25 @@ csr_config = InstructionTypeConfig(required_params={"rd", "rs1", "rs1val", "rs2"
 
 def zicsr_acccess(instr_name: str, rd: int, rs1: int) -> str:
     """Helper function to determine which CSR to use for testing based on supported extensions."""
-    # Use writable unprivileged extension CSRs if any exist,
-    # else use mepc if U is not supported
-    # else use instret (which is not writable, but at least can be accessed)
+    # The CSR is selected once via RVTEST_TEST_CSR (see utils.h), which also pins the test
+    # to a privilege mode where the CSR is writable. A read-only choice (instret) sets
+    # RVTEST_READ_ONLY_TEST_CSR and needs special handling since it cannot be written.
 
     # instret requires special treatment because it is not writable, and the value is not initialized
     if instr_name in ["csrrw", "csrrwi"] or rs1 == 0 or rd == 0:
-        instret_access = f"li x{rd}, 0 # avoid write to read-only instret, or inconsistent result with rs1 or rd = 0"
+        read_only_access = f"li x{rd}, 0 # avoid write to read-only CSR, or inconsistent result with rs1 or rd = 0"
     else:
-        instret_access = (
-            f"{instr_name} x{rs1}, instret, x0\n"
-            f"{instr_name} x{rd}, instret, x0\n"
-            f"sub x{rd}, x{rd}, x{rs1}  # check that instret value has incremented\n"
+        read_only_access = (
+            f"{instr_name} x{rs1}, RVTEST_TEST_CSR, x0\n"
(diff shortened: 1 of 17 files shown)


davidharrishmc:
@karabambus there is a merge conflict.  @jordancarlin aside from that, is this ready?

jordancarlin:
I've updated this and submitted a new version in #2344 (I kept you as a co-author on it). Thanks for getting this moving.

PR #1936, opened 2026-07-22
Create a Simpler Interface for Vector Loads & Provide ifdefs for Fractional LMUL

fourth-bit:
This PR adds `load_vec_regs` in place of `load_vec_reg`. This has a simpler interface, and allows the formatter code to be simpler in the loading register parts. A good example of the simplification it allows is in `formatters/types/vvv_type.py`.


config/whisper/ci.yaml
```diff
@@ -15,7 +15,7 @@
 # InterruptsSSm excluded because whisper supports M mode interrupt delegation and Sail currently only supports S mode interrupt delegation.
 ci_enabled: true
 # TODO: Remove SsstrictV when sail SsstrictV is implemented
```
davidharrishmc, line 18: Add comment with link to issue about when this Vx* can be reenabled.

davidharrishmc (changes requested):

davidharrishmc (approved):

jordancarlin (approved):

PR #1937, opened 2026-07-22
Many PMP test fixes

### LINKED ISSUE #1721: PMP tests assume all implemented entries are functional
`tests/priv/pmp/pmp32/PMPSm/pmpsm_all_entries_check.S` (and the pmp64 variant) walks an `RWX` check across every PMP entry the implementation reports via `UDB_NUM_PMP_ENTRIES`, asserting that each entry is usable (read should succeed, write should fault under L=1, XWR=101).

The problem is that the spec permits a hart to implement a given number of `pmpcfg/pmpaddr` CSRs while making some of them **read-only zero (WARL hardwired to 0)**. For example, a core may expose 16 PMP entries in the CSR space but tie the upper 8 to zero, so there are functionally only 8 usable regions. On such a core these CSRs are not writable, so the configuration written by the test has no effect and the expected access-fault / access-success behavior will not occur — the test will report spurious failures even though the implementation is spec-compliant.

The framework currently has no notion of "number of functional PMP regions" distinct from "number of implemented PMP CSRs", and the test cannot distinguish a hardwired-zero entry from a genuinely broken one.

### LINKED ISSUE #1723: Overlapping-region tests overflow address offsets at high PMP granularity (pmpsm_priority.S, pmpsm_cfg_tor_all.S)
`tests/priv/pmp/pmp32/PMPSm/pmpsm_priority.S` and `tests/priv/pmp/pmp32/PMPSm/pmpsm_cfg_tor_all.S` (and their pmp64 variants) build a series of partially overlapping regions whose sizes grow as multiples of the PMP granule `g` (`g = 1 << UDB_PMP_GRANULARITY`):

* `pmpsm_priority.S` sets up **7 even/odd TOR pairs** where pair `i` has size `i*g`, accessing addresses at `TEST_FOR_EXECUTION + i*g - 4`. The executable `TEST_FOR_EXECUTION` block itself is reserved as **8*g NOPs**.
* `pmpsm_cfg_tor_all.S` defines **15 TOR** regions **TEST_FOR_EXECUTION_0 .. _14** whose cumulative sizes scale with `g` (region `i` contributes `i * g/4`), so the total span grows quadratically with `g`.

As **UDB_PMP_GRANULARITY** increases, `g` grows exponentially and these regions quickly exceed the offsets the test relies on. The data/executable blocks become so large that the trap handler, RVMODEL macros, and failure macros end up too far apart, producing relocation truncation errors (**relocation truncated to fit: R_RISCV_JAL**) and similar link failures.

To let everybody know, as a stop-gap, these tests are currently partially masked to run only at low granularity values via `#if UDB_PMP_GRANULARITY == 2` guards (see pmpsm_priority.S lines ~165–184 and ~207–211). This means the overlapping-priority and TOR-all coverage is lost on implementations with a larger granule.

### LINKED ISSUE #1879: Test generation failures when physical memory protection top of range (PMP TOR) is not supported
**Describe the bug**
If tor_supported is set to false in the sail.json file, there are many test generation failures.

**To Reproduce**
Set "to_supported" to false in the sail.json file in any of the supported configurations.  Run test generation. 
With tor_supported set to true, no fails.

✗ Build failed: 4171 succeeded, 864 failed
  216 task(s) failed (see details above):

**Additional context**

### LINKED ISSUE #1880: PMPS/PMPU tests hardcode PMP entry 15 and abort on targets with 8 usable entries
Affected tests observed
pmps_cfg_A_off.S
pmps_cfg_XWR.S
pmps_mprv_check-01.S
pmps_mprv_check-02.S
pmps_napot_legal_lxwr-01.S
pmps_napot_legal_lxwr-02.S
pmps_tor_legal_lxwr-01.S
pmps_tor_legal_lxwr-02.S
pmpu_cfg_A_off.S
pmpu_cfg_XWR.S

from pmps_cfg_A_off:
The test attempts to configure PMP entry 15 as a locked, executable background region by writing pmpaddr15 and the corresponding configuration byte in pmpcfg2. Because Sail is configured with only eight usable PMP entries, these writes have no effect: Sail records both pmpaddr15 and pmpcfg2 as zero. Consequently, after mret enters S-mode, no PMP region permits execution of the test code, so the first S-mode instruction fetch immediately raises an instruction access fault (mcause=1).

============================

RVCP-SUMMARY: TEST FAILED - Test File "pmps_cfg_A_all.S"
RVCP: ===== TRAP FAILURE DIAGNOSTICS =====
RVCP: Failure: "The trap handler aborted the test before normal completion!"
RVCP: Unrecognized trap failure context. Printing available CSR state:
RVCP: Instruction that trapped: 0x00018293
RVCP: MEPC:    0x0000000080000284
RVCP: MCAUSE:  0x0000000000000000
RVCP: MTVAL:   0x0000000000000000
RVCP: MSTATUS: 0x8000000200006000

FAILURE: 1 (0x00000001)

============================
Several RV64 PMPS/PMPU tests unconditionally configure their global/background PMP region with pmpaddr15 and byte 7 of pmpcfg2. This assumes PMP entry 15 is writable.

A compliant implementation can expose the architectural 16-entry CSR bank while making entries 8-15 read-only zero. SiFive U74 is one example: it has eight usable PMP regions and pmpcfg2 is hardwired to zero. When Sail is configured to model this, the background-region writes have no effect and the tests abort on the first S-mode/U-mode instruction fetch instead of reaching their intended checks.

These tests are not yet ready to run on the board as final self-checking tests. Unexpected instruction access faults cause the Sail reference run to abort, so ACT cannot generate a valid reference signature or final self-checking ELF.

Configuration
UDB currently has only one PMP-entry-count parameter:

params:
  NUM_PMP_ENTRIES: 8
  PMP_GRANULARITY: 12
Sail can separately represent the architectural CSR bank and usable entries:

"memory": {
  "pmp": {
    "grain": 10,
    "count": 16,
    "usable_count": 8,
    "tor_supported": true,
    "na4_supported": false,
    "napot_supported": true
  }
}
The generated header correctly contains:

#define UDB_NUM_PMP_ENTRIES 8
#define UDB_NUM_PMP_ENTRIES_8

[act_pmp_usable_entries_8_core_and_spike_20260716.zip](https://github.com/user-attachments/files/30083820/act_pmp_usable_entries_8_core_and_spike_20260716.zip)

### LINKED ISSUE #1886: PMP test expecting wrong value for PMPADDR
**Test name**
pmpsm_grain.S

**Describe the bug**
The PMPCFG0 byte corresponding to PMPADDR0 is set to 0x1f which implies NAPOT, the PMP grain G is 12. The test expect 0 for PMPADDR0.  The spec says that the least significant 10 bits (G - 2) should read 1.  Here's the spec:

When G ≥ 2 and pmpcfgi.A[1] is set, i.e. the mode is NAPOT, then bits pmpaddri[G-2:0] read as all ones

**Expected behavior**
Least significant 10 bits of PMPADDR0 should be all ones per spec. 

**Additional context**
Test output: 
```
RVCP-SUMMARY: TEST FAILED - Test File "pmpsm_grain.S"
RVCP: Test Info: "test: 3;  cp: cp_pmp_grain"
RVCP: Instruction: 0x3b0023f3
RVCP: Approximate address (failure may be slightly after this): 0x00000000800001a0
RVCP: Register: x7
RVCP: Bad Value:      0x00000000000001ff
RVCP: Expected Value: 0x0000000000000000
```

Relevant instructions (from Whisper log):

```
#579 0  M 000000008000019c 3a031073 c 00000000000003a0 000000000000001f csrrw    x0, pmpcfg0, x6
#580 0  M 00000000800001a0 3b0023f3 r 0000000000000007 00000000000001ff csrrs    x7, pmpaddr0, x0

```

jordancarlin:
Lots of improvements to the PMP tests
- The Python workaround for missing UDB parameters is replaced with the actual UDB parameters.
- The background region now uses the highest writable PMP region instead of always using region 16/64.
- The background region now uses NAPOT when TOR is not supported.
- Tests that relied on TOR but were not specifically testing address matching modes were converted to use NAPOT (which most tests already used).
- PMP tests with large granularity no longer skip parts of the test to avoid linker errors. This was fixed by replacing `j` with `jr` in several places.
- The tests now throw an error is NAPOT is not supported or if there are fewer than 8 writable PMP regions. This can be worked around later, but for now it will make it clear that it is not supported instead of confusingly failing later. We have yet to see a DUT that would not meet these requirements.
- Updated test constraints so only tests that are explicitly testing NAPOT or TOR include that requirement in the YAML header.
- Use NAPOT in SsstrictSm for locked region.
- Lots of other small fixes/cleanups in the PMP tests

Fixes #1721
Fixes #1880
Fixes #1723
Fixes #1879
Fixes #1886


jordancarlin:
I am not yet set up to run coverage (working on having that running soon), so there are likely some further changes needed to the coverpoints after this. That can be part of a follow up PR.

davidharrishmc (approved):
Merging to unblock development.  @UmerShahidengr please make followup PR if you find issues.

PR #1941, opened 2026-07-22
Guard time/timeh CSR accesses behind UDB_TIME_CSR_IMPLEMENTED

### LINKED ISSUE #1924: SsstrictSm-03 ignores TIME_CSR_IMPLEMENTED=false and tests time CSR 0xC01
Test SsstrictSm-03:

The SsstrictSm-03 test checks several CSRs, including the time CSR (0xC01)`.On the VisionFive 2 (SiFive U74), the time CSR is not implemented (TIME_CSR_IMPLEMENTED = false). However, the generated ACT test still executes the instruction:
csrr x31, time

This causes different behavior on different platforms:
Sail and Spike allow the instruction and return a value.
VisionFive 2 treats the time CSR as unsupported and raises an Illegal Instruction exception.

Because the ACT test expected a successful CSR read but VF2 generated an exception, the recorded trap information differs from the expected result, leading to a trap signature pointer offset mismatch.

Logs snippet:

[CASE] START name=SsstrictSm-03
[PAYLOAD] kind=ACT entry=0x0000000080000000 tohost=0x0000000080041100 load_segments=0x0000000000000002
[ACT] sig_begin=0x000000008000bdc0
[ACT] sig_end=0x0000000080041030

RVCP-SUMMARY: TEST FAILED - Test File "SsstrictSm-03.S"
RVCP: Test Info: "Mismatch in trap signature pointer offset! The test likely observed an incorrect number of traps."
RVCP: Instruction: 0x00000013
RVCP: Approximate address (failure may be slightly after this): 0x000000008000445c
RVCP: Register: x6
RVCP: Bad Value:      0x0000000000000000

[act_ssstrictsm03_time_csr_config_issue_20260720.zip](https://github.com/user-attachments/files/30177591/act_ssstrictsm03_time_csr_config_issue_20260720.zip)

karabambus:
This PR fixes #1924 

On a DUT with `TIME_CSR_IMPLEMENTED: false`, reads of the `time` CSR raise an illegal-instruction exception, while the Sail always implements `time` when Zicntr is enabled (there is no sail config option).

- `SsstrictCommon.py`: wrap the CSR sweep blocks for `time` (0xC01) and `timeh` (0xC81) in `#ifdef UDB_TIME_CSR_IMPLEMENTED`.
- `Sm.py`: the `cp_mtime_write` testcase writes `mtime` via MMIO and reads    back the `time` CSR; it was guarded only by `RVMODEL_MTIME_ADDRESS`. A core  with a CLINT but no `time` CSR (e.g. CV32E40P, which defines   `RVMODEL_MTIME_ADDRESS` with `TIME_CSR_IMPLEMENTED: false` in its in-repo config) would hit the same DUT/Sail trap mismatch

**Follow-ups (same bug class, not addressed here)**
ZicntrS/ZicntrU read `time` in their mcounteren/scounteren walking-bit tests without a guard, and InterruptsSstc reads `time` for stimecmp setup. Certifying the trap behavior itself (instead of skipping the access) would need a sail hoption for unimplemented `time` with Zicntr otherwise present.

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
+++ b/generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
@@ -58,6 +58,9 @@ class RawSweep:
 SetupFn = Callable[[list[str], TestData, int], None]
 
 
+# CSRs guarded by a UDB parameter define; compiled out when the CSR is absent.
+PARAM_GUARDED_CSRS = {0xC01: "UDB_TIME_CSR_IMPLEMENTED", 0xC81: "UDB_TIME_CSR_IMPLEMENTED"}
+
 USER_CUSTOM_CSR_RANGES = (range(0x800, 0x900), range(0xCC0, 0xD00))
 S_CUSTOM_CSR_RANGES = (range(0x5C0, 0x600), range(0x9C0, 0xA00), range(0xDC0, 0xE00))
 H_CUSTOM_CSR_RANGES = (range(0x6C0, 0x700), range(0xAC0, 0xB00), range(0xEC0, 0xF00))
@@ -286,6 +289,9 @@ def _generate_csr_sweep_body(
             r1, r2, r3 = test_data.int_regs.get_registers(3)
 
             ih = hex(csr_addr)
+            guard = PARAM_GUARDED_CSRS.get(csr_addr)
+            if guard is not None:
+                lines.append(f"#ifdef {guard}")
             lines.extend(
                 [
                     f"# CSR {ih}",
@@ -304,6 +310,8 @@ def _generate_csr_sweep_body(
                     "",
                 ]
             )
+            if guard is not None:
+                lines.append("#endif")
             test_data.int_regs.return_registers([r1, r2, r3])
         _finalize_chunk(test_data, lines, test_chunks)
     return test_chunks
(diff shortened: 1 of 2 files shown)


generators/testgen/src/testgen/priv/extensions/SsstrictCommon.py
```diff
@@ -58,6 +58,9 @@ class RawSweep:
 SetupFn = Callable[[list[str], TestData, int], None]
 
 
+# CSRs guarded by a UDB parameter define; compiled out when the CSR is absent.
+PARAM_GUARDED_CSRS = {0xC01: "UDB_TIME_CSR_IMPLEMENTED", 0xC81: "UDB_TIME_CSR_IMPLEMENTED"}
```
davidharrishmc, line 62: This approach doesn't test TIME when it is not implemented.  However, Ssstrict requires that TIME trap when not implemented, so Ssstrict has a coverage hole on that requirement.

Is the fundamental issue that Sail should trap when TIME_CSR_IMPLEMENTED is false?  If so, I think the issue should be filed against Sail rather than ACT.

davidharrishmc (changes requested):
Sm change looks good.

It seems Sail should be fixed, rather than putting a waiver on Ssstrict.

jordancarlin:
We are going to catch `time` in the trap handler when TIME_CSR_IMPLEMENTED is false and invisibly emulate it, so I am closing this. The alternate solution will be available soon.

PR #1942, opened 2026-07-22
Added python generator for Sscofpmf

ayeshaanwaar05:
This PR contains tests of  SscofpmfSm ,SscofpmfS and SscofpmfU


ayeshaanwaar05:
Prof @davidharrishmc Now it is passing on both whisper & qemu rv64 & rv32.

jordancarlin:
@ayeshaanwaar05 what is the status of this PR? Is it ready for review?

ayeshaanwaar05:
> ayesha (@ayeshaanwaar05) what is the status of this PR? Is it ready for review?

yes @jordancarlin it is ready for review 
cp_lcofi and cp_lcofi_priority are failing on qemmu for SscofpmfS and SscofpmfU might be a qemu bug because at first they both were failing on whisper as well but when i made changes in trap handler it worked

UmerShahidengr (approved):

config/qemu/ci.yaml
```diff
@@ -23,7 +23,7 @@ ci_enabled: true
 # Smstateen will be added back once sail starts supporting SM1P13 too
 # ZawrsU and ZawrsS fail on qemu due to stval mismatching, filed QEMU issue https://gitlab.com/qemu-project/qemu/-/work_items/4080
 # ZicntrS and ZicntrU fail on qemu because ecall increments instret. Filed QEMU issue: https://gitlab.com/qemu-project/qemu/-/work_items/4087
-exclude_extensions: "Sm,S,Zacas,ZacasZabha,PMPSm,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,SvZicbo,Svinval,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,Zkr,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU,ZicntrS,ZicntrU"
+exclude_extensions: "Sm,S,Zacas,ZacasZabha,PMPSm,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,SvZicbo,Svinval,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,Zkr,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU,ZicntrS,ZicntrU,SscofpmfS,SscofpmfU"
```
davidharrishmc: When adding an exclusion, add a comment of why it is necessary.  The comment preferably will reference an open issue against the simulator that is causing the problem.

config/sail/ci.yaml
```diff
@@ -3,7 +3,7 @@
 # CI configuration for Sail reference model
 
 ci_enabled: true
-exclude_extensions: "SdtrigSm,SdtrigS,SdtrigU"
+exclude_extensions: "SscofpmfSm,SscofpmfS,SscofpmfU,SdtrigSm,SdtrigS,SdtrigU"
```
davidharrishmc: Explain exclusion, preferably with link to issue.

config/spike/ci.yaml
```diff
@@ -8,7 +8,7 @@ ci_enabled: true
 # TODO: Priv tests mismatch due to config issues.
 # TODO: Remove SsstrictV when sail SsstrictV is implemented
 # Smstateen will be added back once sail starts supporting Sscsrind too
-exclude_extensions: "Sm,SsstrictSm,SsstrictS,SsstrictU,SsstrictV,Smstateen,SdtrigSm,SdtrigS,SdtrigU"
+exclude_extensions: "Sm,SsstrictSm,SsstrictS,SsstrictU,SsstrictV,Smstateen,SdtrigSm,SdtrigS,SdtrigU,SscofpmfSm,SscofpmfS,SscofpmfU"
```
davidharrishmc, line 12: Explain exclusion, preferably with link to issue.

config/whisper/whisper-rv64-max/rvmodel_macros.h
```diff
@@ -213,4 +213,9 @@
   li _R2, WHISPER_SSIP_ADDRESS; \
   sw zero, 0(_R2);
 
+#define RVMODEL_MHPMEVENT   mhpmevent3
+#define RVMODEL_MHPMCOUNTER mhpmcounter3
+#define RVMODEL_MHPMEVENT_VAL  14   // use 15 when testing the Store event
```
davidharrishmc, line 218: Define what value 14 means (presumably counts the number of loads?)

generators/testgen/src/testgen/priv/extensions/SscofpmfU.py
```diff
@@ -0,0 +1,84 @@
...
+    march_extensions=[],
+    extra_defines=[
+        "#define RVTEST_TEMP_BOOT_TO_U",
+    ],
+)
+def make_sscofpmfu(test_data: TestData) -> list[TestChunk]:
+    """Generate tests for the SscofpmfU performance-counter-overflow testsuite."""
+    tc = test_data.begin_test_chunk()
+    tc.code.extend(_generate_lcofi_sip_u_tests(test_data))
+    return generate_sscofpmf_suite(test_data, "U")
```
davidharrishmc: This is weird syntax in that it doesn't mention tc.  It would appear it might discard the work done on the existing chunk.  The usual idiom tc.code.extend(_generate_csr_illegal_accesses(test_data)).  Can you do the same?  Applies to all suites.

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -199,7 +199,7 @@ def make_sscofpmfs(test_data: TestData) -> list[TestChunk]:
     """Generate tests for the SscofpmfS performance-counter-overflow testsuite."""
     test_chunks: list[TestChunk] = []
     tc = test_data.begin_test_chunk()
-    tc.code.extend(_generate_lcofi_sip_s_tests(test_data))
+    # tc.code.extend(_generate_lcofi_sip_s_tests(test_data))
```
davidharrishmc: Are you sure you want this commented out?

ayeshaanwaar05: no no my bad

config/qemu/qemu-rv64-max/rvmodel_macros.h
```diff
@@ -263,4 +263,10 @@
 
 #define RVMODEL_CLR_SSW_INT(_R1, _R2)
 
+#define RVMODEL_MHPMCOUNTER    mhpmcounter3
```
davidharrishmc, line 266: I think spike, sail, and imperas should define these RVMODEL macros too so at least they compile, even if they don't do anything.  Leave a comment in the config file that they won'd do anything until HPM counters are able to increment (and reference Sail issue).

generators/testgen/src/testgen/priv/extensions/SscofpmfU.py
```diff
@@ -0,0 +1,116 @@
...
+    test_data.int_regs.return_registers([r_val, r_temp, r_scratch])
+    return lines
+
+
+@add_priv_test_generator(
+    "SscofpmfU",
+    required_extensions=["U", "Sscofpmf"],
+    march_extensions=[],
+    extra_defines=[
+        "#define RVTEST_TEMP_BOOT_TO_U",
```
davidharrishmc, line 106: This should be BOOT_TO_UMODE now after recent T-SBI improvements.

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,247 @@
...
+    test_data.int_regs.return_registers([r1, r_scratch, r_temp])
+    return lines
+
+
+@add_priv_test_generator(
+    "SscofpmfS",
+    required_extensions=["S", "Sscofpmf"],
+    march_extensions=[],
+    extra_defines=[
+        "#define RVTEST_TEMP_BOOT_TO_S",
```
davidharrishmc, line 236: This should be BOOT_TO_SMODE now after recent T-SBI improvements.

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,247 @@
...
+                "mip.LCOFIP (and its delegated sip.LCOFIP view) is a read-only\n"
+                "shadow of the OR of mhpmeventN.OF bits (WARL on direct writes), so\n"
+                "LCOFIP=1 is driven via a real counter overflow rather than\n"
+                "csrs mip, LCOFI_BIT -- a direct write is a silent no-op and never\n"
+                "actually latches the pending bit."
+            ),
+        ),
+        "",
+        "# === M-MODE SETUP ===",
+        "csrw mip, zero      # clear all pending",
```
davidharrishmc, line 40: There can't be M-mode setup, because you've set the S suite to boot into S-mode.  Need to use T-SBI calls if you want to access M-mode from a lower priv mode.

davidharrishmc (commented):
QEMU and whisper aren't generating any LCOFI interrupts and the counters aren't counting.

Make sure the events are configured to occur properly and the inhibits are 0 for the mode where an event should be counted.

Once booting to a mode, the tests need to use tsbi_call to access higher privilege registers.

davidharrishmc:
Check the copilot review above, repeated here.  Uncertain whether mode comment matters.

_generate_scountovf_mcounteren_tests() samples scountovf only once after csr_walk_test(), so it doesn’t check masking for each mcounteren walk setting. Also, the function never switches into the requested mode (S/U), so the "mode" parameter doesn’t reflect the privilege level under test.
        r_scountovf = test_data.int_regs.get_register(exclude_regs=[0, 31])
        lines.append(_csr_access(f"csrr x{r_scountovf}, scountovf   # sample point", mode))
        lines.append("")
generators/testgen/src/testgen/priv/extensions/SscofpmfCommon.py:142

This programs mhpmevent using bit positions 58..63 via a single LI into an integer register. On RV32, those upper bits live in the mhpmevent3h half, so this value will be truncated and won’t set OF/xINH as intended; similarly, writing MHPMCOUNTER with -1 won’t initialize the upper half if the counter is 64-bit on RV32.
                f"LI(x{r_val}, RVMODEL_MHPMEVENT_VAL | (0b11100 << 58) | ({of_initial} << 63))",
            ]
        )

davidharrishmc (changes requested):

PR #1945, opened 2026-07-22
InterruptsS based on testplan, moved a test into InterruptsSSm

Teemooooooooo:
- Both InterruptsS and InterruptsSSm are not at 100% coverage, waiting on SAIL configuration to be adjusted to allow M level interrupt delegation
- Modified InterruptsS test and coverpoints to delegate M level interrupts when testing in U mode, as the undelegated version is already tested in InterruptsU
- Moved cp_priority_mideleg_m into InterruptsSSm


jordancarlin:
> * Both InterruptsS and InterruptsSSm are not at 100% coverage, waiting on SAIL configuration to be adjusted to allow M level interrupt delegation

This should be possible with #1946

davidharrishmc:
Superseded by interrupt refactor.

PR #1949, opened 2026-07-23
Small Touchups to Sail Configs

fourth-bit:
Whisper requires mideleg to have 2AAA as delegatable bits for mideleg.

For coverage, it is also necessary to have 2AAA as delegatable bits for mideleg, so they are set in sail-rv32-max and sail-rv64-max.

### DIFF
--- a/config/sail/sail-RVA20S64/sail.json
+++ b/config/sail/sail-RVA20S64/sail.json
@@ -587,7 +587,7 @@
       "supported": false
     },
     "Zaamo": {
-      "supported": true
+      "supported": false
     },
     "Zabha": {
       "supported": false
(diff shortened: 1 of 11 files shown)


config/sail/sail-RVA20S64/sail.json
```diff
@@ -587,7 +587,7 @@
       "supported": false
     },
     "Zaamo": {
-      "supported": false
+      "supported": true
```
jordancarlin: These had been left as false to match the default sail config. There is also a semantic difference for Sail: if Zaamo is marked as supported, then amos are always enabled irrespective of misa.A. When it is not supported, then they are only enabled when misa.A is set. The same applies to Zalrsc.

fourth-bit: Rolled back those changes

jordancarlin (approved):

PR #1955, opened 2026-07-23
Remove M-mode CSR access in failure code and cleanup after sail 0.13 update

davidharrishmc:
Fixed cvw InterruptsS by increasing RVMODEL_TIMER_INT_SOON_DELAY
Drop unnecessary Zicsr from march_extensions
Drop Zkr exclusion from DUTs that don't support Zkr
Fix xEPC snapshotting

### DIFF
--- a/config/cores/cvw/cvw-rv64gc/rvmodel_macros.h
+++ b/config/cores/cvw/cvw-rv64gc/rvmodel_macros.h
@@ -99,7 +99,12 @@
 
 ##### Machine Timer #####
 
-#define RVMODEL_TIMER_INT_SOON_DELAY 100
+// Wally's mtime advances one tick per core clock, and the code between arming
+// mtimecmp and reaching the lower privilege mode (three CLINT stores plus
+// RVTEST_GOTO_LOWER_MODE) can take well over 100 cycles on a pipelined core
+// with caches.  With a delay of 100 the timer interrupt fires while still in
+// M-mode with MIE=1, so the trap records MPP=M instead of MPP=U/S.
+#define RVMODEL_TIMER_INT_SOON_DELAY 1000
 
 #define RVMODEL_MTIME_ADDRESS  0x0200BFF8  /* Address of mtime CSR */
 
--- a/config/imperas/imperas-rv32gck/imperas.ic
+++ b/config/imperas/imperas-rv32gck/imperas.ic
@@ -60,6 +60,7 @@
 
 # Misaligned access (Zicclsm)
 --override cpu*/unaligned=T
+--override cpu*/Zam=T # misaligned AMOs are supported
 
 # PMP Configuration
 --override cpu*/PMP_registers=16
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsZaamo.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsZaamo.py
@@ -251,7 +251,7 @@ def _generate_amo_access_fault_tests(test_data: TestData) -> list[str]:
 @add_priv_test_generator(
     "ExceptionsZaamo",
     required_extensions=["Zaamo", "Sm"],
-    march_extensions=["I", "Zicsr", "Zaamo", "Zabha", "Zacas"],
+    march_extensions=["I", "Zaamo", "Zabha", "Zacas"],
 )
 def make_exceptionszaamo(test_data: TestData) -> list[TestChunk]:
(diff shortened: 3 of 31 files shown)


config/cores/cvw/cvw-rv64gc/rvmodel_macros.h
```diff
@@ -99,7 +99,12 @@
 
 ##### Machine Timer #####
 
-#define RVMODEL_TIMER_INT_SOON_DELAY 100
+// Wally's mtime advances one tick per core clock, and the code between arming
+// mtimecmp and reaching the lower privilege mode (three CLINT stores plus
+// RVTEST_GOTO_LOWER_MODE) can take well over 100 cycles on a pipelined core
+// with caches.  With a delay of 100 the timer interrupt fires while still in
+// M-mode with MIE=1, so the trap records MPP=M instead of MPP=U/S.
```
jordancarlin, line 106: Is there a way to speed this up? Why are there 3 CLINT stores after arming mtimecmp? Can we do those before arming mtimecmp?

generators/testgen/src/testgen/priv/extensions/ExceptionsZaamo.py
```diff
@@ -251,7 +251,7 @@ def _generate_amo_access_fault_tests(test_data: TestData) -> list[str]:
 @add_priv_test_generator(
     "ExceptionsZaamo",
     required_extensions=["Zaamo", "Sm"],
-    march_extensions=["I", "Zicsr", "Zaamo", "Zabha", "Zacas"],
+    march_extensions=["I", "Zaamo", "Zabha", "Zacas"],
```
jordancarlin: While you are editing this, drop I as well. It will make it easier to support the priv tests on E later.

generators/testgen/src/testgen/priv/extensions/Sstvala.py
```diff
@@ -284,7 +284,7 @@ def _jalr(va: int, suffix: str) -> tuple[str, list[str]]:
 @add_priv_test_generator(
     "Sstvala",
     required_extensions=["Sstvala"],
-    march_extensions=["S", "Zicsr"],
+    march_extensions=["S"],
```
jordancarlin: S shouldn't be in the march either (we actually strip it out before writing the file). How about just making this an empty list?

tests/env/rvtest_trap_handler.h
```diff
@@ -1746,6 +1746,17 @@ tsbi_\__MODE__\()handle_forwarded:
...
 
+        // Snapshot xEPC to saved_mepc before the first TRAP_SIGUPD so
+        // failedtest_print_csr_context reports THIS trap's EPC on any word's
+        // mismatch.  It must be captured here rather than at reporter entry:
+        // the exception path rewrites the live xEPC CSR (adj_\__MODE__\()epc_rtn)
+        // after word 2, and snapshotting only in common_\__MODE__\()excpt_handler
+        // (as before) left word 0/1 mismatches printing the PREVIOUS trap's EPC.
+        // T3 and T4 are dead here (both rewritten in sv_\__MODE__\()vect before use).
+        csrr    T3, CSR_XEPC
+        la      T4, saved_mepc
```
jordancarlin: Use the LA macro.

jordancarlin (changes requested):

config/cores/cvw/cvw-rv64gc/rvmodel_macros.h
davidharrishmc: Not a priority.

jordancarlin: My concern is that this will make it tricky for other DUTs because of the race condition. Fine to revisit later, but I do think it makes these tests less robust.

config/imperas/imperas-rv32gck/imperas.ic
```diff
@@ -60,6 +60,7 @@
 
 # Misaligned access (Zicclsm)
 --override cpu*/unaligned=T
+--override cpu*/Zam=T # misaligned AMOs are supported
```
jordancarlin, line 63: This doesn't seem like the right fix. From the spec:
> The draft Zam extension has been removed, in favor of the definition of a misaligned atomicity granule PMA.

Is there an option to set the MAG PMA for specific memory regions instead?

jordancarlin (approved):
I'm going to merge this for now and we can look at cleaning up the Imperas configs later.

PR #1961, opened 2026-07-23
Allow Certain Mask Producing Operations to be Run at vl=VLEN, SEW=8, LMUL=8

### LINKED ISSUE #1958: Vector tests checking tail elements when tail agnostic is set
**Test name**
Vx32-vmsif.m

**Describe the bug**
The test is checking the tail element of a vector instruction destination register when tail-agnostic is in effect. The expected value of a tail element may be all-ones or the previous value depending on the policy and there is currently no way to configure a tail-agnostic policy.

The problem instruction is "vsmif.m".  The DUT computes all the elements in the destination register (all VLEN bits) as allowed by the spec. There is no way to configure the test-generator for that.

Until we can configure the tail-agnostic policy and the "compute all VLEN bits" option, the vector tests should not be checking any tail element.

**Expected behavior**
Do not check the tail elements when tail-agnostic is in effect or when the DUT computes all VLEN elements of a mask destination register.

**Additional context**

***Test output:***

```
RVCP-SUMMARY: TEST FAILED - Test File "Vx32-vmsif.m-00.S"
RVCP: Test Info: "test: 82; cg: Vx32_vmsif.m_cg; cp: cr_vl_lmul; bin: cp_vl_lmul_vl_vlmax_lmul_0.5"
RVCP: Instruction: 0x50d1a557
RVCP: Approximate address (failure may be slightly after this): 0x0000000080004f48
RVCP: Register: v10
RVCP: Region: TAIL
RVCP: Element Index: 4
RVCP: VL: 0x0000000000000004
RVCP: VTYPE: 0x00000000000000d7
RVCP: Bad Value:      0x00000000
RVCP: Expected Value: 0xdeadbeef
RVCP: Mismatch Mask (one bit per element, up to VLMAX bits):
0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff50
```
Note that VL is 4 and the checked element index is also 4. The test is checking a tail element.

***Spec related to vsmif.m***
The DUT computes all VLEN elements of the destination mask register as allowed by the spec. Consequently the test should not check the tail elements unless the reference model does the same (compute all VLEN bits). Here's the spec:

```
Furthermore, for mask-logical instructions and VMSBF.M, VMSIF.M, VMSOF.M mask-manipulation
instructions, any element in the tail of the result can be written with the value the
mask-producing operation would have calculated with vl=VLEN, SEW=8, and LMUL=8 (i.e., all
bits of the mask register can be overwritten).
```

In addition, for vmsif.f, tail-agnostic is always in effect. Consequently the test should not check tail elements unless the reference model uses the same tail-agnostic policy as the DUT. Here's the spec:
```
The tail elements in the destination mask register are updated under a tail-agnostic policy.
```

fourth-bit:
The spec says that: 
>Furthermore, for mask-logical instructions and vmsbf.m, vmsif.m, vmsof.m mask-manipulation instructions, any element in the tail of the result can be written with the value the mask-producing operation would have calculated with vl=VLEN, SEW=8, and LMUL=8 (i.e., all bits of the mask register can be overwritten).

This patch adds support for running those operations an additional time at vl=VLEN, SEW=8, and LMUL=8, and fixes #1958.

This required a change to sigupd to accommodate always loading and storing masks at their full length, and required a change to mask calculation so that beyond VL, results are consistent between Sail and a DUT setting all ones in a tail-agnostic policy. 

It also updates the whisper config to always test this case.

### DIFF
--- a/config/whisper/ci.yaml
+++ b/config/whisper/ci.yaml
@@ -15,7 +15,7 @@
 # InterruptsSSm excluded because whisper supports M mode interrupt delegation and Sail currently only supports S mode interrupt delegation.
 ci_enabled: true
 # TODO: Remove SsstrictV when sail SsstrictV is implemented
-exclude_extensions: "S,Sm,SmV,SsstrictSm,SsstrictS,SsstrictU,ExceptionsZaamo,ExceptionsSvZalrsc,SsstrictV,Smstateen,InterruptsSSm,Zkr"
+exclude_extensions: "S,Sm,SmV,SsstrictSm,SsstrictS,SsstrictU,ExceptionsZaamo,ExceptionsSvZalrsc,SsstrictV,Smstateen,InterruptsSSm,Zkr,Vx8,Vx16,Vx32"
 install_script: ".github/scripts/install-whisper.sh"
 apt_packages: "libboost-program-options-dev"
 shards: 1
--- a/config/whisper/whisper-rv32-max/whisper.json
+++ b/config/whisper/whisper-rv32-max/whisper.json
@@ -129,7 +129,7 @@
     "trap_non_zero_vstart": true,
     "tt_clear_tval_vl_egs": true,
     "tt_fp_usum_tree_reduction": ["e16", "e32", "e64"],
-    "update_whole_mask": false
+    "update_whole_mask": true
   },
 
   "wfi_timeout": 200
--- a/config/whisper/whisper-rv64-max/whisper.json
+++ b/config/whisper/whisper-rv64-max/whisper.json
@@ -128,7 +128,7 @@
     "trap_non_zero_vstart": true,
     "tt_clear_tval_vl_egs": true,
     "tt_fp_usum_tree_reduction": ["e16", "e32", "e64"],
-    "update_whole_mask": false
+    "update_whole_mask": true
   },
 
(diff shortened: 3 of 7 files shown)


jordancarlin:
AI review:

 Blocking issues

 - generators/testgen/src/testgen/asm/vector_helpers.py:463-467 — random mask loads now use vsetvli ..., e8, m8 before vlm.v, but the random
   mask data generation was not updated. Existing mask labels are sized for the original mask VLMAX (vector_params.py still emits as little
   as one SEW element for base tests, or VLEN_MAX*lmul/sew bits for length tests). With e8,m8, vlm.v reads VLEN/8 bytes, so many tests will
   read past the mask label into following data. That makes the mask contents layout-dependent and can invalidate the intended mask coverage
   / full-mask tail checks. The mask label generation should emit enough bytes for a full mask register when prep_mask_v loads with e8,m8
   (or this path should only use e8,m8 when the data is sized accordingly).

 Nits / non-blocking

 - generators/testgen/src/testgen/asm/vector_helpers.py:344-349 — set_lower_xreg_bits has an unfinished docstring: “Uses the same technique
   as”.

jordancarlin (approved):

PR #1983, opened 2026-07-24
Add C runtime and compilation framework

jordancarlin:
This will enable C tests to be added to the ACTs in the future. It is fully standalone to avoid pulling in any libraries.


davidharrishmc (approved):
LGTM

PR #1987, opened 2026-07-25
Update sail version in Dockerfile

nazar-pc:
Looks like it is not covered by automatic updates, while current `act4` branch already requires Sail 0.13

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@
 # Global ARGs - declared before any FROM so they are overridable across all stages.
 # Each stage that uses them must redeclare them with a bare ARG (no default) to bring them into scope.
 ARG RISCV_TOOLCHAIN_PREFIX=/opt/riscv
-ARG SAIL_VERSION=0.12
+ARG SAIL_VERSION=0.13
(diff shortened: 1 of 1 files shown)


jordancarlin (approved):
Thanks for fixing this. I think you are currently the only person using the Docker image, so it hasn't gotten the attention it probably needs. There have been a variety of higher priority items, but I'm hoping to get back to the Docker image soon and make sure it is integrated into all of the other automation and testing to avoid regression like this.

nazar-pc:
I'm not even using it as such yet since it doesn't contain the source code of the repo due to lack of newer releases, so I'm actually still building container image myself downstream.

There were issues with some tests in the version I used before, so I've upgraded and noticed old Sail version that way. It'd be nice to compile one of the targets with container image in CI as a test to ensure it remains functional.

jordancarlin:
> There were issues with some tests in the version I used before, so I've upgraded and noticed old Sail version that way. It'd be nice to compile one of the targets with container image in CI as a test to ensure it remains functional.

Agreed. That has been on my list for a while, but I haven't gotten to it yet. If you are interested in adding this, feel free. Otherwise, I will try to get to it soon.

PR #1989, opened 2026-07-26
Add pre-commit hook to forbid li/la pseudoinstructions

### LINKED ISSUE #1810: Add pre-commit check to enforce the use of LI and LA macros

pyarchana:
Closes #1810.

`li` and `la` expand to a variable number of instructions depending on the value or address being materialized, making code size and branch offsets target-dependent. This hook requires the fixed-length `LI()`/`LA()` macros from `tests/env/utils.h` instead.

Mirrors `forbid-align-directive` (#1804), as suggested in the issue.

Unlike `.align`, which had no pre-existing occurrences, `li`/`la` appears in 168 files already on `act4`. Those are listed in `exclude` as a baseline so this PR stays hook-only. It's not a permanent exemption, I'd like to convert them in follow-up PRs and delete entries as they're done. New and modified files are enforced immediately: the hook runs on 1887 of the 2428 files matching its filter.

### DIFF
--- a/.github/scripts/check-li-la-pseudoinstructions.sh
+++ b/.github/scripts/check-li-la-pseudoinstructions.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+# check-li-la-pseudoinstructions.sh
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Pre-commit hook: fail if the `li` or `la` assembler pseudoinstructions are
+# used. Both expand to a variable number of instructions depending on the
+# value/address being materialized, which makes code sizes and branch offsets
+# target-dependent. This repo requires the fixed-length `LI(reg, imm)` and
+# `LA(reg, label)` macros from tests/env/utils.h instead.
+#
+# Usage: check-li-la-pseudoinstructions.sh <file>...
+
+set -euo pipefail
+
+# Match `li`/`la` as a standalone mnemonic. The negative lookahead lets the
+# uppercase LI()/LA() macros through, including when written as `LI (reg, imm)`.
+pattern='(^|[[:space:];:"'"'"'])(li|la)(?![[:space:]]*\()[[:space:]]'
+
+if [ -t 1 ] && [ -z "${NO_COLOR:-}" ]; then
+  red=$'\033[31m'
+  bold=$'\033[1m'
+  reset=$'\033[0m'
+else
+  red=''
+  bold=''
+  reset=''
+fi
+
+found=0
+for f in "$@"; do
+  matches=$(grep -nP "$pattern" "$f" || true)
+  if [ -n "$matches" ]; then
+    found=1
+    while IFS= read -r line; do
+      echo "$f:$line"
+    done <<< "$matches"
+  fi
+done
+
+if [ "$found" -ne 0 ]; then
+  echo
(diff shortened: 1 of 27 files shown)


jordancarlin (commented):
@pyarchana it is hard to tell whether this pre-commit hook works or not when all of the failing files are excluded. I would rather we fix the failures in this PR. Please leave the initial commit as it (just the script) and then add additional commits to the PR so that it is easier to review.

pyarchana:
> Archana K. (@pyarchana) it is hard to tell whether this pre-commit hook works or not when all of the failing files are excluded. I would rather we fix the failures in this PR. Please leave the initial commit as it (just the script) and then add additional commits to the PR so that it is easier to review.

makes sense. okay i will leave the first commit as just the hook and add per-area commits on top that fix the failures and drop the matching exclude entries so the exclude block ends up empty by the last commit. starting with `tests/env` then the generators then `tests/priv`.

jordancarlin:
@pyarchana because this PR touches so many files, it is going to keep accumulating merge conflicts unless it is finished all at once and merged. Do you think you are going to be able to finish this soon? If not, would you be ok with me taking it over so that we can get it merged?

pyarchana:
thanks @jordancarlin  yes please take it over. i don't think i can finish it
at the pace it needs and you are right that it will keep accumulating conflicts.

current state: the hook script itself is done and was green on CI. the two
follow-up commits convert `tests/env` and the jal/Zicsr generators plus their
regenerated tests but both now conflict with `act4` and would need redoing on
top of current head.

a few findings that might save you time:

- the hook has to be **case-sensitive** (`grep -nP`, not `-Pi`). case-insensitive
  falsely flags prose in comments, e.g. `Uses LI + add to handle delays` in
  `generators/testgen/src/testgen/asm/interrupts.py`.
- `c.li` is a real compressed instruction and correctly isn't matched.
- as of today there are **113 files** with raw `li`/`la` on `act4`, down from 168
  three weeks ago: 71 `tests/priv`, 20 `generators/testgen`, 16 `tests/rv*`,
  6 `tests/env`.
- these generators emit raw `li`/`la` into generated assembly so converting them
  changes generator output rather than just source:
  `scripts/custom/cp_custom_vfp_state.py`, `scripts/priv/_mstatus_helpers.py`,
  `scripts/priv/_raw_encoding.py`, `scripts/priv/_ssstrictv_helpers.py`,
  `scripts/priv/cp_exceptionsv_{LS,address_fault,ffLS}.py`,
  `scripts/priv/cp_misalignv_ls_family.py`,
  `scripts/priv/cp_ssstrictv_{sew_lmul_family,vstart_ge_vlmax}.py`,
  `scripts/vector-testgen-priv.py`.

unrelated but worth a look: `tests/env/rvtest_trap_handler.h` lines 702 and 712
have `LI    T2, (1<<MPV_LSB)` without parens. `LI` is a function-like macro, so
that never expands and it reaches the assembler as literal `LI T2, ...`. it's
inside a VUmode/VSmode `.if` branch so it may just not be assembling today.

i will leave the branch up in case any of it is useful.

jordancarlin:
Closing. Will open a new PR for this eventually.

PR #1994, opened 2026-07-27
Add linker errors when test exceeds memory size

jordancarlin:
Updates to the linker scripts to define the memory region so that useful errors are given when the test is too large for the provided memory.


davidharrishmc (commented):
I'm not sure I like this.  It adds a lot of complexity to the linker file that every user has to absorb.  Commercial users generally understand their memory constraints, so this mostly benefits small noncommercial designs.  Can we get some more input from the ACT team about whether the pros outweigh the cons?

davidharrishmc (changes requested):
Feedback from ACT about whether this complexity is worthwhile.

UmerShahidengr (approved):

jordancarlin:
> I'm not sure I like this. It adds a lot of complexity to the linker file that every user has to absorb. 

Most of the complexity was added as part of #1983. This is mostly just adding the asserts to get better error messages.

> Commercial users generally understand their memory constraints, so this mostly benefits small noncommercial designs.

They know their memory constraints, but they don't necessarily know the size of all of the ACT ELFs. When the ACT ELF is larger than the allocated memory, it fails signature generation with a very cryptic error (indicating a trap loop). This is trying to make that easier to debug by failing with a clear message indicating there is not enough memory.

davidharrishmc (approved):

PR #2005, opened 2026-07-27
Fix Zkr seed entropy branch labels

jordancarlin:
The Zkr seed entropy generator used numeric local labels around RVTEST_SIGUPD calls. RVTEST_SIGUPD also defines a 1: local label, so the branch to 1f resolved into the signature update macro instead of the intended non-ES16 path.

Use named local labels for the generator control flow to avoid local-label capture.


UmerShahidengr (approved):

davidharrishmc (commented):
Lgtm

PR #2018, opened 2026-07-28
Split CI exclusion lists

jordancarlin:
Convert ci.yaml exclude_extensions to YAML lists with one exclusion per line and move exclusion reasons inline with their entries. This will help avoid merge conflicts.

### DIFF
--- a/.github/scripts/ci_config.py
+++ b/.github/scripts/ci_config.py
@@ -35,7 +35,7 @@ import heapq
 import json
 import os
 import sys
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 from functools import cache
 from pathlib import Path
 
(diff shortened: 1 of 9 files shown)


davidharrishmc:
Minor drawback is that this makes pasting the list of exclusions into
makefile more difficult when trying to run all tests on a dut. Have to
retype them with commas. Any thoughts?

jordancarlin:
That's a fair point. I could go either way on this one. This approach reduces merge conflicts, leaving it as is makes it easier to copy and paste. No strong preference from me. Happy to close instead.

davidharrishmc:
I mostly copy and paste do for my use case the older way is easier. We
could also leave this open and merge when more of the exclusions are
resolved.  I’ll defer to your judgement.

jordancarlin:
Closing for now. We can revisit this later.

PR #2029, opened 2026-07-30
Fix Zkr test and reenable in CI

jordancarlin:
Sail 0.13.1 fixed the nondeterminism in the seed CSR. #2005 fixed a label issue in the Zkr test, which in turn uncovered an issue in the shift to isolate OPST (fixed here).

Zkr is still skipped on QEMU because of https://gitlab.com/qemu-project/qemu/-/work_items/4121.


jordancarlin:
Depends on #2027.

davidharrishmc (approved):

PR #2031, opened 2026-07-30
Full coverage of Ssnpm

UmerShahidengr:
Superseeding #1827


config/qemu/ci.yaml
```diff
@@ -23,7 +23,15 @@ ci_enabled: true
...
 # ZicntrS and ZicntrU fail on qemu because ecall increments instret. Filed QEMU issue: https://gitlab.com/qemu-project/qemu/-/work_items/4087
-exclude_extensions: "Sm,S,Zacas,ZacasZabha,PMPSm,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,SvZicbo,Svinval,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,Zkr,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU,ZicntrS,ZicntrU"
+# Ssnpm: the pointer-masking probes sweep every instruction class that takes a masked
+#        effective address, including the Zicbom CBOs. QEMU's cbo.clean/flush/inval do
+#        not fault on an inaccessible address, so one expected trap is missing and every
+#        later trap-signature entry is shifted -- the same defect already excluding
+#        ExceptionsZicboU/S and SvZicbo (https://gitlab.com/qemu-project/qemu/-/work_items/3501).
+#        Only Ssnpm_bare is affected (under Sv the CBOs page-fault, which QEMU does
+#        implement, so Ssnpm_sv39/sv48/sv57 pass); exclusion is per-extension so the whole
+#        suite is excluded. Verified against Sail 0.13.1 and Spike, which both fault.
```
jordancarlin, line 33: Cut this down to a one-line comment like all of the other exclusions.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -57,7 +57,7 @@
             `ifdef ZACAS_SUPPORTED
                 wildcard bins amocas_w = {AMOCAS_W};
                 wildcard bins amocas_d = {AMOCAS_D};
-                wildcard bins amocas_q = {AMOCAS_Q};
+                //wildcard bins amocas_q = {AMOCAS_Q}; Not supported
```
jordancarlin, line 60: What does "not supported" mean? `amocas.q` is a valid instruction on RV64 when Zacas is implemented.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -128,12 +128,12 @@
         `endif
         // Zicfiss shadow-stack stores and loads
         `ifdef ZICFISS_SUPPORTED
-            wildcard bins sspush      = {SSPUSH};
-            wildcard bins c_sspush    = {C_SSPUSH};
+           // wildcard bins sspush      = {SSPUSH}; in framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh we have localparam [31:0]  SSPUSH_X1  and SSPUSH_X5
+           // wildcard bins c_sspush    = {C_SSPUSH};  in framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh we have localparam [31:0] C_SSPUSH_X1
```
jordancarlin, line 132: Don't comment these out. If they are defined as `SSPUSH_X1` and `SSPUSH_X5`, then use those names. You should be able to do something like the following.
```sv
wildcard bins sspush      = {SSPUSH_X1, SSPUSH_X5};
```

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -128,12 +128,12 @@
...
             wildcard bins ssamoswap_w = {SSAMOSWAP_W};
             wildcard bins ssamoswap_d = {SSAMOSWAP_D};
-            wildcard bins sspopchk    = {SSPOPCHK};
-            wildcard bins c_sspopchk  = {C_SSPOPCHK};
+            //wildcard bins sspopchk    = {SSPOPCHK};
+            // wildcard bins c_sspopchk  = {C_SSPOPCHK};   model doesn’t define this yet, instead in framework/src/act/fcov/coverage/RISCV_imported_decode_pkg.svh we have localparam [31:0] C_SSPOPCHK_X5
```
jordancarlin, line 136: Same as above. Don't comment these out. Use the names defined in riscv-opcodes.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -188,7 +188,7 @@
             bins sv57 = {4'b1010};
         `endif
     }
-    a_upper_bits: coverpoint (ins.current.rs1_val + ins.current.imm)[63:48] {
+    a_upper_bits: coverpoint ((ins.current.rs1_val + ins.current.imm) >> 48)  {
```
jordancarlin, line 191: Why is this change needed?

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -209,7 +209,7 @@
     }
 
     // Misaligned address (e.g. scratch+1); upper 7 bits = 0x01 or 0x00
-    misaligned_addr: coverpoint (ins.current.rs1_val + ins.current.imm)[1:0] {
+    misaligned_addr: coverpoint ((ins.current.rs1_val + ins.current.imm) & 2'b11) {
```
jordancarlin, line 212: Why is this change needed?

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
```diff
@@ -218,9 +218,15 @@
 
     `ifdef RVMODEL_ACCESS_FAULT_ADDRESS
         // Exception should write xtval with masked version of pointer.
-        illegal_addr: coverpoint (ins.current.rs1_val + ins.current.imm)[47:0] {
+        // Match on the UNMASKED low 48 bits only: the probe forms its address as
+        // (tag << 48) | RVMODEL_ACCESS_FAULT_ADDRESS, so every a_upper_bits tag
+        // shares the same low half. Masking the compare to 48 bits is what lets
+        // this coverpoint cross with all 8 a_upper_bits bins; comparing the whole
+        // address (or its high bits) would restrict pm_fault to the upper_0000 bin.
+        // Part-selecting an expression is illegal in SystemVerilog, hence the AND.
```
jordancarlin, line 226: Trim this comment down to just what is necessary to understand the coverpoints. For example, no need to explain what is legal SystemVerilog syntax. This whole comment can probably be dropped and just the original "Exception should write xtval with masked version of pointer" comment.

This is a global note and applies to lots of places in this PR. Please clean up the comments to just what is useful to readers later.

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -36,16 +36,26 @@
...
+    // masking is RV64-only, and the spec makes writing UXL=1 clear the
+    // corresponding PMM bits. Only meaningful on a config where UXL is WARL-
+    // writable — UDB_UXLEN_64 means UXLEN is fixed at 64 (implied by Ssu64xl),
+    // so the bin is structurally unreachable there and is compiled out rather
+    // than left as a permanent hole in the covergroup score.
+    `ifndef UDB_UXLEN_64
     uxl_rv32: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "sstatus", "uxl") {
         bins uxl_01 = {2'b01};
     }
+    `endif // UDB_UXLEN_64
```
jordancarlin, line 49: This doesn't seem right. Pointer masking requires RV64, so `UDB_UXLEN_64` will always be defined if we are running this test. If RV32 is also supported, then `UDB_UXLEN_32` will also be defined. Seems like this should be `` `ifdef UDB_UXLEN_32 ``. The other option would be to change this coverpoint to just check that a CSR write setting `mstatus.UXL` to 0 was attempted. We could always do that whether rv32 is supported or not and the correct result would then be derived from the reference model (clear PMM if rv32 is supported, leave it if not supported).

coverpoints/priv/Ssnpm_coverage.svh
```diff
@@ -36,16 +36,26 @@
...
 
     //Main Crosses
     cp_pmlen_masking : cross priv_mode_u, pmm, satp_mode, a_upper_bits, pm_insn;
     cp_pmlen_misaligned_word: cross priv_mode_u, pm_misalign;
     cp_pmm_mxr: cross priv_mode_u, pmm, a_upper_bits, mxr_bit, satp_mode, sw_lw_insn;
     cp_pmm_jalr: cross priv_mode_u, pmm, a_upper_bits, mxr_bit, satp_mode, jalr_insn;
+    `ifndef UDB_UXLEN_64
     cp_pmm_uxl_clear: cross pmm, uxl_rv32;
+    `endif // UDB_UXLEN_64
```
jordancarlin, line 58: Same comment as above.

tests/env/rvtest_trap_handler.h
```diff
@@ -2001,8 +2001,29 @@ data_adj_\__MODE__\()epc:
         LREG    T2, data_bgn_off(T4)                  // check if EPC is in data segment
         LREG    T6, data_seg_siz(T4)
         add     T6, T6, T2
-        bgeu    T3, T6, abort_test                    // EPC beyond all known segments -> abort
-        bltu    T3, T2, abort_test                    // EPC before data segment -> abort
+        bgeu    T3, T6, oos_\__MODE__\()epc           // EPC beyond all known segments
+        bgeu    T3, T2,  adj_\__MODE__\()epc          // EPC inside data segment -> relocate
+                                                      // else EPC is before the data segment -> fall through
```
jordancarlin, line 2006: This can be simplified to just the second branch that checks if EPC is inside the data segment and jumps to the relocation. Both other cases (before and after the data segment) can just fall through to oos_epc.

tests/env/rvtest_trap_handler.h
```diff
@@ -2001,8 +2001,29 @@ data_adj_\__MODE__\()epc:
...
+// --- Out-of-segment xEPC ---
+// A fetch access/page fault at a deliberately bogus target has an xEPC equal to
+// that raw target address, which cannot be expressed as a segment-relative
+// offset. Pointer-masking tests rely on exactly this to show that masking does
+// NOT apply to instruction fetch (JALR through a tagged pointer), and the Sv/PMP
+// execute-permission probes take the same shape whenever their target lands
+// outside every segment. The address is a test-chosen constant, identical on the
+// DUT and the reference model, so recording it raw is deterministic -- and the
+// fetch-fault return path below rewrites xEPC from ra regardless. Any OTHER
+// cause arriving here is a genuine runaway EPC and still aborts.
```
jordancarlin, line 2017: Trim down comment. This is very long and hard to follow right now.

tests/env/rvtest_trap_handler.h
```diff
@@ -2001,8 +2001,29 @@ data_adj_\__MODE__\()epc:
...
+oos_\__MODE__\()epc:
+        csrr    T2, CSR_XCAUSE
+        LI(     T6, CAUSE_FETCH_ACCESS)
+        beq     T2, T6, sv_\__MODE__\()epc            // fetch access fault -> record raw xEPC
+        LI(     T6, CAUSE_FETCH_PAGE_FAULT)
+        beq     T2, T6, sv_\__MODE__\()epc            // fetch page fault -> record raw xEPC
+        LI(     T6, CAUSE_FETCH_GUEST_PAGE_FAULT)
+        beq     T2, T6, sv_\__MODE__\()epc            // fetch guest-page fault -> record raw xEPC
+        j       abort_test                            // runaway EPC -> abort as before
```
jordancarlin, line 2026: No need to compare to what the code used to do in the comments. They should be about what is currently here. Otherwise, the comment loses meaning as soon as this PR is merged.

```suggestion
        j       abort_test                            // runaway EPC -> abort
```

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Top-level generator
+# ---------------------------------------------------------------------------
+
+
+@add_priv_test_generator(
+    "Ssnpm",
+    required_extensions=["Ssnpm", "Zicsr", "S", "U"],
```
jordancarlin, line 1001: No need to list anything other than Ssnpm as required. Ssnpm already requires S, which in turn requires U and Zicsr.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    # Every extension a probe can emit must be in the march string. Other suites
+    # do the same; a local `.option arch` would be the only use of that directive
+    # in the repo and its extension-name vocabulary differs across toolchains.
```
jordancarlin, line 1004: Remove this comment.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    march_extensions=["I", "A", "F", "D", "C", "V", "Zabha", "Zacas", "Zicbom", "Zicbop", "Zicboz"],
```
jordancarlin, line 1005: Does this need Zicfiss too?

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+)
+def make_ssnpm(test_data: TestData) -> list[TestChunk]:
+    """One TestChunk (hence one test file) per satp mode."""
```
jordancarlin, line 1008: The function-level docstring should describe the overall function, not define how splitting is handled.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+so a trapped probe records a distinguishable value rather than stale state.
+
+Why every trap is taken in M-mode
+---------------------------------
+Running U-mode code under an active satp requires ``PTE_U`` on the pages the
+test executes from, but S-mode may never *fetch* from a ``U=1`` page -- an
+S-mode trap handler living in the same identity-mapped superpage would fault
+forever.  Rather than alias the code at a second VA, these tests clear
+``medeleg`` so every trap is taken in M-mode, which is exempt from translation
+entirely.  ``medeleg`` is restored before the test ends.
```
jordancarlin, line 62: The traps need to go to S-mode for certification. We cannot trap to M-mode.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+===== ============================================================ ==============================
+
+A probe that is expected to fault is not a defect: for a given (PMM, satp) only
+some tag patterns survive masking, and both the surviving value and the trap
+record (cause/epc/tval) are part of the signature, so the reference model pins
+down the whole matrix.  Each probe seeds its destination with a sentinel first
+so a trapped probe records a distinguishable value rather than stale state.
+
+Why every trap is taken in M-mode
+---------------------------------
```
jordancarlin, line 56: Remove most of this comment. This looks like leftover context from an AI and should not be in the final test.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    chk: int  # value handed to RVTEST_SIGUPD
+    tmp: int  # scratch
+    tmp2: int  # scratch
+    fp: int  # FP probe destination
+    fp_c: int  # FP destination for Zcd probes (must be f8-f15)
+
+
+def _li(reg: int, val: int) -> str:
+    """LI() accepts arbitrarily wide hex literals, unlike bare `li`."""
+    return f"LI(x{reg}, {hex(val)})"
```
jordancarlin, line 273: What is the point of this function? Seems like it would be better to just directly use `LI(...)` when generating assembly.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    """
+
+    base: int  # untagged, always-valid base address of the probe page
+    a: int  # tagged probe address (base | tag << 48)
+    data: int  # value fed to store-style probes
+    chk: int  # value handed to RVTEST_SIGUPD
+    tmp: int  # scratch
+    tmp2: int  # scratch
+    fp: int  # FP probe destination
+    fp_c: int  # FP destination for Zcd probes (must be f8-f15)
```
jordancarlin, line 268: Seems like it might be simpler to call get_regs when you actually need registers and return them when you are done in each block of code that needs registers rather than creating this whole data structure to pass around with the list of registers.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+def _fixed(instr: str) -> list[str]:
+    """Emit one instruction with a pinned 32-bit encoding.
+
+
+    Extension availability comes from the suite's ``march_extensions`` rather
+    than a local ``.option arch``: no other suite in the repo uses that
+    directive, and its extension-name vocabulary varies between binutils and
+    LLVM releases, so relying on it makes the tests toolchain-sensitive.
+    """
+    return [".option push", ".option norvc", instr, ".option pop"]
```
jordancarlin, line 285: What is the point of this function? Neither the name `_fixed` nor the docstring explain what it is for. The comment just explains why it doesn't use `.option arch`, which seems unrelated.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    """Emit one instruction that must keep its 16-bit encoding."""
+    return [".option push", ".option rvc", instr, ".option pop"]
+
+
+def _guard_open(macro: str | None) -> list[str]:
+    return [f"#ifdef {macro}"] if macro else []
+
+
+def _guard_close(macro: str | None) -> list[str]:
+    return [f"#endif // {macro}"] if macro else []
```
jordancarlin, line 298: Seems like these could probably be inlined.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    than a local ``.option arch``: no other suite in the repo uses that
+    directive, and its extension-name vocabulary varies between binutils and
+    LLVM releases, so relying on it makes the tests toolchain-sensitive.
+    """
+    return [".option push", ".option norvc", instr, ".option pop"]
+
+
+def _compressed(instr: str) -> list[str]:
+    """Emit one instruction that must keep its 16-bit encoding."""
+    return [".option push", ".option rvc", instr, ".option pop"]
```
jordancarlin, line 290: If you are only calling this with `c.**` instructions, they will already be compressed. Why is this needed?

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+# Zicfiss shadow-stack AMOs are DISABLED. `ssamoswap` is part of the `pm_insn`
+# coverage model, but it is not currently usable as a portable probe:
+# Re-enabling therefore buys 24 of 9600 cross bins (0.25% of cp_pmlen_masking)
+# while costing Ssnpm CI coverage on two of four DUTs. Restore by putting the
+# two entries back once the coverage front-end parses the operands and the DUT
+# bugs are fixed:
+#   [("ssamoswap.w", "lw", 0x2), ("ssamoswap.d", "ld", 0x3)]
+# `_probe_zicfiss` below is kept intact and still emits the `.insn` form, which
+# needs no experimental extension enabled in the assembler.
+_ZICFISS_AMOS: list[tuple[str, str, int]] = []
```
jordancarlin, line 207: What is going on here? Why are these disabled? If they fail on certain DUTs, we should open issues for those DUTs, add the test to the exclusions, and still keep the testcase.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+        "add  a0, a0, a1",
+        "ld   t0, 0(a0)",
+        "li   t1, PTE_U",
+        "or   t0, t0, t1",
+        "sd   t0, 0(a0)",
+    ]
+
+
+# ---------------------------------------------------------------------------
+# M-mode configuration helpers
```
jordancarlin, line 372: This needs to happen in S-mode through T-SBI. Not through M-mode.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+
+
+def _set_mxr(enable: bool, tmp: int) -> list[str]:
+    """MXR gates pointer masking off entirely when set.
+
+    Written through *sstatus*, not mstatus, even though this runs in M-mode and
+    both views share the bit. The coverage front-end reconstructs CSR state from
+    the RVVI write-back stream keyed by CSR address, and `mxr_bit` samples
+    sstatus -- an mstatus write updates only the mstatus entry, leaving the
+    sstatus shadow reading 0 and the MXR=1 bin permanently unreachable.
```
jordancarlin, line 396: Drop this comment.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    ]
+
+
+def _mmode_prelude(mode: str, regs: Regs) -> list[str]:
+    """Everything that has to be true before the first U-mode probe."""
+    cbo_fields = (
+        (0b11 << _SENVCFG_CBIE_SHIFT) | (1 << _SENVCFG_CBCFE_SHIFT) | (1 << _SENVCFG_CBZE_SHIFT) | _ENVCFG_SSE_BIT
+    )
+    lines = [
+        "RVTEST_GOTO_MMODE",
```
jordancarlin, line 412: Since this is a new test, let's not use any of the legacy macros. Switch everything over to T-SBI including switching modes, accessing CSRs, etc.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
```diff
@@ -0,0 +1,1026 @@
...
+    # the allocatable pool and may well be the register holding the mask.
+    lines += [
+        "",
+        "# Take every trap in M-mode. U-mode code running under an active satp needs",
+        "# PTE_U on the identity superpage it executes from, and S-mode may not fetch",
+        "# from a U=1 page -- an S-mode handler in that same superpage would trap",
+        "# forever. M-mode is exempt from translation, so delegation is switched off",
+        "# for the duration of the test and restored at the end.",
+        f"csrr x{regs.tmp2}, medeleg      # stash the framework's delegation mask",
+        "csrw medeleg, zero",
```
jordancarlin, line 446: This is not going to work. We need to trap to S-mode for certification, not M-mode.

jordancarlin (changes requested):
I haven't finished reviewing the Python test generator, but here is an initial round of comments. Once these are addressed, I will review the rest.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
Ammarahwakeel: When I ran this instruction in May, the toolchain was throwing an error, so I commented it out. I'm not certain now if that was due to my toolchain version at the time or a missing zacas flag. I've  re-run amocas.q on my current toolchain and it executes without any errors so I agree it should be uncommented.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
Ammarahwakeel: This change was needed because QuestaSim (vlog 2023.4_4) fails to parse a part-select directly on a parenthesized expression in a coverpoint declaration.

The original line:
a_upper_bits: coverpoint (ins.current.rs1_val + ins.current.imm)[63:48] {
produced:
** at RISCV_coverage_pmm_coverpoints.svh(193): near "[": syntax error, unexpected '[', expecting ';' or '{'.
** at RISCV_coverage_pmm_coverpoints.svh(193): near "6": syntax error, unexpected INTEGER NUMBER.

so I used >>48 expression, it compiled cleanly on Questa, while producing the same top-16-bits value.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
Ammarahwakeel: Same QuestaSim compilation error

coverpoints/priv/Ssnpm_coverage.svh
Ammarahwakeel: I am not sure about sir Umer's thinking on this, but for `ifndef UDB_UXLEN_64 guard, I thought it is according to this spec line :

In particular, an implementation may make UXL be a read-only
field whose value always ensures that UXLEN=MXLEN or UXLEN=SXLEN.

That if MXLEN OR SXLEN=64  so we will put a guard that UXL != 64. Though in work/reports/Ssnpm_report.txt, I could not find this coverpoint or cross at all and I think that's exactly due to the reasoning you mentioned

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
Ammarahwakeel: So in my last PR#1827, I was trapping in S mode. but it was throwing error for non-canonical addresses in SVX cases. 

e.g : IF I am setting upper 48th bit ( upper_pattern = 0x0001 ), with PMM = 01 (PMLEN=7), although pointer masking does work but bit 48 remains set, so for Sv39 the requirement that bits 63:39 equal bit 38 is violated. 

That's why we changed it to M-mode trapping. But ofc since Ssnpm doesnot require M mode, we should come up with other solutions. Like maybe using those upper_pattern which can be masked with PMM=7 /16 in respective cases. But in test plan we agreed to test every possible combination and record any traps.

coverpoints/general/RISCV_coverage_pmm_coverpoints.svh
davidharrishmc: Please push fix and resolve conversation.

generators/testgen/src/testgen/priv/extensions/Ssnpm.py
davidharrishmc: I don’t understand.  When bit 48 is 1 and PMLEN=7, then there should be a trap to S-mode.  The trap handler should
log that trap and continue to the next test bin.

What does “throwing error” mean?  Why is there a problem?

coverpoints/priv/Ssnpm_coverage.svh
davidharrishmc: We can’t guarantee that sstatus.UXL is writable.  All we can do is attempt to write it to 01, and then read back senvcfg,PMM.
So we can’t guarantee this coverpoint is possible to hit in present form.  We need a coverpoint with a csrr to sstatus with 01 in
the UXL field.

The coverpoint can’t be gated by UDB_UXLEN_64 because it is tested on any system that supports Ssnpm.

UmerShahidengr:
PR #2049 superseded this PR, so I am closing this one.

PR #2038, opened 2026-07-31
Edited CRD for clarity and to remove "Custom"

kdockser:
Most commercial RTL features implementation use M-mode 1.13 as a base and add custom features which is allowed (even expected). It is common that these implementations need to make changes to Open-SBI. Calling these custom doesn't differentiate as this type of "custom" is the norm. It is less confusing to call the implementation specific.

The previous custom that I was referring to was something that wildly deviates from M-mode 1.13 such that the developer would need to write their own T-SEE.

### DIFF
--- a/docs/crd/src/crd_common.adoc
+++ b/docs/crd/src/crd_common.adoc
@@ -1,17 +1,12 @@
 === What's a CRD?
 
-Certification Requirements Documents (CRDs) list requirements an implementation MUST meet
-to obtain an associated RVI (RISC-V International) certificate.
+Certification Requirements Documents (CRDs) list test requirements an implementation MUST pass
+as one of the requirements to obtain a RISC-V International certificate.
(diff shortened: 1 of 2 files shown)


jordancarlin:
@kdockser Unfortunately, this seems to have been based off an old version of the repo and an outdated CRD. It is missing all of the changes from the last 2 months. Can you reapply these changes against the latest `act4` branch?

kdockser:
> Kenneth Dockser (@kdockser) Unfortunately, this seems to have been based off an old version of the repo and an outdated CRD. It is missing all of the changes from the last 2 months. Can you reapply these changes against the latest `act4` branch?

Arrrrgh!
Thanks for pointing this out Jordan. I pulled the origin to get the latest version before adding my new edits. I'll check to see if I'm pointing to an old branch. In any case, yes, I'll apply these to the act4 branch.

jordancarlin:
Closing this now that #2041 was opened instead.

PR #2040, opened 2026-07-31
Parse Sail CLINT and SIG bases from sail.json

### LINKED ISSUE #1630: ACT4 assumes that CLINT for sail is at 0x02000000
While this in general should not affect the test flow, if the DUTs start address is aliasing this address, the test cannot be run using sail simulator.

This is only a limitation on the riscv-arch-test part, as riscv-sail already has an option to change the address of Clint.

jordancarlin:
Parse the CLINT and SIG addresses from `sail.json` instead of hardcoding them in `sail_macros.h`. This is required to support users who move the CLINT to a different memory address.

Also add a check to ensure the CLINT and SIG are enabled in sail.json and provide guidance on what to do if your DUT does not support a CLINT.

Fixes #1630

### DIFF
--- a/framework/src/act/build_plan.py
+++ b/framework/src/act/build_plan.py
@@ -12,6 +12,8 @@ import importlib.resources
 from collections import defaultdict
 from pathlib import Path
 
+import pyjson5
+
 from act.build_types import BuildTask, PythonAction, SubprocessAction, SymlinkAction
 from act.config import CompilerType, Config, CoverageSimulator, RefModelType, spike_isa_string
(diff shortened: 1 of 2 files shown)


davidharrishmc (commented):
Should this be mentioned in CTP or setup guide too?

jordancarlin:
I think the setup guide is in need of a bit of an overhaul. There are quite a few subtle things that are not really captured there. I'd rather defer adding this to a follow up PR that tackles it more holistically.

davidharrishmc (approved):

PR #2041, opened 2026-07-31
Updates to rva23-crd.adoc

kdockser:
Clarifications.


docs/crd/src/rva23_crd.adoc
```diff
@@ -25,71 +25,105 @@ a| * Initial draft
...
+implemented. Furthermore, any optional extensions as specified in the profile may be implemented. Any optional portions
+of the base ISA and these extensions may be implemented, but if they are
+implemented, they must be implemented in their entirety.
 
-NOTE: The profiles already explain what is required and what is optional. Restating the requirements here can lead to confusion.
+There is no prohibition against implementing other RISC-V ratified base ISAs or
+extensions, custom extensions, or even non-conforming extensions as
+long as they don't interfere with the proper functioning of the RVA23 mandatory
+and implemented optional extensions. However, these other base ISAs and extensions will not be tested as
+part of RVA23 Certification.
```
davidharrishmc, line 48: Why drop mention of Ssstrict?

docs/crd/src/rva23_crd.adoc
```diff
@@ -25,71 +25,105 @@ a| * Initial draft
...
+
+This `ECALL`'s trap handler executes the
+necessary M-mode code need to complete the requested action. One such need for a
+T-SBI call is necessary to fulfill a requirement of RVA23's defined Sspm
+extension which requires an execution environment to provide a means to select
+PMLEN. Another need is for the tests to be able to request that specific traps are
+delegated away from M- mode.
+
+NOTE: There is no requirement that any traps can be delegated. But if they can
+be, the SEE must provide a means to request that they are and a means to
```
davidharrishmc: T-SEE ?

docs/crd/src/rva23_crd.adoc
```diff
@@ -25,71 +25,105 @@ a| * Initial draft
...
+intention that M-mode code be portable. Therefore, RVA23 Certification does
+not test M-mode instructions or CSRs. Any required interaction with M-mode is
+abstracted away by a Test Supervisor Binary Interface (T-SBI) and a Test
+Supervisor Execution Environment (T-SEE), which are analogous to the RISC-V
+Supervisor Binary Interface and OpenSBI software respectively;
+however, they are developed specifically for certification. The T-SEE is also
+responsible for invoking the implementation's
+boot code as well as setting up the initial M-mode-controlled environment for
+the certification tests. This includes putting the processor into a known
+initial state (for some state, it is sufficient that it holds _any_ legal value).
```
davidharrishmc: This is vague.  How does the developer or test author know which state can hold any legal value?  Should we just drop it, or make it more specific?

docs/crd/src/rva23_crd.adoc
```diff
@@ -25,71 +25,105 @@ a| * Initial draft
...
+
+As part of the certification test collateral, a reference implementation of
+the T-SEE is provided along with a full
+specification of the T-SBI. Just like with OpenSBI,
+it is up to the designer to modify the T-SEE, as necessary, for their
+implementation. The T-SBI specification is especially useful
+for those who have created an M-mode that differs substantially from M-mode 1.13
+as defined in the RISC-V Privileged Architecture.
+
+When the T-SEE receives a trap from M-mode, it invokes the appropriate handler.
```
davidharrishmc: from -> to?

docs/crd/src/rva23_crd.adoc
```diff
@@ -25,71 +25,105 @@ a| * Initial draft
...
 
-Testing M-mode trap behavior is OUT-OF-SCOPE for RVA23 certification with a few exceptions listed below.  As such, the tests delegate almost all traps to S-mode, and do not test M-mode interrupts (MEI/MTI/MSI).  Exceptions are:
+M-mode trap behavior is out of scope for RVA23, with a few exceptions listed
+below.  As such, the certification tests delegate almost all traps to S-mode,
+and do not test M-mode interrupts (MEI/MTI/MSI).  Exceptions are:
 
-* `ecall` from S-mode must not be delegated because the T-SBI must trap to M-mode to handle these calls.
-* Any trap cause that should be handled invisibly by the firmware must not be delegated.
-* Hardware may not support delegating certain traps to S-mode.
+* An `ECALL` from S-mode must not be delegated because the T-SBI must trap to M-mode to handle these calls.
```
davidharrishmc: Instruction names are lower case elsewhere in this document.  Keep it the same for consistency, or change it everywhere.

davidharrishmc (changes requested):

davidharrishmc:
Looks good.  Minor editorial feedback.  I think we are aligned.

docs/crd/src/rva23_crd.adoc
kdockser: Ssstrict is defined in RVA23. I didn't see a need for us to include it here as it might cause confusion.

docs/crd/src/rva23_crd.adoc
kdockser: I agree, it is vague. Partially because I wasn't sure what initial state needs to be specific values and what state is anything legal (i.e., not Xs). 
I was thinking that the setting of the initial state would be documented within the reference T-SEE itself. 
What are your thoughts on this?

docs/crd/src/rva23_crd.adoc
kdockser: Thanks for catching this. I was consistent with other documents, but not this document.
I'm fine with this being changed to lower case.

docs/crd/src/rva23_crd.adoc
kdockser: I agree,"to" makes sense.

docs/crd/src/rva23_crd.adoc
kdockser: Si, T-SEE

docs/crd/src/rva23_crd.adoc
jordancarlin: Agreed that this should be documented either in the description of the T-SEE (which is in the CTP) or in the reference T-SEE itself. I would be in favor of dropping it here to avoid the inevitable questions since it is vague in its current form.

docs/crd/src/rva23_crd.adoc
kdockser: Fixed in next commit.

docs/crd/src/rva23_crd.adoc
kdockser: Fixed in next commit.

docs/crd/src/rva23_crd.adoc
kdockser: Fixed in next commit.

docs/crd/src/rva23_crd.adoc
```diff
@@ -142,18 +174,35 @@ The following M-mode CSRs/fields are OUT-OF-SCOPE and not tested:
...
 
 === Interrupts
+// KAD OUT-OF-SCOPE is a term that we use when something is in scope for RVA23
+// but we are not currently testing it. It should not be used here.
+// Everything in RVA23 is in-scope. We should only list the subset that we are not testing
 
-All supervisor interrupts (SEI, STI, SSI, and LCOFI) are IN-SCOPE for testing. Hypervisor interrupts VSEI, VSTI, VSSI, and SGEI are also IN-SCOPE. Machine interrupts (MEI, MTI, MSI) are OUT-OF-SCOPE for RVA23.
+All supervisor interrupts (SEI, STI, SSI, and LCOFI) are tested.
+Hypervisor interrupts VSEI, VSTI, VSSI, and SGEI are also tested. Machine
+interrupts (MEI, MTI, MSI) are out of scope for RVA23.
```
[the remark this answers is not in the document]
kdockser: Fixed in next commit.

docs/crd/src/rva23_crd.adoc
```diff
@@ -25,71 +25,107 @@ a| * Initial draft
 | July 27, 2026
 | 0.2
 a| * Do not test traps into M-mode except ecall from S-mode
+| July 31, 2026
+| 0.3
+a| Removed "custom M-mode" made clarifications
```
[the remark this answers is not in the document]
kdockser: Fixed in next commit.

kdockser (commented):
A comment

docs/crd/src/rva23_crd.adoc
kdockser: Added NOTE: This document does not restate individual profile constraints such as Ssstrict.
Those requirements are defined normatively in the RVA23 Profiles and are unchanged for certification.

docs/crd/src/rva23_crd.adoc
```diff
@@ -102,21 +138,17 @@ RVA23 certification requires that the T-SBI provide methods of emulating every w
...
-|===
 
+////
 The following CSRs and memory-mapped I/O registers are only read during RVA23 certification.
-
-[options=header]
-[cols="1,5"]
-|===
-| CSR | Notes
+////
```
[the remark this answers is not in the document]
kdockser: Removed

.github/workflows/release-docs.yml
```diff
@@ -127,6 +127,7 @@ jobs:
           fi
 
           gh release create "$tag_name" \
+            --repo "$GITHUB_REPOSITORY" \
```
jordancarlin: Revert this change to the GitHub workflow.

docs/crd/src/rva23_crd.adoc
```diff
@@ -24,72 +24,110 @@ History of documentation changes that eventually lead to releases.
...
 
 This document describes the test requirements for a processor to be eligible for a RVA23 certificate.
 
-The RVA23 processor certificate is based on the https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva23s64_profile[RVA23S64] and RVA23U64 profiles. To
-qualify for a certificate, the processor must implement the RV64I base instruction set and all of the mandatory
-extensions of the profile.  It may implement any of the optional features of
-the profile.  It may implement ratified extensions that are not part of the profile. It may implement other unratified extensions, but must not implement
-unratified extensions outside the custom encoding spaces if the optional Ssstrict extension is implemented. Extensions outside the profile are not tested during
-certification.
+The RVA23 processor certification requires compliance with the https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva23_profiles[RVA23 Profiles], which is comprised of the RVA23S64 and RVA23U64 profiles. As described in these profile specifications,
```
jordancarlin, line 38: I believe this is the official terminology for the overall group of RVA23 profiles.
```suggestion
The RVA23 processor certification requires compliance with the https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva23_profiles[RVA23 Profile Release], which is comprised of the RVA23S64 and RVA23U64 profiles. As described in these profile specifications,
```

jordancarlin, line 38: Should we be more explicit here that certification requires the RVA23S64 profile and not just the RVA23U64 profile? Also, the RVA23S64 profile includes everything in the RVA23U64 profile, so maybe we are better off not even mentioning RVA23U64.

docs/crd/src/rva23_crd.adoc
```diff
@@ -24,72 +24,110 @@ History of documentation changes that eventually lead to releases.
...
-
-Custom M-mode is a supervisor execution environment that deviates from the M-mode specification.  For example, it might have a bug that affects M-mode behavior but is invisible from S-mode.  The firmware could work around this bug by not running M-mode instructions affected by the bug.  Alternatively, custom M-mode might have some security features or custom CSRs that break compatibility with ratified M-mode.
-
-RVA23 does not place any specific requirements on M-mode implementations as it is not required that M-mode code is portable for compliant designs. Instead, in order to be able to run portable code for modes below M-mode, the processor must interface with a Supervisor Execution Environment (SEE).  The RISC-V SBI specification provides an example of a SEE.  Certification tests do not exercise the RISC-V SBI, but provide a simpler Test Supervisor Binary Interface (T-SBI) to access the execution environment.
+NOTE: We encourage those who have implemented other base ISAs or extensions to
+run the appropriate RISC-V ACTs to help ensure compliance.
```
jordancarlin: Might be worth adding a link to the `riscv-arch-test` repo in this note.

docs/crd/src/rva23_crd.adoc
```diff
@@ -24,72 +24,110 @@ History of documentation changes that eventually lead to releases.
...
+for those who have created an M-mode that differs substantially from M-mode 1.13
+as defined in the RISC-V Privileged Architecture.
+
+When the T-SEE receives a trap to M-mode, it invokes the appropriate handler.
+When it is necessary to execute M-mode instructions or access an M-mode
+CSR in order to properly set up the environment for lower privilege tests, a
+T-SBI-defined call is made by placing commands in registers and performing an `ecall`.
+
+This `ecall`'s trap handler executes the
+necessary M-mode code need to complete the requested action. One such need for a
```
jordancarlin:
```suggestion
This `ecall`'s trap handler executes the
necessary M-mode code needed to complete the requested action. One such need for a
```

docs/crd/src/rva23_crd.adoc
```diff
@@ -24,72 +24,110 @@ History of documentation changes that eventually lead to releases.
...
+NOTE: There is no requirement that any traps can be delegated. But if they can
+be, the T-SEE must provide a means to request that they are and a means to
+determine which requested traps have been delegated.
 
-The execution environment must provide a Test Supervisor Binary Interface (T-SBI) that allows S-mode software to make requests of the execution environment.  These requests are made by placing commands in registers and performing an `ECALL`. See the Certification Test Plan (CTP) for details. The certification suite provides a reference T-SBI implementation that runs on standard M-mode, but custom M-mode that differs significantly from standard M-mode might have to provide its own T-SBI implementation.
+=== Related Specifications
 
-The T-SBI also encompasses boot code and an M-mode trap handler.  Both might need to be modified by the user to accommodate custom M-mode behavior.
+* https://riscv.github.io/riscv-isa-manual/snapshot/unprivileged/index.html[Unprivileged ISA Specification] - Latest version
+* https://riscv.github.io/riscv-isa-manual/snapshot/privileged/index.html[Privileged ISA Specification] - Latest version
```
jordancarlin: These links are outdated and no longer work. The correct links seem to be
- https://riscv.github.io/riscv-isa-manual/snapshot/spec/#vol:unpriv
- https://riscv.github.io/riscv-isa-manual/snapshot/spec/#vol:priv

docs/crd/src/rva23_crd.adoc
```diff
@@ -142,18 +172,35 @@ The following M-mode CSRs/fields are OUT-OF-SCOPE and not tested:
...
 
 === Interrupts
+// KAD OUT-OF-SCOPE is a term that we use when something is in scope for RVA23
+// but we are not currently testing it. It should not be used here.
+// Everything in RVA23 is in-scope. We should only list the subset that we are not testing
 
-All supervisor interrupts (SEI, STI, SSI, and LCOFI) are IN-SCOPE for testing. Hypervisor interrupts VSEI, VSTI, VSSI, and SGEI are also IN-SCOPE. Machine interrupts (MEI, MTI, MSI) are OUT-OF-SCOPE for RVA23.
+All supervisor interrupts (SEI, STI, SSI, and LCOFI) are tested.
+Hypervisor interrupts VSEI, VSTI, VSSI, and SGEI are also tested. Machine
+interrupts are subjected to limited testing as described above.
```
jordancarlin, line 199: I don't think we are testing Machine interrupts at all. Should this sentence be updated to say that "Machine interrupts are not tested"?

docs/crd/src/rva23_crd.adoc
```diff
@@ -142,18 +172,35 @@ The following M-mode CSRs/fields are OUT-OF-SCOPE and not tested:
 |===
 
 === Interrupts
+// KAD OUT-OF-SCOPE is a term that we use when something is in scope for RVA23
+// but we are not currently testing it. It should not be used here.
+// Everything in RVA23 is in-scope. We should only list the subset that we are not testing
```
jordancarlin: Can this comment be deleted? Looks like you have already fixed the concern raised here.

jordancarlin (dismissed):
A few minor comments.

docs/crd/src/rva23_crd.adoc
kdockser: You lost me on this, how could this be read as _just_ RVA23U64?

RVA23S64 is a superset of RVA23U64's _mandatory_ baseline only. It does not include the optional extensions of RVA23U64. If we only tested RVA23U64, we would need to exclude the 'U63 options. We need to mention both.

kdockser: The title of the specification is "RVA23 Profiles".  This makes sense as there are two profiles specified.
Where does "Profile Release" come from?

jordancarlin: The RVA23S64 profile says the following:

> RVA23S64 has the same unprivileged options as RVA23U64.

.github/workflows/release-docs.yml
kdockser: I wondered about this. I'm not sure where the change came from.
Is this just a matter of deleting line 130?

jordancarlin: Yes.

docs/crd/src/rva23_crd.adoc
kdockser: That's interesting. I cannot find this in the latest release of the the RVA23 Profiles. Where do you see it?

Also, RVA23S64 Profile states:
> The RVA23S64 profile specifies the ISA features available to a supervisor-mode execution
> environment in 64-bit applications processors.

And, RVA23U64 Profile states:
> The RVA23U64 profile specifies the ISA features available to user-mode execution environments in
> 64-bit applications processor

You can always execute unprivileged instructions in supervisor mode, so it makes sense that RVA23S64 includes "all the mandatory unprivileged extensions in RVA23U64". However, we need both User Mode _and_ Supervisor Mode.

docs/crd/src/rva23_crd.adoc
kdockser: I will update these.
Do we know if these will remain working?

docs/crd/src/rva23_crd.adoc
kdockser: It might be...
I'll add a link.

.github/workflows/release-docs.yml
kdockser: I will make the fix.

docs/crd/src/rva23_crd.adoc
jordancarlin: I got it from https://riscv.github.io/riscv-isa-manual/snapshot/spec/#_rva23s64_optional_extensions.

Good point about user vs supervisor mode. I'm fine with mentioning both of them then. How about something like
> The RVA23 processor certification requires compliance with the RVA23S64 and RVA23U64 profiles.

That way we cut out any possible ambiguity with the more general term "RVA23 Profile".

docs/crd/src/rva23_crd.adoc
jordancarlin: That's probably a @wmat question. Let's use these for now and we can make sure that the version of the CRD that we actually publish after full CSC approval has more permanent links.

docs/crd/src/rva23_crd.adoc
kdockser: It can be, and it will. I just put this in to explain the change.
In hindsight, this would have been better to be placed in the Commit notes.

docs/crd/src/rva23_crd.adoc
kdockser: When we said they are not tested, copilot called us out. It pointed out the text above

> M-mode trap behavior is out of scope for RVA23, with a few exceptions listed below.

jordancarlin: I think Copilot is getting confused between traps and interrupts. There is some M-mode TRAP behavior that is in scope. There is no M-mode INTERRUPT behavior that is in scope.

docs/crd/src/rva23_crd.adoc
kdockser: Added

docs/crd/src/rva23_crd.adoc
kdockser: Updated to the latest "latest" links.

docs/crd/src/rva23_crd.adoc
kdockser: I replaced all OUT-OF-SCOPE comments with out of scope as they are truly out of the scope of RVA23.
The capitalized and hyphenated version was defined (elsewhere) to mean that it is in scope for RVA23 but not going to be tested by the certification test; we need a better term for any such cases.

docs/crd/src/rva23_crd.adoc
jordancarlin: Looks like this one still needs to be resolved. It still says machine interrupt are subjected to limited testing, but it should say that machine interrupts are not tested.

docs/crd/src/rva23_crd.adoc
```diff
@@ -99,30 +138,43 @@ RVA23 certification requires that the T-SBI provide methods of emulating every w
...
+writable fields must be controllable to allow testing of the behavior of lower
+privilege modes when certain features are disabled.
+| `mseccfg` | SSEED and USEED must be controllable if Zkr is supported;
+otherwise they are read-only zero.  PMM, MLPE, RLB, MMWP, MML are out of scope
+ because
+they do not relate to RVA23 features.  Note `mseccfg.PMM` affects Smmpm, which
+is not part of RVA23.
+| `mstateen` | SE0, ENVCFG, CONTEXT may be controllable to allow testing of the
+behavior of lower privilege modes when certain features are disabled. Other
+fields are our of scope because they do not relate to RVA23 features.
```
jordancarlin, line 159:
```suggestion
fields are out of scope because they do not relate to RVA23 features.
```

docs/crd/src/rva23_crd.adoc
```diff
@@ -99,30 +138,43 @@ RVA23 certification requires that the T-SBI provide methods of emulating every w
...
-[options=header]
-[cols="1,5"]
-|===
-| CSR | Notes
 | `mhartid` | Must be readable to influence the behavior of multicore tests.
 | `mtime` | Must be readable to find the current time to set timer interrupts.
 |===
 
-The following M-mode CSRs/fields are OUT-OF-SCOPE and not tested:
+The Certification Test Environment does not need to access the equivalent of the  following M-mode CSRs/fields:
```
jordancarlin, line 166:
```suggestion
The Certification Test Environment does not need to access the equivalent of the following M-mode CSRs/fields:
```

docs/crd/src/rva23_crd.adoc
```diff
@@ -24,72 +24,111 @@ History of documentation changes that eventually lead to releases.
...
+When it is necessary to execute M-mode instructions or access an M-mode
+CSR in order to properly set up the environment for lower privilege tests, a
+T-SBI-defined call is made by placing commands in registers and performing an `ecall`.
+
+This `ecall`'s trap handler executes the
+necessary M-mode code needed to complete the requested action. One such need for a
+T-SBI call is necessary to fulfill a requirement of RVA23's defined Sspm
+extension which requires an execution environment to provide a means to select
+PMLEN. Another need is for the tests to be able to request that specific traps are
+delegated away from M- mode.
```
jordancarlin, line 91:
```suggestion
delegated away from M-mode.
```

docs/crd/src/rva23_crd.adoc
davidharrishmc: Will merge and make this change in the next pass.

docs/crd/src/rva23_crd.adoc
davidharrishmc: Will make this change in the next pass.

davidharrishmc (approved):
Approved to merge.  Remaining discussion can be addressed in minor edits in today's meeting.

davidharrishmc (approved):

PR #2047, opened 2026-08-02
Handle IALIGN=32 xEPC advancement

### LINKED ISSUE #2046: Trap handler assumes C extension support when advancing `xepc` after an exception
**Describe the bug**
The trap handler advances `mepc` by 2 when bits[1:0] of the instruction != 0b11. In`ExceptionsSm.S`, specifically `ExceptionsSm_cg_cp_illegal_instruction_illegal_0x00000000`, this attempts to advance `mepc` by 2 even if the C extension is not supported. From the [spec](https://docs.riscv.org/reference/isa/v20260120/priv/machine.html#2-1-1-14-machine-exception-program-counter-mepc-register): "On implementations that support only IALIGN=32, the two low bits (mepc[1:0]) are always zero." If C is not supported, the +2 is ignored and `mepc` is loaded with the same address that caused the trap, resulting in a trap loop when `mret` is called. For configurations without the C extension, `mepc` should always advance by a multiple of 4 bytes.

In this case, the current implementation interprets the 0x00000000 instruction as two compressed 16-byte instructions.

**To Reproduce**

1. On commit d261a799762ffb94eea486d4adfe7632006d2b4a
2. With `cv32e20` as a base, in `config/cores/cve2/cv32e20/cv32e20.yaml` remove `C` and `Zca` from `implemented extensions` and the `MUTABLE_MISA_C` from `params`. In `sail.json` of the same folder, set `"Zca": {"supported": true}` to false.
3. From `riscv-arch-test/` run:
```
CONFIG_FILES=config/cores/cve2/cv32e20/test_config.yaml DEBUG=True EXCLUDE_EXTENSIONS=I,Zicsr,M,Zifencei make --jobs $(nproc)
```

    begin_signature: 0x128d0
    end_signature: 0x21660
    
    RVCP-SUMMARY: TEST FAILED - Test File "ExceptionsSm-00.S"
    
    RVCP: Trap signature overflow in sig mode. Increase TRAP_SIGUPD_COUNT for this test.
    
    FAILURE: 1 (0x00000001)

**Additional context**
From the trace, we see the +2 being ignored between `adj_Mepc_rtn+40` and `adj_Mepc_rtn+44`, and `mret` returning to the trapping address:
```
mem[X,0x000006268] -> 0x0000
[7115] [M]: 0x00006268 (0x0000) c.illegal 0x0                                       ExceptionsSm_cg_cp_illegal_instruction_illegal_0x00000000+0
handling exc#illegal-instruction at priv M with tval 0x00000000
CSR mstatus (0x300) <- 0x00001800
CSR mstatush (0x310) <- 0x00000000
CSR mcause (0x342) <- 0x00000002
CSR mtval (0x343) <- 0x00000000
CSR mepc (0x341) <- 0x00006268
...
mem[X,0x000007114] -> 0x0433
mem[X,0x000007116] -> 0x0074
[7251] [M]: 0x00007114 (0x00740433) add x8, x8, x7                                  adj_Mepc_rtn+40
x8 <- 0x0000626A

clint mtime 0x0000000000000E2A
mem[X,0x000007118] -> 0x1073
mem[X,0x00000711A] -> 0x3414
[7252] [M]: 0x00007118 (0x34141073) csrrw x0, mepc, x8                              adj_Mepc_rtn+44
CSR mepc (0x341) <- 0x00006268
...
clint mtime 0x0000000000000E38
mem[X,0x000007170] -> 0x0073
mem[X,0x000007172] -> 0x3020
[7280] [M]: 0x00007170 (0x30200073) mret                                            resto_Mrtn+28
CSR mstatus (0x300) <- 0x00001880
CSR mstatush (0x310) <- 0x00000000
ret-ing from M to M

mem[X,0x000006268] -> 0x0000
[7281] [M]: 0x00006268 (0x0000) c.illegal 0x0                                       ExceptionsSm_cg_cp_illegal_instruction_illegal_0x00000000+0
handling exc#illegal-instruction at priv M with tval 0x00000000
CSR mstatus (0x300) <- 0x00001800
CSR mstatush (0x310) <- 0x00000000
CSR mcause (0x342) <- 0x00000002
CSR mtval (0x343) <- 0x00000000
CSR mepc (0x341) <- 0x00006268
```
The instruction width is determined in `riscv-arch-test/tests/env/rvtest_trap_handler.h`. The function is `adj_\__MODE__\()epc_rtn` and the addition is calculated at line `2057`. In the following snippet, T3 = xEPC and T2 = low halfword of the instruction at xEPC:
```
        andi    T2, T2, 3                            // bits[1:0]
        addi    T2, T2, 1                            // ==4 iff bits were 0b11
        srli    T2, T2, 2                            // 1 if 32-bit, 0 if compressed
        addi    T2, T2, 1                            // 2 (32-bit) or 1 (compressed)
        slli    T2, T2, 1                            // advance = 4 or 2
        add     T3, T3, T2                           // next instruction (raw xEPC + width)
        csrw    CSR_XEPC, T3
```

Following a similar pattern from elsewhere in this file, a possible fix is to check if C is supported before checking bits[1:0] of the instruction to determine how to advance `mepc`:

```
        csrr    T6, CSR_MISA                         // check if C extension is supported
        slli    T6, T6, UDB_MXLEN-2-1                // shift MISA.C bit [2] into msb
        bltz    T6, 1f                               // C supported, check if compressed
        li      T2, 4                                // else, advance = 4
        j       2f
1:      andi    T2, T2, 3                            // bits[1:0]
        addi    T2, T2, 1                            // ==4 iff bits were 0b11
        srli    T2, T2, 2                            // 1 if 32-bit, 0 if compressed
        addi    T2, T2, 1                            // 2 (32-bit) or 1 (compressed)
        slli    T2, T2, 1                            // advance = 4 or 2
2:      add     T3, T3, T2                           // next instruction (raw xEPC + width)
        csrw    CSR_XEPC, T3
```
This resolves the trap loop for my DUT config and the modified `cv32e20` config. It also keeps proper behavior when the `cv32e20` config is reverted to its default, and I've checked that the behavior remains the same for `cvw-rv32imc`.

This is assuming an RV32I + Zicsr DUT configuration is supported by the ACTs, without the C extension. If I've misunderstood anything I'd appreciate any clarification.

jordancarlin:
Fixes #2046.

The trap handler previously inferred instruction width from bits [1:0] of the trapping instruction whenever advancing xEPC. On configurations without compressed instruction support, xEPC must advance by 4 bytes because IALIGN=32 masks the low two bits of mepc/sepc.

This changes the width detection to use the 2-byte path only when `ZCA_SUPPORTED` is defined; otherwise xEPC advances by 4 bytes.

Add a `spike-rv32g` config without compressed instruction support so trap-handler regressions on IALIGN=32 configurations can be caught.

### DIFF
--- a/config/spike/spike-rv32g/link.ld
+++ b/config/spike/spike-rv32g/link.ld
@@ -1,0 +1,1 @@
+../spike-rv64-max/link.ld
--- a/config/spike/spike-rv32g/run_cmd.txt
+++ b/config/spike/spike-rv32g/run_cmd.txt
@@ -1,0 +1,1 @@
+spike --instructions=100000000 {debug:-l --log-commits --log=__TRACEFILE__} --isa=rv32imafd_zicclsm_zicsr_zifencei_zicntr_zaamo_zalrsc
--- a/config/spike/spike-rv32g/rvmodel_macros.h
+++ b/config/spike/spike-rv32g/rvmodel_macros.h
@@ -1,0 +1,1 @@
+../spike-RVI20U64/rvmodel_macros.h
--- a/config/spike/spike-rv32g/sail.json
+++ b/config/spike/spike-rv32g/sail.json
@@ -1,0 +1,845 @@
+{
+  "$schema": "/opt/riscv/share/sail-riscv/sail_riscv_config_schema.json",
+  "base": {
+    "xlen": 32,
+    // Whether the base ISA is E, in which case accessing
+    // x16-31 is reserved and in this implementation will
+    // raise an illegal instruction exception.
+    "E": false,
+    "writable_misa": false,
+    "writable_fiom": true,
+    // The top 29 bits in this value control whether the corresponding
+    // HPM counters (hpmcounter31 .. hpmcounter3) are supported
+    // (i.e. are not read-only zero). A set bit specifies that the
+    // corresponding counter is supported. The lowest 3 bits
+    // (writable_hpm_counters[2 .. 0]) are ignored.
+    "writable_hpm_counters": {
+      "len": 32,
+      "value": "0x0"
+    },
+    // The bits in this value control whether the corresponding bits
+    // of the `scounteren` CSR that control access to HPM counters,
+    // and `cycle`, `time` and `instret` registers, are read-only
+    // zero. A set bit specifies that the corresponding bit of
+    // `scounteren` is writable, otherwise the bit is read-only zero.
+    // If Sscounterenw is supported the top 29 bits must be a superset of
+    // the top 29 bits of `writable_hpm_counters`.
+    "scounteren_writable_bits": {
+      "len": 32,
+      "value": "0x0"
+    },
+    // The bits in this value control whether the corresponding bits
+    // of the `mcounteren` CSR that control access to HPM counters,
+    // and `cycle`, `time` and `instret` registers at lower privilege
+    // modes, are read-only zero. A set bit specifies that the
+    // corresponding bit of `mcounteren` is writable, otherwise the
+    // bit is read-only zero.
+    "mcounteren_writable_bits": {
+      "len": 32,
+      "value": "0x0"
+    },
+    // `mtvec.{direct,vectored}.supported` indicates whether the
+    // corresponding mode is supported for `mtvec`.  At least one mode
+    // should be supported.  If both modes are supported, the mode on
+    // reset is unspecified.
+    // The `mtvec.{direct,vectored}.base_alignment` parameters control the alignment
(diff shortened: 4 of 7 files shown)


davidharrishmc (commented):
I think the ifdef could go all the way back to line 2036 to avoid reading the instruction when compressed is not supported.

jordancarlin:
> I think the ifdef could go all the way back to line 2036 to avoid reading the instruction when compressed is not supported.

Updated.

davidharrishmc (approved):

PR #2058, opened 2026-08-04
Decode priv instruction operands

### LINKED ISSUE #2032: Instructions without a testplan entry are silently mis-sampled in operand-derived coverpoints (e.g. `ssamoswap.w/d`)
Per-instruction operand handlers in `coverpoints/coverage/RISCV_instruction_sample.svh` are **generated
from `testplans/*.csv`**. An instruction that is only ever exercised by *privileged* tests has no
testplan row, so it gets no handler and none of its operands are parsed. `ins.current.rs1_val`,
`rs2_val`, `imm`, … then keep their default of `0`.

Any coverpoint derived from those fields therefore samples **0 instead of the real value** — silently,
with no error or warning. The sample lands in a plausible-looking bin (typically the "zero" bin) and the
bins that should have been hit become permanently unreachable. This is not specific to one instruction: it affects **any covergroup that crosses an operand-derived
expression with an instruction that lacks a testplan row**.

## Reproduction

Verified on #2031
1. `framework/src/act/fcov/disassemble.svh` has **no case** for `SSAMOSWAP_W` / `SSAMOSWAP_D`, so
   `disassemble()` falls through to `default: decoded = "illegal";`
   (`framework/src/act/fcov/disassemble.svh:1435`).
2. `RISCV_instruction_sample.svh` dispatches operand handlers **by instruction name**. It is generated
   by `write_instruction_sample_file()` in
   `generators/coverage/src/covergroupgen/generate.py:862`, whose instruction set comes from
   `_merge_instruction_testplans()` — i.e. purely from `testplans/*.csv`. There is no
   `testplans/Zicfiss.csv`, so no `"ssamoswap.w"` / `"ssamoswap.d"` case is ever emitted.
3. With no handler, `add_rs1()` is never called, so `has_rs1 = 0` and `rs1_val = 0`.
4. Coverpoints computing an address as `ins.current.rs1_val + ins.current.imm` therefore see `0`.

Observed in `Ssnpm_cg` (`coverpoints/priv/Ssnpm_coverage.svh`), whose main cross is:

```systemverilog
cp_pmlen_masking : cross priv_mode_u, pmm, satp_mode, a_upper_bits, pm_insn;
```

with (from `coverpoints/general/RISCV_coverage_pmm_coverpoints.svh`):

```systemverilog
a_upper_bits: coverpoint ((ins.current.rs1_val + ins.current.imm) >> 48) { ... 8 bins ... }
```

Because `rs1_val` is always 0 for `ssamoswap`, every sample collapses into the `upper_0000` bin. The
other seven `a_upper_bits` bins are unreachable for both `ssamoswap.w` and `ssamoswap.d`:

| | |
|---|---|
| Missing cross bins | **14** of 9600 (7 tags × 2 instructions) |
| `cp_pmlen_masking` | 98.25% |
| `Ssnpm_cg` overall | **99.80%** — cannot reach 100% |

**The gap is invisible without the cross.** The `pm_insn` coverpoint samples raw `ins.current.insn`,
which needs no operand parsing, so its `ssamoswap_w` / `ssamoswap_d` bins each show a healthy 96 hits
and report 100%. Only the cross exposes that the address dimension was garbage.

## Suggested fixes

1. **Add the missing disassembly cases** in `framework/src/act/fcov/disassemble.svh`:

   ```systemverilog
   SSAMOSWAP_W: $sformat(decoded, "ssamoswap.w %s, %s, (%s)", rd, rs2, rs1);
   SSAMOSWAP_D: $sformat(decoded, "ssamoswap.d %s, %s, (%s)", rd, rs2, rs1);
   ```

2. **Provide a way to emit operand handlers for decode-only instructions** that have no testplan row —
   e.g. a small hand-maintained supplementary table consumed by `write_instruction_sample_file()`. The
   layout needed here is identical to the existing `amoswap.w` entry:

   ```systemverilog
   "ssamoswap.w" : begin
       ins.add_rd(0);
       ins.add_rs2(1);
       ins.add_rs1(2);
   end
   ```

jordancarlin:
Adds an instruction_formats CSV for instructions that are not used in
any unpriv testplans. The CSV specifies the instruction type and which
XLEN it applies to so it can be added to the sample function.

The initial version includes priv instructions from the base ISA,
Svinval, Zicfiss, and Zicfilp.

Fixes #2032


coverpoints/coverage/RISCV_instruction_sample.svh
```diff
@@ -492,6 +492,12 @@
             ins.add_rs1(0);
             ins.add_imm(1);
         end
+        "c.sspopchk"     : begin
```
davidharrishmc, line 495: Are wrs.nto and wrs.sto required?  Hypervisor memory accesses and fences and hinvals?

davidharrishmc (approved):
Approved with query about more instructions.

coverpoints/coverage/RISCV_instruction_sample.svh
jordancarlin: Added. None of the coverpoints seem to depend on them yet, but no harm in adding them.

PR #2065, opened 2026-08-05
Refactor testgen module organization and imports

### LINKED ISSUE #2061: testgen cyclic imports crash on some systems with nondeterministic import order
**Describe the bug**

Following the `README` on the default `act4` branch results in `testgen` failing before any ACT configuration or DUT integration.

[log.txt](https://github.com/user-attachments/files/30716882/log.txt)

The same steps work on the `main` branch.

**To reproduce**

```
https://github.com/riscv/riscv-arch-test
```

Default branch:

```
act4
```

Commands:

```bash
git clone https://github.com/riscv/riscv-arch-test
cd riscv-arch-test
mise trust
uv sync
uv run testgen --help
```

**Expected behavior**

`testgen` prints its help message.

Example (from the `main` branch):

```
Usage: testgen [OPTIONS] TESTPLAN_DIR
...
```

**Actual behavior**

`testgen` terminates with:

```text
ImportError: cannot import name 'extract_instruction_info' from partially initialized module 'testgen.coverpoints.vector.vector_helpers' (most likely due to a circular import) (/home/cpu/tmp/riscv-arch-test/generators/testgen/src/testgen/coverpoints/vector/vector_helpers.py)
```

> [!NOTE]
> Full log with the traceback is attached above.

**Additional context**

- Fresh clone, no local modifications.
- Reproduced on two independent clones.
- The repository default branch is currently `act4`.
- Switching to the `main` branch, deleting `.venv`, running `uv sync`, and executing the same command succeeds and displays the help text.

I originally encountered this while integrating a new DUT into ACT4, but this reproducer does not require any DUT configuration.

Additionally, the `README` currently points new users to the default `act4` branch:

https://github.com/riscv/riscv-arch-test/blob/80254f59309988a1c5faf1a2beed80f4e055bdf9/README.md#get-source-code

As a result, a first-time user following the documented setup on the default `act4` branch encounters this failure immediately.

jordancarlin:
Avoids import loops that can crash on some systems. Also make the import order deterministic to avoid different behavior on different systems in the future.

Fixes https://github.com/riscv/riscv-arch-test/issues/2061


generators/testgen/src/testgen/instructions/vector.py
```diff
@@ -0,0 +1,100 @@
...
+
+from testgen.constants import ELEN_MAX, MIN_SEW_MIN
+
+
+@dataclass
+class InstructionInfo:
+    """
+    Information parsed from an individual vector instruction name.
+
+    Formatter-aware callers add ``widened_regs`` after parsing; it defaults to
```
fourth-bit, line 22: Why was this change made? It seems like a pedantic change for a helper class. The inclusion of `widened_regs` by default in `parse_instruction_info` was intentional as it allows for `get_size_multiplier` to work in all cases without any extra effort from a caller's point of view.

jordancarlin: Setting `widened_regs` in `parse_instruction_info` required importing `get_instr_type_config` from `testgen.formatters.registry`. That ends up creating an import loop. The loop only manifests on some systems because of how Python imports work.

This is the loop:
- testgen.coverpoints.vector.vector_helpers                                                                              
     - testgen.formatters.registry       # imported for get_instr_type_config                                                 
         - testgen.formatters.types.vv_type  # loaded by formatter discovery                                                      
             - testgen.coverpoints.vector.vector_helpers

I couldn't find a good way to avoid that without breaking these up. If you have a better solution, feel free to follow up with another PR, but I'd like to get this merged in the meantime to fix #2061.

mondalsushant:
Thanks for this, @jordancarlin!

I tested this PR on the AlmaLinux 9.8 environment today that consistently reproduced #2061.

With `act4` HEAD, the flow previously failed during `testgen` import with the circular import / partially initialized module exception.

After checking out this PR (https://github.com/jordancarlin/riscv-arch-test/commit/53c253e2b2738fbbdc1b6de83a62384fa22ad642), I can no longer reproduce the import failure. The flow now successfully builds `testgen`, generates covergroups and test suites, and proceeds to the UDB configuration validation.

From my testing, this PR resolves the original import-order issue reported in #2061.

generators/testgen/src/testgen/instructions/vector.py
fourth-bit: Alright, I'll look for a solution and follow up with another PR when this gets merged. Have you found a way to reproduce the issue?

davidharrishmc (approved):

generators/testgen/src/testgen/instructions/vector.py
jordancarlin: I was able to reproduce the issue by stubbing in a custom version of the `rglob` function that goes in the reverse order, but that was very hacky.

PR #2071, opened 2026-08-06
Skip Vx in vector-testgen

jordancarlin:
Exclude suites handled by the main testgen from the standalone vector-testgen. This avoids duplicate generation of Vx tests.


jordancarlin:
CC @fourth-bit.

davidharrishmc (approved):

PR #2092, opened 2026-08-11
Guard reserved PTE tests for S 1.11

### LINKED ISSUE #1875: Sv39 non-leaf D/A/U tests appear to require Privileged ISA 1.12 behavior for a 1.11 DUT
sv39_nleaf_pte_DAU_Smode and sv39_nleaf_pte_DAU_Umode require load, store, and instruction page faults when D, A, or U is set in a non-leaf PTE at level 2 or level 1.

Issue: The VF2/U74 continues the Sv39 page-table walk instead of raising a page fault when the reserved D, A, or U bits are set in a non-leaf PTE.

These tests are selected for a VisionFive 2/U74 configuration declaring:
- { name: S, version: "= 1.11.0" }
- { name: Sm, version: "= 1.11.0" }

However, both tests declare only:    # REQUIRED_EXTENSIONS: [I, Sv39]

Privileged ISA 1.10 and 1.11 state that D, A, and U are reserved in non-leaf PTEs and must be cleared by software for forward compatibility. Their address-translation algorithm does not explicitly say that reaching such a PTE must raise a page fault. The Privileged ISA 1.12 revision history identifies the mandatory page fault for a non-speculatively reached reserved PTE bit as a clarification added since version 1.11, among changes described as not strictly backwards compatible.

Please clarify whether these two tests should:

1-be selected only for Supervisor ISA 1.12 or newer;
2-use version-dependent expected behavior/reference-model configuration; or
3- Should we report this to Sail to handle based on specs version?

Board: StarFive VisionFive 2
SoC: JH7110
Core: SiFive U74
Virtual memory: Sv39
Configured Supervisor/Machine privileged ISA: 1.11

[act_sv39_nleaf_dau_priv_version_issue_20260715.zip](https://github.com/user-attachments/files/30045700/act_sv39_nleaf_dau_priv_version_issue_20260715.zip)

jordancarlin:
Reserved PTEs are only required to trap starting with priv 1.12. Skip tests with reserved PTEs on priv 1.11.

Fixes #1875

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


tests/priv/Sv/sv32_nleaf_pte_DAU_Smode.S
```diff
@@ -25,6 +25,8 @@
 
 RVTEST_BEGIN
 
+#ifndef S1P11P0_SUPPORTED
```
davidharrishmc, line 28: It's not completely obvious why these tests don't run on 1p11.  Do you think a reference to the spec would be helpful to the reader?

davidharrishmc (approved):

davidharrishmc (approved):

PR #2101, opened 2026-08-12
Split U tests to reduce runtime

davidharrishmc:
Also fix imperas YAML, which no longer passed validation


generators/testgen/src/testgen/priv/registry.py
```diff
@@ -59,6 +61,7 @@ def add_priv_test_generator(
     required_extensions: list[str] | None = None,
     march_extensions: list[str] | None = None,
     params: list[str] | None = None,
+    testcases_per_file: int | None = None,
```
jordancarlin: Set the default value here. This removes the need for the check for None later.

```suggestion
    testcases_per_file: int = TESTCASES_PER_PRIV_FILE,
```

generators/testgen/src/testgen/priv/registry.py
```diff
@@ -82,6 +88,7 @@ def decorator(func: PrivTestGenerator) -> PrivTestGenerator:
             required_extensions=required_extensions,
             march_extensions=march_extensions,
             params=params,
+            testcases_per_file=testcases_per_file if testcases_per_file is not None else TESTCASES_PER_PRIV_FILE,
```
jordancarlin:
```suggestion
            testcases_per_file=testcases_per_file,
```

generators/testgen/src/testgen/priv/extensions/U.py
```diff
@@ -55,17 +55,21 @@ def _generate_ucsr_tests(test_data: TestData) -> list[str]:
...
             [
-                test_data.add_testcase(f"{csr}", coverpoint, covergroup),
+                test_data.add_testcase(f"{csr:03x}", coverpoint, covergroup),
                 f"csrr x{temp_reg}, 0x{csr:03x}    # attempt to read CSR {csr:03x}; should get illegal instruction",
                 "",
             ]
         )
+        test_data.int_regs.return_register(temp_reg)
+        tc.code.extend(lines)
+        lines = []
```
jordancarlin: I think we can simplify this to the following (drop the indirection with `lines`):
```py
        tc.code.extend(
            [
                test_data.add_testcase(f"{csr:03x}", coverpoint, covergroup),
                f"csrr x{temp_reg}, 0x{csr:03x}    # attempt to read CSR {csr:03x}; should get illegal instruction",
                "",
            ]
        )
        test_data.int_regs.return_register(temp_reg)
```

generators/testgen/src/testgen/priv/extensions/U.py
```diff
@@ -41,8 +41,8 @@ def _generate_priv_inst_tests(test_data: TestData) -> list[str]:
     return lines
 
 
-def _generate_ucsr_tests(test_data: TestData) -> list[str]:
-    """Generate CSR tests."""
+def _generate_ucsr_tests(test_data: TestData, test_chunks: list[TestChunk]) -> None:
```
jordancarlin, line 44: A cleaner interface would be for this function to return a list of test_chunks that the calling function appends to the main list instead of this function directly mutating the test_chunks argument.

jordancarlin (changes requested):
A few minor notes.

generators/testgen/src/testgen/priv/extensions/U.py
davidharrishmc: Could do that.  _generate_ucsr_tests still requires test_chunks as an input because new_test_chunk mutates test_chunks.  Is it ok to leave as is?

jordancarlin: If you end the existing test chunk before calling `_generate_ucsr_tests`, I think it would work. It can then create its own list of test chunks that you append to the main list.

generators/testgen/src/testgen/priv/registry.py
jordancarlin: Looks like this was marked as resolved but not changed?

generators/testgen/src/testgen/priv/extensions/U.py
```diff
@@ -41,67 +41,79 @@ def _generate_priv_inst_tests(test_data: TestData) -> list[str]:
...
-        ),
+    tc = test_data.new_test_chunk(test_chunks, "ucsr_ro")
+    tc.code.extend(
+        [
+            comment_banner(
+                coverpoint,
+                "Attempt to write read-only CSRs.  Should throw illegal instruction",
+            ),
+        ]
     )
```
jordancarlin: I don't remember off hand why I needed to define `section_header`, but this should go into `section_header` instead of `code`.
```suggestion
    tc.section_header = comment_banner(
        coverpoint,
        "Attempt to write read-only CSRs.  Should throw illegal instruction",
    )
```

generators/testgen/src/testgen/priv/registry.py
davidharrishmc: In flux.  Fixed now.

generators/testgen/src/testgen/priv/extensions/U.py
davidharrishmc: If I end the existing test chunk before calling it, then new_test_chunk isn't very useful.  I'm leaning toward this style.  If you'd strongly prefer a different style, would you mind merging and making a followup with the style that we should apply to other tests when adding T-SBI support?

jordancarlin (approved):

PR #2106, opened 2026-08-12
Integrate Vls into Main Testgen

fourth-bit:
This PR generates Vls in the main testgen module. It gets 100% coverage with slight changes to errors in the covergroup for `cp_custom_ls_indexed`. This covergroup was supposed to cover that index values were truncated with `EEW=64` and `XLEN=32`. The modification is that it only runs on instructions where the index is specified to be 64, rather than all indexed instructions when `SEW=64`. The vs2_edges covergroup still seems incorrect as it checks vs2 for elements using sew rather than the specified index eew, but a fix to this feels out-of-scope for this PR as it requires a logic change rather than a shuffling of constraints.

A few small framework side changes were made as `parse_instruction_info` now populates `widened_regs` and the build timeout was increased to an hour due to vector coverage jobs consistently exceeding the limit.

Every simulator passes the tests.


jordancarlin:
> The vs2_edges covergroup still seems incorrect as it checks vs2 for elements using sew rather than the specified index eew, but a fix to this feels out-of-scope for this PR as it requires a logic change rather than a shuffling of constraints.

Please open an issue to track this so it doesn't get forgotten.

framework/src/act/build.py
```diff
@@ -41,7 +41,7 @@
     SymlinkAction,
 )
 
-BUILD_STEP_TIMEOUT_SECONDS = 300
+BUILD_STEP_TIMEOUT_SECONDS = 60 * 60  # A longer timeout is needed for coverage jobs (Vx8 can take 30-40 minutes)
```
jordancarlin: Can we change this to have a different timeout specifically for coverage? I would rather not increase the build timeout for regular users who are running without coverage. None of the regular steps should take an hour when not running coverage.

generators/testgen/src/testgen/io/templates.py
```diff
@@ -206,6 +206,11 @@ def get_vector_base_extension(testsuite: str, instr_name: str, xlen: int, sew: i
         mapped.remove("Zve32x")
         mapped.append("Zve64x")
 
+    if "Zve32x" in mapped and "64" in instr_name:
+        # This is an unsupported EEW (happens for vle64.v)
+        mapped.remove("Zve32x")
+        mapped.append("Zve64x")
```
jordancarlin, line 212: Fine for now, but I would like to figure out a better way to represent a lot of the information in this file eventually. Ideally encoded in the testplan somehow.

framework/src/act/fcov/coverage/RISCV_coverage_vector.svh
```diff
@@ -416,6 +416,30 @@ function logic[63:0] get_vr_element_zero_widen(int hart, int issue, `VLEN_BITS v
...
+    2'b11:  return val[63:0];
+    `endif
+    default: begin
+      $error("ERROR: SystemVerilog Functional Coverage: Unsupported EEW: %s", eew);
+      $fatal(1);
+    end
+  endcase
+  return 0;
+
+endfunction
```
jordancarlin, line 441: This function doesn't seem right. It takes `eew` as an `int`. When eew=64, it would fall through to the fatal case. Should the cases be 8, 16, 32, and 64 instead?

generators/testgen/src/testgen/data/edges.py
```diff
@@ -578,6 +578,11 @@ class VECTOR_EDGES:
         "sNaN_payload1": 0x7F81,  # signaling NaN with lsb set
     }
 
+    load_store_rs2_e8 = (-2, -1, 0, 1, 2)
+    load_store_rs2_e16 = (-4, -2, 0, 2, 4)
+    load_store_rs2_e32 = (-8, -4, 0, 4, 8)
+    load_store_rs2_e64 = (-16, -8, 0, 8, 16)
```
jordancarlin: Might be cleaner to have a function that takes eew and returns the appropriate set, but I'm fine keeping this as is if you prefer.

generators/testgen/src/testgen/coverpoints/cp_regs.py
```diff
@@ -220,6 +220,13 @@ def make_rs2(instr_name: str, instr_type: str, coverpoint: str, test_data: TestD
     equal_cases = [True, False] if is_zacas else [None]
     all_ones = get_zacas_mask(instr_name, instr_type, test_data)
 
+    is_vector = instr_name.lower().startswith("v")
+    if is_vector:
+        assert test_data.config.sew is not None, "SEW must be set for vector tests"
+        lmul = get_base_lmul(instr_name, instr_type, test_data.config.sew)
+    else:
+        lmul = 1  # Placeholder to keep the type-checker happy
```
jordancarlin, line 228: We should probably update this file so that `make_rs1`, `make_rs2`, and `make_rs3` are all wrappers around a single function so that all of this logic is not duplicated. Fine to leave as is for this PR.

generators/testgen/src/testgen/coverpoints/cp_reg_edges.py
```diff
@@ -53,12 +55,34 @@ def make_rs2_edges(instr_name: str, instr_type: str, coverpoint: str, test_data:
...
+        eew = match.group(1)
+        if eew == "8":
+            edges = VECTOR_EDGES.load_store_rs2_e8
+        elif eew == "16":
+            edges = VECTOR_EDGES.load_store_rs2_e16
+        elif eew == "32":
+            edges = VECTOR_EDGES.load_store_rs2_e32
+        elif eew == "64":
+            edges = VECTOR_EDGES.load_store_rs2_e64
+        else:
```
jordancarlin: If you add a function like I suggested in my other comment, this could simplify to something like
```suggestion
        edges = get_vector_load_store_edges(eew)
```

generators/testgen/src/testgen/coverpoints/vector/cp_vs_edges.py
```diff
@@ -41,11 +42,21 @@ def make_vs_edges(instr_name: str, instr_type: str, coverpoint: str, test_data:
             edges = VECTOR_EDGES.vf_edges
         elif suffix.startswith("ls"):
             edges = VECTOR_EDGES.vls_edges
+            info = parse_instruction_info(instr_name, instr_type)
```
jordancarlin: Should we rename this function `parse_vector_instruction_info` if it is vector specific?

generators/testgen/src/testgen/coverpoints/vector/cp_vs_edges.py
```diff
@@ -41,11 +42,21 @@ def make_vs_edges(instr_name: str, instr_type: str, coverpoint: str, test_data:
...
+            multiplier = info.get_size_multiplier(register, sew)
+
+            if multiplier < 1:
+                suffix += f"_emulf{int(1 / multiplier)}"
+            else:
+                suffix += f"_emul{int(multiplier)}"
```
jordancarlin: Can this reuse the get_lmul_flag function? Would be good not to duplicate that same logic everywhere.

generators/testgen/src/testgen/instructions/vector_params.py
```diff
@@ -43,6 +43,9 @@ def get_register_emul(
     ):
         emul = 1
 
+    if info.whole_registers is not None:
+        emul = info.whole_registers
+
     return emul
```
jordancarlin: Would something like this be simpler?
```
    if info.whole_registers is not None:
        return info.whole_registers

    if (
      (vector_type_config.mask_regs is not None and register_name in vector_type_config.mask_regs)
      or (vector_type_config.scalar_regs is not None and register_name in vector_type_config.scalar_regs)
    ):
        return max(lmul * info.get_size_multiplier(register_name, sew), 1)

    return 1
```

generators/testgen/src/testgen/instructions/vector.py
```diff
@@ -44,6 +55,9 @@ def get_size_multiplier(self, register: str, sew: int) -> int | float:
 
 def parse_instruction_info(instruction: str, instruction_type: str) -> InstructionInfo:
     """Parse vector instruction facts encoded in an instruction name."""
+    # Lazy import here to avoid import loops due to registry imports breaking the import hierarchy with import-time
+    # code execution. If the name is already found, there should not be a performance penalty
+    from testgen.formatters.registry import get_instruction_type_config
```
jordancarlin: It's generally not great practice to use a lazy import for the purposes of avoiding circular imports, but if you feel strongly about keeping it this way, that is ok with me. Here are two other possible approaches from AI:
 - Keep `InstructionInfo` instruction-derived only. Attach `widened_regs` in formatter-aware callers, as before.
 - Or pass formatter metadata into the parser explicitly, instead of importing the registry from `instructions.vector`.

generators/testgen/src/testgen/instructions/vector.py
```diff
@@ -55,6 +69,8 @@ def parse_instruction_info(instruction: str, instruction_type: str) -> Instructi
     # Load/Store EEW: Ends with e<eew>.v, e<eew>ff.v,
     load_store_eew_match = re.search(r"v[ls]\w*e(\d+)(?:ff)?.v", instruction)
     load_store_eew = int(load_store_eew_match.group(1)) if load_store_eew_match is not None else None
+    if instruction in ["vlm.v", "vsm.v"]:
+        load_store_eew = 8  # Special case that needs to be hardcoded
```
jordancarlin, line 68: Should this be encoded in the instruction_type_config or something like that instead?

generators/testgen/src/testgen/instructions/vector.py
```diff
@@ -65,11 +81,22 @@ def parse_instruction_info(instruction: str, instruction_type: str) -> Instructi
...
+
+    widened_regs = instr_type_config.vector_data.widened_regs
+
     return InstructionInfo(
         segments=segments,
         load_store_eew=load_store_eew,
         index_eew=index_eew,
         vext_multiplier=vext_multiplier,
+        whole_registers=whole_registers,
+        widened_regs=widened_regs,
```
jordancarlin: This returns the mutable set from `instr_type_config.vector_data.widened_regs`. We should probably pass a copy of it. Or make it a `frozenset` if nothing will ever modify it.

generators/testgen/src/testgen/formatters/types/vsr_type.py
```diff
@@ -0,0 +1,95 @@
...
+    )
+    assert params.vs3 is not None and params.vs3_val_pointer is not None, (
+        "vs3 and vs3_val_pointer must be provided for VSR-type instructions"
+    )
+    assert params.vd is None and params.vd_val_pointer is None, (
+        "vd and vd_val_pointer must NOT be provided for VSR-type instructions"
+    )
+    assert params.temp_reg is not None, "temp_reg must provided for be VSR-type instructions"
+    assert params.sew is not None, "SEW must provided for be Vector instructions"
+    assert params.lmul is not None, "lmul must provided for be VSR-type instructions"
```
jordancarlin:
```suggestion
    assert params.temp_reg is not None, "temp_reg must be provided for VSR-type instructions"
    assert params.sew is not None, "SEW must be provided for Vector instructions"
    assert params.lmul is not None, "lmul must be provided for VSR-type instructions"
```

jordancarlin: Fix this in all of the other files too.

generators/testgen/src/testgen/coverpoints/vector/cp_custom_vls.py
```diff
@@ -0,0 +1,443 @@
...
+    #     [
+    #         label_line,
+    #         test,
+    #     ]
+    # )
+    # tc.code.extend(check)
+
+    # tc = test_data.end_test_chunk()
+
+    # test_data.int_regs.return_register(check_reg)
```
jordancarlin: Should this all be removed?

generators/testgen/src/testgen/coverpoints/vector/cp_custom_vls.py
```diff
@@ -0,0 +1,443 @@
...
+from testgen.constants import VLEN_MAX
+from testgen.coverpoints.registry import add_coverpoint_generator
+from testgen.data.state import TestData, return_testcase_registers
+from testgen.data.test_chunk import TestChunk
+from testgen.formatters import format_instruction, format_single_testcase, get_instruction_type_config
+from testgen.instructions.vector import parse_instruction_info
+from testgen.instructions.vector_params import generate_random_vector_params
+
+
+@add_coverpoint_generator("cp_custom_ffLS_update_vl")
```
jordancarlin, line 21: AI is saying this specific coverpoint generator is not correct:
> cp_custom_ffLS_update_vl does not test the specified behavior. generators/testgen/src/testgen/coverpoints/vector/cp_custom_vls.py:56-85 leaves the manual faulting sequence and csrr vl signature check commented out, then emits an ordinary fault-only-first load from valid vector_ls_random_base. The generated test never faults and never observes vl, so this coverpoint is falsely reported as covered. Restore a test that masks element 0, faults at a later active element, and signs vl.

generators/testgen/src/testgen/formatters/types/strided_store_vector_type.py
```diff
@@ -0,0 +1,136 @@
...
+    setup.extend(load_code)
+    setup.append(f"LA (x{params.rs1}, {params.rs1val_pointer})")
+    setup.append(f"LI (x{params.rs2}, {params.rs2val})")
+    setup.append(load_test_vtype(params, random_vl_reg))
+
+    # We don't need random_vl_reg anymore
+    if random_vl_reg.startswith("x"):
+        test_data.int_regs.return_register(int(random_vl_reg[1:]))
+
+    equivalent_load = "vl" + instr_str[2:]
```
jordancarlin, line 122: AI comment:

> vector_params.py:157 can select zero stride for multi-element store tests. strided_store_vector_type.py:113 then stores and reloads the result. Strided element writes are unordered, so overlapping zero-stride stores have an unspecified final value; a valid DUT can disagree with Sail. Exclude zero for vl > 1, or use identical source elements.

generators/testgen/src/testgen/formatters/types/indexed_store_vector_type.py
```diff
@@ -0,0 +1,209 @@
...
+    load_code, random_vl_reg = load_vec_regs(to_load, params, test_data)
+    setup.extend(load_code)
+
+    # Ensure that the data in vs2 is within the range [0, 2*vlmax) and sew-aligned
+    setup.extend(
+        [
+            "# Ensure that the data in vs2 is within the range [0, 2*vlmax) and sew-aligned",
+            f"vsetvli x{params.temp_reg}, x0, e{params.sew}, m{get_lmul_flag(params.lmul)}, ta, ma",
+            f"add x{params.temp_reg}, x{params.temp_reg}, x{params.temp_reg}",
+        ]
```
jordancarlin: AI comment:
> indexed_store_vector_type.py:144 reduces random indices modulo 2*VLMAX, making duplicate addresses likely. For unordered vsux* stores, the winning write is unspecified, so the subsequent reload can produce a false failure against Sail. Generate unique indices for unordered stores.

jordancarlin (changes requested):

generators/testgen/src/testgen/coverpoints/vector/cp_custom_vls.py
```diff
@@ -0,0 +1,443 @@
...
+    # Use LMUL=2 (while ensuring EMUL does not exceed 8)
+    lmul = 2
+
+    info = parse_instruction_info(instr_name, instr_type)
+    eew = info.load_store_eew
+    assert eew is not None, f"Could not extract eew from fault-only-first instruction {instr_name}"
+    emul = eew * lmul / test_data.config.sew
+
+    # We might need an ifdef if vl=1 is vlmax (on lmul = 1, which could be required (e.g. 8 segments))
+    ifdef = f"UDB_ZVL{eew * 2}B_SUPPORTED" if emul * info.segments > 8 else ""
```
jordancarlin: Extensions do not have the UDB_ prefix.
```suggestion
    ifdef = f"ZVL{eew * 2}B_SUPPORTED" if emul * info.segments > 8 else ""
```

generators/testgen/src/testgen/coverpoints/vector/helpers.py
```diff
@@ -29,8 +30,13 @@ def make_and_register_edge_label(reg_name: str, edge_name: str, suffix: str, tes
...
-    if not ("random" in label and label in test_data.vector_labels):
+    if label.startswith("vs2_edge_zero_emul8_ls"):
+        # FIXME: Coverage workaround because it requires all zeros in the register
         eew = int(sew * emul)
-        test_data.register_vector_data(label, eew, elements=[get_vector_edge(edge_name, suffix, sew)])
+        max_elements = VLEN_MAX // eew * 8
+        test_data.register_vector_data(label, eew, elements=[0] * max_elements)
+    elif not ("random" in label and label in test_data.vector_labels):
+        eew = int(sew * emul)
+        test_data.register_vector_data(label, eew, elements=[get_vector_edge(edge_name, suffix, eew)])
```
jordancarlin: Why did `get_vector_edge` change from getting sew to eew?

framework/src/act/fcov/coverage/RISCV_coverage_vector.svh
fourth-bit: Fixed. But, this did reveal that we needed sail-rv32-max to have max_index_eew of 64 instead of the XLEN that it defaulted to in order to see this in coverage.

generators/testgen/src/testgen/coverpoints/vector/cp_custom_vls.py
fourth-bit: This is a direct port of what the previous version of testgen had, and satisfies the existing coverage. cp_custom_ffLS is built for the case that AI wants this to cover.

generators/testgen/src/testgen/formatters/types/indexed_store_vector_type.py
fourth-bit: There is a fix for this through a new mechanism in the VectorTypeConfig. I think the logic in it is sound, but needs to be fleshed out more in the future.

generators/testgen/src/testgen/instructions/vector.py
fourth-bit: I decided to remove `widened_regs` from `VectorInstructionInfo`. Now there is no dependence on the registry for anything in the dataclass

generators/testgen/src/testgen/instructions/vector.py
fourth-bit: It seems best to avoid touching instruction_type_config for things in this file

fourth-bit:
The whisper issue has been fixed, so now every sim passes the Vls tests

framework/src/act/build_types.py
```diff
@@ -70,6 +70,7 @@ class BuildTask:
     deps: tuple[Path, ...] = ()  # Primary output paths of predecessor BuildTasks
     label: str | None = None  # Human-readable name for failure messages (defaults to outputs[0].stem)
     intermediate: bool = False  # Only build if needed for another task
+    is_coverage: bool = False  # Is this a coverage task
```
jordancarlin: Let's change this to `timeout` and give it a default of `BUILD_STEP_TIMEOUT_SECONDS`. That gives us more flexibility if other steps end up needing different timeouts.

generators/coverage/src/covergroupgen/generate.py
```diff
@@ -546,7 +546,12 @@ def _resolve_coverpoint(cp: str, arch: str, instr: str) -> str | None:
 
     # cp_custom_ffLS requires LMUL=2; skip where that is infeasible at this SEW.
     if cp == "cp_custom_ffLS" and _is_vector(arch) and not _ffLS_feasible(instr, int(_get_effew(arch))):
-        return None
+        cp = "cp_custom_ffLS_lmul_lt2"
+
+        eew_m = re.search(r"e(\d+)ff", instr)
+        eew = int(eew_m.group(1)) if eew_m else 0
+        if eew >= 32:
+            cp += f"_eew{eew}"
```
jordancarlin: Seems like the comment is no longer true?

generators/testgen/src/testgen/formatters/types/indexed_store_vector_type.py
```diff
@@ -0,0 +1,235 @@
...
+    res = []
+    sew_limit = 2**sew - 1
+
+    for i in range(element_count):
+        vlmax_power_of_2 = i.bit_length()
+
+        # Generate a unique index value under limit for vlmax
+        next_value = min(random_int(vlmax_power_of_2 + 1, signed=False), sew_limit)
+        while next_value in res and i <= sew_limit:
+            next_value = min(random_int(vlmax_power_of_2 + 1, signed=False), sew_limit)
```
jordancarlin: Would this be simpler with `random_range` so you don't need to use `min` after getting the random number?

jordancarlin (changes requested):
A few comments on the latest round of changes.

framework/src/act/build_types.py
```diff
@@ -20,6 +20,10 @@
 # Actions: the work a task performs
 # ---------------------------------------------------------------------------
 
+BUILD_STEP_TIMEOUT_SECONDS = 300
+COVERAGE_STEP_TIMEOUT_SECONDS = 60 * 60  # Vx coverage can take about 40 minutes
+COVERAGE_STEP_TIMEOUT = True  # Enable or disable the timeout for coverage jobs
+
```
jordancarlin: Move these before the `Actions: ...` banner comment.

framework/src/act/build_types.py
```diff
@@ -20,6 +20,10 @@
 # Actions: the work a task performs
 # ---------------------------------------------------------------------------
 
+BUILD_STEP_TIMEOUT_SECONDS = 300
+COVERAGE_STEP_TIMEOUT_SECONDS = 60 * 60  # Vx coverage can take about 40 minutes
+COVERAGE_STEP_TIMEOUT = True  # Enable or disable the timeout for coverage jobs
```
jordancarlin: Can we drop `COVERAGE_STEP_TIMEOUT` and just set `COVERAGE_STEP_TIMEOUT_SECONDS` to `None` to disable the timeout?

jordancarlin (commented):
Two last minor things and then let's get this merged.

jordancarlin (approved):

PR #2107, opened 2026-08-12
Fix UDB Config Error for Imperas Configs without Zihpm

fourth-bit:
The RVI20U32 and RVI20U64 imperas configs do not have Zihpm enabled, with the comment `// haven't found a way to turn this on in imperas.ic in this variant`. Currently, because UDB will not verify the config due to values in `MCOUNTENABLE_EN` and `HPM_COUNTER_EN` that conflict with Zihpm not being supported, resulting in all `make imperas` jobs failing before tests are run. This PR fixes the UDB verification error by disabling the counters that should not be enabled without Zihpm.


jordancarlin:
This is also addressed in #2101.

fourth-bit:
Didn't see that

PR #2109, opened 2026-08-12
Check vxsat for all saturating instructions

jordancarlin:
VVI and VVX saturating instructions were not checking `vxsat`.


jordancarlin:
CC @fourth-bit

davidharrishmc (approved):

PR #2110, opened 2026-08-13
Add some RVA23 privileged normative rules and test plan references to CTP

davidharrishmc:
Also drop unnecessary required extensions


jordancarlin:
Looks like the CTP build is failing in CI. I think because of the following:
```
asciidoctor: WARNING: norm/Smstateen_norm_rules.adoc: line 9: id assigned to block already in use: t-Smstateen-normative-rules
asciidoctor: WARNING: norm/Smstateen_norm_rules.adoc: line 9: id assigned to block already in use: t-Smstateen-normative-rules
```

generators/testgen/src/testgen/priv/extensions/Smstateen.py
```diff
@@ -561,7 +561,7 @@ def _generate_fcsr_lower_fp_instrs(test_data: TestData) -> list[str]:
 
 @add_priv_test_generator(
     "Smstateen",
-    required_extensions=["S", "Zicsr", "Smstateen"],
+    required_extensions=["S", "Smstateen"],
```
jordancarlin: Why does Smstateen depend on S?

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
```diff
@@ -446,7 +446,7 @@ def _generate_fcsr_lower_fp_instrs(test_data: TestData) -> list[str]:
 
 @add_priv_test_generator(
     "Ssstateen",
-    required_extensions=["S", "Zicsr", "Smstateen", "Ssstateen"],
+    required_extensions=["S", "Ssstateen"],
```
jordancarlin: Ssstateen implies S.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
davidharrishmc: I had deleted S earlier.  See the above copilot feedback that I resolved:

```
required_extensions no longer includes a privilege-mode marker (Sm/S/U). This causes Smstateen tests to be treated as non-privileged in act.select_tests (it skips priv tests only when REQUIRED_EXTENSIONS intersects {Sm,S,U}), so they may still run when include_priv_tests is false.
```

If copilot is truthful about select_tests behavior, the S is needed as a privilege-mode marker.

jordancarlin: `include_priv_tests` is really intended to skip Sm tests on an rv32i core because UDB requires that all configs have Sm. Once we are done with the T-SBI refactor, I intend to clean that up. It should be fine to remove S for now.

jordancarlin (approved):

PR #2111, opened 2026-08-13
Added support to continue catching errors 

AAgarwal04:
Instead of catching test errors one at a time, this will allow all errors to be outputted at once.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -167,7 +167,7 @@ help:
 .DEFAULT_GOAL := elfs
 .PHONY: elfs
 elfs: tests
-	@$(UV_RUN) act $(CONFIG_FILES) \
+	@$(UV_RUN) act $(CONFIG_FILES)  -k \
 		--workdir $(WORKDIR) \
 		--test-dir $(TESTDIR) \
 		--jobs $(JOBS) \


jordancarlin (commented):
We need to decide if we want this to be the default or if it should be a variable set at the top of the Makefile.

allenjbaum:
This still only catches one error per test, right? This makes perfect
sense. I think default is OK, though there should be an easy way to force
only failing tests to be re-run
(and that means something slightly easier than manually collecting the
names of the failing tests into a file and pointing make at it,
e.g. a CLI string that parses the test result file for test failures and
makes all the magic happen.

jordancarlin:
This is only related to compiling the tests, not running them on a DUT.

davidharrishmc:
Errors in the compile process happen either during test development or because of a bad configuration file.

I find it can be difficult to interpret Python compiler errors because they are so verbose and the root cause line is mixed in with many other lines in a complex set of behind-the-scene calls.  If multiple errors are reported, I think it could become even more cumbersome to locate the root causes.  

I think this is a specialized use case to want to see multiple errors at once.  I vote for it to be a non default behavior.  In fact, I think the need is rare enough that it might be worth just being a comment in the file ("Add -k to the make statement if you want to continue compiling past the first error") rather than a variable, because the variable may distract from other more commonly used variables and make the common cases of DEBUG or VERBOSE more difficult for a casual user to find and use.

jordancarlin:
> Errors in the compile process happen either during test development or because of a bad configuration file.
> 
> I find it can be difficult to interpret Python compiler errors because they are so verbose and the root cause line is mixed in with many other lines in a complex set of behind-the-scene calls. If multiple errors are reported, I think it could become even more cumbersome to locate the root causes.
> 
> I think this is a specialized use case to want to see multiple errors at once. I vote for it to be a non default behavior. In fact, I think the need is rare enough that it might be worth just being a comment in the file ("Add -k to the make statement if you want to continue compiling past the first error") rather than a variable, because the variable may distract from other more commonly used variables and make the common cases of DEBUG or VERBOSE more difficult for a casual user to find and use.

That is the status quo. The option exists, but needs to be manually added by a user.

@AAgarwal04, what was the motivation behind this?

AAgarwal04:
I found it to be much more digestible to view all config file caused errors all at once. Instead of observing one error at a time (which I found to be quite frustrating) and addressing it with a little visibility to other possible issues later down the line, it was more useful to view all issues at once and make changes based on all this information.

I understand if this is meant to be manually added by the user. I just found it significantly more pleasant to view all errors created during the compile process and thought that it should be default behavior.

jordancarlin:
Thanks for opening this. We discussed it and decided that for most errors, it is easier to terminate after the first error. Many tests fail for the same reason, and seeing 100+ error messages makes it harder to sort out what went wrong. This flag can still be added manually by users if they find it useful.

PR #2113, opened 2026-08-13
Add tests for Svukte extension

AlexanderPortland:
Adds tests & coverpoints for the Svukte extension; sail support implemented in riscv/sail-riscv#1872.

The tests are passing, but the coverpoints for our extension and some others are not fully covered. Specifically, it seems that coverpoints using an instruction's memory access (through `ins.current.read_access`, `ins.current.write_access` or similar) or virtual address (through `ins.current.virt_adr_i` or `ins.current.virt_adr_d`) fail to be covered. As far as we can tell, those values are parsed from the RVVI trace files [here](https://github.com/riscv/riscv-arch-test/blob/56add40aa51f0d328d864476ba8363e731204f74/framework/src/act/fcov/testbench.sv#L163-L228), but the required fields are not generated by the Sail model in our traces.

Because of this, we wanted to ask for more context & advice on how you'd approach this. Thank you in advance!


jordancarlin:
Thanks for working on this!

> The tests are passing, but the coverpoints for our extension and some others are not fully covered. Specifically, it seems that coverpoints using an instruction's memory access (through `ins.current.read_access`, `ins.current.write_access` or similar) or virtual address (through `ins.current.virt_adr_i` or `ins.current.virt_adr_d`) fail to be covered. As far as we can tell, those values are parsed from the RVVI trace files [here](https://github.com/riscv/riscv-arch-test/blob/56add40aa51f0d328d864476ba8363e731204f74/framework/src/act/fcov/testbench.sv#L163-L228), but the required fields are not generated by the Sail model in our traces.
> 
> Because of this, we wanted to ask for more context & advice on how you'd approach this. Thank you in advance!

We are in the process of switching over to [RVVI-Text](https://github.com/riscv-verification/RVVI/blob/master/RVVI-TEXT/README.md) and the latest version of [RVVI-Trace](https://github.com/riscv-verification/RVVI/blob/master/RVVI-TRACE/README.md). The virtual memory signals will all be renamed as part of this, but once complete they should be available. For now, go ahead and use the existing names and it is ok that it is not getting full coverage. Once the RVVI updates are complete, we can update the coverpoints.

config/cores/cvw/cvw-rv64gc/sail.json
```diff
@@ -810,6 +810,9 @@
       "supported": true,
       "sfence_vma_illegal_if_svbare_only": true
     },
+    "Svukte": {
+      "supported": true
+    },
```
jordancarlin, line 815: Don't mark Svukte supported as true for anything that does not explicitly support it. CVW does not claim support for it.

config/qemu/qemu-rv64-max/sail.json
```diff
@@ -805,6 +805,9 @@
       "supported": true,
       "sfence_vma_illegal_if_svbare_only": true
     },
+    "Svukte": {
+      "supported": true
+    },
```
jordancarlin, line 810: Does QEMU already support Svukte? If so, update the UDB YAML config for it as well. If not, set this to false.

config/sail/sail-RVA20S64/sail.json
```diff
@@ -805,6 +805,9 @@
       "supported": true,
       "sfence_vma_illegal_if_svbare_only": true
     },
+    "Svukte": {
+      "supported": true
+    },
```
jordancarlin, line 810: RVA20S64 does not include Svukte.

coverpoints/priv/Svukte_coverage_init.svh
```diff
@@ -0,0 +1,11 @@
+///////////////////////////////////////////
+//
+// RISC-V Architectural Functional Coverage Covergroups Initialization File
+//
+// Copyright (C) 2024 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
```
jordancarlin: Use your own copyright for the new files.

coverpoints/priv/Svukte_coverage.svh
```diff
@@ -0,0 +1,138 @@
...
+    svukte_qualified_write_fault: cross ukte_set, priv_mode_u, satp_not_bare, va_high_bit_d, write_acc, store_page_fault;
+
+    // -----------------------------------------------------------------------
+    // Machine Svukte-Qualified Accesses (using MPRV)
+    // -----------------------------------------------------------------------
+
+    priv_m_effective_u: cross priv_mode_m, mprv_mstatus_set, mpp_mstatus_u;
+
+    svukte_qualified_mprv_read_fault:  cross ukte_set, priv_m_effective_u, satp_not_bare, va_high_bit_d, read_acc,  load_page_fault;
+    svukte_qualified_mprv_write_fault: cross ukte_set, priv_m_effective_u, satp_not_bare, va_high_bit_d, write_acc, store_page_fault;
```
jordancarlin: All of the machine mode tests need to move into a separate SvukteSm test suite. We are in the process of doing this for existing suites. This enables testing for profiles that do not require the standard RISC-V machine mode.

tests/priv/Svukte/sv39_Svukte.S
jordancarlin, line 1: We've been finding that maintaining handwritten tests is significantly harder than maintaining tests that use the priv test generator already in this repo. Would it be possible to use a Python generator for this suite? There are lots of examples in `generators/testgen/src/testgen/priv/extensions`.

jordancarlin (changes requested):
Here are a few high-level comments. I haven't looked through the coverpoints or tests in detail yet, but will try to take a more thorough look soon. Please also add a normative rule to coverpoint mapping in `coverpoints/norm` and a CTP entry in `docs/ctp/src`. Look at the other files in those directories for examples and reach out with any questions. Thanks again for working on this extension.

coverpoints/priv/Svukte_coverage.svh
AlexanderPortland: Okay! So we should have a separate set of `SvukteSm_coverage*` files and a separate test suite specifically for testing only the machine mode functionality & its coverage?

jordancarlin: Exactly. The Svukte set can cover S and U mode behavior. The SvukteSm suite will cover machine mode behavior. We also probably need a separate SvukteH suite since you can implement Svukte without hypervisor but need to test the extra hypervisor behavior if it is implemented.

AlexanderPortland: Okay, so add the SvukteH suite now? Is the best way to gate those H tests to use `#ifdef H_SUPPORTED`? Also, if we're writing hypervisor tests should we add the hypervisor crosses to the coverpoints?

tests/priv/Svukte/sv39_Svukte.S
AlexanderPortland: We can switch to that if its preferred! We'd originally done handwritten tests because that was what the other Sv extensions seemed to do, and there was already an outline to start with from [this previously closed PR](https://github.com/riscv/riscv-arch-test/pull/767)

coverpoints/priv/Svukte_coverage.svh
jordancarlin: If you are writing the tests, definitely include them in the coverpoints. The next release of the Sail model will have Hypervisor support, so it should be available before Svukte support is merged into Sail. In terms of gating them, there are two approaches you could use. If you want to include it in the same test as the regular Svukte tests, you can use `#ifdef H_SUPPORTED`. If you want to write them as separate tests (which will likely be easier), then you can just add `H` to the list of `REQUIRED_EXTENSIONS` in the YAML header at the top of the test files.

tests/priv/Svukte/sv39_Svukte.S
jordancarlin: We are likely going to be switching the Sv tests over to use Python soon. If this gets merged first, I am ok with it being handwritten and we can migrate it when we switch over the other virtual memory tests. Otherwise, it would be better for it to use Python from the start.

AlexanderPortland:
For the CTP, it looks like almost all the other Sv entries are in [this google sheet](https://docs.google.com/spreadsheets/d/1LGwD2e01j4SN9tg85nt6gPwpp7nXyXMF7hqbw3jvXko/edit?gid=375458051#gid=375458051). Should we request access to keep ours in the same spot as the rest?

davidharrishmc:
Permanent homes haven't been selected, but that would be fine to request edit access and create a new tab.

nwrousell:
An update on our end:

We've ported the tests to python generators and written hypervisor tests. We ran into some infrastructure hurdles that might require some non-Svukte changes. 

1. Per feedback on https://github.com/riscv/sail-riscv/pull/1872, the Sail model now gates Svukte behind an `--enable-experimental-extensions` CLI flag. We tested by adding this flag to the sail call in `framework/src/act/build_plan.py`, but didn't commit this as we assumed the maintainers would want to decide how to expose this to the user. 
2. We ran into several errors on non-hypervisor tests when uncommenting the `undef` of `H_SUPPORTED` in `tests/env/riscv_arch_test.h`. I think it's due to how `H_SUPPORTED` changes the trap handler even for non-hypervisor tests (as `H_SUPPORTED` is determined by the config I think). SvukteS and SvukteSm still pass with `H_SUPPORTED` unset and SvukteH passes with `H_SUPPORTED` set. 
3. When getting SvukteH running, we ran into a hypervisor bug with the trap handler where the `lhu` used to grab the faulting instruction (to determine if it's 16 or 32 bits) executed the access via satp instead of vsatp. We added a patch that uses `hlvx.lhu` to execute the access with an effective mode the same as the trapping address. We gated this with the existing `RVTEST_HYPERVISOR` macro (which gets set by the SvukteH test).

davidharrishmc:
Agreed that H support is incomplete and untested, and will just be a source of frustration until it matures.  @UmerShahidengr is the lead and may have insight on timeframe.

AlexanderPortland:
@jordancarlin, based on what @davidharrishmc said above, is it a viable path to merge this but keep the hypervisor tests gated for now (as they currently are)? Or, is hypervisor support & testing going to be made robust in the near-enough future that it's worth holding off on these changes until then? Thanks, and appreciate any guidance!

jordancarlin:
A version of Sail with Hypervisor support will be coming later this week (or next week at the latest). Once that is released, the ACT infrastructure will be updated soon after to support Hypervisor. Since this work will be completed in the next few weeks, I would rather wait for it and then revisit this PR.

AlexanderPortland:
Hi @jordancarlin, how have you been? Any updates on H support within the ACT infrastructure or changes we can make to get this merged?

jordancarlin:
The updated version of sail-riscv was released and the changes needed in the ACT trap handler and boot code are in #2363. Hoping to get that merged soon, and once we do, you should be able to update this PR.

Also worth noting that the existing Sv tests were ported to used Python generators recently. You may be able to reuse some of that work.

Finally, I just added support for unratified extensions behind an experimental flag (see #2433), so once this is ready we will be able to merge it even though it is not yet ratified.

PR #2115, opened 2026-08-14
Adding coverpoints of Sscofpmf

ayeshaanwaar05:
This PR contains coverpoints of SscofpmfSm,  SscofpmfS and SscofpmfU


config/imperas/imperas-rv64-max/rvmodel_macros.h
```diff
@@ -151,4 +151,11 @@
   li _R2, CLINT_SSIP_ADDRESS;              \
   sw zero, 0(_R2);
 
+##### Sscofpmf #####
+
+#define RVMODEL_MHPMEVENT   mhpmevent3
+#define RVMODEL_MHPMCOUNTER mhpmcounter3
+#define RVMODEL_MHPMEVENT_VAL               // Values will be assigned once there is a way to increment counters
```
davidharrishmc: Set it to 0 for now so the compilation doesn't crash, leave a note that it needs to change.  Applies to all configs that don't increment counters.  Set RVMODEL_MHPMEVENT_CODE to nop for now with similar note.

coverpoints/priv/SscofpmfSm_coverage.svh
```diff
@@ -0,0 +1,51 @@
...
+    `else
+        cp_overflow_hw_only:   cross priv_mode_m, mip_clear, mie_clear, mhpmcounter_write_extremes, mhpmevent_all_zero, mhpmevent_base_zero;
+    `endif
+    cp_scountovf_mcounteren:   cross priv_mode_m, of_write_pattern, mcounteren_write_pattern, mcounteren_walking_one;
+    cp_sscofpmf_access:        cross priv_mode_m, csr_access_pattern, hpm_csr_target ;
+    cp_lcofi:                  cross priv_mode_m, lcofi_ip, lcofi_ie, lcofi_mideleg, mstatus_mie_clear, mstatus_sie_set;
+    cp_lcofip_priority:        cross priv_mode_m, mstatus_mie_set, sstatus_sie_set, mie_clear, lcofi_ip_one, mip_other_pending;
+endgroup
+
+function void sscofpmfSm_sample(int hart, int issue, ins_t ins);
```
davidharrishmc: Case differs from sscofpmfsm_sample where it is called.

coverpoints/priv/RISCV_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,205 @@
...
+            bins yes = {1};
+    }
+    mhpmcounter_write_extremes: coverpoint ins.current.insn {
+        wildcard bins write_ones  = {CSRRW} iff (ins.current.insn[31:20] == RVMODEL_MHPMCOUNTER && ins.current.rs1_val == '1);
+        wildcard bins write_zeros = {CSRRW} iff (ins.current.insn[31:20] == RVMODEL_MHPMCOUNTER && ins.current.rs1_val == '0);
+   }
+
+    // Pack the 29 OF bits (mhpmevent3..mhpmevent31) into one expression via macro
+    `ifdef UDB_MXLEN_64
+        `define OF_VEC {ins.current.csr[CSR_MHPMEVENTH31][63], ins.current.csr[CSR_MHPMEVENTH30][63], \
```
davidharrishmc: There should be no H suffix on RV64 and H suffix in RV32.  Applies multiple places.

coverpoints/priv/RISCV_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,205 @@
...
+    `else
+        // On RV32, MINH/SINH/UINH/VSINH/VUINH live in mhpmevent*h[30:26] (address + 0x400)
+        mhpmevent_inhibits_all_set: coverpoint ins.current.insn {
+                wildcard bins write_pattern = {CSRRW} iff (ins.current.insn[31:20] == RVMODEL_MHPMEVENT + 12'h400 && ins.current.rs1_val[30:26] == 5'b11100);
+        }
+    `endif
+
+    hpm_csr_target: coverpoint ins.current.insn[31:20] {
+            bins scountovf   = {CSR_SCOUNTOVF};
+            bins mhpmevent[] = {CSR_MHPMEVENTH3,  CSR_MHPMEVENTH4,  CSR_MHPMEVENTH5,
```
davidharrishmc: The H registers should only apply when RV32 is supported.

davidharrishmc (changes requested):

davidharrishmc:
I asked Claude for a review of coverpoints vs. testplan because I've read this too many times.  It raised some concerns.

Note that the coverpoints should just be testing the input stimulus.  The tests check that the proper output occurs.

---

The four review comments from 2026-08-24 have been addressed in the current revision: macro values set to 0 / `nop`
with notes, sample-function name case fixed, H-suffix CSRs used only under RV32.

## Verdict

The file structure, config wiring (`RISCV_coverage_config.svh`, `_base_init.svh`, `_base_sample.svh`), and the
coverpoint names follow the plan, and the CSR-access, scountovf, and inhibit stimulus coverpoints are close to it.
But most of the crosses cannot be hit as written, several check a different condition than the plan states, and
three of the five `rvmodel_macros.h` edits are syntactically broken. Needs another revision before merge.

## Plan coverage summary

| Plan coverpoint | PR | Status |
|---|---|---|
| cp_xinh_inhibits_xmode (32 combos, per mode) | cp_minh/sinh/uinh_inhibits_xmode | Sm/S: cross partly unreachable; U: no combos, no RV32 |
| cp_of_set_on_overflow | cp_of_set_on_overflow | hits at setup only; outcome (OF=1, LCOFIP, counter value) not covered |
| cp_overflow_hw_only | cp_overflow_hw_only | unreachable cross (two instruction-based coverpoints) |
| cp_lcofip_hw_only | cp_lcofip_hw_only | trivial (mode × OF); LCOFIP read not covered |
| cp_scountovf_shadow | cp_scountovf_shadow | unreachable cross (pattern_class vs walking_one) |
| cp_scountovf_mcounteren | cp_scountovf_mcounteren | mostly unreachable; `scountovf_of_match` unused |
| cp_sscofpmf_access | cp_sscofpmf_access | matches plan |
| cp_lcofi_m (8 combos from M, S, U) | cp_lcofi_m | MIE=0 instead of 1; only M mode; enable bits from ins.current |
| cp_lcofi_sip_s | cp_lcofi_sip_s | enable bits from ins.current (unreachable when trap taken) |
| cp_lcofi_sip_u | cp_lcofi_sip_u | SIE=0 and mip/mie instead of SIE=1 and sip/sie |
| cp_lcofip_priority_m / _s | cp_lcofip_priority_m / _s (+ U) | wrong pending set for M; mie==0 at wrong point; taken interrupt not observed |
| RVMODEL macros in all configs | added | 3 broken continuations, 1 wrong signature, no RVMODEL_MHPMCOUNTERH |
| Normative rule mapping | `coverpoints/norm/Sscofpmf.yaml` (from PR 2100) | not updated: cites `cp_lcofi` and `cp_lcofip_priority`, which this PR does not define; one rule name is a typo |

## Blocking: crosses that are unreachable

- **cp_overflow_hw_only** (all three files) crosses `mhpmcounter_write_extremes` with `mhpmevent_all_zero`. Both are
  bins on `ins.current.insn` for different CSRRW targets, so no single instruction satisfies both (same for the RV32
  variant with `mhpmevent_base_zero`). Split into per-instruction crosses, or make one a CSR-state coverpoint
  (event register == 0).
- **cp_scountovf_shadow** crosses `of_pattern_class` (bins only for 0 and 29 OF bits) with `of_walking_one` (valid only
  when exactly one OF bit is set). They can never be true together. Add a `walking = {1}` bin or cross separately.
- **cp_scountovf_mcounteren** crosses `mcounteren_write_pattern` (0, 1, or 29 ones) with `mcounteren_walking_one`
  (exactly one), so only the walking_one row is reachable, and both come from the same instruction.
  `scountovf_of_match` is defined but not used in any cross, yet it is the coverpoint that expresses the plan's
  expectation (M reads OF; S reads OF & mcounteren). Cross it at the scountovf read.
- **cp_minh_inhibits_mmode / cp_sinh_inhibits_smode** cross `mhpmevent_xinh_combos` (32 values of bits 62:58) with
  `mhpmevent_minh` / `mhpmevent_sinh` (a bit inside that field), so half the cross is impossible. Crossing with
  `hpmcounter_nonzero` adds more impossible rows (MINH=1 must give zero). Cross the 32 combos with
  `mhpmevent_of_zero` and a read of the counter; let the signature check the outcome.
- **cp_lcofi_m, cp_lcofi_sip_s, cp_lcofip_priority_\*** sample enable bits from `ins.current`. The Sstc and Interrupts
  covergroups sample `mstatus.MIE` / `SIE` from `ins.prev` because the trace shows them cleared on the instruction
  that takes the trap, so the enabled-and-pending rows will not hit. Follow that precedent.

## Blocking: crosses that check something other than the plan

- **cp_lcofi_m** uses `mstatus_mie_clear` (MIE = 0); the plan says MIE = 1. The plan asks for the 8 combinations
  from each of M, S, and U mode; only `priv_mode_m` is crossed. `mstatus_sie_set` and `sstatus_sie_set` are identical
  coverpoints on `CSR_SSTATUS`; one should be on `CSR_MSTATUS`.
- **cp_lcofi_sip_u** uses `sstatus_sie_clear` (SIE = 0) and the mip/mie bits; the plan says sstatus.SIE = 1 with
  sip/sie when S is supported, and the mip/mie fallback only when it is not.
- **cp_lcofip_priority_m** crosses `mip_other_pending` with six bins including SEIP/STIP/SSIP; the plan's M variant is
  {MEIP, MTIP, MSIP, none}. The Sm file defines `sip_other_pending_m` with exactly those bins and never uses it. It also
  crosses `mie_clear` (mie == 0), but the plan's trigger point is mie = all 1s, and `sstatus_sie_set` is irrelevant
  in M mode. None of the priority crosses observe which interrupt was taken; `InterruptsSm/cp_priority` crosses
  mcause for this, and the plan's expectation ("highest priority fires; LCOFIP only if none other pending") is not
  covered without it.
- **cp_of_set_on_overflow** hits at the `csrw mhpmevent` instruction for both OF = 0 and OF = 1, before any counting
  happens, so it covers setup but not the expectations: OF becomes 1, the counter is neither all 0s nor all 1s, and
  LCOFIP = 1 iff OF was initially 0. Add the counter all-ones write, and cross `lcofi_ip` at the mip/sip read after
  the latency wait. `mip_clear` (mip == 0) cannot hold once LCOFIP is set, which is the state of interest.
- **cp_lcofip_hw_only** is just `priv_mode` × `mhpmevent_of`. The plan's observable is LCOFIP staying 0 after a
  software set and clear of OF; cross the set/clear instruction with `lcofi_ip_zero` at the mip/sip read.
- Crosses with no instruction qualifier (inhibit, overflow, priority) hit on any instruction whose CSR state happens
  to match, not at the plan's sample points. The testgen marks explicit sample points (`csrr RVMODEL_MHPMCOUNTER`,
  `csrr mip`); the crosses should include those reads.

## RV32 and U-mode gaps

- `SscofpmfU_coverage.svh` has no RV32 handling: `mhpmevent_uinh` reads bit 60 of `mhpmevent3`, which on RV32 is bit
  28 of `mhpmevent3h`. The U suite also lacks `mhpmevent_xinh_combos`, so it does not cover the plan's 32 combinations.
- `mhpmcounter_write_extremes` and `hpmcounter_nonzero` only look at `mhpmcounter3`; on RV32 the plan's all-ones write
  includes `mhpmcounter3h`, and no `RVMODEL_MHPMCOUNTERH` macro is defined (the plan notes one is needed).
- The coverpoints hardcode counter 3 while the plan lets each target choose `RVMODEL_MHPMCOUNTER` /
  `RVMODEL_MHPMEVENT`. Either document that coverage assumes counter 3 or parameterize the CSR index.
- `CSR_MHPMEVENT3 + 12'h400` works, but `CSR_MHPMEVENT3H` exists and reads better; `hpm_csr_target` already uses
  the H names.

## rvmodel_macros.h bugs

- Imperas, Sail, and Spike define `RVMODEL_MHPMEVENT_CODE(_R1, _R2) \  // comment` with the comment after the
  backslash. A backslash followed by a comment is not a line continuation, so the macro body is a stray `\` and the
  `nop` on the next line is emitted at the include site. Move the comment above the define.
- QEMU defines `RVMODEL_MHPMEVENT_CODE` with no parameters; the testgen invokes it as
  `RVMODEL_MHPMEVENT_CODE(xA, xB)`, which would expand to `nop(xA, xB)`. Use the same two-parameter signature.
- Whisper's `lw _R2, 0(_R1)` requires `_R1` to hold a readable address in every privilege mode; state that where the
  macro is documented so the tests set it.
- The plan's "if any macro is undefined and SSCOFPMF_SUPPORTED, error" check and the Sail-specific variant it mentions
  are not in this PR (they may belong with the tests PR, but nothing references them yet).

## Missing per repo convention

- `coverpoints/norm/Sscofpmf.yaml` already exists on act4 (PR 2100, "Adding ctp of Sscofpmf") and covers 15 of the
  16 rules the plan lists, but it was written before this PR's coverpoint names settled and this PR does not update
  it. Against the coverpoints actually defined here:
  - `{SscofpmfSm_cg,SscofpmfS_cg,SscofpmfU_cg}/{cp_lcofi}` — no covergroup defines `cp_lcofi`; the PR has
    `cp_lcofi_m` (Sm), `cp_lcofi_sip_s` (S), `cp_lcofi_sip_u` (U). The rules `mip_lcofip_acc` and part of
    `sip_sie_Sscofpmf` therefore point at nothing.
  - `{SscofpmfSm_cg,SscofpmfS_cg,SscofpmfU_cg}/{cp_lcofip_priority}` — only the U file uses that exact name; Sm and S
    define `cp_lcofip_priority_m` / `_s`, so `intr_M_mode_pri` and `sip_sie_priority_bit_order` resolve only for U.
  - `count_overflow_triggermip_lcofip_mie_lcofie_op` is two rule names run together, so `mip_lcofip_mie_lcofie_op`
    (the only plan rule not otherwise present) is unmapped and a nonexistent rule name is introduced. The generated
    `Sscofpmf_norm_rules.adoc` reproduces the typo.
  - `LCOFIP_op -> SscofpmfS_cg/cp_lcofip_hw_only` and `scountovf_op -> SscofpmfS_cg/cp_scountovf_shadow` cite S only;
    the PR defines `cp_lcofip_hw_only` in all three covergroups and `cp_scountovf_shadow` in Sm and S, matching the
    plan.
- U mode names `cp_lcofip_priority` without the `_u` suffix used elsewhere; the plan only defines `_m` and `_s`, so
  either add the suffix or drop it.

## Bottom line

Stimulus-side coverage (CSR accesses, write patterns, inhibit combinations) is mostly there; outcome-side coverage
(OF set, LCOFIP, taken interrupt, scountovf value) is largely absent, and the crosses that try to combine them are
built from mutually exclusive coverpoints. Rework the crosses around the test's sample instructions using `ins.prev`
for enable bits, fix the U-mode RV32 handling, repair the four macro headers, and bring the existing norm YAML in
line with the final coverpoint names.

davidharrishmc:
Next Claude review.  Don't assume everything it says is true, but if you disagree, justify your approach.


Partially. The 2026-09-07 commit "Implemented feedback" fixes about half of the items from the 2026-09-03 review, leaves the
  rest untouched, and introduces new problems in the priority crosses. It still needs another revision.
     
  Resolved

  - cp_overflow_hw_only is now reachable. The event-register-zero check became a CSR-state coverpoint instead of a second
    instruction bin.
  - The two scountovf crosses no longer combine mutually exclusive bins. A single stimulus pattern coverpoint (zero, all ones,
    29 walking ones) replaced the pattern-class and walking-one pair, and the previously unused match coverpoint is now crossed.
  - cp_lcofi_m now requires MIE=1, and the duplicated SIE coverpoint was split into mstatus and sstatus versions.
  - cp_lcofi_sip_u now selects sstatus/sip/sie when S is supported and falls back to mip/mie otherwise, as the plan says.
  - The M-mode priority cross uses the {MEIP, MTIP, MSIP, none} set.
  - cp_lcofi_sip_u now selects sstatus/sip/sie when S is supported and falls back to mip/mie otherwise, as the plan says.
  - The M-mode priority cross uses the {MEIP, MTIP, MSIP, none} set.
  - The U covergroup reads UINH from mhpmevent3h on RV32.
  - The macro rename to CSR_MHPMEVENT3 works, since riscv_arch_test.h includes encoding.h before rvmodel_macros.h. The Whisper
    event value and the new perf event map entry agree.

  Not addressed

  - The three broken macro headers (Imperas, Sail, Spike) still have a comment after the line-continuation backslash, so the
    macro body is a stray backslash and the nop leaks into the include site. The QEMU macro still takes no parameters while the
    generator invokes it with two.
  - The inhibit crosses still cross MINH or SINH with the 5-bit combo field that contains it, and still include the
    counter-nonzero outcome. The U covergroup still lacks the 32 combinations.
  - cp_of_set_on_overflow and cp_lcofip_hw_only are unchanged.
  - The norm YAML was not touched. It is now worse: the U covergroup no longer defines any priority cross, so the
    cp_lcofip_priority citation resolves nowhere, and cp_lcofi and the run-together rule name remain.
  - OF_VEC is still redefined three times.

  New problems in this revision

  - The rewritten priority crosses go against the "coverpoints test only stimulus" comment on the PR. They cross a taken-trap
    flag, the post-trap privilege mode, and a scountovf-equals-OF match. The InterruptsSm precedent crosses only mode, MIE, mie
    bits, and mip bits.
  - cp_lcofip_priority_m cannot hit. It crosses an instruction bin (the csrw to mhpmevent3) with the trap flag, so the interrupt
    would have to land on that exact instruction. It also requires prev MIE=0, but an M-mode interrupt needs MIE=1, and the
    mie-all-zeros row with a taken LCOFI is impossible.
  - The S and U covergroups cross their privilege mode with M-only CSR writes. The generator in PR 1942 routes those writes
    through a T-SBI ecall, so the csrw retires in M mode and the S or U mode qualifier is false. This makes
    cp_of_set_on_overflow, cp_overflow_hw_only, both scountovf crosses, and cp_lcofip_priority_s unreachable in S and U. The S/U
    sample points should be the reads the test makes in that mode.
  - cp_lcofi_sip_s now mixes prev-epoch sie with current-epoch sstatus and sip.

ayeshaanwaar05:
- The norm YAML was not touched. It is now worse: the U covergroup no longer defines any priority cross, so the
 cp_lcofip_priority citation resolves nowhere, and cp_lcofi and the run-together rule name remain.
- OF_VEC is still redefined three times.
These two are remaining to be addressed . Will make separate pr for ctp and OF_VEC is correctly defined one time

davidharrishmc:
Claude review:
  Fixed in this revision

  - The three broken macro headers (Imperas, Sail, Spike) now put the comment on the value line, and the QEMU macro takes two parameters.
  - The inhibit crosses drop the MINH/SINH bit and the counter outcome, and the U covergroup now covers all 32 combinations.
  - The priority crosses no longer use trap_taken, post-trap mode, or the scountovf match. They now have the InterruptsSm shape.

  The two items the author lists as remaining

  1. Norm YAML. coverpoints/norm/Sscofpmf.yaml still cites cp_lcofi and {Sm,S,U}/cp_lcofip_priority, which do not exist under those names, and
     still has the run-together rule name. Deferring to a separate PR is acceptable, but the generated CTP table is wrong until it lands, so the
     two should merge together.
  2. OF_VEC. The author says it is defined once. That is true of the source file, but the shared file is included into three covergroups, so the
     compiler sees the same define three times. It is at most a warning. A one-line ifndef guard settles it.

  New or still open

  1. Blocking. In the RV64 branch, mhpmevent_of_one reads bit 31 of CSR_MHPMEVENT3 + 0x400, which is mhpmevent3h. That CSR does not exist on
     RV64, the trace never writes that index, and the coverpoint is constant zero. cp_of_set_on_overflow is therefore unreachable in all three
     covergroups on RV64. It should read bit 63 of mhpmevent3.
  2. cp_of_set_on_overflow crosses mip==0 with OF==1. When overflow sets OF from 0, hardware also sets LCOFIP, so only the OF-initially-1 row can
     hit. The plan wants both OF cases and the "LCOFIP=1 iff OF was 0" relation. Cross lcofi_ip with both values instead of mip_clear.
  3. cp_lcofi_m, cp_lcofi_sip_s, cp_lcofi_sip_u, and both priority crosses sample MIE or SIE from the current epoch. The InterruptsSm,
     InterruptsS, and Sstc covergroups sample them from the previous epoch, with a comment that RVVI shows post-trap state on the trapping
     instruction. Every row where the interrupt actually fires therefore reads the enable as 0 and never hits. For the S-mode crosses the firing
     instruction is the mret into S, so the prev-mode qualifier is also false. InterruptsS handles this with priv_mode_s_after plus prev SIE.
  4. The 32 inhibit combinations include VSINH and VUINH. Without H those bits are read-only zero, so 24 of 32 bins can never hit on non-H
     configs. The plan says "exclude unsupported modes." Gate the bins on H_SUPPORTED.
  5. The checkerboard bin names are swapped relative to the generator in PR 2181, since OF_VEC bit 0 is mhpmevent3. Both patterns still get hit.
     Rename only.
  6. Generator-side note for PR 2181: it only writes the 0b11100 inhibit pattern, so four of the five pattern bins will stay uncovered until it
     sweeps them.

davidharrishmc:
Claude review of `4f0632093`, evaluated together with PR #2181 at `0adec4a16`. Coverage was measured with the two
branches merged, on `sail-rv64-max` and `sail-rv32-max`. Sail does not count HPM events (sail-riscv#1733), so bins
that need a real overflow are expected to miss there and are not listed as problems.

## What this commit does

It re-adds the four coverpoints that had been deleted from the shared file (`sip_lcofi`, `sie_lcofi`,
`sstatus_sie_set`, `sie_state`), now locally in the S and U covergroups. Both covergroups compile again under
Questa for RV64 and RV32, which clears the blocking item from the last review. Nothing else changed in this PR.

## Coverage now, merged with #2181

The tests PR moved a lot: the S suite boots to S, the lcofi tests write `sstatus`/`sip` instead of
`mstatus`/`mip`, and the inhibit sweep covers all 32 combinations in every mode. That is what made most of the
previously zero crosses hit.

| Cross | Last review | Now (rv64 / rv32) |
|---|---|---|
| S `cp_sinh_inhibits_smode`, U `cp_uinh_inhibits_umode` | 5/8 | 100% |
| S `cp_scountovf_mcounteren` | 2/93 | 100% |
| S `cp_lcofi_sip_s` | 0% | 100% |
| U `cp_lcofi_sip_u` | 0% | 75% |
| S `cp_lcofip_priority_s` | 0% | 0% |
| Sm `cp_lcofip_priority_m` | 1/8 | 1/8 |
| rv32 S `cp_sscofpmf_access` | 5/150 | 5/150 |
| Covergroup totals Sm / S / U | 93 / 79 / 70 | 93 / 85 / 81 |

## Still open, in priority order

1. **`mie_state` and `sie_state` `all_ones` bins are unreachable.** `mie`/`sie` are WARL; after `csrw mie, -1`
   Sail reads back `0x2AAA`, and every real hart will drop its unimplemented bits too. This alone holds
   `cp_lcofip_priority_m` at 1/8 and `cp_lcofip_priority_s` at 0 on every model. Use the InterruptsSm idiom,
   `coverpoint ins.current.csr[CSR_MIE][15:0] { wildcard bins ones = {16'b??1?1???1???1???}; }` (LCOFIE plus the
   three M enables), or a concatenation of the four bits, for both `mie` and `sie`.

2. **`cp_lcofi_sip_u` misses its enabled-and-pending row.** The interrupt fires on the `mret` that enters U, so the
   trap record is the `mret`, retired in M, and `priv_mode_u` (`ins.prev.mode`) is false. Use
   `priv_mode_s_after`-style qualifiers as `cp_user_sei_handled_s` does in `InterruptsS_coverage.svh`.

3. **RV32 S `cp_sscofpmf_access` still carries 145 impossible bins.** `hpm_csr_target` adds the 29 `mhpmeventNh`
   targets on RV32 and the S cross inherits them, but the plan and the generator sweep those only from M. Put
   the H targets in an Sm-only coverpoint, or give the S cross a scountovf-only target.

4. **`cp_of_set_on_overflow` in S and U still crosses `mip_clear` with OF = 1.** After a real overflow LCOFIP = 1,
   so `mip != 0`; the 20% shown is one incidental hit from leftover state. Sm already uses `lcofi_ip` here; S and U
   should match.

5. **The CSR alias problem was worked around from the test side rather than fixed.** The RVVI `csr[]` array only
   sees the address Sail logs, and Sail logs one address per access, so `csrs sip` never updates `CSR_MIP` and
   `csrsi mstatus` never updates `CSR_SSTATUS`. #2181 now writes `sstatus`/`sip` so these coverpoints sample, which
   works for this exact test pairing, but the coverpoints still cannot see LCOFIP set by hardware or by an `mip`
   write. Fine if the two PRs merge together; worth a comment in the covergroups so the dependency is visible.

6. **Norm YAML** still cites `cp_lcofi` and `{Sm,S,U}/cp_lcofip_priority`, which do not exist under those names.
   No follow-up PR has appeared yet.

## For #2181

The new S and U lcofi code hops back to M with the legacy `RVTEST_GOTO_MMODE`. Converted suites should use
`RVTEST_TSBI_GOTO_MMODE` or T-SBI CSR calls instead (`docs/tsbi-changes.md`).

davidharrishmc:
# Fourth review, 2026-09-09 (head 72fc0f156 "Implemented feedback")

Diff since c4ff5efa7 touches the four rvmodel_macros.h and the four Sscofpmf svh files. CI green.
This round is backed by a real coverage run: PR 2115 merged with PR 2181 (tests), `make coverage
EXTENSIONS=SscofpmfSm,SscofpmfS,SscofpmfU` on sail-rv64-max and sail-rv32-max. Both XLENs agree.

## Fixed since the third review
- All four rvmodel_macros.h bugs (imperas/sail/spike `\ // comment` continuation, qemu missing (_R1,_R2)).
- mhpmevent_of_one RV64 branch now reads csr[CSR_MHPMEVENT3][63], not the nonexistent mhpmevent3h. Was blocking.
- mhpmevent_xinh_combos gated on H_SUPPORTED: 8 combos (MINH/SINH/UINH) without H instead of 32 with 24 dead.
- mstatus_mie_set / mstatus_sie_set now sample ins.prev.
- U adopted priv_mode_u_after (weight 0), citing the InterruptsS precedent.
- cp_overflow_hw_only is 100% in all three covergroups on both XLENs.

## Measured coverage
| Covergroup | rv64 | missing bins: unused coverpoints | crosses |
|---|---|---|---|
| SscofpmfSm | 94.40% | 0 | 13 |
| SscofpmfS  | 85.75% | 7 | 13 |
| SscofpmfU  | 79.46% | 38 | 7 |

## Open
1. Unused helper coverpoints never close. RISCV_coverage_sscofpmf.svh is included whole into all three
   covergroups but each uses a subset; unused ones still count. U: of_stimulus_pattern 28, csr_access_pattern 2,
   mip_other_pending 2, plus 6 single-bin (lcofi_ip_one, lcofi_ip, lcofi_ie, mstatus_mie_set, mstatus_sie_set,
   mie_state) = 38 of 45 misses. S: same six minus mstatus_sie_set = 7. Fix with type_option.weight = 0 (the file
   already uses that idiom for priv_mode_u_after). Takes U to ~95%, S to ~90%.
2. Every row needing OF=1 together with LCOFIP=1 is empty: cp_of_set_on_overflow 0/5 in all three;
   cp_lcofip_hw_only missing exactly <*,OF=1,LCOFIP=1>; cp_lcofi_sip_u missing <*,*,LCOFIE=1,LCOFIP=1>.
   Cause: sail (and spike, imperas) have RVMODEL_MHPMEVENT_VAL 0 / CODE nop, so no counter increments and
   hardware never sets OF or LCOFIP (sail-riscv#1733). Coverage is measured only on Sail, so these cannot close
   in CI. Either have the tests set OF and mip.LCOFIP by software, or mark them as awaiting model support.
3. cp_lcofip_priority_m/_s at 1/8: only <mode, inhibits=0, MIE=1, mie=all_zeros, LCOFIP=1, other=none>.
   Missing the all_ones enable row (4) and every other-pending row. PR 2181 never raises a second interrupt or
   sets the other enables, so the plan's actual expectation is unverified.
4. U has no cp_sscofpmf_access (hence csr_access_pattern unused). Decide whether the plan wants a U-mode
   scountovf access check [DH: yes].
5. norm/Sscofpmf.yaml still cites cp_lcofi and cp_lcofip_priority (neither exists) and the run-together rule
   name count_overflow_triggermip_lcofip_mie_lcofie_op. Unchanged from the first review.

## Minor, unchanged
- OF_VEC has no `ifndef guard; defined three times.
- read_only bin matches any CSRRS with rs1 value 0, not only csrr.
- of_write_pattern checker_even holds OF_VEC bits 0,2,4... = mhpmevent3,5,7 = odd-numbered counters.
- sstatus_sie_clear (U) and mstatus_mie_clear (shared) are defined but used in no cross.

ayeshaanwaar05:
https://github.com/riscv/riscv-arch-test/pull/2294
this pr contains changes in ctp

davidharrishmc (approved):
LGTM!

config/whisper/whisper-rv64-max/rvmodel_macros.h
```diff
@@ -194,4 +194,21 @@
...
   sw      _R2, 0(_R1); /* clear source 2 interrupt */
 
+#define RVMODEL_SET_SSW_INT(_R1, _R2) \
+  li _R1, 1; \
+  li _R2, WHISPER_SSIP_ADDRESS; \
+  sw _R1, 0(_R2);
+
+#define RVMODEL_CLR_SSW_INT(_R1, _R2) \
+  li _R2, WHISPER_SSIP_ADDRESS; \
+  sw zero, 0(_R2);
```
jordancarlin: Why is this adding software interrupt macros? Seems unrelated to Sscofpmf.

jordancarlin (commented):
I pulled this out of the merge queue because of the changes to the rvmodel macro files, especially the `RVMODEL_SET_SSW_INT` macros. To avoid further delaying this PR, how about moving all of the RVMODEL changes into the corresponding test PR since that is where they are actually used. They are not needed for the coverpoints.

config/whisper/whisper-rv64-max/rvmodel_macros.h
ayeshaanwaar05: while resolving mergfe conflicts something like this would have happened

ayeshaanwaar05:
But removing macros from here would cause error because there are conditions defined ifdef SSCOFPMF_SUPPORTED these rvmodel macros must be define otherwise error

jordancarlin:
> But removing macros from here would cause error because there are conditions defined ifdef SSCOFPMF_SUPPORTED these rvmodel macros must be define otherwise error

I don't see those conditions being added anywhere in this PR.

ayeshaanwaar05:
> > But removing macros from here would cause error because there are conditions defined ifdef SSCOFPMF_SUPPORTED these rvmodel macros must be define otherwise error
> 
> I don't see those conditions being added anywhere in this PR.
Oh that was defined in the previous pr of coverpoints that was garbled and was closed. So should i remove these rvmodel macros from this pr ?

jordancarlin:
I think it would be a good idea to remove them for now since they will just confuse users and they are not needed until we add the tests. When we actually add the new RVMODEL macros, we will need to document them in the README and CTP.

coverpoints/priv/RISCV_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,132 @@
...
+                }
+        `else
+                // VSINH/VUINH (bits 59:58) hardwired 0 without H-ext -- only MINH/SINH/UINH vary
+                mhpmevent_xinh_combos: coverpoint ins.current.csr[CSR_MHPMEVENT3][62:60] {
+                bins combo[] = {[0:7]};
+                }
+        `endif
+    `else
+        `ifdef H_SUPPORTED
+                mhpmevent_xinh_combos: coverpoint ins.current.csr[CSR_MHPMEVENT3 + 12'h400][30:26] {
```
jordancarlin: Why not just use `CSR_MHPMEVENT3H` to make this easier to understand?

coverpoints/priv/RISCV_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,132 @@
...
+        mhpmevent_inhibits_pattern_state: coverpoint (ins.current.csr[CSR_MHPMEVENT3][62:58]) {
+                bins none_set  = {5'b00000};
+                bins msu_set   = {5'b11100};
+                bins minh_only = {5'b10000};
+                bins sinh_only = {5'b01000};
+                bins uinh_only = {5'b00100};
+        }
+    `else
+        // On RV32, MINH/SINH/UINH/VSINH/VUINH live in mhpmevent*h[30:26] (address + 0x400)
+        mhpmevent_inhibits_pattern_state: coverpoint (ins.current.csr[CSR_MHPMEVENT3 + 12'h400][30:26]) {
```
jordancarlin: CSR_MHPMEVENT3H?

coverpoints/priv/RISCV_coverage_sscofpmf.svh
```diff
@@ -0,0 +1,132 @@
...
+        mhpmevent_all_zero: coverpoint (ins.current.csr[CSR_MHPMEVENT3] == '0) {
+                bins yes = {1};
+        }
+    `else
+        mhpmevent_all_zero: coverpoint (ins.current.csr[CSR_MHPMEVENT3 + 12'h400] == '0) {
+                bins yes = {1};
+        }
+        mhpmevent_base_zero: coverpoint (ins.current.csr[CSR_MHPMEVENT3] == '0) {
+                bins yes = {1};
+        }
```
jordancarlin: Seems cleaner to combine this into one coverpoint that concatenates both CSRs and checks that the whole thing is 0.

coverpoints/priv/RISCV_coverage_sscofpmf.svh
jordancarlin: Applies lots of other places too.

jordancarlin (commented):
A few small comments on the coverpoints since I was looking anyway because of the macros.

jordancarlin (approved):

ayeshaanwaar05:
> I think it would be a good idea to remove them for now since they will just confuse users and they are not needed until we add the tests. When we actually add the new RVMODEL macros, we will need to document them in the README and CTP.

Removed. And in tests pr do i need to document in README and CTP rvmodel macros or will you do that ?

PR #2126, opened 2026-08-14
Fixed testbench and made debug register tests conditional on Sdext

davidharrishmc:
Removed stale interrupt signals from testbench so make coverage passes
Only test debug CSRs are inaccessible from M-mode if Sdext is supported; otherwise they are reserved and unpredictable.

### DIFF
--- a/framework/src/act/fcov/testbench.sv
+++ b/framework/src/act/fcov/testbench.sv
@@ -166,11 +166,6 @@ module testbench;
           "PC":             num = $sscanf(val, "%h", pc_rdata);
           "MODE":           num = $sscanf(val, "%d", mode);
           "MODE_VIRT":      num = $sscanf(val, "%d", mode_virt);
-          // Interrupts
-          //"M_EXT_INTR":     num = $sscanf(val, "%b", m_ext_intr);
-          //"S_EXT_INTR":     num = $sscanf(val, "%b", s_ext_intr);
-          //"M_TIMER_INTR":   num = $sscanf(val, "%b", m_timer_intr);
-          //"M_SOFT_INTR":    num = $sscanf(val, "%b", m_soft_intr);
           // Virtual Memory
           "VIRT_ADR_I":     num = $sscanf(val, "%h", virt_adr_i);
           "VIRT_ADR_D":     num = $sscanf(val, "%h", virt_adr_d);
--- a/framework/src/act/fcov/testbench.sv
+++ b/framework/src/act/fcov/testbench.sv
@@ -167,10 +167,10 @@ module testbench;
           "MODE":           num = $sscanf(val, "%d", mode);
           "MODE_VIRT":      num = $sscanf(val, "%d", mode_virt);
           // Interrupts
-          "M_EXT_INTR":     num = $sscanf(val, "%b", m_ext_intr);
-          "S_EXT_INTR":     num = $sscanf(val, "%b", s_ext_intr);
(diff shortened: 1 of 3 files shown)


framework/src/act/fcov/testbench.sv
```diff
@@ -167,10 +167,10 @@ module testbench;
...
           "MODE_VIRT":      num = $sscanf(val, "%d", mode_virt);
           // Interrupts
-          "M_EXT_INTR":     num = $sscanf(val, "%b", m_ext_intr);
-          "S_EXT_INTR":     num = $sscanf(val, "%b", s_ext_intr);
-          "M_TIMER_INTR":   num = $sscanf(val, "%b", m_timer_intr);
-          "M_SOFT_INTR":    num = $sscanf(val, "%b", m_soft_intr);
+          //"M_EXT_INTR":     num = $sscanf(val, "%b", m_ext_intr);
+          //"S_EXT_INTR":     num = $sscanf(val, "%b", s_ext_intr);
+          //"M_TIMER_INTR":   num = $sscanf(val, "%b", m_timer_intr);
+          //"M_SOFT_INTR":    num = $sscanf(val, "%b", m_soft_intr);
```
jordancarlin: You can go ahead and just delete these. Looks like I had this change locally and forgot to commit it in my previous PR.

jordancarlin:
> Only test debug CSRs are inaccessible from M-mode if Sdext is supported; otherwise they are reserved and unpredictable.

What is motivating this change? The spec seems to say otherwise. The following is from the Machine mode chapter and says nothing about changing if debug mode is not implemented.

> Machine-mode standard read-write CSRs 0x7A0‍–‍0x7BF are reserved for use by the debug system. Of these CSRs, 0x7A0‍–‍0x7AF are accessible to machine mode, whereas 0x7B0‍–‍0x7BF are only visible to debug mode. Implementations should raise illegal-instruction exceptions on machine-mode access to the latter set of registers.

davidharrishmc:
Good catch.  Allen had said that they are reserved during his CTP review back in April.  But I agree the spec says otherwise.

davidharrishmc:
@allenjbaum you had mentioned back in an April testplan review that debug registers shouldn't be tested unless debug is supported because they are reserved and unpredictable.  But Jordan points out that the spec says "Implementations should raise illegal-instruction exceptions on machine-mode access to [them]."  

Is there something more subtle (e.g. Should != MUST)?  Or should we test the debug registers are always inaccessible from M-mode?

allenjbaum:
Hmm, the spec is pretty clear there. 
In general, access to unimplemented CSRs is reserved - which was a change from an earlier version of the spec. So (I am guessing) when I made that statement, I was unaware of that explicit statement (which has always been there). 

On the other hand ... "should" not must. 
Sounds like we need another parameter, or we need the spec updated.

jordancarlin:
> Hmm, the spec is pretty clear there. In general, access to unimplemented CSRs is reserved - which was a change from an earlier version of the spec. So (I am guessing) when I made that statement, I was unaware of that explicit statement (which has always been there).
> 
> On the other hand ... "should" not must. Sounds like we need another parameter, or we need the spec updated.

This _hopefully_ seems like an instance where a spec clarification is needed vs a new parameter. Let's add it to the list for our ongoing spec clarification meetings.

davidharrishmc:
Agree that spec could be clarified but that this test is not needed.

PR #2129, opened 2026-08-17
S improvements

davidharrishmc:
Whisper now passes S in CI
Support all scause values in Ss1p12
Split CSR tests into chunks for faster runtime
Drop m-mode m/sret tests that are already in Sm and don't belong in S
Fix race condition in run_tests causing new line not to be printed
Turn on Ssqosid in Whisper since it is now supported in Sail (but Whisper doesn't yet enable mstateen0.srmcfg https://github.com/tenstorrent/whisper/issues/40)
Fast Trap handler defers to ordinary handler for traps during boot time.  This is needed because Sail traps on CSR access to mnstatus that is not yet implemented, but Whisper needs the register during boot.
Fixed whisper config file for nonzero stval on OP-P instructions https://github.com/tenstorrent/whisper/issues/43


config/whisper/whisper-rv32-max/whisper.json
```diff
@@ -4,7 +4,7 @@
   "address_translation_pmms": ["off"],
 
   // restore smmpm_smnpm_ssnpm when supported by Sail; affects mseccfg.PMM
```
jordancarlin: Drop this comment. Smmpm, Smnpm, and Ssnpm are all supported by Sail (and have been for a while).

config/spike/ci.yaml
```diff
@@ -3,7 +3,8 @@
 # CI configuration for Spike simulator
 # Shared across all Spike configs
 
-# SsstrictSm, SsstrictS, SsstrictU excluded because of Sail bugs and not all instructions in spike-max supported by Sail
+# SsstrictSm, SsstrictS, SsstrictU excluded because Spike does not presently aim to support Ssstrict.
+# https://github.com/riscv-software-src/riscv-isa-sim/issues/1976#issuecomment-5335375358
```
jordancarlin: Rather than excluding it here, just remove Ssstrict from the Spike UDB configs so that the tests are not selected in the first place. The UDB config should only include extensions that the DUT actually supports.

coverpoints/priv/S_coverage.svh
```diff
@@ -49,48 +49,50 @@ covergroup S_scause_cg with function sample(ins_t ins);
...
+            bins b_16_double_trap = {16}; // never delegated to S mode
+            bins b_17_reserved = {17};
+            bins b_18_software_check = {18}; // only can be triggered in hardware if SOFTWARE_CHECK_SUPPORTED but scause must be able to hold 0-31
+            bins b_19_hardware_error = {19}; // scause must be able to hold 0-31
+            // These can only be triggered in hardware if H supported, but scause must be able to hold 0-31
             bins b_20_instr_guest_page_fault = {20};
             bins b_21_load_guest_page_fault = {21};
             bins b_22_virtual_instruction = {22};
             bins b_23_store_guest_page_fault = {23};
+            bins b_31_24_custom[] = {[31:24]}; // scause must be able to hold 0-31
```
jordancarlin, line 69: A DUT that supports H or Ssdbltrp but not S 1.12 would not test some of the causes that it is required to support. Probably doesn't matter because it is unlikely that such a DUT would actually exist, but figured it was worth at least mentioning.

coverpoints/priv/S_coverage.svh
```diff
@@ -289,7 +278,7 @@ covergroup S_scsr_cg with function sample(ins_t ins);
     }
 
     cp_scsr_access:           cross priv_mode_s, csrname, csraccesses;
-    cp_scsrwalk:              cross priv_mode_s, csrname, csrop, walking_ones;
+    cp_scsrwalk:              cross priv_mode_s, csrwalk, csrop, walking_ones;
```
jordancarlin, line 281: Why are sepc and stval dropped? The testplan needs to be updated.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -434,27 +302,47 @@ def _generate_srets_tests(test_data: TestData) -> list[str]:
...
     # Standard S-mode CSRs
     # Format: (CSR Name, Mask).  Mask specifies a set of bits to check
 
+    # Create bit masks.  WPRI fields should be 0 to ignore reads.
+
+    # sstatus bit mask
+    sstatus_mask = (
+        (1 << 1)  # SIE:  Supervisor Interrupt Enable
+        | (1 << 5)  # SPIE: Supervisor Previous Interrupt Enable
+        | (0 << 6)  # UBE not yet supported by Sail; change to 1 when supported
```
jordancarlin: Make this comment a TODO (or better yet, open an issue) so we don't forget to update this later.

tests/env/rvtest_trap_handler.h
```diff
@@ -2610,6 +2610,22 @@ excpt_\__MODE__\()hndlr_tbl:
...
+        // that does not). Ask whether xEPC is in the test code segment rather than
+        // inferring it from x2's contents. This compares unrelocated addresses, which is
+        // already the fast path's assumption: RVTEST_SIGUPD_FAST_TRAP writes through x2 raw.
+        SREG a2, rvmodel_sv_off+3*REGWIDTH(a0)  // stash caller's a2 (rvmodel_sv slot 3 is free)
+        csrr a1, mepc
+        LREG a2, code_bgn_off(a0)       // a2 = rvtest_code_begin
+        sub  a1, a1, a2                 // a1 = mepc - code_begin (wraps if mepc is below it)
+        LREG a2, code_seg_siz(a0)       // a2 = code segment size
+        bgeu a1, a2, fast_Mbootrap      // outside the test code — use the standard handler
+        LREG a2, rvmodel_sv_off+3*REGWIDTH(a0)  // restore caller's a2
```
jordancarlin, line 2628: This adds a lot of instructions to the fast handler. What about just moving the initialization of x2 earlier so it is initialized before any of these possible traps? Then the fast trap handler doesn't need to change at all.

run_tests.py
```diff
@@ -86,8 +86,14 @@ def _simulator_error_lines(log_file: Path, limit: int = 3) -> list[str]:
...
     command: str, log_dir: Path, elf_dir: Path, elf_path: Path, verbose: bool, timeout: int
-) -> tuple[bool, Path, str]:
-    """Run a single ELF and return (failed, elf_path, rvcp_summary_line)."""
+) -> tuple[bool, Path, str, str]:
+    """Run a single ELF and return (failed, elf_path, rvcp_summary_line, fail_message).
+
+    fail_message is returned rather than printed here: this runs in a Pool worker, and all
+    workers share the parent's stdout. print() emits the message and its terminating newline
+    as two separate writes, so on a line-buffered stdout (i.e. a terminal) another worker can
+    flush between them and glue its message onto this one's last line. The parent prints.
```
jordancarlin: This shouldn't be in the function docstring.

jordancarlin (changes requested):

coverpoints/priv/S_coverage.svh
davidharrishmc: Seems too obscure to worry about.  In the meantime, a few points don't get hit here but they will be tested anyway by the H suite.

coverpoints/priv/S_coverage.svh
davidharrishmc: Dropped because it is hard to express what values are legal.  Test plan updated.

tests/env/rvtest_trap_handler.h
davidharrishmc: Spent a good deal of time trying that.  x2 also gets changed when initializing registers, and it was a pain patching it all up.  Moved on in the interest of expedience.

jordancarlin: What about loading x2 twice when the fast trap handler is enabled? Once at the beginning before loading the trap handler and once where it is currently being loaded.

jordancarlin:
Looks like pre-commit is failing.

jordancarlin (approved):
Approving this. I might look at backing out the trap handler changes later.

PR #2143, opened 2026-08-20
Sm full coverage on CVW and Spike, improved coverage elsewhere

davidharrishmc:
* CVW and Spike now passes Sm in CI
* Many fixes on Whisper, QEMU, Imperas
* Downgraded spike-rv32-max to Sm1.12 because it does not support medelegh, filed issue
* Broke RO tests into chunks for faster runtime
* Avoid writing mstatus/menvcfg/mseccfg bits that lock up machine
* Fix many configuration errors for reference model to match DUTs
* Fix UDB config issue for Imperas
* Exclude testing reserved values of CSR fields (MPP/PMM/CBIE)
* Defined Imperas CLINT_address to support mtime and msip, which should enable some interrupt tests.
* Restore misa after attempting to change it, so it is not left in an unsafe state


config/sail/sail-rv32-max/sail.json
```diff
@@ -96,7 +96,7 @@
       // does not exist.
       "delegatable_bits": {
         "len": "xlen",
-        "value": "0x0000_2AAA"
```
jordancarlin, line 99: Why did this need to change?

config/sail/sail-rv64-max/sail.json
```diff
@@ -96,7 +96,7 @@
       // does not exist.
       "delegatable_bits": {
         "len": "xlen",
-        "value": "0x0000_0000_0000_2AAA"
+        "value": "0x0000_0000_0000_2222"
```
jordancarlin, line 99: Why did this need to change?

config/whisper/whisper-rv32-max/whisper-rv32-max.yaml
```diff
@@ -289,30 +289,32 @@ params:
...
+      false,
+      false,
+      false,
+      false,
+      false,
     ]
   HPM_EVENTS: [0]
   MCOUNTINHIBIT_IMPLEMENTED: true
+  # Whisper's mcountinhibit mask is 0xFFFFFFFD regardless of num_mmode_perf_regs, but UDB requires
+  # COUNTINHIBIT_EN[n] to be false whenever HPM_COUNTER_EN[n] is false, so bits 11-31 are listed as false.
```
jordancarlin, line 317: Open an issue on UDB. The spec does not require this so the parameters should not either.

config/spike/spike-rv64-max/sail.json
```diff
@@ -15,7 +15,7 @@
     // (writable_hpm_counters[2 .. 0]) are ignored.
     "writable_hpm_counters": {
       "len": 32,
-      "value": "0xFFFF_FFFF"
+      "value": "0x0000_00007"
     },
```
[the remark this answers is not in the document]
davidharrishmc: This presents a bit of a conflict.  Sscofpmf requires at least one true HPM_COUNTER_EN in UDB.  Spike doesn't yet support HPM counters.  Leave as is for now while Spike is in flux.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -426,88 +497,107 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
+        coverpoint,
+        "Set and clear each bit individually in all writable M-mode CSRs",
+    )
+
+    tc = test_data.new_test_chunk(test_chunks)
+    warl_fields = [("mpp", 11, 2, 0b10)]
+    tc.code.extend(
+        csr_walk_test(
+            test_data, ("mstatus", mstatus_mask), covergroup, coverpoint, warl_fields=warl_fields, maskedwrites=True
+        )
```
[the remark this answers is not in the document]
davidharrishmc: works as is

config/whisper/whisper-rv32-max-spike-ref/whisper.json
```diff
@@ -31,6 +31,8 @@
   },
 
   "csr": {
+    "misa": { "reset": "0x4034112f" }, // acdfimsuv + B (bit 1). Whisper only derives misa from the
+    // ISA string under --newlib/--linux, so set it explicitly.
```
jordancarlin, line 35: Might be worth an issue on Whisper. It should probably default to setting the correct `misa` string.

config/whisper/whisper-rv32-max-spike-ref/whisper.json
```diff
@@ -31,6 +31,8 @@
   },
 
   "csr": {
+    "misa": { "reset": "0x4034112f" }, // acdfimsuv + B (bit 1). Whisper only derives misa from the
```
jordancarlin, line 34: Why is B called out separate from the other extensions?

config/whisper/ci.yaml
```diff
@@ -2,9 +2,10 @@
 # Jordan Carlin jcarlin@hmc.edu April 2026
 # CI configuration for Whisper simulator
 
-# Sm lacks sufficient parameters
-# SsstrictSm, SsstrictS, SsstrictU excluded because not all instructions in whisper-max supported by Sail
-# Smstateen https://github.com/tenstorrent/whisper/issues/40 mstateen0.SRMCFG read-only 0 should be writable when Ssqosid supported
+# Sm: menvcfg.SSE=0 (whisper#46), mstatush not writable (whisper#47)
```
jordancarlin: Joe said this was fixed.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -178,12 +178,18 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
-        wildcard bins csrrc_all = {CSRRC} iff (ins.current.rs1_val == '1); // csrc all ones
-        wildcard bins csrrw0    = {CSRRW} iff (ins.current.rs1_val ==  0); // csrw all zeros
-        wildcard bins csrrw1    = {CSRRW} iff (ins.current.rs1_val == '1); // csrw all ones
-        wildcard bins csrrs_all = {CSRRS} iff (ins.current.rs1_val == '1); // csrs all ones
-        wildcard bins csrr      = {CSRR}  iff (ins.current.rs1_val ==  0); // csrr
+        wildcard bins csrrc_all = {CSRRC} iff (ins.current.rs1_val != 0); // csrc mask or all ones
+        wildcard bins csrrw0    = {CSRRW} iff (ins.current.rs1_val == 0); // csrw all zeros
+        wildcard bins csrrw1    = {CSRRW} iff (ins.current.rs1_val != 0); // csrw mask or all ones
+        wildcard bins csrrs_all = {CSRRS} iff (ins.current.rs1_val != 0); // csrs mask or all ones
+        wildcard bins csrr      = {CSRR}  iff (ins.current.rs1_val == 0); // csrr
```
jordancarlin: Changing these to `!= 0` makes them functionally pointless because they will be covered by the walking-ones tests. Let's split this into two coverpoint variants instead, `csraccess` for most CSRs and `csraccess_masked` (or something similar) for the special CSRs.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -422,7 +428,47 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
 
     cp_mcsr_access:             cross priv_mode_m, mcsrname, csraccesses;
     cp_mcsr_access_ro:          cross priv_mode_m, mcsrname_ro, csraccesses;
-    cp_mcsrwalk :               cross priv_mode_m, mcsrname, csrop, walking_ones;
+    // The Sm CSR walk tests use masked writes for the CSRs whose reserved/unimplemented bits are
+    // dangerous to poke (see csr_walk_test(..., maskedwrites=True) in
+    // generators/testgen/src/testgen/priv/extensions/Sm.py).  The walking one is ANDed with the
+    // CSR's writable mask before the csrs/csrc, so a bit outside the mask is written as zero,
+    // rs1_val is not one-hot, and walking_ones never samples.  Those cross bins are therefore
+    // unreachable by construction.  Keep the lists below in sync with the masks in Sm.py.
```
jordancarlin: Trim this comment down. The internal details of how csr_walk_test works are not relevant to the coverpoint and are likely to get out of sync.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -422,7 +428,47 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+    cp_mcsrwalk :               cross priv_mode_m, mcsrname, csrop, walking_ones {
+        // mstatus_mask covers bits 1, 3, 5, 7:23, 31, 41 and 63.
+        ignore_bins mstatus_unwritable = binsof(mcsrname.mstatus) &&
+            binsof(walking_ones) intersect {0, 2, 4, 6, [24:30], [32:40], [42:62]};
```
jordancarlin: - Why are UBE (6), SBE (36), and MBE (37) excluded?
- Why are SDT (24) and MDT (42) excluded?
- Why are GVA (38) and MPV (39) excluded?

If the only reason for excluding those bins is that they do not match Sail because Sail doesn't have the required extension yet, then exclude them in the test but keep them in the coverage. That makes it impossible to forget that we need to come back and reenable those bits in the test once Sail supports them.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -422,7 +428,47 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        `ifdef SM1P12P0_OR_LATER_SUPPORTED
+            // menvcfg_mask covers bits 0, 2:7, 32, 33, 61, 62 and 63.
+            ignore_bins menvcfg_unwritable = binsof(mcsrname.menvcfg) &&
+                binsof(walking_ones) intersect {1, [8:31], [34:60]};
```
jordancarlin: Same question about DTE (59) and CDE (60). Seems like we should be able to test those.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -422,7 +428,47 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        `endif
+        `ifdef UDB_MXLEN_32
+            // mstatush is masked to bit 9 (MPELP) only.
+            ignore_bins mstatush_unwritable = binsof(mcsrname.mstatush) &&
+                binsof(walking_ones) intersect {[0:8], [10:31]};
```
jordancarlin: Seems like we should be able to test 4:7 and 9:10.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -422,7 +428,47 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+            `ifdef SM1P12P0_OR_LATER_SUPPORTED
+                // menvcfgh is the top half of menvcfg_mask: bits 0, 1, 29, 30, 31.
+                ignore_bins menvcfgh_unwritable = binsof(mcsrname.menvcfgh) &&
+                    binsof(walking_ones) intersect {[2:28]};
```
jordancarlin: Should be able to test DTE and CDE (27 and 28).

coverpoints/priv/Sm_coverage.svh
```diff
@@ -422,7 +428,47 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        // medelegh has no writable bits in these configs.
+        `ifdef MSECCFG_SUPPORTED
+            ignore_bins mseccfg_not_walked = binsof(mcsrname.mseccfg);
+            `ifdef UDB_MXLEN_32
+                ignore_bins mseccfgh_not_walked = binsof(mcsrname.mseccfgh);
+            `endif
+        `endif
+        `ifdef UDB_MXLEN_32
+            `ifdef SM1P13P0_OR_LATER_SUPPORTED
+                ignore_bins medelegh_not_walked = binsof(mcsrname.medelegh);
```
jordancarlin: Should we be testing bits 32-47 (the non-custom ones)? We aren't excluding the upper half of `medeleg` on RV64 here, so we shouldn't exclude `medelegh` either.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -422,7 +428,47 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+            `endif
+        `endif
+
+        // mseccfg/mseccfgh and medelegh appear in the access tests but are deliberately not
+        // walked: mseccfg's Smepmp bits (MML/MMWP) lock the machine out of its own memory, and
+        // medelegh has no writable bits in these configs.
+        `ifdef MSECCFG_SUPPORTED
+            ignore_bins mseccfg_not_walked = binsof(mcsrname.mseccfg);
+            `ifdef UDB_MXLEN_32
+                ignore_bins mseccfgh_not_walked = binsof(mcsrname.mseccfgh);
```
jordancarlin: Why is `mseccfg `fully excluded unlike the other CSRs that only skip specific fields? Seems like we could test MLPE, PMM, USEED, and SSEED.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -313,29 +315,85 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
+        | (0 << 38)  # GVA not supported by Sail; TODO change to 1 when H is implemented
+        | (0 << 39)  # MPV not supported by Sail; TODO change to 1 when H is implemented
+        | (1 << 41)  # MPELP: Machine Previous Expect Landing Pad
+        | (0 << 42)  # MDT:   not yet supported by Sail; TODO change to 1 when Smdbltrp implemented
+        | (1 << 63)  # SD for RV64
+    )
+    mseccfg_mask = (
+        (0 << 0)  # Smepmp MML not supported TODO: change these to 1 when Sail implements & boot code sets it up
+        | (0 << 1)  # Smepmp MMWP not supported
+        | (0 << 2)  # Smepmp RLB not supported
```
jordancarlin, line 357: I don't think we want to mess with these three bits even once Sail supports them. They can lock us out of M-mode memory.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -313,29 +315,85 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
+        ("mie", 0x3EEE),  # limit to standard interrupt bits
         ("mtvec", 0b10),  # mtvec.MODE[1] must be 0. Legal values for BASE are hard to describe with a reference model
         ("mcounteren", None),
         ("mscratch", None),
         ("mepc", None),
         #        ("mcause", None), # WLRL fields can't be handled with masks.  Use cp_mcause_* instead
         ("mtval", None),
-        ("mip", 0xFFFF),  # limit to standard interrupt bits
-        ("mcountinhibit", None),
+        ("mip", 0x3EEE),  # limit to standard interrupt bits
```
jordancarlin: Why is the `mip` mask changing? The spec defines bits 0, 4, 8, and 14-15 as 0 (not reserved), so I think we should be testing them.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -313,29 +315,85 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
-        # TODO: sail does not yet support sstatus.S/M/UBE; mask it until available to avoid mismatches.  Delete mask when Sail has endian support.
-        ("mstatus", 0xFFFFFFCFFFFFFFBF),
         (
             "medeleg",
             0xDBBFE,
         ),  # mask off custom bits and reserved bits; instr misaligned [0] depends on ZCA_SUPPORTED so don't check it
-        ("mideleg", 0xFFFF),  # limit to standard interrupt bits
-        ("mie", 0xFFFF),  # limit to standard interrupt bits
+        ("mideleg", 0x3EEE),  # limit to standard interrupt bits
+        ("mie", 0x3EEE),  # limit to standard interrupt bits
```
jordancarlin: Why is the `mie` mask changing? The spec defines bits 0, 4, 8, and 14-15 as 0 (not reserved), so I think we should be testing them.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -313,29 +315,85 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
     csrs = [
-        # TODO: sail does not yet support sstatus.S/M/UBE; mask it until available to avoid mismatches.  Delete mask when Sail has endian support.
-        ("mstatus", 0xFFFFFFCFFFFFFFBF),
         (
             "medeleg",
             0xDBBFE,
         ),  # mask off custom bits and reserved bits; instr misaligned [0] depends on ZCA_SUPPORTED so don't check it
-        ("mideleg", 0xFFFF),  # limit to standard interrupt bits
-        ("mie", 0xFFFF),  # limit to standard interrupt bits
+        ("mideleg", 0x3EEE),  # limit to standard interrupt bits
```
jordancarlin: I think we should be able to test the full 0xFFFF here.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -313,29 +315,85 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
         #        ("mcause", None), # WLRL fields can't be handled with masks.  Use cp_mcause_* instead
         ("mtval", None),
-        ("mip", 0xFFFF),  # limit to standard interrupt bits
-        ("mcountinhibit", None),
+        ("mip", 0x3EEE),  # limit to standard interrupt bits
+        # Only check CY (bit 0) and IR (bit 2). Bit 1 (TM) is always read-only zero. The HPMn inhibit bits are
+        # WARL and may be writable even for unimplemented counters (priv spec 3.1.12 does not require them to be
+        # read-only zero); Spike and Whisper implement all 29 inhibit bits while hardwiring most counters to zero.
+        # Sail derives the writable mcountinhibit bits from writable_hpm_counters, so it cannot model that.
+        ("mcountinhibit", 0b101),
```
jordancarlin: If bit 1 is always read-only zero, we should be testing it because it has clearly defined behavior.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -366,57 +424,70 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
         ("mhpmevent31", 0),  # mask all bits because they are WARL and can all be ROZ
     ]
-    csr_menvcfg = ("menvcfg", None)
+    csr_menvcfg = ("menvcfg", menvcfg_mask)
     # RV32-only high CSRs
-    csr_mstatush = ("mstatush", None)
-    csr_menvcfgh = ("menvcfgh", None)
+    csr_mstatush = (
+        "mstatush",
+        (mstatus_mask >> 32) & 0x7FFFFFF0,
```
jordancarlin: Why is there an additional mask being applied here? Shouldn't it just be the right-shift by 32?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -366,57 +424,70 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
-    lines.append("#endif //  SM1P12P0_OR_LATER_SUPPORTED")
-    lines.append("\n#ifdef MSECCFG_SUPPORTED")
-    lines.extend(csr_access_test(test_data, ("mseccfgh", None), covergroup, coverpoint))
-    lines.append("#endif // MSECCFG")
-    lines.append("\n#ifdef SM1P13P0_OR_LATER_SUPPORTED")
-    lines.extend(csr_access_test(test_data, ("CSR_MEDELEGH", None), covergroup, coverpoint))
-    lines.extend(
+    tc.code.append("\n#ifdef SM1P12P0_OR_LATER_SUPPORTED")
+    tc.code.extend(
+        csr_access_test(test_data, ("menvcfgh", menvcfg_mask >> 32), covergroup, coverpoint, maskedwrites=True)
```
jordancarlin: You defined `csr_menvcfgh` as the tuple `("menvcfgh", menvcfg_mask >> 32)` above. Either use that here or delete `csr_menvcfgh`.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -636,25 +726,25 @@ def _generate_mcsr_tests(test_data: TestData) -> list[str]:
...
     ######################################
     coverpoint = "cp_misa_clear_c"
     ######################################
 
-    lines.append(
-        comment_banner(
-            coverpoint,
-            "Try to clear misa.C.  Should not change if PC is at 2-byte aligned address",
-        ),
+    tc = test_data.new_test_chunk(test_chunks, "misa_clear_c")
```
jordancarlin: It probably makes sense to just use the suffix "misa" for all of the misa related test_chunks so that they are not split into too many files.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -1104,12 +1194,11 @@ def make_sm(test_data: TestData) -> list[TestChunk]:
...
-    tc = test_data.begin_test_chunk("mcsr")
-    tc.code.extend(_generate_mcsr_tests(test_data))
-    test_chunks.append(test_data.end_test_chunk())
-
     tc = test_data.begin_test_chunk("mcsr_cntr")
     tc.code.extend(_generate_mcsr_cntr_tests(test_data))
     test_chunks.append(test_data.end_test_chunk())
 
+    _generate_mcsr_tests(test_data, test_chunks)
+    test_chunks.append(test_data.end_test_chunk())
```
jordancarlin: Probably makes more sense for the `test_data.end_test_chunk()` to be inside `_generate_mcsr_tests`.

jordancarlin (changes requested):

config/sail/sail-rv32-max/sail.json
davidharrishmc: Most of the sims including Sail don't allow delegating M-mode interrupts to S-mode.  The spec doesn't forbid it, but the example in the spec doesn't show it.  Also, the CRD says not to test M-mode interrupts.  Seemed best to just not test this delegation than to patch up many simulators to test an obscure case people don't use.

config/sail/sail-rv64-max/sail.json
davidharrishmc: see above

config/whisper/whisper-rv32-max-spike-ref/whisper.json
davidharrishmc: Because that is the part of the MISA string that is not automatically generated correctly from the ISA string by Whisper.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: SD is not in bit 31 of mstatush, so masking that.  Latest code already masks sxl/uxl so taking the 0 out.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: If Smepmp is supported,  I think we want the boot code to conditionally configure it to a safe state to allow testing these bits.

config/whisper/whisper-rv32-max-spike-ref/whisper.json
davidharrishmc: https://github.com/tenstorrent/whisper/issues/48
Will delete this misa when Whisper issue is closed.

coverpoints/priv/Sm_coverage.svh
davidharrishmc: UBE/SBE/MBE are excluded because they crash the system because the framework isn't designed to switch endianness midstream.  Reenabled the others.

coverpoints/priv/Sm_coverage.svh
davidharrishmc: Agreed.

coverpoints/priv/Sm_coverage.svh
davidharrishmc: 32-47 are reserved, and we haven't been testing reserved fields.

config/whisper/whisper-rv32-max/whisper-rv32-max.yaml
davidharrishmc: https://github.com/riscv/riscv-unified-db/issues/2511

coverpoints/priv/Sm_coverage.svh
```diff
@@ -421,8 +435,42 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
     cp_mcsr_access:             cross priv_mode_m, mcsrname, csraccesses;
+    cp_mcsr_access_masked:      cross priv_mode_m, mcsrname_masked, csraccesses_masked;
     cp_mcsr_access_ro:          cross priv_mode_m, mcsrname_ro, csraccesses;
     cp_mcsrwalk :               cross priv_mode_m, mcsrname, csrop, walking_ones;
+    // The masked-write CSRs are walked through their writable mask only, so the bits outside the
+    // mask can never be one-hot sources.  Keep the lists below in sync with the masks in Sm.py.
+    cp_mcsrwalk_masked :        cross priv_mode_m, mcsrname_masked, csrop, walking_ones {
+        // mstatus_mask covers bits 1, 3, 5, 7:23, 31, 41 and 63.
+        ignore_bins mstatus_unwritable = binsof(mcsrname_masked.mstatus) &&
+            binsof(walking_ones) intersect {0, 2, 4, 6, [25:30], [32:37], 40, [43:62]};
```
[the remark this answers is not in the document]
davidharrishmc: Intentional.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -336,8 +336,8 @@ def _generate_scsr_tests(test_data: TestData, test_chunks: list[TestChunk]) -> N
         ("stvec", 0b10),
         ("scounteren", None),
         ("sscratch", None),
-        ("sip", 0x3EEE),  # only test standard non-reserved portion
-        ("sie", 0x3EEE),  # only test standard non-reserved portion
+        ("sip", 0xFFFF),  # only test standard non-reserved portion
+        ("sie", 0xFFFF),  # only test standard non-reserved portion
```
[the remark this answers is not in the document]
davidharrishmc: These are standard.

config/sail/sail-rv32-max/sail.json
jordancarlin: The other sims don't need to support it, but I don't see any harm in leaving those delegation bits writable in Sail since it does support it.

coverpoints/priv/Sm_coverage.svh
jordancarlin: Then we should be consistent about it for rv64 and not test those bits in `medeleg` either.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -421,8 +435,42 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        `endif
+        `ifdef MSECCFG_SUPPORTED
+            // mseccfg is accessed but deliberately not walked: its Smepmp bits (MML/MMWP) would lock
+            // the machine out of its own memory.  mseccfgh (RV32) is walked, masked to PMM.
+            ignore_bins mseccfg_not_walked = binsof(mcsrname_masked.mseccfg);
+            `ifdef UDB_MXLEN_32
+                // mseccfgh is the top half of mseccfg_mask: bits 0 and 1 (PMM).
+                ignore_bins mseccfgh_unwritable = binsof(mcsrname_masked.mseccfgh) &&
+                    binsof(walking_ones) intersect {[2:31]};
+            `endif
```
jordancarlin: This seems inconsistent. Why would we only walk those bits in mseccfgh and not in mseccfg? Also, we should be able to walk USEED and SSEED.

coverpoints/priv/Sm_coverage.svh
davidharrishmc: Deleted medelgh_not_walked.  The mask prevents checking, but there is no harm attempting the walk.

coverpoints/priv/Sm_coverage.svh
davidharrishmc: Coverpoint fixed.

jordancarlin (approved):
LGTM

PR #2154, opened 2026-08-22
T-SBI conversion of S, UF, UV: run in S-mode, move M-mode coverpoints to Sm

davidharrishmc:
**Stacked on #2143** (`sm2`). Only the last commit, `TSBI support for S, including moving some parts to Sm`, is new here; the rest is #2143 and will drop out of the diff once it merges.

## What this does

The `S` suite now boots to S-mode and stays there, using T-SBI for the one privileged operation it still needs. The coverpoints that genuinely require M-mode move to `Sm`, gated on `S_SUPPORTED`.

**`S` (`S.py`, `S_coverage.svh`)**
- `cp_sret_s` no longer sweeps `mstatus.TSR` or touches `medeleg`. It runs the 8-case `spp × spie × sie` sweep in S-mode and returns after each `sret` with a single `RVTEST_TSBI_GOTO_SMODE` instead of `RVTEST_GOTO_MMODE` + `RVTEST_GOTO_LOWER_MODE Smode`. The `sstatus` readback is masked with `S_SSTATUS_MASK`.
- `cp_shadow`, `cp_scsr_from_m`, and the TSR dimension of `cp_sret_s` are removed from `S`. `cp_shadow` cannot be expressed through T-SBI at all: it samples an M-mode `csrw` *immediately* followed by the shadow `csrr`.
- `S_SSTATUS_MASK`, `S_CSRS`, `S_CSRS_NOWALK`, `S_CSR_SENVCFG` are module-level so `Sm` imports them instead of retyping.

**`Sm` (`Sm.py`, `Sm_coverage.svh`)** — each `#ifdef S_SUPPORTED` / `` `ifdef S_SUPPORTED ``:
- `_generate_sret_s_tests`: the full `sret` sweep including TSR (direct CSR access, M-mode start/end), in the `xret` chunk; `cp_sret_s` cross in `Sm_mprivinst_cg` with `old_mstatus_tsr`.
- `_add_shadow` + a new `shadow` chunk; `cp_shadow` cross in `Sm_mcsr_cg`.
- `cp_scsr_from_m` cross in `Sm_mcsr_cg` (new `scsrname` coverpoint, reusing Sm's `csraccesses`); the test itself was already in `Sm`.

* UF and UV: trivial changes to boot to U mode.
* Chunked UV to reduce file sizes

## Validation

- `S` (8 files) and `Sm` (27 files), both XLENs: Sail, Spike, CVW all pass; `S` also passes on Whisper, Imperas, and both `sail-*-clang` configs. QEMU fails `S` on its known `sstatus.UXL`, `sip.LCOFIP`, and reserved-`senvcfg.CBIE` defects — identical to before this change.
- `Sm` failures on Whisper/Imperas/QEMU are exactly the already-documented set; the two moved files fail on QEMU only for its known `LCOFIP`/`LCOFIE` (bit 13) defect, as they did when they lived in `S`.
- `make coverage EXTENSIONS=S,Sm`: 100% on all eight covergroups, both XLENs. `cp_sret_s` is 8/8 bins in `S` and 16/16 in `Sm`.
- Dynamic cost: `S-00` drops from 7,034 → 3,931 instructions and 33 → 12 traps on Sail RV64 (TSR half moved out; one T-SBI ecall per `sret` case instead of two mode hops). Every other `S` file has identical instruction and trap counts on every model.

https://claude.ai/code/session_01JmYvYtC2wx93qJ7pH5YPQV

### DIFF
--- a/coverpoints/norm/S.yaml
+++ b/coverpoints/norm/S.yaml
@@ -18,7 +18,11 @@ normative_rule_definitions:
   # mode if the SPP bit is 0, or supervisor mode if the SPP bit is 1; SPP is then
   # set to 0.
   - names: [sstatus_spp_sz, sstatus_spp_acc, sstatus_spp_op]
-    coverpoint: ["S_sprivinst_cg/cp_sret_s"]
+    coverpoint:
+      [
+        "S_sprivinst_cg/cp_sret_s",
+        "Sm_mprivinst_cg/cp_sret_s (sweep with mstatus.TSR, runs in Sm)",
+      ]
 
   # The SIE bit enables or disables all interrupts in supervisor mode. When SIE is
   # clear, interrupts are not taken while in supervisor mode. When the hart is
@@ -35,7 +39,7 @@ normative_rule_definitions:
   - names: [sstatus_spie_sz, sstatus_spie_acc, sstatus_spie_op]
     coverpoint:
       [
-        "InterruptsS_cg/cp_user_mti, ExceptionsS_cg/cp_medeleg_msu_illegalinstruction, S_sprivinst_cg/cp_sret_s",
+        "InterruptsS_cg/cp_user_mti, ExceptionsS_cg/cp_medeleg_msu_illegalinstruction, S_sprivinst_cg/cp_sret_s, Sm_mprivinst_cg/cp_sret_s",
       ]
 
   # The UXL field controls the value of XLEN for U-mode, termed UXLEN, which may
@@ -356,7 +360,11 @@ normative_rule_definitions:
   # although CSRs and instructions are associated with one privilege level, they are
   # also accessible at all higher privilege levels.
   - name: Zicsr_higher_priv
-    coverpoint: ["S_scsr_cg/cp_scsr_from_m", "S_scsr_cg/cp_ucsr_from_s"]
+    coverpoint:
+      [
+        "Sm_mcsr_cg/cp_scsr_from_m (S-mode CSRs from M-mode; moved to the Sm suite)",
+        "S_scsr_cg/cp_ucsr_from_s",
+      ]
 
   # The top two bits (csr[11:10]) indicate whether the register is read/write
   # (00,01, or 10) or read-only (11).
--- a/coverpoints/norm/S.yaml
+++ b/coverpoints/norm/S.yaml
@@ -360,11 +360,7 @@ normative_rule_definitions:
   # although CSRs and instructions are associated with one privilege level, they are
   # also accessible at all higher privilege levels.
   - name: Zicsr_higher_priv
-    coverpoint:
-      [
-        "Sm_mcsr_cg/cp_scsr_from_m (S-mode CSRs from M-mode; moved to the Sm suite)",
-        "S_scsr_cg/cp_ucsr_from_s",
-      ]
+    coverpoint: ["Sm_mcsr_cg/cp_scsr_from_m", "S_scsr_cg/cp_ucsr_from_s"]
 
   # The top two bits (csr[11:10]) indicate whether the register is read/write
   # (00,01, or 10) or read-only (11).
--- a/generators/testgen/src/testgen/priv/extensions/S.py
+++ b/generators/testgen/src/testgen/priv/extensions/S.py
@@ -15,7 +15,7 @@ from testgen.data.state import TestData
 from testgen.data.test_chunk import TestChunk
 from testgen.priv.registry import add_priv_test_generator
 
-# Standard S-mode CSRs, shared with the Sm suite (cp_scsr_from_m); import rather than retype.
+# Standard S-mode CSRs, shared with the Sm suite (cp_scsr_from_m)
 # Format: (CSR Name, Mask).  Mask specifies a set of bits to check
 
 # Create bit masks.  WPRI fields should be 0 to ignore reads.
@@ -28,7 +28,7 @@ S_SSTATUS_MASK = (
     | (1 << 8)  # SPP:  Supervisor Previous Privilege
     | (3 << 9)  # VS:   Vector Status
     | (3 << 13)  # FS:   Floating-Point Status
-    | (3 << 15)  # XS:   User-Mode Extension Status
+    | (3 << 15)  # XS:   Custom Extension Status
     | (1 << 18)  # SUM:  Supervisor User Memory Access
     | (1 << 19)  # MXR:  Make eXecutable Readable
     | (1 << 23)  # SPELP: Supervisor Previous Expect Landing Pad
@@ -261,11 +261,7 @@ def _generate_priv_inst_tests(test_data: TestData) -> list[str]:
 
 
 def _generate_srets_tests(test_data: TestData) -> list[str]:
-    """
-    Generate sret from S-mode with spp, spie, sie sweep (no TSR: that needs M-mode and lives in Sm).
-
-    Runs entirely in S-mode; the only privileged operation is the T-SBI GOTO_SMODE after each sret.
-    """
+    """Generate sret from S-mode with spp, spie, sie sweep (no TSR: that needs M-mode and lives in Sm)."""
     ######################################
     covergroup = "S_sprivinst_cg"
     coverpoint = "cp_sret_s"
@@ -330,10 +326,6 @@ def _generate_scsr_tests(test_data: TestData, test_chunks: list[TestChunk]) -> N
     """Generate CSR tests, one test chunk per CSR so they can be split across files."""
     covergroup = "S_scsr_cg"
 
-    # S-mode CSR lists are module-level (shared with Sm.py); bind the local names used below
-    csrs = S_CSRS
-    csrs_nowalk = S_CSRS_NOWALK
-    csr_senvcfg = S_CSR_SENVCFG
     # Floating-point CSRs
     csrf = [("fflags", None), ("frm", None), ("fcsr", None)]
     # Vector CSRs
@@ -356,13 +348,13 @@ def _generate_scsr_tests(test_data: TestData, test_chunks: list[TestChunk]) -> N
         "Read, write all 1s, write all 0s, set all 1s, set all 0s, restore all S-mode CSRs",
     )
 
-    for csr in csrs + csrs_nowalk:
+    for csr in S_CSRS + S_CSRS_NOWALK:
         tc = test_data.new_test_chunk(test_chunks)
         tc.code.extend(csr_access_test(test_data, csr, covergroup, coverpoint))
 
     tc = test_data.new_test_chunk(test_chunks)
     tc.code.extend(["", "#ifdef S1P12P0_OR_LATER_SUPPORTED"])
-    tc.code.extend(csr_access_test(test_data, csr_senvcfg, covergroup, coverpoint))
+    tc.code.extend(csr_access_test(test_data, S_CSR_SENVCFG, covergroup, coverpoint))
     tc.code.extend(["", "#endif"])
 
     ######################################
@@ -396,7 +388,7 @@ def _generate_scsr_tests(test_data: TestData, test_chunks: list[TestChunk]) -> N
         "Set and clear each bit individually in all writable S-mode CSRs",
     )
 
-    for csr in csrs:
+    for csr in S_CSRS:
         tc = test_data.new_test_chunk(test_chunks)
         tc.code.extend(csr_walk_test(test_data, csr, covergroup, coverpoint))
 
@@ -407,7 +399,7 @@ def _generate_scsr_tests(test_data: TestData, test_chunks: list[TestChunk]) -> N
     # legalize to any legal value, so those iterations check that the field is legal
     # instead of exact-matching the reference model.
     warl_fields = [("cbie", 4, 2, 0b10), ("pmm", 32, 2, 0b01)]
-    tc.code.extend(csr_walk_test(test_data, csr_senvcfg, covergroup, coverpoint, warl_fields=warl_fields))
+    tc.code.extend(csr_walk_test(test_data, S_CSR_SENVCFG, covergroup, coverpoint, warl_fields=warl_fields))
     tc.code.extend(["", "#endif"])
 
     # cp_csr_satp waived because behavior of other fields is UNSPECIFIED when satp.MODE = Bare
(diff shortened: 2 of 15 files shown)


generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -313,29 +316,213 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
+        | (3 << 15)  # XS:   User-Mode Extension Status
+        | (1 << 17)  # MPRV: Modify Privilege
+        | (1 << 18)  # SUM:  Supervisor User Memory Access
+        | (1 << 19)  # MXR:  Make eXecutable Readable
+        | (1 << 20)  # TVM:  Trap Virtual Memory
+        | (1 << 21)  # TW:   Timeout Wait
+        | (1 << 22)  # TSR:  Trap SRET
+        | (1 << 23)  # SPELP: Supervisor Previous Expect Landing Pad
+        | (0 << 24)  # SDT: not yet supported by Sail; TODO change to 1 when Ssdbltrp implemented
+        | (1 << 31)  # SD for RV32 (probably shouldn't be tested for RV64, but seems to work ok)
```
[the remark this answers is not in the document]
davidharrishmc: working for now.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -230,74 +275,53 @@ def _generate_srets_tests(test_data: TestData) -> list[str]:
...
         ),
         "",
         "# Setup",
         f"csrr x{save_reg}, sstatus        # read and save sstatus",
-        f"LI(x{reg1}, 1 << 2)",
-        f"csrc medeleg, x{reg1}          # turn off delegating illegal instruction exceptions so TSR won't cause a trap loop on sret",
         f"{INDENT}# set up x{reg1} with sstatus except SPP, SPIE, SIE cleared",
         f"LI(x{reg2}, 0x122)          # x{reg2} has all SPP, SPIE, SIE bits set (bits [8], [5], [1] respectively)",
         f"not x{reg2}, x{reg2}              # x{reg2} has all but SPP, SPIE, SIE bits set",
         f"and x{reg1}, x{save_reg}, x{reg2}          # clear SPP, SPIE, SIE bits",
```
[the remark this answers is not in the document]
davidharrishmc: priv.py already reserves these.

generators/testgen/src/testgen/priv/extensions/UV.py
```diff
@@ -14,18 +14,12 @@
...
-    """In M-mode prelude: set mstatus.VS=Dirty and configure a legal vtype/vl."""
+def _enable_vector(temp_reg: int) -> list[str]:
+    """configure a legal vtype/vl."""
     return [
-        f"LI(x{temp_reg}, {_VS_MASK})",
-        f"csrc mstatus, x{temp_reg}  # clear VS",
-        f"LI(x{temp_reg}, {3 << 9})",
-        f"csrs mstatus, x{temp_reg}  # VS=Dirty",
         f"vsetivli x{temp_reg}, 1, e32, m1, tu, mu",
         "csrw vstart, x0",
```
jordancarlin: Is it worth moving this into the boot code so that the vector unit is always in a usable state when a test starts?

generators/testgen/src/testgen/priv/extensions/UV.py
```diff
@@ -37,54 +31,55 @@ def _enable_vs_and_vector(temp_reg: int) -> list[str]:
...
         "#define RVTEST_SEW 0",
         "#define VDSEW 0",
     ],
+    testcases_per_file=512,
 )
 def make_uv(test_data: TestData) -> list[TestChunk]:
     """Generate UV tests (vector CSR access from U-mode)."""
     test_chunks: list[TestChunk] = []
-    tc = test_data.begin_test_chunk()
+    test_data.begin_test_chunk("vucsr")
```
jordancarlin: There is inconsistent use of `vucsr` and `uvcsr` throughout this file. I think we should standardize on one.

coverpoints/norm/S.yaml
```diff
@@ -356,7 +360,11 @@ normative_rule_definitions:
   # although CSRs and instructions are associated with one privilege level, they are
   # also accessible at all higher privilege levels.
   - name: Zicsr_higher_priv
-    coverpoint: ["S_scsr_cg/cp_scsr_from_m", "S_scsr_cg/cp_ucsr_from_s"]
+    coverpoint:
+      [
+        "Sm_mcsr_cg/cp_scsr_from_m (S-mode CSRs from M-mode; moved to the Sm suite)",
```
jordancarlin:
```suggestion
        "Sm_mcsr_cg/cp_scsr_from_m (S-mode CSRs from M-mode; in the Sm suite)",
```

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -15,6 +15,47 @@
 from testgen.data.test_chunk import TestChunk
 from testgen.priv.registry import add_priv_test_generator
 
+# Standard S-mode CSRs, shared with the Sm suite (cp_scsr_from_m); import rather than retype.
```
jordancarlin:
```suggestion
# Standard S-mode CSRs, shared with the Sm suite (cp_scsr_from_m).
```

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -15,6 +15,47 @@
...
+
+# sstatus bit mask
+S_SSTATUS_MASK = (
+    (1 << 1)  # SIE:  Supervisor Interrupt Enable
+    | (1 << 5)  # SPIE: Supervisor Previous Interrupt Enable
+    | (0 << 6)  # UBE not yet supported by Sail; test in Endian
+    | (1 << 8)  # SPP:  Supervisor Previous Privilege
+    | (3 << 9)  # VS:   Vector Status
+    | (3 << 13)  # FS:   Floating-Point Status
+    | (3 << 15)  # XS:   User-Mode Extension Status
```
jordancarlin:
```suggestion
    | (3 << 15)  # XS:   Custom Extension Status
```

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -15,6 +15,47 @@
...
+    ("stvec", 0b10),
+    ("scounteren", None),
+    ("sscratch", None),
+    ("sip", 0xFFFF),  # only test standard non-reserved portion
+    ("sie", 0xFFFF),  # only test standard non-reserved portion
+]
+# skip walking 1s on this because valid virtual addresses is not described adequately
+S_CSRS_NOWALK = [
+    ("sepc", None),  # only has to be able to hold all valid virtual addresses
+    ("stval", None),  # only has to be able to hold all valid virtual addresses and 0
```
jordancarlin, line 54: We could eventually test this more rigorously by testing writing valid Sv57, Sv48, and Sv39 addresses to it and making sure they are read back correctly.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -220,7 +261,11 @@ def _generate_priv_inst_tests(test_data: TestData) -> list[str]:
 
 
 def _generate_srets_tests(test_data: TestData) -> list[str]:
-    """Generate sret from S-mode with spp, mprv, spie, sie, tsr sweep."""
+    """
+    Generate sret from S-mode with spp, spie, sie sweep (no TSR: that needs M-mode and lives in Sm).
+
+    Runs entirely in S-mode; the only privileged operation is the T-SBI GOTO_SMODE after each sret.
```
jordancarlin: Drop this from the docstring.

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -306,46 +330,10 @@ def _generate_scsr_tests(test_data: TestData, test_chunks: list[TestChunk]) -> N
...
-    csrs_nowalk = [
-        ("sepc", None),  # only has to be able to hold all valid virtual addresses
-        ("stval", None),  # only has to be able to hold all valid virtual addresses and 0
-    ]
-    # senvcfg CBIE/PMM reserved values are handled with warl_fields in the walk test below
-    csr_senvcfg = ("senvcfg", None)
+    # S-mode CSR lists are module-level (shared with Sm.py); bind the local names used below
+    csrs = S_CSRS
+    csrs_nowalk = S_CSRS_NOWALK
+    csr_senvcfg = S_CSR_SENVCFG
```
jordancarlin: Seems cleaner to use the global names directly. This just adds indirection.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -223,7 +224,11 @@ def _generate_mret_tests(test_data: TestData) -> list[str]:
         f"and x{reg1}, x{save_reg}, x{reg2}         # clear MPP, MPRV, MPIE, MIE bits",
     ]
 
-    for mpp in (3,):  # only M-mode; this will expand in other tests
+    # MPP selects the mode mret returns to; the S and U cases only exist when those modes do
+    mpp_guard = {3: None, 1: "S_SUPPORTED", 0: "U_SUPPORTED"}
+    for mpp in (3, 1, 0):
```
jordancarlin:
```suggestion
    for mpp in mpp_guard.keys():
```

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -315,6 +323,133 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
     return lines
 
 
+def _generate_sret_s_tests(test_data: TestData) -> list[str]:
+    """
+    Generate sret from S-mode with spp, spie, sie, tsr sweep (cp_sret_s).
+
+    Moved here from the S suite: it needs M-mode to set mstatus.TSR and to turn off medeleg
+    delegation of illegal instructions, so it runs in the Sm suite, gated on S_SUPPORTED.
+    Starts and ends in M-mode.
```
jordancarlin: Remove from docstring. A shorter version of this as a comment after the docstring would be fine.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -315,6 +323,133 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
+            coverpoint,
+            "Execute sret from S-mode while sweeping cross-product of sstatus.spp, spie, sie; mstatus.tsr\n"
+            "Go to S or U mode depending on SPP.  SIE <- SPIE.  SPIE <- 1.  "
+            "MPRV <- 0. SPP <- 0 (U-mode).  TSR causes illegal instruction.",
+        ),
+        "",
+        "# Setup",
+        f"csrr x{save_reg}, sstatus        # read and save sstatus",
+        f"LI(x{reg1}, 1 << 2)",
+        f"csrc medeleg, x{reg1}          # turn off delegating illegal instruction exceptions so TSR won't cause a trap loop on sret",
```
jordancarlin: Simplify to a `csrci`?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -315,6 +323,133 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
+                            gen_csr_read_sigupd(check_reg, ("sstatus", S_SSTATUS_MASK), test_data, reg3),
+                        ]
+                    )
+
+    lines.extend(
+        [
+            f"\ncsrw sstatus, x{save_reg}    # restore CSR",
+            "RVTEST_TSBI_GOTO_MMODE      # back to M-mode to touch medeleg",
+            f"LI(x{reg1}, 1 << 2)",
+            f"csrs medeleg, x{reg1}           # restore delegating illegal instructions",
```
jordancarlin: Simplify to a `csrci`?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -315,6 +323,133 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
+                            f"or x{check_reg}, x{check_reg}, x{reg1}          # value to write to sstatus with SPP/SPIE/SIE bits set/clear",
+                            f"LA(x{reg3}, 1f)             # return address after sret",
+                            f"csrw sepc, x{reg3}          # set sepc to return address.",
+                            f"csrw sstatus, x{check_reg}       # write sstatus with SPP/SPIE/SIE bits set/clear",
+                            test_data.add_testcase(f"{binname}_wval", coverpoint, covergroup),
+                            "sret                   # test sret instruction",
+                            f"addi x{check_reg}, zero, -1              # should not be executed",  # should not be executed
+                            "1:                         # sret should return to here",
+                            write_sigupd(check_reg, test_data),
+                            "RVTEST_TSBI_GOTO_SMODE      # We might be coming from U-mode, so to get back to S-mode, macros may have to go through M",
```
jordancarlin: Comment doesn't make sense. How is coming from U related to going through M?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -315,6 +323,133 @@ def _generate_sret_tests(test_data: TestData) -> list[str]:
...
+                            # Test sstatus was updated properly, masked the same way as the S suite's cp_sret_s.
+                            # x{reg3} is free again (sepc consumed it); split the load because the mask has bits above 31.
+                            "#if __riscv_xlen == 64",
+                            f"LI(x{reg3}, {S_SSTATUS_MASK:#x})    # sstatus mask",
+                            "#else",
+                            f"LI(x{reg3}, {S_SSTATUS_MASK & 0xFFFFFFFF:#x})    # sstatus mask (low 32 bits)",
+                            "#endif",
+                            gen_csr_read_sigupd(check_reg, ("sstatus", S_SSTATUS_MASK), test_data, reg3),
+                        ]
+                    )
```
jordancarlin, line 400: Might be worth factoring this block out into a function that is used in both S and Sm instead of duplicating it. Then Sm would just call the function inside its loop over TSR. Fine to leave this way if you prefer though.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/S.py
davidharrishmc: The spec never defines valid virtual address. The spec does say "stval is a WARL register that must be able to hold all valid virtual addresses and the value 0."  That suggests 0 might not be a valid virtual address, even though it is a well-formed address.  I'll post a query.  Let's merge in the meantime.

davidharrishmc: https://github.com/riscv/riscv-isa-manual/issues/3342

jordancarlin (approved):

PR #2166, opened 2026-08-23
T-SBI for ExceptionsS/Sm

davidharrishmc:
Moved tests involving delegation to ExceptionsSm
Add T-SBI


coverpoints/norm/ExceptionsS.yaml
jordancarlin, line 1: ExceptionsS not covering any normative rules doesn't seem right. If it doesn't link to any normative rules we open ourselves up to the question of why it exists.

coverpoints/norm/ExceptionsSm.yaml
```diff
@@ -255,3 +255,46 @@ normative_rule_definitions:
       [
         "ExceptionsSm_cg/{cp_misaligned_priority_load, cp_misaligned_priority_store, cp_misaligned_priority_fetch}",
       ]
+
+  # --- moved from ExceptionsS: the medeleg walk runs from M-, S- and U-mode and lives in ExceptionsSm ---
```
jordancarlin: Drop this comment. No need to explain history.

generators/testgen/src/testgen/priv/extensions/ExceptionsS.py
```diff
@@ -63,359 +64,100 @@ def _generate_illegal_instruction_csr_tests(test_data: TestData) -> list[str]:
...
-                        f"LI(x{mask_mie}, 0x88)",
-                        f"LI(x{mask_sie}, 0x2)",
-                        f"{'csrs' if medeleg_b8 else 'csrc'} medeleg, x{medeleg_reg}",
-                        # Set MPIE and MIE so mret goes to the proper MIE in the next mode
-                        f"{'csrs' if mie else 'csrc'} mstatus, x{mask_mie}",
-                    ]
-                )
+    for mie in (0, 1):
+        for sie in (0, 1):
+            tag = f"{mode_tag}_mdlg_mie_{mie}_sie_{sie}"
```
jordancarlin:
```suggestion
            tag = f"{mode_tag}_mie_{mie}_sie_{sie}"
```

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -34,6 +34,189 @@
...
+)
+
+
+def _generate_medeleg_msu_tests(test_data: TestData, mode_tag: str, priv_mode: int) -> list[str]:
+    """
+    Runs 10 exception tests x 17 medeleg values for one privilege mode.
+    Starts and ends in M-mode.  For each medeleg value: write medeleg in M-mode, drop to the mode
+    under test with RVTEST_TSBI_GOTO_SMODE / RVTEST_TSBI_GOTO_UMODE (nothing for M-mode), run the
+    tests, and return with RVTEST_TSBI_GOTO_MMODE (via RVTEST_TSBI_GOTO_SMODE first when returning
+    from U-mode with ecall-from-U delegated).
```
jordancarlin: Too many internal details in the docstring. This should just say what the function does, not how.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -34,6 +34,189 @@
...
+        lines.extend(
+            [
+                "#ifdef RVMODEL_ACCESS_FAULT_ADDRESS",
+                test_data.add_testcase(f"instrmisaligned_{tag}", coverpoint, covergroup),
+                f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+                f"jalr x1, 0(x{addr_reg})  # aligned target",
+                f"jalr x1, 2(x{addr_reg})  # misaligned target",
+                "#endif",
+            ]
+        )
```
jordancarlin, line 75: If this is supposed to test instruction misaligned, why is it using ACCESS_FAULT_ADDRESS?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -34,6 +34,189 @@
...
+            [
+                "#if __riscv_xlen == 64",
+                f" sd x{data_reg}, 0(x{addr_reg})",
+                "#endif",
+                "#endif",
+            ]
+        )
+
+        # Ecall (uses the T-SBI ECALL_TEST protocol: a bare ecall with a stale a0 would be
+        # misinterpreted by the trap handler as a T-SBI instruction-table request)
```
jordancarlin: Probably no need for the comment about the bare ecall.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -34,6 +34,189 @@
...
+            ]
+        )
+
+        # Return to M-mode.  With ecall-from-U delegated (medeleg bit 8), a GOTO_MMODE from U-mode is
+        # forwarded by the S-mode handler and the caller resumes in U-mode (handler returns into the
+        # forwarding stub, whose sret drops to SPP=U), so hop to S-mode first and go to M from there.
+        if priv_mode == 0 and medeleg_val & (1 << 8):
+            lines.extend(["RVTEST_TSBI_GOTO_SMODE", "RVTEST_TSBI_GOTO_MMODE"])
+        else:
+            lines.extend(goto_back)
```
jordancarlin, line 204: Seems like we should fix the T-SBI handler so that `RVTEST_TSBI_GOTO_MMODE` from U-mode actually works instead of working around it with the chained calls here. This will likely cause us issues elsewhere when we forget this otherwise.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -63,9 +246,66 @@ def _generate_mstatus_ie_tests(test_data: TestData) -> list[str]:
...
+
+    test_data.int_regs.return_registers([save_reg, mask_mie, mask_sie, medeleg_reg])
+    return lines
+
+
 @add_priv_test_generator(
     "ExceptionsSm",
     required_extensions=["Sm"],
+    extra_defines=[
+        "#define TRAP_SIGUPD_COUNT 3000",  # words; each medeleg walk file records 306 x up to 6-word trap signatures
```
jordancarlin: Comment is confusing. Also, there are lots of testcases other than just the medeleg walks.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -34,6 +34,189 @@
...
+        # Return to M-mode.  With ecall-from-U delegated (medeleg bit 8), a GOTO_MMODE from U-mode is
+        # forwarded by the S-mode handler and the caller resumes in U-mode (handler returns into the
+        # forwarding stub, whose sret drops to SPP=U), so hop to S-mode first and go to M from there.
+        if priv_mode == 0 and medeleg_val & (1 << 8):
+            lines.extend(["RVTEST_TSBI_GOTO_SMODE", "RVTEST_TSBI_GOTO_MMODE"])
+        else:
+            lines.extend(goto_back)
+
+    # Clear medeleg (in M-mode)
+    lines.extend([f"LI(x{medeleg_reg}, 0)", f"csrw medeleg, x{medeleg_reg}"])
```
jordancarlin: Why are we clearing `medeleg` at the end of this testcase? Isn't our default state to have everything delegated?

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -97,6 +337,18 @@ def make_exceptionssm(test_data: TestData) -> list[TestChunk]:
...
+    tc.code.append("#endif // S_SUPPORTED")
     test_chunks.append(test_data.end_test_chunk())
+
+    # medeleg only exists with S-mode; walk it from M-, S- and U-mode.  One file per mode: each walk
+    # records 306 trap signatures of up to 6 words, ~1.8k words of the TRAP_SIGUPD_COUNT area.
+    for mode_tag, priv_mode in (("mode_m", 3), ("mode_s", 1), ("mode_u", 0)):
+        tc = test_data.begin_test_chunk(f"medeleg_{mode_tag[-1]}")
+        tc.code.append("#ifdef S_SUPPORTED")
+        tc.code.extend(_generate_medeleg_msu_tests(test_data, mode_tag, priv_mode))
+        tc.code.append("#endif // S_SUPPORTED")
```
jordancarlin, line 345: I don't love the idea of having test files where the entire test is ifdefed out but the test still runs. We can leave this as is for now, but it might be a good reason to prefer the approach with ExceptionsSSm as a separate suite.

jordancarlin (changes requested):

coverpoints/norm/ExceptionsS.yaml
davidharrishmc: Need to revisit normative rules later.  Some of these are stubs.

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
davidharrishmc: Fixed in 2170, which is dependent on this one being merged.

jordancarlin (approved):

generators/testgen/src/testgen/priv/extensions/ExceptionsSm.py
```diff
@@ -34,6 +34,182 @@
...
+def _generate_medeleg_msu_tests(test_data: TestData, mode_tag: str, priv_mode: int) -> list[str]:
+    """Runs 10 exception tests x 17 medeleg values for one privilege mode."""
+    covergroup = _CG
+    coverpoint = "cp_medeleg_msu"
+
+    addr_reg, data_reg, check_reg, medeleg_reg, medeleg_orig = test_data.int_regs.get_registers(5)
+    goto_mode = {3: [], 1: ["RVTEST_TSBI_GOTO_SMODE"], 0: ["RVTEST_TSBI_GOTO_UMODE"]}[priv_mode]
+    goto_back = ["RVTEST_TSBI_GOTO_MMODE"] if priv_mode != 3 else []
+
+    lines = ["csrr x{medeleg_orig}, medeleg  # save original medeleg value"]
```
jordancarlin:
```suggestion
    lines = [f"csrr x{medeleg_orig}, medeleg  # save original medeleg value"]
```

PR #2167, opened 2026-08-23
Fix for CVW RV32IMC

davidharrishmc:
- Avoid testing mstatus.MPP = 01 when S is not implemented
- Update Sail.json to reflect FIOM not supported on RV32IMC
- Remove Sm from Makefile exclusion now that it works robustly
- Remove Svpbmt from exclusions; it now runs on RV64GC.  Root cause of why it was failing or what change fixed it has not been found.  Could be a memory interlock issue similar to https://github.com/openhwgroup/cvw/issues/1766 that is still failing on Sv.


jordancarlin (approved):
I have some ideas on how to restructure our CSR tests to remove a lot of the complexity that is growing in these functions and make it much easier to gate specific fields. I will follow up with another PR implementing that later. This approach is fine for now.

PR #2170, opened 2026-08-24
T-SBI Handler Fixes

davidharrishmc:
rvtest_setup.h: use RVTEST_TSBI_GOTO_MMODE in epilogue that works from any mode
rvtest_trap_handler.h: fixes to forwarding requests to M-mode
ExceptionsSm.py (simplify going to M-mode)
Sm: medelegh support for Clang, robust mtime_poll


davidharrishmc:
This break RVTEST_GOTO_MMODE.  Pause until some of the other PRs are merged including Jordan's boot code, and then solve more gracefully, possibly replacing with RVTEST_TSBI_GOTO_MMODE.

UPDATE: only breaks RVTEST_GOTO_MMODE for tests that don't boot with BOOT_TO_MMODE.  InterruptsU needed fixing because it boots to U mode.  Other suites with RVTEST_GOTO_MMODE  intended to operate below M mode will need to be ported during SBI conversion booting to lower modes.

UmerShahidengr (approved):

PR #2171, opened 2026-08-24
SmV speedup and reduce default TESTCASES_PER_PRIV_FILE

davidharrishmc:
Chunked slowest coverpoint to reduce to < 100k instructions per SmV_vl_walk-0*.S file.


generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -538,30 +523,38 @@ def _gen_vl_walking1s_sew_lmul(test_data: TestData, temp_reg: int) -> list[str]:
         "#define RVTEST_SEW 0",
         "#define VDSEW 0",
     ],
+    testcases_per_file=512,  # csrw vl always traps, so split the vl_walk tests for runtime
```
jordancarlin: We seem to be setting this in most of the priv files. Should we change the default and then just increase it for Ssstrict instead of decreasing it manually in most suites?

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -62,54 +62,34 @@ def _gen_vcsrrswc(test_data: TestData, temp_reg: int) -> list[str]:
...
-            lines.append(f"csrc {csr}, x{mask_reg}  # clear all bits")
-            lines.append(test_data.add_testcase(f"{csr}_bit_{i}", coverpoint, _CG))
-            lines.append(f"csrw {csr}, x{walk_reg}  # walking-1 bit {i}")
-            lines.append(f"slli x{walk_reg}, x{walk_reg}, 1")
-        lines.append("#endif")
+            tc.code.append(f"csrc {csr}, x{mask_reg}  # clear all bits")
+            tc.code.append(test_data.add_testcase(f"{csr}_bit_{i}", coverpoint, _CG))
+            tc.code.append(f"csrw {csr}, x{walk_reg}  # walking-1 bit {i}")
+            tc.code.append(f"slli x{walk_reg}, x{walk_reg}, 1")
+        tc.code.append("#endif")
```
jordancarlin, line 81: Any reason this can't use the standard csr_walk function?

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -62,54 +62,34 @@ def _gen_vcsrrswc(test_data: TestData, temp_reg: int) -> list[str]:
...
             )
-        )
-        + 1
-    ]
-    lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
-    lines.extend(_vector_setup(temp_reg))
-    for csr in _VECTOR_CSRS_WR:
-        lines.append(f"# walking-1s on {csr}")
-        lines.append(f"LI(x{mask_reg}, -1)  # all 1s")
-        lines.append(f"LI(x{walk_reg}, 1)   # one-hot starting at bit 0")
```
jordancarlin, line 98: Wow. This test was very messed up before. Surprised we didn't catch this.

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -501,32 +481,37 @@ def _gen_vstart_oob(test_data: TestData, temp_reg: int) -> list[str]:
...
-                lines.append(f"csrw vl, x{walk_reg}  # bit {i}")
-                lines.append(f"slli x{walk_reg}, x{walk_reg}, 1")
-            lines.append("#endif")
-    test_data.int_regs.return_registers([walk_reg])
-    return lines
+                tc.code.append(f"vsetivli x{temp_reg}, 1, {sew_name}, {lmul_name}, tu, mu")
+                tc.code.append(test_data.add_testcase(f"vl_walk_{sew_name}_{lmul_name}_b{i}", coverpoint, _CG))
+                tc.code.append(f"csrw vl, x{walk_reg}  # bit {i}")
+                tc.code.append(f"slli x{walk_reg}, x{walk_reg}, 1")
+            tc.code.append("#endif")
```
jordancarlin, line 498: csr_walk?

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/SmV.py
davidharrishmc: This has a vsetvli before each access.  Doesn't play nice with walk.

generators/testgen/src/testgen/priv/extensions/SmV.py
davidharrishmc: subtle differences.  Might be possible with some work, but not worth the bother.

generators/testgen/src/testgen/constants.py
```diff
@@ -29,7 +29,7 @@ def indent_asm(line: str) -> str:
 # Max testcases per test file before splitting into multiple files. Individual test
 # chunks won't be split, so if one test chunk exceeds this, the file will exceed this limit.
 TESTCASES_PER_FILE = 1000
-TESTCASES_PER_PRIV_FILE = 6000
+TESTCASES_PER_PRIV_FILE = 512
```
jordancarlin, line 32: Ssstrict* should probably override this to a larger number so that it doesn't result in hundreds of tests.

jordancarlin (changes requested):
One last comment and then there will be merge conflicts to sort out with #2154.

generators/testgen/src/testgen/constants.py
davidharrishmc: Ssstrict instruction count is a work in progress.  I've changed it to 8 for the time being because that helped me find and fix a ton of tool issues that were otherwise masked.  I'd rather come back to Ssstrict later after some issues are closed.

jordancarlin: Merging this PR as is will increase the number of tests for Ssstrict from around a dozen to close to 150. I'm fine with that for the moment, but we will need to get that sorted out later.

jordancarlin (approved):

PR #2178, opened 2026-08-24
Imperas override to fix MISA dependency

davidharrishmc:
Imperas requires a d_requires_f override to describe MISA D and F dependency correctly.


jordancarlin (approved):
Strange that this isn't the default since it is required behavior.

PR #2179, opened 2026-08-24
Sv test generator

davidharrishmc:
Starting to review the code, but this might do what we want.

The ELFs are instruction-identical to the originals,
except that SIGUPD_COUNT/TRAP_SIGUPD_COUNT have been computed exactly, causing some addresses to shift.

The .S files differ in white space and some punctuation.

The process found some anomalies in the original .S files that ought to be addressed:

  What the final rounds surfaced and encoded faithfully: Svadu's custom per-svmode code VAs (new va_code_override field) and the sv39-U-mode
  a4-vs-t0 register quirk; SvZicbo's manual bug where every mid-level DAU A-case writes SUPERPAGE_…LEVEL0 (macro name kept, level wrong);
  zicboz U-mode dropping the execute-only case; per-case S1P12P0 and RVMODEL_ACCESS_FAULT_ADDRESS guards inline in bodies.


UmerShahidengr (approved):

jordancarlin:
Closing. There will be a new version of this PR coming soon.

PR #2181, opened 2026-08-24
Adding tests of Sscofpmf

ayeshaanwaar05:
This PR contains tests of SscofpmfSm, SscofpmfS and SscofpmfU. 
Running on whisper and qemu.
Sail, Spike and Imperas do not have way to icrement counters .
Sail issue : https://github.com/riscv/sail-riscv/issues/1733
Spike issue : https://github.com/riscv-software-src/riscv-isa-sim/issues/2370


davidharrishmc:
Claude review of PR #2181 at `d8d9f77c3` (generators at `0adec4a16`; the head is a merge of act4). Built and run
together with #2115 on Spike, Sail, Whisper, and QEMU, rv32 and rv64. Coverage was reviewed on #2115 and is
not repeated here. Sail does not count HPM events (sail-riscv#1733), so every result that depends on a real
overflow is expected to differ between Sail and a counting DUT; those are noted but not treated as defects.

## Results as run

| Config | SscofpmfSm | SscofpmfS | SscofpmfU |
|---|---|---|---|
| sail (rv32/rv64-max, clang, RVA22/23, RVB23, optionals) | pass | pass | pass |
| spike-rv32-max, spike-rv64-max | -00 pass, -01 fails at `xinh_combo_00100` | fails at `xinh_combo_00100` | fails at `xinh_combo_00100` |
| whisper-rv64-max | -00 mip mismatch (MEIP), -01 OF set | `lcofip_priority_s_seip` OF set | `of_overflow_u` OF set |
| qemu-rv64-max | -00 mcause mismatch (no LCOFI), -01 OF set | `lcofip_priority_s_seip` OF set | `of_overflow_u` OF set |

The Whisper/QEMU "OF set" rows are the DUT counting while Sail does not. The Spike rows are a config problem,
item 3 below. ruff, ruff format, and pyright are clean.

## Blocking

1. **Nothing runs in CI, and the one place it would run cannot assemble.** `config/{sail,spike,whisper,qemu}/ci.yaml`
   all exclude the three suites, so the PR body's "running on whisper and qemu" is not what CI will do. The
   imperas configs (rv64-max, rv32gck, rv32gcv) are not excluded and declare Sscofpmf, so CI will build these
   suites there, and `config/imperas/imperas-rv64-max/rvmodel_macros.h` defines
   `RVMODEL_MHPMEVENT_CODE 0x00000000` with no parameter list. `RVMODEL_MHPMEVENT_CODE(x9, x5)` then expands to
   `0x00000000(x9, x5)`, which fails with "junk at end of line" (verified with the toolchain). Define it as a
   two-parameter `nop` like the other configs, or exclude the suites on imperas too.

2. **RV32 is not handled in the overflow and OF tests.** In `_generate_of_set_on_overflow_tests` the inhibit
   pattern is built as `RVMODEL_MHPMEVENT_VAL | (pattern << 58)` and loaded with `LI`, which truncates to 32
   bits, so on RV32 the pattern never reaches `mhpmevent3h` and the test always runs with no inhibits. The
   all-ones counter preload writes only `mhpmcounter3`; `mhpmcounter3h` is never written anywhere in the S test,
   so the 64-bit counter cannot overflow on RV32 at all. `_generate_lcofip_hw_only_tests` uses
   `OF_BIT = 1 << 63`, which is 0 after truncation, so OF is never set on RV32. Use `mhpmevent3h[31]` and an
   `RVMODEL_MHPMCOUNTERH` write on RV32, as the test plan already calls for.

3. **The Spike reference config has no HPM counters.** `config/spike/spike-rv*-max/sail.json` sets
   `writable_hpm_counters` to `0x7`, so Sail treats `mhpmevent3` as read-only zero while Spike implements it.
   Every inhibit-combination bin therefore fails on both Spike configs, starting at the first bit the test does
   not mask (`00100`). Whisper's json allows counters 3 to 10 and QEMU's allows all, which is why they get past
   this point. The PR should update the Spike json (or choose a counter both models implement).

4. **The T-SBI path silently ignores the platform's counter choice.** `_resolve_tsbi_csr` hard-codes
   `RVMODEL_MHPMEVENT` to `0x323` and `RVMODEL_MHPMCOUNTER` to `0xb03`, and the handler table only has
   `mhpmcounter3`, so in S and U mode a platform that sets `RVMODEL_MHPMEVENT` to any other counter tests
   counter 3 anyway. Either state in the docs that counter 3 is the only supported choice, or use
   `RVTEST_TSBI_CSR_*` with the platform macro. Related: Spike and Imperas define the macros as bare names
   (`mhpmevent3`) while Sail, Whisper, and QEMU use CSR numbers (`CSR_MHPMEVENT3`); pick one convention.

5. **The S and U suites are written M-centric, against the T-SBI conversion rules.** Each section starts with
   an "M-MODE SETUP" block that only works because the first chunk of the file did a legacy `RVTEST_GOTO_MMODE`
   and never returned to the boot mode, and every section hops down with `RVTEST_GOTO_LOWER_MODE` and back with
   `RVTEST_GOTO_MMODE`. `docs/tsbi-changes.md` asks converted suites to boot to their own mode and stay there,
   using `RVTEST_TSBI_GOTO_*` and T-SBI CSR calls for M-mode work. Today this holds together only because
   SscofpmfS and SscofpmfU each fit in one file; the Sm suite already splits into two, and the moment S or U
   splits, its second file boots to S or U and the M-mode setup traps. `SscofpmfU` also still defines
   `RVTEST_TEMP_BOOT_TO_U`, which no longer exists on act4 (#2216); it works only because U is the default
   boot mode.

6. **S-mode views are read from M.** `_csr_access` routes every non-Sm access through T-SBI, including reads
   of `sip` and `scountovf` in the S suite (`_generate_lcofip_hw_only_tests`, `_generate_of_set_on_overflow_tests`,
   `_generate_scountovf_mcounteren_tests`, `_generate_scountovf_shadow_tests`). Those reads execute in the
   M-mode handler, so the S-mode `scountovf` value masked by `mcounteren`, which is the whole point of the
   plan's "in S-mode" rows, is never observed from S. Read S-level CSRs directly in S mode.

7. **The single-bit inhibit toggles check nothing.** `_generate_xinh_inhibits_tests` emits
   `csrr RVMODEL_MHPMCOUNTER` for `xinh = 0/1` but never writes it to the signature; only the 32-combination
   sweep records anything, and it records `mhpmevent`, not the counter. The plan's expectation is "counter
   nonzero iff xinh = 0". Record a normalized value (for example `snez`) so the check is independent of how
   many events each model counts.

## Should fix

8. Dead code: `_generate_lcofi_tests` (`cp_lcofi`) is never called; `_mode_suffix` wraps `.lower()`; the
   `# In SscofpmfCommon.py` line at the top is a leftover. The long inline comments about Sail and H in the
   generated assembly could be one line each (`AGENTS.md`: keep generated comments about what the code does).

9. The Sm priority test uses the legacy interrupt macro names (`RVTEST_SET_MEXT_INT`, `RVTEST_SET_MSW_INT`,
   `set_mtimer_int`). They exist on act4 today, but #2213 replaces them with `_M/_S/_U` flavors, so expect a
   conflict at merge.

10. The trap-handler change adds a `clr_Lcofi_int` routine through a `.set`/`.if` inside the reserved-cause
    `.rept`. #2213 adds its own LCOFI clearing in the same table; the two will need reconciling. The routine
    itself is right (`csrc sip` from the S handler, `csrc mip` from M).

11. `.codespellrc` adds `min` to the ignore list, which will hide real typos repo-wide. `inh` and `sinh` are
    enough for these files.

12. Whisper `SscofpmfSm-00` fails on `mip` (expects MEIP 0x800; Whisper routes SEIP/MEIP through mvip) and QEMU
    `SscofpmfSm-00` fails on `mcause` (QEMU does not raise LCOFI from a software `mip` write, as the ci.yaml
    comment says). Both are platform issues already known elsewhere in the repo; worth listing in the PR body
    so the exclusions have a stated reason per platform.

## What is fine

- Boot defines for Sm (`BOOT_TO_MMODE`) and S (`BOOT_TO_SMODE`) are now correct, and the lcofi tests write
  `sstatus`/`sip` so the #2115 coverpoints sample them.
- All three suites pass on every Sail configuration, RV32 and RV64.
- The 32-combination inhibit sweep now runs in all three modes, with VS/VU bits masked out of the signature
  because Sail has no H support.
- The handler table additions for `mhpmevent3..31`, `mhpmevent3h..31h`, `mhpmcounter3`, and `scountovf` are
  correct and were exercised by the S and U suites through T-SBI.

ayeshaanwaar05:
Spike is failing on cp_xinh because of mask in sail.json on :
"writable_hpm_counters": {
      "len": 32,
      "value": "0x0000_0007"
    },
    If i try to modify it to make cp_xinh work  like this 
    "writable_hpm_counters": {
      "len": 32,
      "value": "0x0000_000F"
    },
    Sm tests starts failing

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,262 @@
...
+            "sip.LCOFIP x sie.LCOFIE.\n",
+        ),
+        "",
+        _csr_access("csrw mip, zero      # clear all pending", "S"),
+        _csr_access("csrw mie, zero      # disable all interrupts", "S"),
+        _csr_access("csrw RVMODEL_MHPMEVENT, zero", "S"),
+        # mideleg is deliberately excluded from the T-SBI dispatch table (see
+        # docs/tsbi-changes.md) -- it needs an actual, one-time mode change, not T-SBI.
+        f"LI(x{r_val}, {hex(LCOFI_BIT)})",
+        "RVTEST_TSBI_GOTO_MMODE",
```
davidharrishmc: S and U tests can't use GOTO_MMODE because we don't know that a conforming M-mode exists.  LCOFI should already be 1 so it shouldn't have to be set here.  If it isn't 1, change it in the BOOT_TO_S_MODE rvtest_setup.S file.

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,262 @@
...
+                *prime_counter_overflow(r_val, r_temp2, r_temp, r_addr, "S"),
+                "# the overflow sets OF and raises LCOFIP; sie = 0, so nothing fires yet",
+            ]
+        )
+
+        if other_int == "seip":
+            lines.append("RVTEST_SET_SEXT_INT")
+
+        elif other_int == "stip":
+            # set_stimer_mmode writes mip directly and must run at M.
```
davidharrishmc: Can't run directly in M.  Use macro to RVTEST_SET_MTIMER_INT_S or RVTEST_SET_MTIMER_INT_SOON_S.

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,262 @@
...
+                write_sigupd(r_temp2, test_data),
+                "",
+            ]
+        )
+
+        if other_int == "seip":
+            lines.append("RVTEST_CLR_SEXT_INT")
+
+        elif other_int == "stip":
+            lines.append("RVTEST_TSBI_GOTO_MMODE")
```
davidharrishmc: See other comment

generators/testgen/src/testgen/priv/extensions/SscofpmfU.py
```diff
@@ -0,0 +1,112 @@
...
+            "mideleg.LCOFI=1 and sstatus.SIE=0 held fixed; sweep sip.LCOFIP x sie.LCOFIE.\n"
+            "Sample point is the T-SBI delegate's sret back to U (sstatus.SPP=0).\n",
+        ),
+        "",
+        _csr_access("csrw mip, zero      # clear all pending", "U"),
+        _csr_access("csrw mie, zero      # disable all interrupts", "U"),
+        _csr_access("csrw RVMODEL_MHPMEVENT, zero", "U"),
+        f"LI(x{r_val}, {hex(LCOFI_BIT)})",
+        "RVTEST_TSBI_GOTO_MMODE",
+        f"csrs mideleg, x{r_val}   # mideleg.LCOFI = 1 ",
```
davidharrishmc: See other comment.

generators/testgen/src/testgen/priv/extensions/SscofpmfU.py
```diff
@@ -0,0 +1,112 @@
...
+            )
+
+    lines.extend(
+        [
+            "",
+            f"LI(x{r_temp}, {hex(LCOFI_BIT)})",
+            _csr_access(f"csrc sip, x{r_temp}      # clear LCOFIP", "U"),
+            _csr_access(f"csrc sie, x{r_temp}      # clear LCOFIE", "U"),
+            "RVTEST_TSBI_GOTO_MMODE",
+            f"csrc mideleg, x{r_temp}  # clear mideleg.LCOFI",
```
davidharrishmc: Clearing mideleg.LCOFI can't be part of a S/U mode suite.  Changing mideleg is only in a Sm suite.  Is the test plan wrong?

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+
+    lines.extend(
+        [
+            "",
+            f"LI(x{r_temp}, {hex(LCOFI_BIT)})",
+            f"csrc sip, x{r_temp}      # clear LCOFIP",
+            f"csrc sie, x{r_temp}      # clear LCOFIE",
+            "RVTEST_TSBI_GOTO_MMODE",
+            f"csrc mideleg, x{r_temp}  # clear mideleg.LCOFI",
```
davidharrishmc: See comment in U

davidharrishmc (changes requested):

ayeshaanwaar05:
Prof @davidharrishmc did you review those tsbi changes ? if yes should i ask jordan for review

config/qemu/ci.yaml
```diff
@@ -24,7 +24,9 @@ ci_enabled: true
 # TODO: Remove SsstrictV when sail SsstrictV is implemented
 # Smstateen will be added back once sail starts supporting SM1P13 too
 # ZawrsU and ZawrsS fail on qemu due to stval mismatching, filed QEMU issue https://gitlab.com/qemu-project/qemu/-/work_items/4080
-exclude_extensions: "Sm,S,SsstrictSm,SsstrictS,SsstrictU,Zacas,ZacasZabha,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,ExceptionsZicboSm,SvZicbo,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU"
+# ZicntrS and ZicntrU fail on qemu because ecall increments instret. Filed QEMU issue: https://gitlab.com/qemu-project/qemu/-/work_items/4087
+# Sscofpmf : qemu isn't setting lcofi bit in mie will come back once sail has a way to increment counter
+exclude_extensions: "Sm,S,SsstrictSm,SsstrictS,SsstrictU,Zacas,ZacasZabha,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,ExceptionsZicboSm,SvZicbo,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,ZkrSm,ZkrS,ZkrU,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU,ZicntrS,ZicntrU,SscofpmfSm,SscofpmfS,SscofpmfU"
```
davidharrishmc: It looks like your list of exclusions has gotten out of sync with the latest in ACT4.  Start with the latest in ACT4 and add just your Sscofpmf* changes.  Otherwise CI won't be testing the right things.

config/spike/spike-rv64-max/rvmodel_macros.h
```diff
@@ -182,4 +182,11 @@
   lw _R1, 0(_R2);                               \
   sw _R1, 0(_R2);
 
+
+#define RVMODEL_MHPMEVENT   mhpmevent3
+#define RVMODEL_MHPMCOUNTER mhpmcounter3
+#define RVMODEL_MHPMEVENT_VAL 0x0  // will be updated once there is a way to increment counters Spike issue: https://github.com/riscv/sail-riscv/issues/1733
+#define RVMODEL_MHPMEVENT_CODE(_R1, _R2) \
+    nop // will be updated once there is a way to increment counters Spike issue: https://github.com/riscv/sail-riscv/issues/1733
```
davidharrishmc: This is the Sail issue.  Is there also an issue filed against Spike?

config/imperas/imperas-rv64-max/rvmodel_macros.h
```diff
@@ -111,5 +111,13 @@
 #define RVMODEL_SET_SEXT_INT(_R1, _R2)        nop
 #define RVMODEL_CLR_SEXT_INT(_R1, _R2)        nop
 
+##### Sscofpmf #####
+
+#define RVMODEL_MHPMEVENT   CSR_MHPMEVENT3
+#define RVMODEL_MHPMCOUNTER CSR_MHPMCOUNTER3
+#define RVMODEL_MHPMEVENT_VAL 0x00000000  // will be updated once there is a way to increment counters
+#define RVMODEL_MHPMEVENT_CODE(_R1, _R2) \
+    nop // will be updated once there is a way to increment counters
```
davidharrishmc: Looks like an exclusion is needed in impoeras/ci.yaml

config/whisper/whisper-rv64-max/rvmodel_macros.h
```diff
@@ -195,4 +195,12 @@
   sw      _R2, 0(_R1); /* clear source 2 interrupt */
 
 
+##### Sscofpmf #####
+
+#define RVMODEL_MHPMEVENT   CSR_MHPMEVENT3
+#define RVMODEL_MHPMCOUNTER CSR_MHPMCOUNTER3
+#define RVMODEL_MHPMEVENT_VAL  11
```
davidharrishmc: Add a comment that this event code counts loads

config/whisper/ci.yaml
```diff
@@ -7,8 +7,9 @@
 # Smstateen Sail does not yet support 1P13 bit
 # ExceptionsSvZalrsc: Whisper reports page faults where Sail expects access faults.
 # Sdtrig* not yet implemented in Sail
+# Sscofpmf: Counters aren't incrementing
```
davidharrishmc: Make this more specific.  Is it the Sail issue, or a different whisper issue.  Link issue.

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,225 @@
...
+    lines = [
+        comment_banner(
+            coverpoint,
+            "Interrupt pending and enable, mode = S.\n"
+            "mideleg.LCOFI=1 from the boot-to-S setup (required to reach S-mode with\n"
+            "LCOFI visible), sstatus.SIE=1 held fixed per testplan; sweep is\n"
+            "sip.LCOFIP x sie.LCOFIE.\n",
+        ),
+        "",
+        _csr_access("csrw mip, zero      # clear all pending", "S"),
```
davidharrishmc: Is this necessary?  Writing mip doesn't clear all interrupts.  We have RVTEST_CLR_<>_INT_S macros that would clear interrupts, but you probably don't have any pending and you weren't going to enable them anyway.

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,225 @@
...
+                "sstatus.SIE=1; LCOFIP is raised by a real hpmcounter overflow through\n"
+                "RVMODEL_MHPMEVENT_CODE, together with one of {SEIP,STIP,SSIP,none}.\n"
+                "Each case holds with sie = all 0s (nothing fires), then with sie = all 1s:\n"
+                "the competing interrupt fires first and LCOFI only after it (lowest priority).\n"
+                "Other pending bits go through mip, which cp_lcofip_priority_s samples;\n"
+                "enables stay in sie, written only here."
+            ),
+        ),
+        "",
+        _csr_access("csrw mip, zero      # clear all pending", "S"),
```
davidharrishmc: Global comment on mip

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,225 @@
...
+                "",
+                test_data.add_testcase(f"{binname}_sie_off", coverpoint, covergroup),
+                "# sie = all 0s: LCOFIP and the competing interrupt stay pending, nothing fires",
+                f"RVTEST_IDLE_FOR_INTERRUPT(x{r_temp})",
+                "",
+                f"LI(x{r_temp}, -1)",
+                test_data.add_testcase(binname, coverpoint, covergroup),
+                _csr_access(f"csrs sie, x{r_temp}   # sie = all 1s: competing interrupt fires first, then LCOFI", "S"),
+                "",
+                # Already at S throughout -- interrupt fires immediately or on timer maturity.
```
davidharrishmc: What is "timer maturity"?

generators/testgen/src/testgen/priv/extensions/SscofpmfS.py
```diff
@@ -0,0 +1,225 @@
...
+        elif other_int == "stip":
+            lines.append("RVTEST_CLR_STIME_INT_S")
+
+        elif other_int == "ssip":
+            lines.append("RVTEST_CLR_SSW_INT_S")
+
+        lines.extend(
+            [
+                f"LI(x{r_val}, {hex(LCOFI_BIT)})",
+                _csr_access(f"csrc mip, x{r_val}   # clear LCOFIP for next iteration", "S"),
```
davidharrishmc: global change.  Disable LCOFI with write to sip.

generators/testgen/src/testgen/priv/extensions/SscofpmfSm.py
```diff
@@ -0,0 +1,245 @@
...
+    lines = [
+        comment_banner(
+            coverpoint,
+            "Interrupt pending and enable, mode = M.\n"
+            "mstatus.MIE=1, mstatus.SIE=1 held fixed per testplan; sweep is\n"
+            "mip.LCOFIP x mie.LCOFIE x mideleg.LCOFI.\n",
+        ),
+        "",
+        "# === M-MODE SETUP ===",
+        "csrw mip, zero      # clear all pending",
```
davidharrishmc: Carry over applicable changes from S.

generators/testgen/src/testgen/priv/extensions/SscofpmfSm.py
```diff
@@ -0,0 +1,245 @@
...
+                        "csrw mie, zero        # disable LCOFIE before next iteration",
+                    ]
+                )
+
+    lines.extend(
+        [
+            "",
+            "# === M-MODE CLEANUP ===",
+            f"LI(x{r_temp}, {hex(LCOFI_BIT)})",
+            f"csrc mip, x{r_temp}      # clear LCOFIP",
```
davidharrishmc, line 91: Ok to clear LCOFIP in mip in this suite because sip might not exist.

generators/testgen/src/testgen/priv/extensions/SscofpmfU.py
```diff
@@ -0,0 +1,106 @@
...
+    lines = [
+        comment_banner(
+            coverpoint,
+            "Interrupt pending and enable, mode = U.\n"
+            "mideleg.LCOFI=1 (from the boot setup) and sstatus.SIE=0 held fixed;\n"
+            "sweep sip.LCOFIP x sie.LCOFIE.\n"
+            "Sample point is the T-SBI delegate's sret back to U (sstatus.SPP=0).\n",
+        ),
+        "",
+        _csr_access("csrw mip, zero      # clear all pending", "U"),
```
davidharrishmc: Carry over changes from S

generators/testgen/src/testgen/priv/extensions/SscofpmfU.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+    test_data.int_regs.return_registers([r_val, r_temp])
+    return lines
+
+
+@add_priv_test_generator(
+    "SscofpmfU",
+    required_extensions=["U", "Sscofpmf"],
+    march_extensions=[],
```
davidharrishmc: These aren't necessary if blank.

generators/testgen/src/testgen/priv/extensions/SscofpmfCommon.py
```diff
@@ -0,0 +1,781 @@
...
+            (
+                "OF being set by software alone must never make LCOFIP pend --\n"
+                "only a real hardware counter-register increment through overflow\n"
+                "may set OF (and therefore LCOFIP). No mip/mie sweep here: this\n"
+                f"coverpoint tests the absence of a software-only path to LCOFIP.\n"
+                f"mode = {priv_mode}."
+            ),
+        ),
+        "",
+        _csr_access("csrw mip, zero   # clear pending", priv_mode),
```
davidharrishmc: Doubt this is necessary.

generators/testgen/src/testgen/priv/extensions/SscofpmfCommon.py
```diff
@@ -0,0 +1,781 @@
...
+    ######################################
+
+    r_val, r_of = test_data.int_regs.get_registers(2, exclude_regs=[0, 31])
+
+    lines = [
+        comment_banner(
+            coverpoint,
+        ),
+        "",
+        _csr_access("csrw mip, zero   # clear LCOFIE", priv_mode),
```
davidharrishmc: Doubt this is necessary.  Comment is incorrect.  Globally check mip.

generators/testgen/src/testgen/priv/extensions/SscofpmfCommon.py
```diff
@@ -0,0 +1,781 @@
...
+    ]
+
+
+def prime_counter_overflow(r_val: int, r_hval: int, r_temp: int, r_addr: int, priv_mode: str) -> list[str]:
+    """Overflow RVMODEL_MHPMCOUNTER through RVMODEL_MHPMEVENT_CODE (OF 0 -> 1, raising LCOFIP)."""
+    return [
+        *write_event_pattern(r_val, r_hval, 0, priv_mode),
+        *write_counter_all_ones(r_temp, priv_mode),
+        f"LA(x{r_addr}, scratch)",
+        "# Overflow must occur only via RVMODEL_MHPMEVENT_CODE; run at least twice per spec",
```
davidharrishmc: Explain why it has to be run twice, or change to once.

generators/testgen/src/testgen/priv/extensions/SscofpmfCommon.py
```diff
@@ -0,0 +1,781 @@
...
+
+    r_val, r_temp = test_data.int_regs.get_registers(2, exclude_regs=[0, 31])
+
+    lines = [
+        comment_banner(
+            coverpoint,
+            f"{inh_prefix.upper()}INH bit (mhpmevent[{inh_bit_pos}]) inhibits counting in {priv_mode}-mode.",
+        ),
+        "",
+        _csr_access("csrw mip, zero   # clear pending before toggle/sweep", priv_mode),
```
davidharrishmc: Don't think this is necessary.  Global change.

davidharrishmc:
Reviewed at head `87dfc32ff`. Generated and ran all three suites on 9 Sail configs, spike-rv32/rv64-max, and imperas.

## Results as run

| Config | SscofpmfSm-00 | SscofpmfSm-01 | SscofpmfS | SscofpmfU |
|---|---|---|---|---|
| 9 sail configs (rv32/rv64-max, clang, RVA22/23, RVB23, optionals) | pass | pass | pass | pass |
| spike-rv32-max, spike-rv64-max | pass | fail `xinh_combo_00100` | fail `xinh_combo_00100` | fail `xinh_combo_00100` |
| imperas-rv32gcv | — | — | reference ELF does not build | — |

## T-SBI conversion: correct

- Legacy `RVTEST_GOTO_MMODE`, `RVTEST_GOTO_LOWER_MODE` and `RVTEST_TEMP_BOOT_TO_U` are gone from all three suites. Each boots to its own mode (`BOOT_TO_MMODE` / `BOOT_TO_SMODE` / default-lowest) and stays there.
- `_S_ACCESSIBLE_CSRS = ("sip", "sie", "sstatus", "scountovf")` reads S-level CSRs directly in S.
- The single-bit inhibit toggles normalise the counter (`snez`) and write it to the signature.
- `imperas-rv64-max`'s `RVMODEL_MHPMEVENT_CODE` now takes two parameters and assembles.
- The `tsbi_instr_table` additions (mhpmevent3..31, mhpmevent3h..31h, mhpmcounter3, scountovf) are correct; `tsbi_call` strips the trailing comment before encoding.
- Dispatch is right. In the U suite the `ecall` is delegated to S and `csrrs x10, scountovf, x0` retires in S mode at `Srvmodel_sv`; the reads return 33 distinct values across the `mcounteren` walk, so the masking is genuinely exercised.

## 1. The T-SBI round trip is counted by the counter under test

`mhpmcounter3` is M-level, so both the zeroing and the read are serviced in M (57 accesses in the S suite, 52 in the U suite, all `mode=M`). Measured from the Sail traces, between zeroing and reading the counter:

| Suite | zero→read windows | median window | summed by mode |
|---|---|---|---|
| SscofpmfS-00 | 7 | 1,300 instructions | M 7,513 / S 252 |
| SscofpmfU-00 | 7 | 1,442 instructions | M 7,513 / S 852 / U 250 |

The intended workload is one `lw` (`RVMODEL_MHPMEVENT_CODE`). No write in the S or U suite sets MINH — `grep -c "<< 62"` on the generated S test returns 0 — so M-mode counting stays enabled, and in the U suite SINH is clear as well. `cp_sinh_inhibits_smode` / `cp_uinh_inhibits_umode` expect the counter to read 0 when the inhibit bit is set; on a model that counts, it reads the ~1,300 intervening instructions instead.

Whole-test mode split, for the instruction/trap delta `docs/tsbi-changes.md` asks for:

| Suite | retired | M | S | U | M entries | median M instrs per entry |
|---|---|---|---|---|---|---|
| SscofpmfSm-00 | 2,437 | 100% | 0 | 0 | 1 | — |
| SscofpmfSm-01 | 6,825 | 100% | 0 | 0 | 1 | — |
| SscofpmfS-00 | 1,060,024 | 98.4% | 16,824 | 0 | 1,254 | 865 |
| SscofpmfU-00 | 225,550 | 81.8% | 34,640 | 6,351 | 331 | 650 |

## 2. `scountovf` is never checked

124 `scountovf` reads in the S suite and 186 in the U suite produce **zero** signature writes. The value is loaded into a register and overwritten; the labels exist only as coverage sample points. A DUT returning any value passes both suites. The stimulus is correct (item above) — only the checking is missing.

## 3. Duplicate `clr_Sext_int` in `tests/env/rvtest_trap_handler.h`

A fourth `\__MODE__\()clr_Sext_int` is added at line 2612, inside the `#else` of `#ifdef STANDARD_SM_SUPPORTED` (line 2598), where `clr_Sext_int` is already defined at line 2604 — a duplicate label in the same branch. The added body is the pre-#2170 version using `RVTEST_GOTO_LOWER_MODE Smode` and the legacy `li a0, 0; ecall` GOTO_MMODE, which act4 replaced with `RVTEST_TSBI_CSR_READ` / `RVTEST_TSBI_CSR_CLEAR`.

It does not break today because that branch's handler macro is never instantiated: `EXTENSIONS=I` on `spike-RVI20U64` (no `STANDARD_SM_SUPPORTED`) builds 255 ELFs clean. It looks like a merge artifact and should be reverted.

## 4. CI fails on the only platform that is not excluded

`config/{sail,spike,whisper,qemu}/ci.yaml` exclude all four suite names; `config/imperas/ci.yaml` excludes none. `imperas-rv32gcv` declares Sscofpmf in its UDB yaml but its `sail.json` has `"Sscofpmf": {"supported": false}`, so the reference model rejects `csrw mhpmevent3h`:

```
RVCP: Instruction that trapped: 0x72359073      # csrw 0x723, x11
RVCP: XCAUSE:  0x00000002                       # illegal instruction
✗ Build failed: 3 succeeded, 21 up-to-date, 1 failed
```

`imperas-rv64-max` and `imperas-rv32gck` have `supported: true` and build. Either fix the `imperas-rv32gcv` config/UDB mismatch or exclude the suites on imperas too.

## 5. Spike reference config has no HPM counters

`config/spike/spike-rv*-max/sail.json` sets `writable_hpm_counters` to `0x0000_0007` (counters 0-2), so Sail treats `mhpmevent3` as read-only zero while Spike implements it. Every inhibit-combination bin diverges from `xinh_combo_00100` on. whisper uses `0x0000_07F8`, qemu and sail `0xFFFF_FFFF`.

## 6. Counter selection is hardcoded

`_FIXED_TSBI_ALIASES` maps `RVMODEL_MHPMEVENT` to `0x323` and `RVMODEL_MHPMCOUNTER` to `0xb03`, so in S and U a platform that points those macros at a different counter is tested on counter 3 regardless.

## Context

Sail does not count HPM events (sail-riscv#1733), and spike and imperas set `RVMODEL_MHPMEVENT_VAL 0x0` with a `nop` workload. Every counter-value expectation is therefore vacuous against the reference model: the suites pass on all 9 Sail configs because nothing counts anywhere. Items 1 and 2 are latent until that changes.

config/whisper/ci.yaml
ayeshaanwaar05: Need to remove this comment now because counters started incrementing when we changed val for EVMODEL_MHPMEVENT_VAL 14 to 11

config/imperas/imperas-rv64-max/rvmodel_macros.h
ayeshaanwaar05: Can add that but imperas isn't included in CI that's why i didn't add it in exclusion list of ci

config/spike/spike-rv64-max/rvmodel_macros.h
ayeshaanwaar05: Yeah there is separate issue filed on spike will update that

config/imperas/imperas-rv64-max/rvmodel_macros.h
davidharrishmc: We still maintain a list of tests that are failing imperas, to help track progress.

davidharrishmc: You can indicate that there are no plans for Imperas to support hpm events and so this will not be closed in the foreseeable future.

davidharrishmc:
Claude review of PR #2181 at head `525a7c05a`. Generated all three suites and built on sail (13 configs) and spike.

## Status: the PR does not build at head

Two independent integration breaks, both from the act4 merges since the 2026-09-11 review (which built and passed on 9 sail configs).

### 1. Blocking — `SscofpmfSm` uses interrupt macros act4 has renamed

`SscofpmfSm.py:152,165,195,206` emit the unsuffixed `RVTEST_SET_MEXT_INT`, `RVTEST_CLR_MEXT_INT`, `RVTEST_SET_MSW_INT`, `RVTEST_CLR_MSW_INT`. act4's `tests/env/utils.h` now defines only the privilege-suffixed forms (`_M`, `_S`, `_U`); there is no unsuffixed spelling left.

```
SscofpmfSm-00.S:234: Error: unrecognized opcode `rvtest_set_mext_int'
SscofpmfSm-00.S:250: Error: unrecognized opcode `rvtest_clr_mext_int'
SscofpmfSm-00.S:347: Error: unrecognized opcode `rvtest_set_msw_int'
SscofpmfSm-00.S:363: Error: unrecognized opcode `rvtest_clr_msw_int'
```

The suite runs in M-mode, so the fix is `_M` on all four.

### 2. Blocking — `clr_Lcofi_int` is defined in only one `#ifdef` branch

The new cause-13 dispatch entry is emitted unconditionally, but the routine it points at is defined only inside the `#else` of `#ifdef STANDARD_SM_SUPPORTED` (`rvtest_trap_handler.h:2518` / `:2611`). Every config that defines `STANDARD_SM_SUPPORTED` therefore links against a symbol that does not exist:

```
SscofpmfSm-00.S:415: undefined reference to `Mclr_Lcofi_int'
SscofpmfSm-00.S:415: undefined reference to `Sclr_Lcofi_int'
```

`clr_Sext_int` next door has three definitions, one per branch; `clr_Lcofi_int` has one. It needs the same treatment. After fixing item 1 locally, this is what every sail config fails on.

Scope: only the Sscofpmf suites are affected — `ZicntrSm` still builds with this handler — so it is not a repo-wide regression, but the Sm suite cannot build anywhere until it is fixed.

### 3. Blocking — `config/qemu/ci.yaml` re-disables five unrelated suites

The line adds `ZkrSm, ZkrS, ZkrU, ZicntrS, ZicntrU` on top of the Sscofpmf entries. act4 commit `a99bc0c16d` ("Update exclusions for QEMU 11.1.1", #2247, 2026-09-04) deliberately **removed** those five after QEMU 11.1.1 fixed them. Merging as-is silently turns them back off in CI.

This is the same thing flagged on 2026-09-11 ("start with the latest in ACT4 and add just your Sscofpmf* changes"). sail, spike and whisper are now clean — qemu is the only one still out of sync.

## Still open from the 2026-09-11 review

**`scountovf` is never checked.** 125 reads in `SscofpmfS-00`, 35 in `SscofpmfSm-01`, and not one reaches the signature — the value is read into a register and overwritten. The labels say what the intent was:

```
Sscofpmf_cg_cp_scountovf_shadow_scountovf_shadow_s_all_0s:
  csrr x8, scountovf   # sample point -- must match OF pattern
```

Nothing matches it. `cp_scountovf_shadow` and `cp_scountovf_mcounteren` pass on a DUT that returns any value.

**The T-SBI round trip is counted by the counter under test.** Unchanged, and the `snez` normalisation sharpens it. In `cp_sinh_inhibits_smode` the sequence is `csrw mhpmcounter3, zero` (T-SBI ecall into M) → `csrr x15, mhpmcounter3` (a second ecall into M) → `snez` → signature. Nothing sets MINH, so M-mode counting is enabled across both round trips. The `sinh = 1` case expects 0 but reads nonzero on any DUT that counts, because SINH inhibits S-mode only. It passes today solely because Sail does not count (sail-riscv#1733).

## Smaller review comments not yet addressed

- **imperas `ci.yaml`** — no exclusion added. Requested twice, with the note that there are no plans for Imperas to support hpm events so it will not be closed soon.
- **`config/spike/spike-rv64-max/rvmodel_macros.h:190`** — the comment reads `Spike issue: https://github.com/riscv/sail-riscv/issues/1733`, which is the *Sail* issue. Either link the real Spike issue or call it the Sail issue.
- **`config/whisper/ci.yaml`** — `# Sscofpmf: Counters aren't incrementing` still has no issue link. It may also be stale: per the comment thread, counters started incrementing once `RVMODEL_MHPMEVENT_VAL` changed 14 → 11.
- **`config/whisper/whisper-rv64-max/rvmodel_macros.h:210`** — `RVMODEL_MHPMEVENT_VAL 11` still has no comment saying event 11 counts loads.
- **`SscofpmfS.py:197`** — "timer maturity" is still there.
- **`csrw mip, zero`** — still emitted 10× in the S suite and 18× in the U suite.
- **S and U still switch to M-mode and write `mideleg` there.** The legacy `RVTEST_GOTO_MMODE` is gone, but `RVTEST_TSBI_GOTO_MMODE` followed by a direct `csrs/csrc mideleg` is the same mode hop (`SscofpmfS.py:41,90,141`; `SscofpmfU.py:38,85`). `InterruptsS/U` do this too, but those are the suites #2263 replaces, so it is not a pattern to copy.

## Fixed since the last review

- The duplicate `clr_Sext_int` merge artifact is gone, replaced by a real `clr_Lcofi_int` wired into cause 13 that clears via `sip` below M — the right shape, modulo the guard bug above.
- Raw `mtimecmp` manipulation is gone from the S suite; only Sm touches it.
- LCOFI is disabled through `sip` in S.
- The "run at least twice" sequence now carries an explanatory comment.
- sail, spike and whisper `ci.yaml` exclusions are Sscofpmf-only and match act4.
- `spike` `sail.json` `writable_hpm_counters` was left at act4's `0x0000_0007`; the `xinh_combo` divergence stands, but spike is excluded from CI so it does not block.

ayeshaanwaar05:
Prof @davidharrishmc I think you reviewed 525a7c05a  your own local merge of act4 into your branch while the actual PR head f134bed6a had since been amended. like most of the fixes have been already made

davidharrishmc:
@ayeshaanwaar05 you are right.  This is ready for @jordancarlin review.

davidharrishmc (approved):

PR #2183, opened 2026-08-24
ExceptionsVx/SsstrictV fixes

davidharrishmc:
Replace opcodes that Clang rejects with .insn in SsstrictV so sail-clang can compile
Select legal registers on ExceptionsVx for vmv* so sail-clang can compile and tests run as intended
Cite imperas vmv*v with vill not trapping issue on ExceptionsVx


davidharrishmc:
@fourth-bit can you review this one?

generators/testgen/scripts/vector-testgen-priv.py
```diff
@@ -365,6 +365,10 @@ def writePrivTestPrep(description, instruction, instruction_data=None, lmul = 1,
             vd_sig_lmul = lmul if isinstance(lmul, int) else 1
         vd_emul  = max(1, int(lmul * vec_data['vd' ].get('size_multiplier', 1) * vec_data['vd' ].get('segments', 1)), vd_sig_lmul)
         vs2_emul = max(1, int(lmul * vec_data['vs2'].get('size_multiplier', 1) * vec_data['vs2'].get('segments', 1)))
+        if instruction in whole_register_move:
```
fourth-bit: It seems like this should also include whole register load/stores. Same goes for the changes to vector_testgen_common.

generators/testgen/scripts/priv/_ssstrictv_helpers.py
```diff
@@ -135,6 +180,10 @@ def build_testline(instruction: str, instruction_data: list, *,
         testline += ", "
 
     testline = testline[:-2]
+    if common.extension == "SsstrictV":
```
fourth-bit: This if statement seems unnecessary for a SsstrictV helper

generators/testgen/scripts/vector-testgen-priv.py
davidharrishmc: done

generators/testgen/scripts/priv/_ssstrictv_helpers.py
davidharrishmc: removed

jordancarlin:
@fourth-bit please take another look at this when you get a chance.

fourth-bit (approved):
LGTM

jordancarlin (approved):

PR #2184, opened 2026-08-24
Reverse engineered PMP python generators from .S files

davidharrishmc:
Done by AI.  Needs review.

AI sanity checks:
Same number of traps generated
Same ELF (modulo SIGUPD count changes)
Source is the same except for reasonable changes in comments

Reduced from 35,505 lines of manually generated .S files to 8,584 lines of Python.  Python is still long because of inconsistencies across files.  Improve this with a followup PR.


davidharrishmc:
Closed in favor of 2186, which is less identical but shorter and fixes bugs.

PR #2186, opened 2026-08-25
More aggressive PMP Python Generator

davidharrishmc:
As compared to the earlier PR #2184, this version is more aggressive about fixing bugs in PMP and unifying inconsistent code sequences to simplify the generator.

The results are no longer identical, but are better.  100% coverage, pass all sims except Imperas, which now has a bug that a fetch from illegal address 0 with CLINT enabled does not throw a fetch fault.

This PR replaces the per-file verification macros that the pmp2gen generators reproduced verbatim
  with a single set of 20 macros in tests/env/rvtest_pmp_macros.h, built from two primitives
  (PMP_PROBE, PMP_PROBE_X) and selected by XLEN with .if; the 47 distinct local VERIFICATION_RWX
  bodies and the pmp_templates/ directory are gone, and every generated test now calls a header
  macro. The generators shrink from 6,764 Python + 1,820 template lines to 2,478 lines: PMPS and
  PMPU become one module parameterised by mode, the three PMPSm modules merge into one, and shared
  factories (walk_file, entry_walk, csr_write) cover the LXWR-walk and every-entry families. Test
  content changes are limited to correctness and consistency fixes, each checked against the pmp2gen
  baseline by per-file Sail trap comparison: the TOR -01 files use entries 11/9/7 as the
  coverpoints require, and PMP_VERIFICATION_RWX_LEGAL probes the TOR granule instead of the NAPOT
  minimum, which takes PMPS, PMPU and PMPSm from 95–99% to 100% coverage on both Sail configs (all
  eight suites are now 100%); store probes write a ret instead of a NOP so an overwritten pad no
  longer causes order-dependent fetch faults; PMPZalrsc moves to the page-aligned region so its rv64
  lr.d/sc.d are no longer accidentally misaligned; the rv32 pmpcfg walk covers all 16 CSRs with a
  new odd-CSR cross; and the redundant mstatus.FS enables, the duplicated cfg_A_off_all case, the
  inert RVTEST_PRIV_TEST define, the in-M-mode RVTEST_GOTO_MMODE calls and the mismatched
  coverpoint-name strings are removed. All 158 tests pass on spike-rv32-max and spike-rv64-max.


UmerShahidengr (approved):

jordancarlin:
@UmerShahidengr I’m planning to review this and the virtual memory PR later today. Let’s wait to merge until after that.

generators/testgen/src/testgen/priv/pmp/suites/__init__.py
```diff
@@ -0,0 +1,2 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Declarative PmpFile tables for the generated pure PMP suites."""
```
jordancarlin: What does this mean? I have no clue what a "Declarative PmpFile table" is.

jordancarlin:
Closing in favor of #2218.

PR #2190, opened 2026-08-26
Test valid values in Sm and S CSRs

davidharrishmc:
Also refactors to use PrivCommon for Sm, S, U.


generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -8,6 +8,8 @@
 
 """Sm privileged extension test generator."""
 
+from __future__ import annotations
+
```
jordancarlin: Remove this.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -451,10 +451,85 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+    `ifdef SM_VADDR_WALK
+        `ifdef SV57_SUPPORTED
+            `define SM_VADDR_WALK_MSB 55
+        `elsif SV48_SUPPORTED
+            `define SM_VADDR_WALK_MSB 46
+        `elsif SV39_SUPPORTED
+            `define SM_VADDR_WALK_MSB 37
+        `else
+            `define SM_VADDR_WALK_MSB 31
+        `endif
```
jordancarlin: Simplify this to something like the following:

```suggestion
    `ifdef SV57_SUPPORTED
        `define SM_VADDR_WALK_MSB 55
    `elsif SV48_SUPPORTED
        `define SM_VADDR_WALK_MSB 46
    `elsif SV39_SUPPORTED
        `define SM_VADDR_WALK_MSB 37
    `elsif SV32_SUPPORTED
        `define SM_VADDR_WALK_MSB 31
    `endif

    `ifdef SM_VADDR_WALK_MSB 
```

coverpoints/priv/Sm_coverage.svh
```diff
@@ -451,10 +451,85 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        mepc: coverpoint ins.current.insn[31:20] {
+            bins mepc = {CSR_MEPC};
+        }
+        // mnepc walks are disabled until the Sail reference model supports Smrnmi
+        // `ifdef SMRNMI_SUPPORTED
+        //     mnepc: coverpoint ins.current.insn[31:20] {
+        //         bins mnepc = {CSR_MNEPC};
+        //     }
+        // `endif
```
jordancarlin: This should eventually go in a separate Smrnmi suite, not as part of Sm.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -16,6 +18,30 @@
...
+# which the canonical-address walks in _generate_vaddr_walk_tests cover.
+SM_CSRS_NOWALK = ["mepc", "mtval"]
+
+# Address CSRs that must hold every valid virtual address. The value written is the address
+# itself: mepc and mnepc hold bit 0 at 0 and bit 1 at 0 unless Zca makes 2-byte alignment legal;
+# mtval holds any byte address. Each entry is (csr, held-low mask, {bit: gate define}, gate define).
+SM_VADDR_CSRS = [
+    ("mepc", 0b01, {1: "ZCA_SUPPORTED"}, None),
+    ("mtval", 0b00, {}, None),
+    # ("mnepc", 0b01, {1: "ZCA_SUPPORTED"}, "SMRNMI_SUPPORTED"),  # until the Sail reference model supports Smrnmi
```
jordancarlin: Drop. Will be part of Smrnmi suite.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -16,6 +18,30 @@
...
+]
```
jordancarlin: Seems like you could drop `SM_CSRS_NOWALK` and just use `SM_VADDR_CSRS.keys()`

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -16,6 +18,30 @@
 from testgen.priv.extensions.S import S_CSR_SENVCFG, S_CSRS, S_CSRS_NOWALK, S_SSTATUS_MASK
 from testgen.priv.registry import add_priv_test_generator
 
+# Address CSRs not walked with csr_walk_test: they only have to hold valid virtual addresses,
+# which the canonical-address walks in _generate_vaddr_walk_tests cover.
+SM_CSRS_NOWALK = ["mepc", "mtval"]
+
+# Address CSRs that must hold every valid virtual address. The value written is the address
+# itself: mepc and mnepc hold bit 0 at 0 and bit 1 at 0 unless Zca makes 2-byte alignment legal;
+# mtval holds any byte address. Each entry is (csr, held-low mask, {bit: gate define}, gate define).
```
jordancarlin: Simplify comment. What does "The value written is the address itself" mean?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -442,6 +468,152 @@ def _add_shadow(
...
+    )
+    if gate is not None:
+        lines.append(f"#endif // {gate}")
+    test_data.int_regs.return_registers([save_reg, ones_reg, walk_reg, check_reg])
+    return lines
+
+
+def _generate_vaddr_walk_tests(test_data: TestData, test_chunks: list[TestChunk]) -> None:
+    """Generate the mtval value tests and the valid virtual address walks for mepc, mtval, and mnepc."""
+    tc = test_data.new_test_chunk(test_chunks, "mcsr_vaddr")
```
jordancarlin: Chunk and function name do not align with the actual contents. This includes mtval_ilen tests too.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -442,6 +468,152 @@ def _add_shadow(
...
+            gen_csr_read_sigupd(check_reg, csr, test_data),
+            "#endif // UDB_REPORT_ENCODING_IN_MTVAL_ON_ILLEGAL_INSTRUCTION",
+            f"csrw mtval, x{save_reg}            # restore CSR",
+        ]
+    )
+    test_data.int_regs.return_registers([save_reg, walk_reg, check_reg])
+    return lines
+
+
+def _vaddr_walk_test(
```
jordancarlin: Might be worth moving this function into a common file of some kind. I think we need to add the same test for `sepc` and `stval`.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -442,6 +468,152 @@ def _add_shadow(
...
+        "Write 0 to mtval, and every ILEN-bit value as a walking 1 and all 32 1s when the illegal instruction\n"
+        "encoding is reported in mtval. Write every valid virtual address as a walking 1 (0s in the msbs) and a\n"
+        "walking 0 (1s in the msbs) to mepc, mtval, and mnepc (Smrnmi). Canonical addresses have bits\n"
+        "XLEN-1:VALEN-1 equal, so bits low..VALEN-2 are walked; requires Sv32 or Sv39 and extends the walk to\n"
+        "Sv48/Sv57 when supported",
+    )
+    tc.code.extend(_mtval_value_tests(test_data))
+    for csr_name, held_low, gated_bits, gate in SM_VADDR_CSRS:
+        tc = test_data.new_test_chunk(test_chunks)
+        tc.code.extend(_vaddr_walk_test(test_data, csr_name, held_low, gated_bits, gate))
```
jordancarlin: The spec says
> When address translation is not in effect, virtual addresses and physical addresses are equal. Hence, the set of addresses mepc must be able to represent includes the set of physical addresses that can be used as a valid pc or effective address.

How about also testing a few physical addresses (current address, base of signature, etc)?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -442,6 +468,152 @@ def _add_shadow(
...
+        )
+    lines.extend(
+        [
+            "",
+            "# Testcase: mtval = all 32 encoding bits set",
+            f"LI(x{walk_reg}, 0xFFFFFFFF)",
+            test_data.add_testcase("ones", "cp_mtval_ilen_ones", covergroup),
+            f"csrw mtval, x{walk_reg}",
+            gen_csr_read_sigupd(check_reg, csr, test_data),
+            "#endif // UDB_REPORT_ENCODING_IN_MTVAL_ON_ILLEGAL_INSTRUCTION",
```
jordancarlin: Seems like this could pretty easily use `csr_walk_test` just by changing that function to take an optional `end_bit` instead of hardcoding 64.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -442,6 +468,152 @@ def _add_shadow(
...
+            f"csrw mtval, x{save_reg}            # restore CSR",
+        ]
+    )
+    test_data.int_regs.return_registers([save_reg, walk_reg, check_reg])
+    return lines
+
+
+def _vaddr_walk_test(
+    test_data: TestData, csr_name: str, held_low: int, gated_bits: dict[int, str], gate: str | None
+) -> list[str]:
```
jordancarlin: This function is very convoluted and hard to follow as a human. It comes across as very AI generated and I am struggling to make sense of what it is trying to do. Please find ways of simplifying/refactoring it.

generators/testgen/src/testgen/priv/extensions/Sm.py
jordancarlin: `mtval` only needs to support all of the canonical virtual addresses if it actually reports virtual addresses on access or page faults. Making `mtval` read-only 0 is a perfectly legal implementation.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/Sm.py
jordancarlin: Just saw #2185 which adds the same tests for sepc and stval. Please combine the two PRs and factor the common logic out into a shared function.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: I think this should be captured with a Sail parameter because we want to test virtual addresses if they are allowed.  Is it on your list of UDB parameters?

jordancarlin: There are already both UDB and Sail parameters for it. There are separate parameters for if each kind of exception records a value in xtval.

davidharrishmc: ok, so it's fine to test the walking ones and expect 0 will be written in both Sail and DUT so long as Sail is properly configured?

jordancarlin: Should be. Not sure what I was thinking before.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: Turns out not to be so easy.  Depends on SV parameter that is not known at code generation time.

generators/testgen/src/testgen/priv/extensions/U.py
```diff
@@ -8,99 +8,43 @@
...
+    tc.code.extend(
+        priv_inst_trap_tests(
+            test_data,
+            "U_uprivinst_cg",
+            "cp_uprivinst",
+            "Execute privileged instructions",
+            [
+                ("ebreak", "ebreak              # test ebreak instruction"),
+                ("mret", "mret                  # test mret instruction"),
+                ("sret", "sret                  # test sret instruction"),
```
jordancarlin: This seems very redundant. Can we simplify the interface to just take a list of instructions and construct the comment in the function?

generators/testgen/src/testgen/priv/extensions/S.py
```diff
@@ -45,14 +51,20 @@
...
     ("sie", 0xFFFF),  # only test standard non-reserved portion
 ]
-# skip walking 1s on this because valid virtual addresses is not described adequately
-S_CSRS_NOWALK = [
-    ("sepc", None),  # only has to be able to hold all valid virtual addresses
-    ("stval", None),  # only has to be able to hold all valid virtual addresses and 0
-]
+# Address CSRs that must hold every valid virtual address: {csr: (held-low mask, {bit: gate define})}.
+# stvec carries the address in BASE with MODE = Direct, so its two low bits are held at 0;
+# sepc holds bit 0 at 0 and bit 1 at 0 unless Zca makes 2-byte alignment legal;
```
jordancarlin: Are we testing somewhere else that `sepc[1]` is read-only 0 when Zca is supported?

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -13,9 +13,20 @@
 from testgen.constants import INDENT
 from testgen.data.state import TestData
 from testgen.data.test_chunk import TestChunk
-from testgen.priv.extensions.S import S_CSR_SENVCFG, S_CSRS, S_CSRS_NOWALK, S_SSTATUS_MASK
+from testgen.priv.extensions.PrivCommon import (
+    addr_csr_tests,
+    csr_insufficient_priv_tests,
+    csr_ro_write_tests,
+)
+from testgen.priv.extensions.S import S_CSR_SENVCFG, S_CSRS, S_SSTATUS_MASK
```
jordancarlin: Now that we have a common file, how about moving these to PrivCommon and be renaming them to XENVCFG, XSTATUS, etc. That way Sm is not importing from S.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -442,6 +453,52 @@ def _add_shadow(
     )
 
 
+def _mtval_value_tests(test_data: TestData) -> list[str]:
+    """Write 0 and, when illegal-instruction encodings are reported, every ILEN-bit value to mtval."""
```
jordancarlin: These same requirements apply to stval too. Why is this mtval specific?

> stval is a WARL register that must be able to hold all valid virtual addresses and the value 0. It need not be capable of holding all possible invalid addresses. Prior to writing stval, implementations may convert an invalid address into some other invalid address that stval is capable of holding. If the feature to return the faulting instruction bits is implemented, stval must also be able to hold all values less than 2N, where N is the smaller of SXLEN and ILEN.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -442,6 +453,52 @@ def _add_shadow(
...
+    lines = [
+        "",
+        "# mtval must hold 0",
+        f"csrr x{save_reg}, mtval      # Save CSR",
+        "",
+        "# Testcase: mtval = 0",
+        f"LI(x{walk_reg}, 0)",
+        test_data.add_testcase("zero", "cp_mtval_zero", covergroup),
+        f"csrw mtval, x{walk_reg}",
+        gen_csr_read_sigupd(check_reg, csr, test_data),
```
jordancarlin: Could this be simplified to use `gen_csr_write_sigupd`?

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+    """One walk iteration: write a canonical address with `bit` set (or clear) and check the readback."""
+    csr = (csr_name, None)
+    if walking_ones:
+        lines = [
+            "",
+            f"# Testcase: {csr_name} = bit {bit} set, 0s elsewhere",
+            f"LI(x{walk_reg}, {1 << bit:#x})",
+            test_data.add_testcase(f"walking1_{bit}", f"cp_{csr_name}_vaddr_walk1", covergroup),
+            f"csrw {csr_name}, x{walk_reg}",
+            gen_csr_read_sigupd(check_reg, csr, test_data),
```
jordancarlin, line 85: `gen_csr_write_sigupd`?

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+def priv_inst_trap_tests(
+    test_data: TestData,
+    covergroup: str,
+    coverpoint: str,
+    description: str,
+    instrs: list[tuple[str, str]],
+) -> list[str]:
+    """An ecall answered by the T-SBI environment, then privileged instructions that must trap."""
```
jordancarlin: Confusing comment.

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+    test_chunks: list[TestChunk],
+    covergroup: str,
+    csr_ranges: list[range],
+    chunk_name: str,
+) -> None:
+    """Attempt to write each read-only CSR; every write must raise illegal instruction."""
+    coverpoint = "cp_csr_ro"
+    tc = test_data.new_test_chunk(test_chunks, chunk_name)
+    tc.section_header = comment_banner(coverpoint, "Attempt to write read-only CSRs.  Should throw illegal instruction")
+    for csr in (csr for csr_range in csr_ranges for csr in csr_range):
```
jordancarlin: This is a very difficult line to parse. Suggest either renaming some of these variables or more likely splitting this up somehow.

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+        tc = test_data.new_test_chunk(test_chunks, chunk_name)
+        temp_reg = test_data.int_regs.get_register()
+        tc.code.extend(
+            [
+                test_data.add_testcase(f"{csr:03x}", coverpoint, covergroup),
+                f"LI(x{temp_reg}, -1)          # x{temp_reg} = all 1s",
+                f"csrw 0x{csr:03x}, x{temp_reg}    # attempt to write read-only CSR {csr:03x}; should get illegal instruction",
+                "",
```
jordancarlin: Do we test anywhere that the register that a CSR write reads back into doesn't get updated when it traps from an illegal instruction?

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+    tc = test_data.new_test_chunk(test_chunks, chunk_name)
+    tc.section_header = comment_banner(coverpoint, description)
+    for csr in (csr for csr_range in csr_ranges for csr in csr_range):
+        tc = test_data.new_test_chunk(test_chunks, chunk_name)
+        temp_reg = test_data.int_regs.get_register()
+        tc.code.extend(
+            [
+                test_data.add_testcase(f"{csr:03x}", coverpoint, covergroup),
+                f"csrr x{temp_reg}, 0x{csr:03x}    # attempt to read CSR {csr:03x}; should get illegal instruction",
+                "",
```
jordancarlin: Should we check that the register this is supposed to be read into doesn't change? Would require changing it to not read into x0 like it does right now.

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+    """Real-address and canonical-address walk tests for each address CSR, one chunk per CSR."""
+    names = ",".join(vaddr_csrs)
+    tc = test_data.new_test_chunk(test_chunks, chunk_name)
+    tc.section_header = comment_banner(
+        f"cp_{{{names}}}_vaddr_walk{{1,0}}, cp_{{{names}}}_vaddr_{{pc,scratch}}",
+        "Write every valid virtual address as a walking 1 (0s in the msbs) and a walking 0 (1s in the msbs)\n"
+        "registers that hold addresses. Canonical addresses have bits XLEN-1:VALEN-1 equal, so bits\n"
+        "low..VALEN-2 are walked; requires Sv32 or Sv39 and extends the walk to Sv48/Sv57 when supported.\n"
+        "Also write two real addresses — the current pc and the scratch area — which are valid even when\n"
+        "address translation is off, so those tests are not gated on Sv* support",
```
jordancarlin, line 208:
```suggestion
        "Write every valid virtual address as a walking 1 (0s in the msbs) and a walking 0 (1s in the msbs)\n"
        "to registers that hold addresses. Canonical addresses have bits XLEN-1:VALEN-1 equal, so bits\n"
        "low..VALEN-2 are walked; requires Sv32 or Sv39 and extends the walk to Sv48/Sv57 when supported.\n"
        "Also write two physical addresses — the current pc and the scratch area — which are valid even when\n"
        "address translation is off, so those tests are not gated on Sv* support",
```

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+    lines += [
+        f"csrw {csr_name}, x{save_reg}            # restore CSR",
+        f"#endif // {VADDR_GATE.split(' ', 1)[1]}",
+    ]
+    test_data.int_regs.return_registers([save_reg, ones_reg, walk_reg, check_reg])
+    return lines
+
+
+def vaddr_value_tests(test_data: TestData, csr_name: str, covergroup: str) -> list[str]:
+    """Write real addresses (the current pc and scratch) to csr_name and check they read back exactly.
```
jordancarlin:
```suggestion
    """Write physical addresses (the current pc and scratch) to csr_name and check they read back exactly.
```

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,247 @@
...
+        f"csrr x{save_reg}, {csr_name}      # Save CSR",
+        "",
+        f"# Testcase: {csr_name} = current pc",
+        f"auipc x{walk_reg}, 0",
+        test_data.add_testcase("pc", f"cp_{csr_name}_vaddr_pc", covergroup),
+        f"csrw {csr_name}, x{walk_reg}    # directly follows the auipc so the value is this csrw's pc - 4",
+        gen_csr_read_sigupd(check_reg, csr, test_data),
+        "",
+        f"# Testcase: {csr_name} = address within scratch",
+        f"la x{walk_reg}, scratch",
```
jordancarlin: Use the LA macro.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -451,10 +451,92 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+    mtval_zero: coverpoint ins.current.rs1_val {
+        bins zero = {0};
+    }
+    // Real addresses are valid even when address translation is off, so these are not gated on
+    // Sv* support (the walk coverpoints below are). The generator emits the csrw immediately
+    // after `auipc xN, 0`, so the pc value written equals the csrw's pc - 4; the scratch value
+    // is scratch + 0xA8, an offset chosen so no walk, zero, or all-ones pattern shares its low byte.
+    mepc_addr: coverpoint ins.current.insn[31:20] {
+        bins mepc = {CSR_MEPC};
+    }
```
jordancarlin:
```suggestion
    mepc: coverpoint ins.current.insn[31:20] {
        bins mepc = {CSR_MEPC};
    }
```

coverpoints/priv/Sm_coverage.svh
```diff
@@ -451,10 +451,92 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+        `define SM_VADDR_WALK_MSB 46
+    `elsif SV39_SUPPORTED
+        `define SM_VADDR_WALK_MSB 37
+    `elsif SV32_SUPPORTED
+        `define SM_VADDR_WALK_MSB 31
+    `endif
+    `ifdef SM_VADDR_WALK_MSB
+        mepc: coverpoint ins.current.insn[31:20] {
+            bins mepc = {CSR_MEPC};
+        }
```
jordancarlin: Duplicate with coverpoint above.

coverpoints/priv/S_coverage.svh
```diff
@@ -260,15 +260,102 @@ covergroup S_scsr_cg with function sample(ins_t ins);
...
+    // is scratch + 0xA8, an offset chosen so no walk, zero, or all-ones pattern shares its low byte.
+    stvec_addr: coverpoint ins.current.insn[31:20] {
+        bins stvec = {CSR_STVEC};
+    }
+    sepc_addr: coverpoint ins.current.insn[31:20] {
+        bins sepc = {CSR_SEPC};
+    }
+    stval_addr: coverpoint ins.current.insn[31:20] {
+        bins stval = {CSR_STVAL};
+    }
```
jordancarlin: Inconsistent naming with other files.
```suggestion
    stvec: coverpoint ins.current.insn[31:20] {
        bins stvec = {CSR_STVEC};
    }
    sepc: coverpoint ins.current.insn[31:20] {
        bins sepc = {CSR_SEPC};
    }
    stval: coverpoint ins.current.insn[31:20] {
        bins stval = {CSR_STVAL};
    }
```

coverpoints/priv/S_coverage.svh
```diff
@@ -260,15 +260,102 @@ covergroup S_scsr_cg with function sample(ins_t ins);
...
+    `ifdef S_VADDR_WALK_MSB
+        stvec: coverpoint ins.current.insn[31:20] {
+            bins stvec = {CSR_STVEC};
+        }
+        sepc: coverpoint ins.current.insn[31:20] {
+            bins sepc = {CSR_SEPC};
+        }
+        stval: coverpoint ins.current.insn[31:20] {
+            bins stval = {CSR_STVAL};
+        }
```
jordancarlin: Duplicates with above coverpoints.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: Moved.  Should remain S, not X, because they are supervisor CSRs.

generators/testgen/src/testgen/priv/extensions/S.py
davidharrishmc: Changed mask to test bits 0 and 1, whose behavior should be DUT-dependent.

generators/testgen/src/testgen/priv/extensions/Sm.py
davidharrishmc: It could, but then loses its parallel structure with the lines above.  Doesn't seem worth messing with.

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
davidharrishmc: Good point.  Added that.

coverpoints/priv/S_coverage.svh
```diff
@@ -260,15 +260,110 @@ covergroup S_scsr_cg with function sample(ins_t ins);
...
+            sepc_vaddr_walk0: coverpoint $clog2(~(ins.current.rs1_val | 1))
+                              iff ($onehot(~(ins.current.rs1_val | 1))) {
+                bins b_0[] = { [1:`S_VADDR_WALK_MSB] };
+            }
+        `else
+            sepc_vaddr_walk0: coverpoint $clog2(~(ins.current.rs1_val | 3))
+                              iff ($onehot(~(ins.current.rs1_val | 3))) {
+                bins b_0[] = { [2:`S_VADDR_WALK_MSB] };
+            }
+        `endif
```
jordancarlin: I thought we were always going to walk bit 1 and the behavior will just be DUT dependent. We should try walking bit 0 too and it will always read back as 0.

coverpoints/priv/Sm_coverage.svh
```diff
@@ -451,10 +451,89 @@ covergroup Sm_mcsr_cg with function sample(ins_t ins);
...
+            xepc_vaddr_walk0: coverpoint $clog2(~(ins.current.rs1_val | 1))
+                              iff ($onehot(~(ins.current.rs1_val | 1))) {
+                bins b_0[] = { [1:`SM_VADDR_WALK_MSB] };
+            }
+        `else
+            xepc_vaddr_walk0: coverpoint $clog2(~(ins.current.rs1_val | 3))
+                              iff ($onehot(~(ins.current.rs1_val | 3))) {
+                bins b_0[] = { [2:`SM_VADDR_WALK_MSB] };
+            }
+        `endif
```
jordancarlin: Same comment about bit 1.

generators/testgen/src/testgen/priv/extensions/PrivCommon.py
```diff
@@ -0,0 +1,341 @@
...
+    lines += [
+        f"csrw {csr_name}, x{save_reg}            # restore CSR",
+        f"#endif // {VADDR_GATE.split(' ', 1)[1]}",
+    ]
+    test_data.int_regs.return_registers([save_reg, ones_reg, walk_reg, check_reg])
+    return lines
+
+
+def vaddr_value_tests(test_data: TestData, csr_name: str, covergroup: str) -> list[str]:
+    """Write physical addresses (the current pc and scratch) to csr_name and check they read back exactly.
```
jordancarlin: Maybe rename this function since it is about physical addresses, not virtual addresses (as implied by vaddr in the name).

jordancarlin (changes requested):
A few last minor things, but looking very close.

coverpoints/priv/S_coverage.svh
jordancarlin: Looks like the tests now include this and the coverpoints just need to be updated.

jordancarlin (approved):

PR #2194, opened 2026-08-26
Port Zawrs* to T-SBI

davidharrishmc:
Ports the Zawrs suites to T-SBI, and simplifies the per-privilege test shells to just call into common code.

- Deleted `interrupts.py`, which is obsolete now that RVTEST macros exist to set and clear each interrupt source.
- Modified `ZawrsCommon` to use the RVTEST macros to trigger and clear interrupts.
- Had to import the new delays from the Interrupts PR to give timers enough time for T-SBI round trips.

**The `Interrupts*` deletions are intentional, not an accident.** Those suites relied on `interrupts.py`, and they break with the new delays and without it. Dropping them for now and restoring when #2263 merges.

## Review updates

- `ZawrsCommon` now disarms the Sstc comparator with `RVTEST_CLR_SSTC_INT_M/S/U` instead of a local helper that wrote `stimecmp` directly. Both call sites were already inside `#ifdef SSTC_SUPPORTED`, the same guard the routines live behind. In U-mode this replaces a hand-built T-SBI `ecall` sequence per `stimecmp` write with a single `jal`.
- The `RVMODEL_TIMER_INT_SOON_DELAY` values were raised across all 19 configs earlier in this branch, but the CTP table still described the old requirement. It now says the delay must cover returning from the `RVTEST_SET_*_INT_SOON_*` routine and reaching the instruction under test, including the `ecall`, handler and return of a T-SBI round trip when the test runs below machine mode.

The SSI and SEI writes in `wrs_no_mie_helper` are deliberately left as direct `mip` writes rather than `RVTEST_SET_SSW_INT`/`RVTEST_SET_SEXT_INT`. Those macros only fall back to `mip` when the model defines no platform interrupt controller, and sail defines `RVMODEL_SET_SSW_INT` while spike does not — so the two models take different numbers of T-SBI round trips and disagree on the trap count this test records in its signature. Swapping them makes `ZawrsS` and `ZawrsU` fail on spike with a trap count of 23 against an expected 22.

## Test plan

- [x] `ZawrsS`, `ZawrsU` and `ZawrsSm` pass on `spike-rv64-max`, `spike-rv32-max`, every sail config that runs them, `whisper-rv32-max` and `whisper-rv64-max`
- [x] `qemu-rv64-max` fails `ZawrsS` and `ZawrsU` both before and after this change (pre-existing)
- [x] `ruff check` and `pyright` clean


generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -251,21 +254,26 @@ def set_stimer_int_soon_sstc(r_mtime: int, r_temp1: int, r_temp2: int, r_temp3:
...
-        f"addi x{r_temp3}, x{r_temp1}, RVMODEL_TIMER_INT_SOON_DELAY",
+        f"csrw stimecmph, x{r_temp2}",
+        f"lw x{r_temp1}, 0(x{r_mtime})",
+        f"lw x{r_temp4}, 4(x{r_mtime})",
+        f"bne x{r_temp2}, x{r_temp4}, 7b",
+        f"addi x{r_temp3}, x{r_temp1}, {delay_val}",
         f"sltu x{r_temp4}, x{r_temp3}, x{r_temp1}",
-        f"add x{r_temp2}, x{r_temp2}, x{r_temp4}",
+        f"bnez x{r_temp4}, 7b",
         f"csrw stimecmp, x{r_temp3}",
```
jordancarlin: The spec provides the following example code for setting mtimecmp/mtimecmph. Is there a reason it doesn't work here?

> Sample code for setting the 64-bit time comparand in RV32 assuming a little-endian memory system and that the registers live in a strongly ordered I/O region. Storing -1 to the low-order bits of mtimecmp prevents mtimecmp from temporarily becoming smaller than the lesser of the old and new values.
> ```
>             # New comparand is in a1:a0.
>             li t0, -1
>             la t1, mtimecmp
>             sw t0, 0(t1)     # No smaller than old value.
>             sw a1, 4(t1)     # No smaller than new value.
>             sw a0, 0(t1)     # New value.
> ```

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -8,18 +8,82 @@
...
+    if priv == "M" or match is None:
+        return instr
+    mnemonic, first, second, third = match.groups()
+    csr = (first if mnemonic in ("csrw", "csrs", "csrc") else second).lower()
+    if csr not in _M_CSRS and not (priv == "U" and csr in _S_CSRS):
+        return instr
+    if mnemonic in _RMW_TO_WRITE:
+        # No Zawrs helper uses the value read back, so the T-SBI form only writes.
+        instr = f"{_RMW_TO_WRITE[mnemonic]} {second}, {third}"
+    return tsbi_call(instr.strip())
```
jordancarlin: Doing this with regex seems very brittle.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -8,18 +8,82 @@
...
+
+
+def _csrs(priv: str, lines: list[str]) -> list[str]:
+    return [_csr(priv, line) for line in lines]
```
jordancarlin: Unclear why this function is needed.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -8,18 +8,82 @@
...
+
+
+def _mstatus(priv: str, r_temp: int, mask: int, set_bits: bool, comment: str) -> list[str]:
+    return [comment, f"LI(x{r_temp}, 0x{mask:X})", _csr(priv, f"{'csrs' if set_bits else 'csrc'} mstatus, x{r_temp}")]
```
jordancarlin: A lot of these functions seem like they are just making it harder to follow what is happening. I think AI went a bit too far in trying to eliminate duplicate code at the cost of readability.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -358,35 +433,71 @@ def wrs_no_res_helper(test_data: TestData, priv: str, covergroup: str) -> list[s
...
+    priv: str,
     coverpoint: str,
     covergroup: str,
 ) -> list[str]:
-    """WRS ops timeout behavior"""
+    """WRS ops timeout behavior; cp_wrs_nto_timeout_h runs the test in VS and VU mode from an S-mode test."""
     r_cause, r_temp, r_temp2 = test_data.int_regs.get_registers(3)
-    lines = []
     op = "WRS.STO" if coverpoint == "cp_wrs_sto_timeout" else "WRS.NTO"
+    hyp = coverpoint == "cp_wrs_nto_timeout_h"
```
jordancarlin: Don't drive the function based on the coverpoint name. How about an argument named `timeout` with valid values `short` or `no` (or something along those lines).

jordancarlin (changes requested):
A lot of the AI generated code in ZawrsCommon seems brittle and is overly convoluted or hard to follow. Please try to pull it back into something more digestible for a human.

generators/testgen/src/testgen/asm/interrupts.py
davidharrishmc: Restored the spec's three-write order; the loop wasn't needed.
This is all going away when we move to the new interrupt framework.

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -236,56 +242,77 @@ def clr_stimer_int(r_temp: int, r_stimecmp: int, r_scratch: int, r_stce: int) ->
...
         f"{INDENT}# Set supervisor timer interrupt to fire soon with Sstc extension",
         f"LA(x{r_mtime}, RVMODEL_MTIME_ADDRESS)  # NOTE: This will need to be replaced by a SBI call because MTIME might not exist or be accessible",
         "#if __riscv_xlen == 64",
         f"{INDENT}# Disable comparator first",
         f"LI(x{r_temp1}, -1)",
-        f"csrw stimecmp, x{r_temp1}",
+        _csr_instr(f"csrw stimecmp, x{r_temp1}", tsbi),
         f"{INDENT}# Read current time and add delay",
+        f"LI(x{r_temp2}, {delay_val})",
         f"ld x{r_temp1}, 0(x{r_mtime})",
```
jordancarlin: Doesn't this need to use T-SBI too because `mtime` may not be accessible to S-mode?

generators/testgen/src/testgen/asm/interrupts.py
```diff
@@ -236,56 +242,77 @@ def clr_stimer_int(r_temp: int, r_stimecmp: int, r_scratch: int, r_stce: int) ->
...
     ]
 
 
-def set_stimer_mmode(r_scratch: int) -> list[str]:
+def set_stimer_mmode(r_scratch: int, tsbi: bool = False) -> list[str]:
     """Set supervisor timer interrupt in M-mode (direct mip write).
 
     This function directly writes mip.STIP without mode transitions.
-    Must be called from M-mode.
+    Must be called from M-mode, or with tsbi=True to make the mip write a T-SBI call.
```
jordancarlin: Function name should be changed if this can now be called from modes other than M.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -8,24 +8,77 @@
...
 
 
+def m_csr(priv: str, instr: str) -> str:
+    """M-mode CSR instruction; a T-SBI call when the test runs below M-mode."""
+    return instr if priv == "M" else tsbi_call(instr)
+
+
+def s_csr(priv: str, instr: str) -> str:
+    """S-mode CSR instruction; a T-SBI call when the test runs in U-mode."""
+    return instr if priv != "U" else tsbi_call(instr)
```
jordancarlin, line 23: Fine for now, but we should standardize versions of these functions in a common file because there are almost identical versions of this scattered across lots of tests at this point.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -8,24 +8,77 @@
...
+        )
+    else:
+        lines.extend(
+            [
+                "# mstatus.MPIE and MIE = 0",
+                f"LI(x{r_temp}, 0x88)",
+                m_csr(priv, f"csrc mstatus, x{r_temp}"),
+                *_sstatus_sie(priv, r_temp, 0),
+            ]
+        )
```
jordancarlin: What is the point of using a separate `__status_sie` function if we are writing to `mstatus` right before it anyway?

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -39,23 +92,40 @@ def wrs_resume_helper(
...
+            else []
+        ),
+        "cross lr instruction to set up reservation.",
+        "mstatus.TW = 0" if lower else "mstatus.TW = {0/1}",
+        "cross with mie.MTIE = 1" + (" (if Sstc supported use STIP, cross menvcfg.STCE = 1)" if lower else ""),
+        "mstatus.MIE = {0/1}",
+        *([f"mstatus.SIE = {{0/1}}{' (if S supported)' if priv == 'U' else ''}"] if lower else []),
+        "Set up timer to interrupt soon",
+        f"execute {{WRS.NTO/WRS.STO}} in {priv} mode",
+        "2 x 2 x 2 bins",
```
jordancarlin: This is a very confusing comment and I don't think is all that useful in the generated tests in its current form.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -193,29 +262,56 @@ def wrs_no_mie_helper(
...
+                m_csr(priv, "csrw mie, zero"),
             ]
         )
+        if lower:
+            lines.extend(
+                [
+                    "# mstatus.MIE and MPIE = 1",
+                    f"LI(x{r_temp}, 0x88)",
+                    m_csr(priv, f"csrs mstatus, x{r_temp}"),
+                    *_sstatus_sie(priv, r_temp, 1),
```
jordancarlin: Seems like no need to make these separate writes.

jordancarlin (changes requested):
A few small things, but looks close.

generators/testgen/src/testgen/asm/interrupts.py
davidharrishmc: Moving to the RVTEST macros that you recently merged and taking all this code out of interrupts.py.

generators/testgen/src/testgen/asm/interrupts.py
davidharrishmc: moving to RVTEST macros.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
davidharrishmc: Agreed.  There are a bunch of versions across a bunch of PRs, some of which are still open.  Let's revisit after merging.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
davidharrishmc: Simplified

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
davidharrishmc: deleted

tests/env/sail_macros.h
```diff
@@ -122,7 +126,7 @@
 #define RVMODEL_INTERRUPT_LATENCY 1
 
 #undef RVMODEL_TIMER_INT_SOON_DELAY
-#define RVMODEL_TIMER_INT_SOON_DELAY 100
+#define RVMODEL_TIMER_INT_SOON_DELAY 5000 // Sail ticks once per instruction; T-SBI round trips in the *_INT_SOON macros cost up to ~1000 ticks
```
jordancarlin, line 129: Sound like we need to update the documentation. It still says that `RVMODEL_TIMER_INT_SOON_DELAY` "should be long enough that the core can run a few instructions before the interrupt fires".

coverpoints/priv/InterruptsS_coverage.svh
jordancarlin, line 1: Something seems to have gone wrong? This PR is now removing all of the interrupt coverpoints and tests.

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
```diff
@@ -32,6 +23,34 @@ def s_csr(priv: str, instr: str) -> str:
...
+def _disable_stimecmp(priv: str, r: int) -> list[str]:
+    """stimecmp = -1 to disarm the Sstc timer; an S-mode CSR reached via T-SBI from U-mode."""
+    return [
+        "# Disable Sstc timer: stimecmp = -1",
+        f"LI(x{r}, -1)",
+        "#if __riscv_xlen == 32",
+        s_csr(priv, f"csrw stimecmph, x{r}"),
+        "#endif",
+        s_csr(priv, f"csrw stimecmp, x{r}"),
+    ]
```
jordancarlin: Seems like we can just use the `RVTEST_CLR_SSTC_INT_*` macros?

jordancarlin (changes requested):

coverpoints/priv/InterruptsS_coverage.svh
davidharrishmc: Intentional

generators/testgen/src/testgen/priv/extensions/ZawrsCommon.py
davidharrishmc: Done

tests/env/sail_macros.h
davidharrishmc: Updated abstraction.adoc

jordancarlin (approved):

PR #51, opened 2019-03-05
Initial definition of targets for the Sail C and OCaml emulators.

pmundkur:
The Sail formal model generates two emulators, C and OCaml.  This adds a target for each of them.  Both are passing all current tests.

PR #2198, opened 2026-08-27
Test generator and coverpoints of Zicfiss extension

UmerShahidengr:
Adds test generators and coverpoints for Zicfiss (shadow stack), split by the privilege mode the tests run in:

- **ZicfissU** (U-mode): shadow stack instructions, the `ssp` CSR, target page and PMA behaviour, and the U-mode half of the SSE enable chain.
- **ZicfissS** (S/HS-mode): SS page encoding, U/SUM/MXR priority, and SSE gating seen from S-mode.
- **ZicfissSm** (M-mode control plane): menvcfg.SSE gating, the envcfg read-only-zero chain, SSAMOSWAP in M-mode, PMP permissions, and SS instructions with `satp.MODE=Bare` below M.
- **ZicfissV**: vector loads and stores to an SS page.

`ZicfissH_coverage.svh` holds the VS/VU coverpoints but has no generator yet.

Functional coverage is 100% for all four covergroups on `sail-rv64-max` and `sail-rv32-max`. Normative-rule mappings are in `coverpoints/norm/Zicfiss.yaml`.

## Framework change: Sail trace converter

The SS page coverpoints need the PTE of each translated access, so Sail now runs with `--trace-ptw`, and `sail_to_rvvi.py` adds `TRAP`, `PTE_I` and `PTE_D` to the RVVI trace. Existing fields are unchanged; only the new tokens are added.

This changes the reference trace for **every** suite's coverage run. The PTE coverpoints in `Sv`, `Svadu`, `Svade` and `SvZicbo` never filled against the Sail trace before. Measured on `sail-rv64-max`, base `2dea4b02c` vs this branch:

| covergroup | base | this PR |
|---|---|---|
| `Svade_cg` | 6.83% | 32.91% |
| `Svadu_cg` | 7.85% | 37.85% |

No bin regressed and traces grow by about 1%. Merging this moves the published coverage numbers for the VM suites, so the coverage and CTP baselines should be regenerated.

## Simulator exclusions

All four suites pass on Sail. The other simulators exclude suites for model bugs, documented in each `config/<sim>/ci.yaml`:

- **Spike**: ZicfissU. SSPOPCHK from an SS page with `pte.D=0` is treated as a store under Svade.
- **Whisper**: ZicfissU/S/Sm. SSAMOSWAP in M-mode raises an illegal-instruction exception, and U=1/SUM=0 gives an access fault instead of a page fault.
- **QEMU**: all four suites.
- **Imperas** (not in CI): ZicfissU/S/Sm, three model bugs.


config/qemu/ci.yaml
```diff
@@ -26,7 +26,16 @@ ci_enabled: true
 # Smstateen will be added back once sail starts supporting SM1P13 too
 # ZawrsU and ZawrsS fail on qemu due to stval mismatching, filed QEMU issue https://gitlab.com/qemu-project/qemu/-/work_items/4080
 # ZicntrS and ZicntrU fail on qemu because ecall increments instret. Filed QEMU issue: https://gitlab.com/qemu-project/qemu/-/work_items/4087
-exclude_extensions: "Sm,S,SsstrictSm,SsstrictS,SsstrictU,Zacas,ZacasZabha,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,SvZicbo,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,Zkr,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU,ZicntrS,ZicntrU"
+# ZicfissU, ZicfissS, ZicfissSm excluded due to QEMU bugs (to be reported upstream):
```
davidharrishmc: Please insert links to issue numbers (and put them in the bug tracking sheet) before merging this.  Otherwise it becomes hard to track.

config/whisper/ci.yaml
```diff
@@ -8,9 +8,15 @@
 # ExceptionsSvZalrsc: Whisper reports page faults where Sail expects access faults.
 # Vx excluded because of https://github.com/tenstorrent/whisper/issues/35
 # Sdtrig* not yet implemented in Sail
+# ZicfissU, ZicfissS, ZicfissSm excluded due to a Whisper bug (to be reported upstream):
```
davidharrishmc: You may be able to work around this with a mask on menvcfg.  However, Joe has been fixing these things with one day turnaround when reported.

coverpoints/param/Zicfiss.yaml
```diff
@@ -0,0 +1,32 @@
+# Parameter mappings
```
davidharrishmc, line 1: Good to have parameters.

Could you add the normative rules to this PR, and the CTP section with a link to the test plan?

davidharrishmc (changes requested):
Will do a substantive review agains the test plan soon.

generators/testgen/src/testgen/priv/extensions/ZicfissSm.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Top-level generator
+# ---------------------------------------------------------------------------
+
+
+@add_priv_test_generator(
+    "ZicfissSm",
+    required_extensions=["S", "U", "Zicfiss", "Zimop", "Zaamo", "Zicsr"],
```
davidharrishmc: Now that Jordan's boot PR is in, these boot to U mode unless otherwise specified.  Add decorator to boot to the necessary mode, and use T-SBI.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,309 @@
...
+# sail-riscv 0.13.1 aborts with "Assertion failed: sys/vmem_pte.sail:148.24-148.25"
+# when its page-table walker resolves such a PTE, so those testcases cannot currently
+# have a reference signature generated. Everything else in the Zicfiss suites — the
+# ssp CSR, the enable-chain gating, the Zimop-revert behaviour, and SS instructions
+# aimed at non-SS pages — is unaffected and runs today.
+#
+# The guard is deliberately NOT defined anywhere in-tree. Define it (or delete the
+# guard entirely) once the reference model handles the SS page encoding. See
+# sail-zicfiss-bug/BUG_REPORT.md for a standalone reproducer.
+SS_PAGE_GUARD = "ZICFISS_SS_PAGE_REF_MODEL_OK"
```
davidharrishmc, line 79: Let's wait until new Sail release fixes this and the guard can be removed before merging.

config/qemu/ci.yaml
```diff
@@ -26,7 +26,12 @@ ci_enabled: true
...
+#  1. Zicfiss: sspopchk raises a software-check exception where a store/AMO access fault must take priority
+#  (https://gitlab.com/qemu-project/qemu/-/work_items/4379)
+#  2. mtval is reported as 0 on the M-mode SSAMOSWAP access fault, disagreeing with
```
davidharrishmc, line 30: Is this a bug, or just a parameter choice?  Does this test stval, which Sstvala says should hold the address?

coverpoints/priv/ZicfissU_coverage.svh
```diff
@@ -0,0 +1,272 @@
...
+//
+// Zicfiss (shadow stack) — U-mode coverage
+//
+// Copyright (C) 2026 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+////////////////////////////////////////////////////////////////////////////////////////////////
+
+`define COVER_ZICFISSU
```
davidharrishmc, line 13: Claude claims tests are not reaching full coverage.

coverpoints/priv/ZicfissU_coverage.svh
```diff
@@ -0,0 +1,272 @@
...
+    cp_ssp_low_bits_ro_zero:       cross priv_mode_u, csrops, ssp_csr, ssp_wr_low_bits, ssp_rd_low_bits;
+    cp_sspush:                     cross priv_mode_u, ss_push_instr, ssp_write_pattern;
+    cp_sspopchk_match:             cross priv_mode_u, ss_pop_instr, sspopchk_outcome {
+        ignore_bins mismatch = binsof(sspopchk_outcome.mismatched);
+    }
+    cp_sspopchk_mismatch:          cross priv_mode_u, ss_pop_instr, sspopchk_outcome {
+        ignore_bins match = binsof(sspopchk_outcome.matched);
+    }
+    cp_sspopchk_fault_priority:    cross priv_mode_u, ss_pop_instr, ssp_fault_address;
+    cp_ss_call_return:             cross priv_mode_u, ss_push_instr, ss_pop_instr;
```
davidharrishmc: This doesn't make sense - the current instruction can't be a push and a pop simultaneously.  Should push be based on inst.prev?

coverpoints/param/Zicfiss.yaml
```diff
@@ -0,0 +1,37 @@
...
+        cannot distinguish the cause from the tval alone.",
+      ]
+  - name: REPORT_CAUSE_IN_STVAL_ON_SHADOW_STACK_SOFTWARE_CHECK
+    coverpoint: ["cp_sspopchk_mismatch, cp_sspopchk_mismatch_s"]
+    effect:
+      [
+        "Same as the mtval parameter, for a software-check exception delegated to S-mode.",
+      ]
+  - name: REPORT_CAUSE_IN_VSTVAL_ON_SHADOW_STACK_SOFTWARE_CHECK
+    coverpoint: ["cp_ssp_csr_gating_vs, cp_ssp_csr_gating_vu"]
```
davidharrishmc: Does this coverpoint cause a software check exception?

coverpoints/param/Zicfiss.yaml
```diff
@@ -0,0 +1,37 @@
...
+    effect:
+      [
+        "Same as the mtval parameter, for a software-check exception delegated to VS-mode.",
+      ]
+  - name: MISALIGNED_LDST
```
davidharrishmc: Should this be MISALIGNED_AMO?

coverpoints/priv/ZicfissU_coverage.svh
```diff
@@ -0,0 +1,272 @@
...
+    ssp_wr_low_bits: coverpoint ins.current.rs1_val[1:0] {
+        // auto fills 00 through 11
+    }
+    // ssp[1:0] as actually read back — must always be zero.
+    ssp_rd_low_bits: coverpoint ins.current.csr[CSR_SSP][1:0] {
+        bins read_only_zero = {2'b00};
+    }
+
+    // ── Alignment building blocks ─────────────────────────────────────────
+    // DH review: sweep the bottom 3 bits over all 8 values rather than
```
davidharrishmc: What is DH review?

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -0,0 +1,719 @@
...
+    return [comment_banner(coverpoint, "SSAMOSWAP.W/.D swap, sign-extension, and narrow store"), *both_xlens(build)]
+
+
+# ---------------------------------------------------------------------------
+# cp_ss_address_alignment
+# ---------------------------------------------------------------------------
+
+
+def _generate_alignment(test_data: TestData) -> list[str]:
+    """DH review: sweep addr[2:0] over all 8 values rather than aligned/misaligned cases."""
```
davidharrishmc: ? DH

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -0,0 +1,719 @@
...
+
+def _generate_ssp_access(test_data: TestData) -> list[str]:
+    coverpoint = "cp_ssp_access"
+
+    def build(xlen: int) -> list[str]:
+        ss_va, _, _ = va_for(xlen)
+        val_reg, rd_reg = test_data.int_regs.get_registers(2)
+        lines = _umode_prologue(test_data, xlen)
+
+        # DH idiom: read, write all-ones, write all-zeros, set all, clear all, restore.
```
davidharrishmc: ? DH

generators/testgen/src/testgen/priv/extensions/ZicfissS.py
```diff
@@ -0,0 +1,414 @@
...
+        lines.extend(save_lines)
+
+        for mnemonic, compressed, name in _PUSH_FORMS:
+            reg = "x1" if "x1" in mnemonic else "x5"
+            lines.extend(
+                [
+                    f"LI(x{addr_reg}, {hex(ssp_top)})",
+                    f"csrw ssp, x{addr_reg}",
+                    f"LI({reg}, 0xA5A5A5A5)",
+                    test_data.add_testcase(f"{name}_s_rv{xlen}", "cp_sspush_s", _CG),
```
davidharrishmc, line 202: Does this read back the value pushed to confirm it worked?  Global check.

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -0,0 +1,719 @@
...
+        # RV64 SSAMOSWAP.W: sign-extension (MSB 0 and 1) and the rs2[63:32]-ignored case.
+        cases = [("msb0", "0x7FFFFFFF", "0x12345678"), ("msb1", "0x80000000", "0xFFFFFFFF")]
+        if xlen == 64:
+            cases.append(("rs2_upper_ignored", "0x12345678", "0xAAAAAAAAFFFFFFFF"))
+
+        for name, mem_val, rs2_val in cases:
+            lines.extend(
+                [
+                    f"LI(x{addr_reg}, {hex(ss_va)})",
+                    f"LI(x{seed_reg}, {mem_val})",
```
davidharrishmc, line 415: Looks like seed reg is loaded but never used, so amoswap doesn't have a defined value in memory?

generators/testgen/src/testgen/priv/extensions/ZicfissS.py
```diff
@@ -0,0 +1,414 @@
...
+        comment_banner(coverpoint, "menvcfg.SSE gates ssp at S/HS; senvcfg.SSE must not"),
+    ]
+
+    for menvcfg in (0, 1):
+        for senvcfg in (0, 1):
+            tag = f"m{menvcfg}s{senvcfg}"
+
+            def build(xlen: int, menvcfg: int = menvcfg, senvcfg: int = senvcfg, tag: str = tag) -> list[str]:
+                rd_reg, val_reg = test_data.int_regs.get_registers(2)
+                block = _smode_prologue(test_data, xlen, menvcfg=menvcfg, senvcfg=senvcfg)
```
davidharrishmc, line 96: Read back, confirm senvcfg is always 0 if menvcfg = 0.  The m0s1 coverpoint can never be hit.

davidharrishmc (changes requested):
Low hanging fruit.  Let's postpone close review until coverage is 100% and TSBI is implemented.

config/qemu/ci.yaml
UmerShahidengr: It is a bug, it is not behaving the way regular store/amo behaves, and specs say that fault will have same behavior as store/amo fault.

generators/testgen/src/testgen/priv/extensions/ZicfissS.py
UmerShahidengr: Yes, it was pushed

generators/testgen/src/testgen/priv/extensions/ZicfissV.py
```diff
@@ -0,0 +1,166 @@
...
+
+@add_priv_test_generator(
+    "ZicfissV",
+    required_extensions=["S", "U", "V", "Zicfiss", "Zimop", "Zaamo", "Zcmop", "Zca", "Zicsr"],
+    march_extensions=["V", "Zicfiss", "Zimop", "Zaamo", "Zcmop", "Zca"],
+    extra_defines=[
+        "#define RVTEST_VECTOR",
+        "#define RVTEST_SEW 0",
+        "#define VDSEW 0",
+        # TODO: Remove BOOT_TO_MMODE when converting this test to T-SBI.
```
davidharrishmc: This can be removed now.  It will boot to U mode if unspecified.

config/whisper/ci.yaml
```diff
@@ -8,9 +8,11 @@
 # ExceptionsSvZalrsc: Whisper reports page faults where Sail expects access faults.
 # Vx excluded because of https://github.com/tenstorrent/whisper/issues/35
 # Sdtrig* not yet implemented in Sail
+# ZicfissU, ZicfissS, ZicfissSm, ZicfissV excluded due to a Whisper bug (https://github.com/tenstorrent/whisper/issues/51):
```
davidharrishmc, line 10: Whisper #51 is reported as fixed.  Try the latest and see if this exclusion can be removed.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,311 @@
...
+def guard_ss_page(lines: list[str], *, reason: str) -> list[str]:
+    """Wrap a block that performs a translated access through an SS page (pte.xwr=010)."""
+    return [
+        f"#ifdef {SS_PAGE_GUARD}  // blocked on sail-riscv vmem_pte.sail:148 — {reason}",
+        *lines,
+        f"#endif  // {SS_PAGE_GUARD}",
+    ]
+
+
+GOTO_UMODE = "RVTEST_GOTO_LOWER_MODE Umode  # enter U-mode"
```
davidharrishmc: Update to use T-SBI

config/whisper/ci.yaml
```diff
@@ -8,9 +8,11 @@
...
+# menvcfg.SSE / senvcfg.SSE are not writable, so Zicfiss can never be activated
 ci_enabled: true
 # TODO: Remove SsstrictV when sail SsstrictV is implemented
-exclude_extensions: "SsstrictSm,SsstrictS,SsstrictU,ExceptionsSvZalrsc,SsstrictV,Smstateen,Vx8,Vx16,Vx32,Vx64,SdtrigSm,SdtrigS,SdtrigU"
+exclude_extensions: "SsstrictSm,SsstrictS,SsstrictU,ExceptionsSvZalrsc,SsstrictV,Smstateen,Vx8,Vx16,Vx32,Vx64,SdtrigSm,SdtrigS,SdtrigU,ZicfissU,ZicfissS,ZicfissSm,ZicfissV"
```
davidharrishmc: Imperas also needs to exclude ZicfissU/S.

docs/ctp/src/privmisc23.adoc
```diff
@@ -122,11 +122,30 @@ In Privileged section because it has both unpriv and privileged behaviors.
...
+suites: ZicfissU (U-mode), ZicfissS (S/HS-mode), ZicfissSm (M-mode control plane) and
+ZicfissV (vector accesses to a shadow stack page). The VS/VU-mode suite is documented in
+the Hypervisor chapter, see <<t-ZicfissH-coverpoints>>.
+
 [[t-Zicfiss-coverpoints]]
 .Zicfiss Coverpoints
-//,===
-//include::{testplansdir}/Zicfiss.adoc[]
-//,===
+https://docs.google.com/spreadsheets/d/1ED9ontJZP-Dg2MAF-4mkhqKFgpjf6Fp6yeXKuW61FOY/edit?usp=sharing[Google Sheet Testplan]
```
davidharrishmc, line 132: These links look identical.  They should point to the right tabs.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -269,9 +302,19 @@ def satp_setup(xlen: int) -> list[str]:
...
+    ``mode`` is the boot mode of the suite, and every section is expected to both start
+    and end there. ``sfence.vma`` is not available below S-mode and cannot be marshalled
+    through T-SBI, so a U-mode suite makes an M-mode excursion to tear translation down
+    and returns to U-mode afterwards.
+    """
+    if mode == "S":
+        return ["csrwi satp, 0", "sfence.vma", ""]
+    if mode == "M":
+        return [GOTO_MMODE, "csrwi satp, 0", "sfence.vma", ""]
+    return [GOTO_MMODE, "csrwi satp, 0", "sfence.vma", GOTO_UMODE, ""]
```
davidharrishmc: U-mode programs can't go to MMODE because there may not be a standard MMODE.  How about making this GOTO_SMODE?

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -11,23 +11,28 @@
...
 """
 
 from __future__ import annotations
 
 from testgen.asm.helpers import comment_banner, write_sigupd
+from testgen.asm.tsbi import tsbi_call
 from testgen.data.state import TestData
 from testgen.data.test_chunk import TestChunk
 from testgen.priv.extensions.ZicfissCommon import (
+    GOTO_MMODE,
```
davidharrishmc: Nothing in RVA23 can use GOTO_MMODE.  Nothing intended to run in S or U should use GOTO_MMODE.

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -57,20 +62,26 @@
...
     therefore fetch-page-fault inside the S-mode handler and loop. M-mode does not
     translate, so its handler runs regardless of PTE_U.
+
+    That is also why this setup keeps an M-mode excursion after the T-SBI conversion
+    rather than running from U-mode through ``tsbi_call``: medeleg is deliberately
+    absent from ``tsbi_instr_table``, and the page-table stores and ``sfence.vma`` are
+    not reachable from U-mode either once the mappings carry PTE_U.
     """
     return [
+        GOTO_MMODE,
```
davidharrishmc: Violates our no M-mode rules

generators/testgen/src/testgen/priv/extensions/ZicfissV.py
```diff
@@ -33,10 +38,11 @@
 
 def _umode_prologue(test_data: TestData, xlen: int) -> list[str]:
     return [
+        GOTO_MMODE,
```
davidharrishmc: Not allowed

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -605,10 +616,11 @@ def build(xlen: int, perms: str = perms, u_bit: int = u_bit) -> list[str]:
             # user=False keeps the identity map supervisor-only; the SS page's own U bit
             # is what is being swept here.
             block = [
+                GOTO_MMODE,
```
davidharrishmc: Not allowed

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -605,10 +616,11 @@ def build(xlen: int, perms: str = perms, u_bit: int = u_bit) -> list[str]:
...
                 *satp_setup(xlen),
                 *map_zicfiss_pages(xlen, ss_perms=perms, user=True),
-                *set_envcfg_sse("menvcfg", 1, test_data),
-                *set_envcfg_sse("senvcfg", 1, test_data),
+                *set_envcfg_sse("menvcfg", 1, test_data, mode="M"),
+                *set_envcfg_sse("senvcfg", 1, test_data, mode="M"),
                 "csrw medeleg, x0",
```
davidharrishmc: We need to refactor so only ZicfissSm tests medeleg.  None of the lower privilege tests can touch it, and they stick with the default from boot time that exceptions are delegated.

generators/testgen/src/testgen/priv/extensions/ZicfissV.py
```diff
@@ -33,10 +38,11 @@
...
         *satp_setup(xlen),
         *map_zicfiss_pages(xlen),
-        *set_envcfg_sse("menvcfg", 1, test_data),
-        *set_envcfg_sse("senvcfg", 1, test_data),
+        *set_envcfg_sse("menvcfg", 1, test_data, mode="M"),
+        *set_envcfg_sse("senvcfg", 1, test_data, mode="M"),
         "csrw medeleg, x0",
```
davidharrishmc: Not allowed

davidharrishmc (changes requested):

davidharrishmc:
Let's not mix Sail 0.14 issues into this PR.  Let's wait until one of the Sail 0.14 PRs is merged.

In particular, let's not create exclusions in the Zicfiss PR related to Sm or S tests.

UmerShahidengr:
Yes, it was pushed by mistake, I am cleaning this up

davidharrishmc (commented):
Substantive review against the test plan, now that coverage is at 100% and T-SBI is in. I built and ran everything rather than reading only.

**Verified**

- `make coverage` with `EXTENSIONS=ZicfissU,ZicfissS,ZicfissSm,ZicfissV` on `sail-rv64-max` and `sail-rv32-max`: **4 covergroups at 100% on both**, 114 build tasks succeeded, 16 ELFs per XLEN.
- All 32 normative-rule names in `coverpoints/norm/Zicfiss.yaml` resolve against the canonical `norm-rules.json` snapshot. (`Zicfilp_ssp_below_m_mode` looks like a wrong-extension typo but is the genuine upstream name — all its tags are `norm:zicfiss_*`.)
- My earlier comments are all addressed: `GOTO_MMODE` survives only in the `BOOT_TO_MMODE` suite and `teardown_vm("M")`, no suite writes `medeleg`, `ZicfissU` boots to U by default, the `SS_PAGE_GUARD` Sail workaround is gone, `m0s1` is an `illegal_bins`, and the CTP links point at distinct tabs.
- **Imperas**: I ran the suites on ImperasDV under licence — 5/6 fail on `imperas-rv64-max`, 5/5 on `imperas-rv32gck`, from three distinct model bugs. Details in the `config/imperas/ci.yaml` comment.

**What needs fixing**

The inline comments below, of which three matter: the `pte.U` sweep in `ZicfissU.py` is dead (both legs map the page user-accessible, confirmed in the generated asm), the `pte_u` coverpoint that should have caught it is never crossed, and `cp_envcfg_sse_rdonly0_henvcfg` is neither compiled nor stimulated while a normative rule is mapped to it. Each of those makes the 100% number cover less than it appears to. The rest are latent or cosmetic.

Nice work — the generators are unusually well documented and the T-SBI integration reads cleanly.

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -0,0 +1,1087 @@
...
+
+        def build(xlen: int, perms: str = perms, u_bit: int = u_bit) -> list[str]:
+            ss_va, _, _ = va_for(xlen)
+            ssp_top = ss_va + 0x800
+            addr_reg, mask_reg = test_data.int_regs.get_registers(2)
+            # user=False keeps the identity map supervisor-only; the SS page's own U bit
+            # is what is being swept here.
+            block = [
+                GOTO_SMODE,
+                *map_zicfiss_pages(xlen, ss_perms=perms, user=True),
```
davidharrishmc: The `pte.U` sweep is dead — both legs map the SS page user-accessible.

`perms` varies `PTE_U` on line 612, but this call passes `user=True`, and `map_zicfiss_pages` appends `| PTE_U` to `ss_perms` unconditionally when `user=True` (`ZicfissCommon.py:433`). The generated assembly confirms it:

```
# --- pte.U = 0 ---
PTE_SETUP_SV39(rvtest_zicfiss_ss_page, (PTE_D | PTE_A | PTE_W | PTE_V | PTE_U), ...)
# --- pte.U = 1 ---
PTE_SETUP_SV39(rvtest_zicfiss_ss_page, (PTE_D | PTE_A | PTE_W | PTE_V | PTE_U | PTE_U), ...)
```

`PTE_U | PTE_U` appears twice in `ZicfissU-00.S` (RV64 and RV32). The two legs are identical stimulus and the supervisor-page case is never tested. The comment on line 618 says `user=False`, which is what the code should be doing — `ZicfissS._generate_perm_priority_s` passes `user=False` and does produce a correct U=0/U=1 split.

coverpoints/priv/ZicfissU_coverage.svh
```diff
@@ -0,0 +1,443 @@
...
+        bins read_only       = {3'b001};
+        bins ss_page         = {3'b010};
+        bins read_write      = {3'b011};
+        bins exec_only       = {3'b100};
+        bins exec_read       = {3'b101};
+        bins rsvd_wx         = {3'b110};
+        bins read_write_exec = {3'b111};
+    }
+    // pte.U is bit 4; pte.A is bit 6 and pte.D is bit 7.
+    pte_u: coverpoint ins.current.pte_d[4] {
```
davidharrishmc, line 239: `pte_u` is declared here but never appears in any cross — `cp_ss_instr_target_page` is `priv_mode_u x ss_mem_instr x pte_xwr` only.

It still reports 100%, but only because it fills incidentally from S-mode handler traffic: in my `sail-rv64-max` run `bin supervisor` was hit 79,393 times and `bin user` 2,414. So it measures nothing about the pte.U sweep, which is why the `ZicfissU.py:622` bug above doesn't show up as a coverage hole. `ZicfissS_coverage.svh` crosses `pte_u` in `cp_ss_page_perm_priority`, which is presumably what was intended here.

coverpoints/priv/ZicfissSm_coverage.svh
```diff
@@ -0,0 +1,206 @@
...
+        ignore_bins csrrw_reads_back_written =
+            binsof(csr_write_ops.csrrw) && binsof(menvcfg_sse.sse_on) &&
+            ((binsof(sse_bit_written.wrote_zero) && binsof(senvcfg_sse_readback.reads_one)) ||
+             (binsof(sse_bit_written.wrote_one) && binsof(senvcfg_sse_readback.reads_zero)));
+        ignore_bins csrrs_set_reads_one =
+            binsof(csr_write_ops.csrrs) && binsof(menvcfg_sse.sse_on) &&
+            binsof(sse_bit_written.wrote_one) && binsof(senvcfg_sse_readback.reads_zero);
+    }
+    `ifdef H_SUPPORTED
+        cp_envcfg_sse_rdonly0_henvcfg: cross priv_mode_m, csr_write_ops, henvcfg_csr, menvcfg_sse,
```
davidharrishmc, line 216: `cp_envcfg_sse_rdonly0_henvcfg` is neither compiled nor stimulated, yet `norm/Zicfiss.yaml:30` maps `menvcfg_sse_rdonly0` to it.

Two independent problems:

1. It is the **only** cross in this file missing from the coverage report. On `sail-rv64-max` the report lists `cp_envcfg_sse_rdonly0_senvcfg`, `cp_menvcfg_sse_gating`, `cp_ss_instr_inactive_m`, `cp_ss_instr_inactive_s`, `cp_ss_pmp_permissions`, `cp_ssamoswap_mmode_fault` — and not this one. The `` `ifdef H_SUPPORTED `` isn't taking, even though `work/<cfg>/rvtest_config.svh` defines `H_SUPPORTED` and `framework/src/act/fcov/testbench.sv:15` includes it. So the "100% coverage" result excludes this coverpoint entirely.
2. No generator ever writes `henvcfg`: `grep -c henvcfg` over all six generated `.S` files returns 0.

It also lacks the `ignore_bins` its senvcfg twin (line 183) has, so several of its bins are architecturally unreachable as well.

coverpoints/norm/Zicfiss.yaml
```diff
@@ -3,75 +3,128 @@ normative_rule_definitions:
...
+  - name: menvcfg_sse_op
+    coverpoint:
+      [
+        "{ZicfissSm_cg}/{cp_menvcfg_sse_gating,cp_ss_instr_inactive_m,cp_ss_instr_inactive_s}",
+        "{ZicfissS_cg}/{cp_ssp_csr_gating_s,cp_ss_page_enc_load,cp_ss_page_enc_store}",
+        "{ZicfissU_cg}/{cp_ss_instr_inactive_u,cp_ssamoswap_sse_gating}",
+      ]
+
+  # When menvcfg.SSE is 0, the henvcfg.SSE and senvcfg.SSE fields are read-only zero.
+  - name: menvcfg_sse_rdonly0
```
davidharrishmc, line 30: `mtval_swchk` isn't mapped to a coverpoint anywhere in `coverpoints/norm/`.

It's the rule that `mtval` holds 3 (Shadow Stack Fault) on a software-check exception — exactly what `cp_sspopchk_mismatch` / `cp_sspopchk_mismatch_s` exercise, and what the new `REPORT_CAUSE_IN_MTVAL_ON_SHADOW_STACK_SOFTWARE_CHECK` parameter is about. It lives in `normative_rule_defs/machine.yaml` rather than `priv-cfi.yaml`, which is probably why it was missed. Everything else lines up: I checked all 32 rule names in this file against the canonical `norm-rules.json` snapshot and all 32 resolve, and of the 45 CFI-chapter rules the 18 unmapped ones are all Zicfilp.

coverpoints/norm/Zicfiss.yaml
```diff
@@ -3,75 +3,128 @@ normative_rule_definitions:
...
   # VU levels.
   - name: ssmp_henvcfg_sse
-    coverpoint: [""]
+    coverpoint: ["{ZicfissH_cg}/{cp_ss_page_enc_virt}"]
 
   # If satp.MODE (or vsatp.MODE when V=1) is set to Bare and the effective privilege
   # mode is less than M, shadow stack instructions raise a store/AMO access-fault
   # exception.
   - name: satp_mode_bare
+    # TODO no test yet: SS instruction with satp.MODE=Bare below M-mode is not generated
```
davidharrishmc: This one looks reachable today, unlike the `Zicfiss_smode_xsse` TODO above (which genuinely needs a build with S-mode absent).

`ZicfissSm` already boots to M-mode and drops to S-mode through T-SBI, so an SS instruction executed below M with `satp.MODE=Bare` is a short addition to that suite. Worth filling before merge rather than shipping as a documented hole, since it's one of the few Zicfiss rules with no coverage at all.

coverpoints/priv/ZicfissH_coverage.svh
```diff
@@ -0,0 +1,280 @@
...
+        bins sse_on  = {1'b1};
+    }
+    // Exception flavour differs by which field gates: menvcfg.SSE=0 raises
+    // ILLEGAL-instruction even when V=1, while henvcfg.SSE/senvcfg.SSE gating with V=1
+    // raises VIRTUAL-instruction. The inactive_men_off bin below is therefore the
+    // illegal-instruction case, not a virtual-instruction one.
+    // VS-mode needs menvcfg.SSE and henvcfg.SSE.
+    vs_sse_state: coverpoint {(get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "menvcfg", "sse") == 1),
+                              ins.prev.csr[CSR_HENVCFG][3]} {
+        bins inactive_men_off = {2'b01};
```
davidharrishmc: `inactive_men_off = {2'b01}` is unreachable.

The concatenation is `{menvcfg.SSE == 1, henvcfg.SSE}`, so `2'b01` means menvcfg.SSE=0 **and** henvcfg.SSE=1 — impossible, because menvcfg.SSE=0 forces henvcfg.SSE read-only zero (the rule this same file maps at `norm/Zicfiss.yaml:30`). The reachable "menvcfg off" state is `2'b00`, which `inactive_both` already claims, so `cp_ssp_csr_gating_vs` carries a permanent hole for exactly the illegal-instruction case the comment above says this bin exists to cover.

Latent today since ZicfissH is never compiled (no generator), but it will bite whoever lands one.

coverpoints/priv/ZicfissH_coverage.svh
```diff
@@ -0,0 +1,280 @@
...
+    csr_write_ops: coverpoint ins.current.insn {
+        wildcard bins csrrw = {CSRRW};
+        wildcard bins csrrs = {CSRRS};
+    }
+    // Full three-bit enable state. Which bits actually apply depends on the mode.
+    sse_chain: coverpoint {(get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "menvcfg", "sse") == 1),
+                           (get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "senvcfg", "sse") == 1),
+                           ins.prev.csr[CSR_HENVCFG][3]} {
+        bins m0s0h0 = {3'b000};
+        bins m0s0h1 = {3'b001};
```
davidharrishmc: `bins m0s0h1 = {3'b001}` contradicts the comment four lines below it.

`3'b001` is menvcfg.SSE=0 with henvcfg.SSE=1 — a set child under menvcfg.SSE=0, which the comment correctly says is unreachable. But `illegal_bins m0_child_set` lists only `{3'b010, 3'b011}`, leaving `3'b001` as a normal bin that can never fill. Either add `3'b001` to the illegal bins and drop `m0s0h1`, or the comment is wrong.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,540 @@
...
+        "# Round rvtest_data_begin UP: the S-mode save area sits just below it and the",
+        "# T-SBI S-mode handler executes its CSR instruction out of that save area, so its",
+        "# page must stay supervisor-only or S-mode cannot fetch the scratch code.",
+        "LA(t4, rvtest_data_begin)",
+        "add t4, t4, t6",
+        "addi t4, t4, -1",
+        "srli t4, t4, 12",
+        "slli t4, t4, 12          # first page of the user-writable data region",
+        "1:",
+        "bgeu t0, t1, 2f",
```
davidharrishmc, line 340: This PTE-fill loop has no bound on the leaf table.

`t0` runs from the image start to `_end` advancing `t2` one entry per page, with nothing checking `t2` against the end of `rvtest_uimg_lvl0_pg_tbl` (4 KiB = 512 entries = 2 MiB on Sv39, 1024 = 4 MiB on Sv32). `page_table_data_section()` declares `rvtest_uimg_mapped` immediately after that table, then `rvtest_zicfiss_ss_page` / `_rw_page` / `_ro_page` — so an overflow silently corrupts the pages under test rather than leaving them unmapped, which is what the docstring at line 305 promises.

Current headroom from the generated ELFs: 110 of 512 pages used on Sv39, 85 of 1024 on Sv32. So it's latent, not broken — but the suite is already four chunks and growing. A bounds check (or at least an assembler-time `.if` on the span) would make it safe.

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -0,0 +1,1087 @@
...
+# ---------------------------------------------------------------------------
+# cp_ssp_csr_gating_u / cp_ssamoswap_sse_gating
+# ---------------------------------------------------------------------------
+
+
+def _generate_sse_gating(test_data: TestData) -> list[str]:
+    """Sweep the (menvcfg.SSE, senvcfg.SSE) enable chain from U-mode."""
+    lines: list[str] = [comment_banner("cp_ssp_csr_gating_u", "SSE enable chain seen from U-mode")]
+
+    for menvcfg, senvcfg in [(1, 1), (1, 0), (0, 1), (0, 0)]:
```
davidharrishmc: The `(menvcfg=0, senvcfg=1)` leg is architecturally identical to `(0, 0)`.

`_umode_prologue` writes menvcfg first, so by the time it does `csrs senvcfg` the SSE field is already read-only zero and the write is a no-op. Both `m0s0_*` and `m0s1_*` testcase tags appear in `ZicfissU-01.S` describing the same state, so the labels misdescribe what ran.

`ZicfissSm._generate_instr_inactive` already handles this correctly by enumerating only `(0,0) / (1,0) / (1,1)`. Either do the same here, or reorder the writes so the leg means something.

coverpoints/priv/ZicfissS_coverage.svh
```diff
@@ -0,0 +1,260 @@
...
+    sstatus_mxr: coverpoint ins.prev.csr[CSR_SSTATUS][19] {
+        bins mxr_clear = {1'b0};
+        bins mxr_set   = {1'b1};
+    }
+    // What was written into senvcfg bit 3, and what it reads back as.
+    sse_bit_written: coverpoint ins.current.rs1_val[3] {
+        bins wrote_zero = {1'b0};
+        bins wrote_one  = {1'b1};
+    }
+    senvcfg_sse_readback: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "senvcfg", "sse") {
```
davidharrishmc, line 177: A question rather than a defect: this and `cp_senvcfg_sse_rdonly0_s` (line 239) seem to assume different things about the same value.

The comment at line 100 says the trace does not re-log senvcfg when menvcfg.SSE=0, which is why `s_sse_state` ANDs the two fields. But `cp_senvcfg_sse_rdonly0_s` samples `senvcfg_sse_readback` raw and relies on it reading 0 under menvcfg.SSE=0, discarding the opposite with `ignore_bins rdonly0_cannot_read_one`. Both can't be describing the same quantity.

The architectural behaviour is still checked through the signature comparison either way, so this is a coverage-accounting question, not a hole in the test — but if line 100 is right then the AND in `s_sse_state` / `ZicfissU.u_sse_active` / `ZicfissSm.sse_state` is masking a real axis. Same pair at `ZicfissSm_coverage.svh:132` / `:183`.

config/imperas/ci.yaml
```diff
@@ -13,6 +13,9 @@
 # - InterruptsSm/S/SSm/U: Imperas does not implement external interrupts
 # - Smstateen: Sail does not implement all state yet to match
 # - ExceptionsVx: https://solvnetplus.synopsys.com/s/case/500WQ000023l1CPYAY/vmvnrrv-does-not-raise-an-illegalinstruction-exception-when-vtypevill-1
+# - ZicfissU/S/Sm/V: not yet verified against Imperas. The suites are new and Imperas cannot run
```
davidharrishmc: I ran these suites against Imperas so this exclusion can cite real bugs instead of "not yet verified". ImperasDV `Y-2026.03-DEV-20260612`, `Zicfiss=T` confirmed by default on the `RVA23S64` variant, Sail 0.14.1 as the reference:

- `imperas-rv64-max`: **5 of 6 fail** (only ZicfissV passes)
- `imperas-rv32gck`: **5 of 5 fail**

Three distinct model bugs, all of which look like Imperas is wrong and Sail is right:

1. **A memory fault must outrank the shadow-stack software check.** `ZicfissSm-00` at `sspopchk ra` in `cp_ss_pmp_permissions_sspopchk_x1_pmp_r_rv64`: Sail raises store/AMO access fault (mcause=7), Imperas raises software check (mcause=18, mtval=3). Same at `ZicfissS-00`, `cp_ss_instr_target_page_s_sspopchk_x1_xwr001_s_rv64`, where Sail raises store/AMO page fault (scause=15) and Imperas again raises software check. This is the same class as the QEMU bug already cited above (qemu work-item 4379).
2. **A SUM violation is a page fault, not an access fault.** `ZicfissS-01` at `sspush ra` in `cp_ss_page_perm_priority_sspush_x1_u1_sum0_mxr0_rv64`: S-mode store to a U=1 page with SUM=0. Sail raises store/AMO page fault (scause=15, stval=0x1403007f8), Imperas raises store/AMO access fault (scause=7). This is the same bug Whisper has, already described in `config/whisper/ci.yaml`.
3. **The SS-page PTE encoding isn't modelled.** `ZicfissU-00/01` produce no signature at all (actual offset 0x0 vs expected 0x2d0) and the log fills with `Warning (RISCV_PTWE) Page table entry non-leaf and U!=0 [address=0x140300800 PTEAddress=0x8001a800 access=R]` — that address is the shadow stack page, so Imperas is classifying `pte.xwr=010` as non-leaf instead of as an SS page.

So the exclusion is justified, but I'd rewrite the comment to name these three and file them with Synopsys the way the Ssstrict and ExceptionsVx entries above do — otherwise "not yet verified" becomes permanent.

framework/src/act/sail_to_rvvi.py
```diff
@@ -25,29 +25,69 @@ def sailLog2Trace(inputLogFile: Path, outputTraceFile: Path) -> None:
         "V": re.compile(r"v(\d+) <- 0x([0-9a-fA-F]+)"),
     }
 
+    # Page-table walks (Sail --trace-ptw): one "PTW: Step" per level, ending in "PTW: Success" or "PTW: failed".
```
davidharrishmc, line 28: Worth flagging how much this does for suites outside Zicfiss, since nothing in the PR description mentions it.

`--trace-ptw` plus the TRAP / PTE_I / PTE_D emission changes the reference trace for **every** extension's coverage run — the PTE coverpoints in `Sv`, `Svadu`, `Svade` and `SvZicbo` have been dead against the Sail trace until now. Measured on `sail-rv64-max`, base `2dea4b02c` vs this branch:

| covergroup | base | this PR |
|---|---|---|
| `Svade_cg` | 6.83% | **32.91%** |
| `Svadu_cg` | 7.85% | **37.85%** |

I checked it doesn't cost anything or break anything:

- no bin regressed — the set of zero bins strictly shrank in both covergroups
- trace files grow about 1%; `--trace-ptw` is present in the pinned Sail 0.14.1
- running the old and new parsers over all 455,258 instruction records in the Svade / Svadu / Zicfiss traces gives byte-identical output apart from the new tokens, so the restructured line loop drops no register writes

Just be aware that merging this moves the published coverage number for every VM suite, so the CTP/coverage baselines will want regenerating.

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
UmerShahidengr: Fixed in 6ceca23b1. `map_zicfiss_pages` now takes `ss_page_user`, and this sweep passes `user=True, ss_page_user=False`: the image stays user-mapped for the U-mode testcases, and the SS page's U bit comes only from `perms`. The generated asm now has `(PTE_D | PTE_A | PTE_W | PTE_V)` for the U=0 leg and a single `PTE_U` for U=1. The sweep also runs SSAMOSWAP now, so it covers every SS memory instruction.

coverpoints/priv/ZicfissU_coverage.svh
UmerShahidengr: Crossed in 6ceca23b1: `cp_ss_instr_target_page_u_mxr: cross priv_mode_u, ss_mem_instr, pte_ss_page, pte_u, sstatus_mxr`. The sweep's testcases are tagged with it, and it's mapped under `ss_page_exceptions`. It's at 100% on sail-rv64-max (32 bins) and sail-rv32-max (28), and the `supervisor` bins are now hit by U-mode SS instructions, not by S-mode handler traffic.

coverpoints/priv/ZicfissSm_coverage.svh
UmerShahidengr: On (1): this is deliberate and repo-wide. `framework/src/act/fcov/riscv_arch_test.sv` undefines `H_SUPPORTED` right after including `rvtest_config.svh`, and `riscv_arch_test.h` does the same for the tests, until Sail supports the hypervisor extension. Every H-gated coverpoint in the repo (Sm, Smstateen, Sscofpmf, ...) is dormant for the same reason, and I've added a comment saying so above the cross. On (2): fixed in 6ceca23b1. `_generate_envcfg_rdonly0` now runs the same csrrw/csrrs write and read-back leg against henvcfg, under `#ifdef H_SUPPORTED`. The cross now has its senvcfg twin's `ignore_bins`, with the read-one-under-menvcfg.SSE=0 case as an `illegal_bins`. The yaml no longer counts the henvcfg half as covered; it names the cross as dormant until H is enabled.

coverpoints/norm/Zicfiss.yaml
UmerShahidengr: I don't think `mtval_swchk` can be reached from Zicfiss without writing medeleg. Shadow stack instructions are never active in M-mode, and the ACT4 boot sets `medeleg[18]=1`, so every shadow stack software check from S or U is taken in S-mode with stval=3. The rule that actually applies is `stval_exception_info`, which 6ceca23b1 maps to `cp_sspopchk_mismatch` and `cp_sspopchk_mismatch_s`. `mtval_swchk` is now listed in the yaml with that explanation rather than left unmapped. ZicfilpSm covers it for landing pad faults, which are taken in M-mode (#2382).

coverpoints/norm/Zicfiss.yaml
UmerShahidengr: Added in 6ceca23b1: `cp_ss_satp_bare` in ZicfissSm. With satp=0 and menvcfg.SSE=senvcfg.SSE=1, the suite drops to S-mode and then to U-mode and runs every SS memory instruction (sspush, c.sspush, sspopchk, c.sspopchk, ssamoswap.w/.d) with ssp on the SS page. Cross: `priv_mode_s_u, ss_mem_instr, ss_active_below_m, satp_bare`. `satp_mode_bare` is mapped to it, and it's at 100% on both XLENs. Sail and Spike both raise the store/AMO access fault. Whisper still fails ZicfissSm earlier, on its existing M-mode SSAMOSWAP bug, so it never reaches this leg.

coverpoints/priv/ZicfissH_coverage.svh
UmerShahidengr: Fixed in 6ceca23b1. `vs_sse_state` now samples henvcfg.SSE as its effective value (`menvcfg.SSE && henvcfg.SSE`), the same way the S/U/Sm files handle the stale senvcfg log. `inactive_men_off` is `{2'b00}`, the illegal-instruction case, and the duplicate `inactive_both` is gone.

coverpoints/priv/ZicfissH_coverage.svh
UmerShahidengr: Fixed in 6ceca23b1 as you suggested: `m0s0h1` is gone and `3'b001` joins `m0_child_set`. `sse_chain` also samples both children as effective values, so a stale log after menvcfg.SSE is cleared can't trip the illegal bins.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
UmerShahidengr: Fixed in 6ceca23b1. `_end` is a linker symbol, so an assembler-time `.if` can't see it. Instead the loop checks `t2` against `rvtest_uimg_lvl0_pg_tbl + 4096` on every iteration and stops there. Anything past the table is left unmapped, as the docstring says, instead of overwriting the pages under test.

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
UmerShahidengr: Removed in 6ceca23b1. `_generate_sse_gating` now sweeps only (1,1), (1,0) and (0,0), like ZicfissSm, and the docstring says why (0,1) is the same state as (0,0).

coverpoints/priv/ZicfissS_coverage.svh
UmerShahidengr: Good question. I checked the Sail traces, and the two sample different moments, so they're consistent:

- An explicit senvcfg write (csrrw/csrrs) is logged with its legalized value. In both ZicfissS files, every senvcfg write made with menvcfg.SSE=0 logs SSE=0. That write is what `cp_senvcfg_sse_rdonly0_s` samples.
- Clearing menvcfg.SSE does not re-log senvcfg. In each ZicfissS file, menvcfg.SSE goes 1→0 once while senvcfg.SSE is logged as 1, and the logged value stays 1 until the next senvcfg write. The AND in `s_sse_state` / `u_sse_active` / `sse_state` corrects exactly that stale value. It's the effective value, not a masked axis, because senvcfg.SSE really does read 0 then.

In 6ceca23b1 I rewrote those comments to say this, and made `rdonly0_cannot_read_one` an `illegal_bins` in S and Sm, so a read-back that wasn't legalized would now fail the coverage run instead of being dropped.

config/imperas/ci.yaml
UmerShahidengr: Thanks for running these. In 6ceca23b1 I rewrote the comment to name the three bugs, with the suite and coverpoint that exposes each one. I also removed ZicfissV from the exclusion, since it passes in your run and none of the three bugs affect it. I'll file the three with Synopsys and add the case links, like the Ssstrict and ExceptionsVx entries.

framework/src/act/sail_to_rvvi.py
UmerShahidengr: Thanks for measuring this. I've added a PR description, and it has a section on the converter change and its effect on the Sv/Svadu/Svade/SvZicbo coverage numbers, including your before/after table, so the baselines get regenerated when this merges.

davidharrishmc:
Claude review:

Z1 (medium) — the mtval_swchk "not reachable" justification is wrong, and the rule is reachable today. The yaml claims "the ACT4 boot delegates software-check exceptions to S-mode (medeleg[18]=1)". That mask is set only by RVTEST_BOOT_TO_SMODE (rvtest_setup.h:1203+); RVTEST_BOOT_TO_MMODE writes csrw medeleg, zero (:949) — and ZicfissSm-00.S defines BOOT_TO_MMODE. Its traps already land in M-mode (Sail gave mcause=7 for the ZicfissSm PMP case vs scause=15 for ZicfissS). ZicfissSm already runs SS instructions in S-mode; it just never generates a value mismatch there, which would give cause 18 with mtval=3. Related: neither Zicfiss covergroup samples CSR_STVAL or CSR_MTVAL anywhere, so the new stval_exception_info mapping is also traced by association. One xtval coverpoint plus one S-mode mismatch case in ZicfissSm closes both rules.

Z2 (low) — the ZicfissH fix covered two of four coverpoints. vs_sse_state and sse_chain now AND with menvcfg.SSE, but vu_sse_state and vu_sse_inactive still sample henvcfg[3]/senvcfg.SSE raw and can report active while menvcfg.SSE=0. Latent (no generator); I compile-checked the file standalone — 0 errors.

Z3 (low) — cp_ss_satp_bare goes through both_xlens, so it's gated on SV39_SUPPORTED/SV32_SUPPORTED. The one test whose premise is satp.MODE=Bare is skipped on a Bare-only DUT, which is the configuration the rule is about. It uses no page-table helper, so the guards aren't needed.

davidharrishmc:
Filed https://solvnetplus.synopsys.com/s/case/500WQ000028qjufYAA/zicfiss-discrepancies for imperas mismatches.

UmerShahidengr:
All three points are resolved now. Applied the following Zicfiss changes: an explicit S-mode mismatch case plus xtval coverpoints (Z1), the VU state fix (Z2), and dropping the Sv guards from the satp=Bare test (Z3).

davidharrishmc:
LGTM. Fixed a minor coverpoint mapping issue. Ready for Jordan's review.

davidharrishmc (approved):
LGTM

config/qemu/ci.yaml
```diff
@@ -24,7 +24,12 @@ ci_enabled: true
 # TODO: Remove SsstrictV when sail SsstrictV is implemented
 # Smstateen will be added back once sail starts supporting SM1P13 too
 # ZawrsU and ZawrsS fail on qemu due to stval mismatching, filed QEMU issue https://gitlab.com/qemu-project/qemu/-/work_items/4080
-exclude_extensions: "Sm,S,SsstrictSm,SsstrictS,SsstrictU,Zacas,ZacasZabha,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,ExceptionsZicboSm,SvZicbo,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU"
+# ZicfissU, ZicfissS, ZicfissSm, ZicfissV excluded due to QEMU bugs:
+#  1. sspopchk raises a software-check exception where a store/AMO access fault must take priority
+#     (https://gitlab.com/qemu-project/qemu/-/work_items/4379)
+#  2. mtval is reported as 0 on the M-mode SSAMOSWAP access fault, disagreeing with
+#     the reference model (https://gitlab.com/qemu-project/qemu/-/work_items/4380)
```
jordancarlin: Isn't this legal? Reporting either zero or the faulting address are both allowed.

config/spike/ci.yaml
```diff
@@ -6,7 +6,9 @@
 ci_enabled: true
 # Smstateen will be added back once sail starts supporting Sscsrind too
 # Sdtrig* not yet supported in Sail reference
-exclude_extensions: "Smstateen,SdtrigSm,SdtrigS,SdtrigU"
+# ZicfissU: Spike treats every shadow stack access as a store, so SSPOPCHK from a shadow stack page
+# with pte.D=0 raises a store/AMO page fault under Svade instead of reading it
```
jordancarlin: Issue link?

config/whisper/ci.yaml
```diff
@@ -5,8 +5,11 @@
 # SsstrictSm, unknown
 # Smstateen Sail does not yet support 1P13 bit
 # Sdtrig* not yet implemented in Sail
+# ZicfissU, ZicfissS, ZicfissSm excluded due to Whisper bugs: SSAMOSWAP in M-mode raises an
+# illegal-instruction exception instead of a store/AMO access fault, and an S-mode ordinary store to a
+# U=1 shadow stack page with SUM=0 raises a store/AMO access fault instead of the U/SUM page fault
```
jordancarlin: Issue link?

jordancarlin: I am a little hesitant to merge these tests until we know they are passing on at least one DUT. Looks like right now they are excluded on all DUTs so we don't know that they are working correctly.

framework/src/act/sail_to_rvvi.py
jordancarlin: This whole script will be going away soon so I am not reviewing that closely. Nothing jumps out as particularly problematic.

coverpoints/priv/ZicfissH_coverage_init.svh
```diff
@@ -0,0 +1,11 @@
+///////////////////////////////////////////
+//
+// RISC-V Architectural Functional Coverage Covergroups Initialization File
+//
+// Copyright (C) 2026 Harvey Mudd College, 10x Engineers, UET Lahore, Habib University
```
jordancarlin: Copyright for all of these new files should probably be either you or RVI since you are the one who developed them.

generators/testgen/src/testgen/priv/extensions/ZicfissU.py
```diff
@@ -0,0 +1,1091 @@
...
+
+Covers the ZicfissU sheet of the simplified Zicfiss testplan: shadow stack
+instruction behaviour, the ssp CSR, page/PMA behaviour, and the U-mode half of
+the SSE enable chain.
+
+The suite boots to U-mode. Each block makes a short T-SBI excursion to M-mode to set
+up translation and the SS page, runs the testcases in U-mode, then makes a second
+excursion to tear translation down and returns to U-mode. Privileged CSR accesses made
+from inside the U-mode body go through T-SBI instead. See ZicfissCommon for the x1/x5
+and encoding-width constraints.
```
jordancarlin: Clean up comment. What "simplified" testplan? Why does it mention PMAs for a U mode test? etc.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,547 @@
...
+``RVTEST_SIGUPD``. Every SS sequence therefore brackets itself with
+``save_link_regs`` / ``restore_link_regs`` and only calls ``write_sigupd`` once
+x1/x5 have been restored.
+
+Encoding width
+--------------
+With Zcmop enabled the assembler compresses ``sspush x1`` to ``c.sspush x1`` and
+``sspopchk x5`` to ``c.sspopchk x5``. The covergroups have distinct bins for the
+32-bit and 16-bit encodings, so the 32-bit forms must be emitted inside
+``.option norvc`` or those bins can never fill. ``ss_insn`` handles this.
```
jordancarlin: Seems like it is probably simpler to go the other direction. Remove `zcmop` from `march` and explicitly enable it only around the instructions that you want compressed.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,547 @@
...
+Page-table layout
+-----------------
+Three leaf pages share one Sv39/Sv32 leaf table so a single PTE chain covers all
+of them:
+  _VA_SS   — the shadow stack page,  pte.xwr = 010
+  _VA_RW   — an ordinary read/write page, pte.xwr = 011
+  _VA_RO   — a read-only page, pte.xwr = 001
+"""
+
+from __future__ import annotations
```
jordancarlin: Remove from all files.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,547 @@
...
+    "101": "PTE_D | PTE_A | PTE_X | PTE_R | PTE_V",
+    "110": "PTE_D | PTE_A | PTE_X | PTE_W | PTE_V",
+    "111": "PTE_D | PTE_A | PTE_X | PTE_W | PTE_R | PTE_V",
+}
+
+# SSAMOSWAP follows the A-extension alignment rules, but the reference models differ on whether a
+# misaligned SSAMOSWAP that stays inside one misaligned atomicity granule executes, as an AMO, or
+# faults. The alignment sweep starts 8 bytes into a 16-byte granule so that every misaligned access
+# crosses it and must fault. SSAMOSWAP.W with addr[2:0] of 1-3 cannot cross, so it is left out.
+SSAMOSWAP_SWEEP_BASE = 0x408
```
jordancarlin: This seems like it is avoiding a bug? We should not skip testing something just because the simulators are doing different things. Either one of them has a bug or Sail needs an additional config option.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,547 @@
...
+    """addr[2:0] values swept for SSAMOSWAP.W (``w``) or SSAMOSWAP.D (``d``)."""
+    return [0, 4, 5, 6, 7] if width == "w" else list(range(8))
+
+
+# menvcfg/senvcfg/henvcfg SSE field is bit 3.
+SSE_BIT = 3
+
+GOTO_UMODE = "RVTEST_TSBI_GOTO_UMODE  # enter U-mode"
+GOTO_SMODE = "RVTEST_TSBI_GOTO_SMODE  # enter S-mode"
+GOTO_MMODE = "RVTEST_TSBI_GOTO_MMODE  # return to M-mode"
```
jordancarlin: Inline these.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,547 @@
...
+SSAMOSWAP_SWEEP_BASE = 0x408
+
+
+def ssamoswap_sweep_offsets(width: str) -> list[int]:
+    """addr[2:0] values swept for SSAMOSWAP.W (``w``) or SSAMOSWAP.D (``d``)."""
+    return [0, 4, 5, 6, 7] if width == "w" else list(range(8))
+
+
+# menvcfg/senvcfg/henvcfg SSE field is bit 3.
+SSE_BIT = 3
```
jordancarlin: Is this needed? Could we use the value from encoding.h instead?

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,547 @@
...
+# SSAMOSWAP follows the A-extension alignment rules, but the reference models differ on whether a
+# misaligned SSAMOSWAP that stays inside one misaligned atomicity granule executes, as an AMO, or
+# faults. The alignment sweep starts 8 bytes into a 16-byte granule so that every misaligned access
+# crosses it and must fault. SSAMOSWAP.W with addr[2:0] of 1-3 cannot cross, so it is left out.
+SSAMOSWAP_SWEEP_BASE = 0x408
+
+
+def ssamoswap_sweep_offsets(width: str) -> list[int]:
+    """addr[2:0] values swept for SSAMOSWAP.W (``w``) or SSAMOSWAP.D (``d``)."""
+    return [0, 4, 5, 6, 7] if width == "w" else list(range(8))
```
jordancarlin: The testplan says to sweep across all 8 values for both instructions. Why does this diverge? Probably indicates an issue with the coverpoints too.

generators/testgen/src/testgen/priv/extensions/ZicfissCommon.py
```diff
@@ -0,0 +1,547 @@
...
+def ss_insn(mnemonic: str, *, compressed: bool = False) -> list[str]:
+    """Emit one instruction, pinning its encoding width.
+
+    The assembler will happily compress ``sspush x1`` / ``sspopchk x5``, or an ordinary
+    ``lw``/``sd`` on x8-x15, which would leave the 32-bit covergroup bins unreachable. Wrap
+    the uncompressed forms in ``.option norvc`` and spell the compressed forms explicitly.
+    """
+    if compressed:
+        return [f"{mnemonic}"]
+    return [".option push", ".option norvc", mnemonic, ".option pop"]
```
jordancarlin: Drop this and selectively enable compressed instructions instead.

PR #2199, opened 2026-08-27
Extend test selection metadata

jordancarlin:
- Adds support for a list of prohibited extensions for each test
- Adds nested lists to indicate one extension out of a list is required
- Updates ExceptionsZicbo{U/S} to require at least one of Zicbom, Zicboz, Zicbop

### DIFF
--- a/docs/DeveloperGuide.md
+++ b/docs/DeveloperGuide.md
@@ -136,15 +136,15 @@ REQUIRED_EXTENSIONS:
 
 Each top-level item is required. A string requires that extension. A nested
 list requires at least one extension in that list. More than one nested list
-can be used. For example, `[I, [A, B], [F, D]]` means `I AND (A OR B) AND
-(F OR D)`.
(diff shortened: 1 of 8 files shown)


docs/DeveloperGuide.md
```diff
@@ -110,28 +110,45 @@ assembly code in the file.
...
+REQUIRED_EXTENSIONS:
+  - I
+  - [Sm, U]
+```
 
-# Unquoted style (common in hand-written tests)
-REQUIRED_EXTENSIONS: [I, S, Zicsr, Sm]
+Each top-level item is required. A string requires that extension. A nested
+list requires at least one extension in that list. More than one nested list
+can be used. For example, `[I, [A, B], [F, D]]` means `I AND (A OR B) AND
```
davidharrishmc: Can this be more of a real-world example with a reason OR is needed.  Maybe give Zicbo example and SmnpmS/U example to motivate?

davidharrishmc (approved):

PR #2207, opened 2026-08-28
Draft Certification Testing Information

davidharrishmc:
For review. 

Before merging this, @jordancarlin will implement the CERTIFICATE=RVA23 flag to limit extensions to the required and optional extensions of the RVA23 profile.  The flag should also ensure that at least the required extensions are defined in the UDB configuration.


README.md
```diff
@@ -315,24 +315,25 @@ Once all [dependencies](#prerequisites) are installed and the [configuration fil
...
-| `FAST`                | _(empty)_                                       | Set to `True` to skip objdump generation for faster builds. Makes debugging mismatches harder. Mutually exclusive with `DEBUG`.                                                                                                       |
-| `CLEAN_INTERMEDIATES` | _(empty)_                                       | Set to `True` to delete each config's intermediate `build/` directory (`.sig.elf`/`.sig`/`.results`/logs) after a successful build, keeping only the ELFs. Saves disk space (useful for CI caching). Mutually exclusive with `DEBUG`. |
-| `JOBS`                | Auto-detected from `make -j` flag, or CPU count | Number of parallel build jobs for test compilation. Set to `1` for debugging test hangs.                                                                                                                                              |
```
jordancarlin: This should be retained. The `JOBS` flag is how the default of using the full number of available cores is overridden.

docs/RVA23Cert.md
```diff
@@ -0,0 +1,45 @@
+# RVA23 Certification
+
+RISC-V International is developing an RVA23 Certification Program. The program is not yet live, but a significant part of obtaining an RVA23 certificate will involve running the subset of ACTs relevant to the RVA23 profile. This page documents that process for those who would like to begin running RVA23 tests now.
```
kdockser, line 3: Some ACTs can serve as certification tests, but it is not necessarily the case that all certification tests are ACTs. It is misleading and confusing to refer to the ACTs as certification tests. We need to separate the two.
Here is a proposed rephrasing:
a significant part of _applying for_ an RVA23 certificate will involve _passing a series of architectectural tests, including some_ ACTs relevant to the RVA23 profile.

docs/RVA23Cert.md
```diff
@@ -0,0 +1,45 @@
...
+implemented. Furthermore, any optional extensions as specified in the profile may be implemented. Any optional portions
+of the base ISA and these extensions may be implemented, but if they are
+implemented, they must be implemented in their entirety. The instruction set and extensions are defined in the [Ratified ISA Specifications](https://riscv.org/specifications/ratified/).
+
+There is no prohibition against implementing other RISC-V ratified base ISAs or
+extensions, custom extensions, or even non-conforming extensions as
+long as they don't interfere with the proper functioning of the RVA23 mandatory
+and implemented optional extensions. However, these other base ISAs and extensions are not part of RVA23 testing.
+
+Remember that certification testing is not verification. Certification tests check that you read and understood the specification, and are intended to promote software interoperability. They do not probe for microarchitectural or logic bugs and are not a substitute for your own comprehensive verification suite.
```
kdockser: We don't want to over promise what the certification tests do. Here is a proposed revision:

`Certification tests _are intended to help_ check that you read and understood the specification`

docs/RVA23Cert.md
```diff
@@ -0,0 +1,45 @@
...
+
+The RVA23 profile does not specify machine-mode requirements for an RVA23 profile. The ACTs use a Test Supervisor Binary Interface (T-SBI) to abstract access to machine mode, facilitating testing of systems with non-standard machine mode. The ACTs come with a default T-SBI implementation that works for standard M-mode. If your system does not implement standard machine mode, consult the [Certification Requirements Document](https://github.com/riscv/riscv-arch-test/blob/act4/docs/crd/src/rva23_crd.adoc) for information about what your T-SBI must provide.
```
kdockser, line 18:
```suggestion
The RVA23 profile does not specify machine-mode requirements. The certification tests use a Test Supervisor Binary Interface (T-SBI) to abstract machine mode, in a similar manner to the RISC-V SBI. The Certification Tests  come with a default T-SBI implementation. If your system includes implementation defined behaviors that need to be accounted for when running code in M-mode, consult the [Certification Requirements Document](https://github.com/riscv/riscv-arch-test/blob/act4/docs/crd/src/rva23_crd.adoc) for information about what your T-SBI must provide.
```

docs/RVA23Cert.md
```diff
@@ -0,0 +1,45 @@
...
+
+To run the RVA23 tests, follow the [Getting Started Guide](https://github.com/riscv/riscv-arch-test#getting-started) to install the toolchain, clone the ACT repository, and create configuration files for your Device Under Test. Then generate the tests and compile them into self-checking ELFs in `$WORKDIR/<config_name>/elfs` with:
+
+`CONFIG_FILES=<your_config_directory>/test_config.yaml make CERTIFICATE=RVA23`
+
+The `CERTIFICATE` argument limits the generated tests to those applicable to the certificate. In particular, it excludes machine-mode tests and extensions that are not required or optional in the profile. Your UDB configuration must contain at least all of the required extensions for the certificate.
```
kdockser:
```suggestion
The `CERTIFICATE` argument limits the generated tests to those applicable to the certificate. In particular, it excludes machine-mode tests and extensions that are neither mandatory nor optional in the profile. Your UDB configuration must contain at least all of the required extensions for the certificate.
```

kdockser (commented):
To help avoid confusion, we need to avoid conflating ACTs with Certification Tests.
Also, the RISC-V SBI doesn't talk about "non-standard M-mode", we should be consistent with its wording to avoid creating the impression that certification is abstracting M-mode for different reasons. This is the case even though certification expressly accesses more of the M-mode CSRs than would normally be available to [H]S mode software. These other CSRs would still need to be available to M-level code in a non-certification environment to enable control of the architected [H]S-mode defined features.

allenjbaum:
Why would't it be the case that all Certification tests be ACTs?
That implies that there are ISA certification tests that are not ACTs, and
ACTs that are not part of a certification
I can't think of a case where the a Cetification test wouldn't be an ACT,
though I think it is obvious that an ACT wouldn't be part of Certification
(e.g. not a required feature)

kdockser:
@allenjbaum The ACTs are a great source of tests that can be individually selected by the CSC to be certification tests. However, there is no reason to limit ourselves to ACTs as there are other more flexible tests generators that the CSC should be able to choose from.

I think that the more important issue is having a common user interface such that those who are running the tests do not need to modify their input files (e.g. parameters) for different test sources. It should be transparent to them.

docs/RVA23Cert.md
davidharrishmc: Updated per Friday's Requirements WG discussion.  @kdockser please review

PR #2209, opened 2026-08-28
Fix PMP coverpoints for VCS

jordancarlin:
Signed-off-by: Jordan Carlin <jcarlin@qti.qualcomm.com>

PR #2218, opened 2026-08-30
PMP Test Python Generator

jordancarlin:
Replace handwritten PMP tests with Python generator.

Built on top of #2186 with the following additional changes:
- Remove PMP specific code paths and integrate PMP generators into the standard priv test generation pipeline. This should make long term maintenance easier and avoids all of the custom PMP logic.
  - This required a few changes to the general priv test generation (which should be helpful for Sv tests as well):
    - Priv tests now (optionally) support generating XLEN-specific variants.
    - Adds support for appending `raw_data` to the generated tests.
    - Adds support for registering multiple generators for a single extension. Each will produce a different file in the same output directory.
- Inline the PMP verification macros. This makes the resulting assembly code easier to follow, let's us use the standard Python `add_testcase` and `write_sigupd` functions, and results in more descriptive debug strings.
- Fix PMP coverpoints for VCS (supersedes #2209)
- Fixes a bunch of other bugs found in the PMP tests:
  - Integer registers were being checked instead of floating point registers after `flw` and related instructions.
  - The wrong register was checked after `jalr` instructions.

I'm not sure why GitHub is showing the PMP tests as fully new, but they are still getting 100% coverage and I have reviewed the diff compared to what was in #2186 and it all looks correct to me.


UmerShahidengr (approved):

.gitignore
```diff
@@ -36,6 +36,7 @@ tests/rv64i/Zv*
 # priv tests
 tests/priv/headers
 tests/priv/**/*.S
+!tests/priv/rv*/**/*.S
 !tests/priv/pmp/**/*.S
```
davidharrishmc: Is this still needed?

tests/priv/PMPS/PMPS_cfg_A_off-00.S
```diff
@@ -0,0 +1,116 @@
...
+  LI(a4, RVTEST_PMP_RET_ENCODING)
+PMPS_cp_cfg_A_off_off_2_sw:
+  sw a4, 0(a5)
+  # Check if x14 contains the expected result. x2 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x2, x5, x4, x14, PMPS_cp_cfg_A_off_off_2_sw, PMPS_cp_cfg_A_off_off_2_sw_str)
+PMPS_cp_cfg_A_off_off_3_lw:
+  lw a4, 0(a5)
+  # Check if x14 contains the expected result. x2 is the signature ptr, x5 is the link ptr, x4 is a temp reg.
+  RVTEST_SIGUPD(x2, x5, x4, x14, PMPS_cp_cfg_A_off_off_3_lw, PMPS_cp_cfg_A_off_off_3_lw_str)
+  RVTEST_TSBI_GOTO_MMODE
```
davidharrishmc, line 76: Do we still want this?  I think RVTEST_CODE_END takes care of it.

davidharrishmc (approved):

.gitignore
jordancarlin: Already removed. I think you were looking at an old version.

tests/priv/PMPS/PMPS_cfg_A_off-00.S
jordancarlin: It is inserted after each iteration and isn't worth special casing the end just to avoid it.

PR #2219, opened 2026-08-30
Fix: Zkr test needs to account for OPST != ES16

AAgarwal04:
`cp_zkr_seed_entropy_zero_non_es16` treats entropy being present (`OPST == ES16`) as the only passing signal by writing `0xB0BA` in the scorecard when the match is found. 

Initially, if `OPST == ES16`, the test wrote `0xB0BA` into the signature. Otherwise it just wrote the entropy bits. An empty source (WAIT has an entropy 0) therefore wrote 0. That is spec-legal, but Sail is a virtual ES and always returns ES16, so it writes `0xB0BA`. The signatures mismatch and the test fails.

The Zkr spec allows seed to stay in WAIT/BIST/DEAD, and requires entropy bits to be zero if OPST is in the WAIT state.

[Section 31.1.4.1](https://docs.riscv.org/reference/isa/v20260120/unpriv/scalar-crypto.html#crypto_scalar_seed_csr):

> The status bits seed[31:30] = OPST may be ES16 (10), indicating successful polling, or one of three entropy polling failure statuses BIST (00), WAIT (01), or DEAD (11), discussed below.

> When OPST is not ES16, entropy must be set to 0. An implementation may safely set reserved and custom bits to zeros.

> 01 - WAIT means that a sufficient amount of entropy is not yet available. This is not an error condition and may (in fact) be more frequent than ES16 since physical entropy sources often have low bandwidth.

The change is only in `_generate_seed_entropy_zero_non_es16_tests` in `generators/testgen/src/testgen/priv/extensions/ZkrCommon.py`. It still does two `csrrw` seed polls and still scores `cp_zkr_seed_entropy_zero_non_es16`. What changed is what gets written to the signature.

With this change, entropy is initially masked.

- `OPST == ES16` or not-ES16 with entropy 0: expected behavior so write `0xB0BA` to the signature register.
- `OPST != ES16` and `entropy != 0`: Send the entropy to the signature register and compare against `0xB0BA`. 

A DUT that stays in WAIT/BIST/DEAD with entropy 0 now matches. A real leak still fails.


davidharrishmc:
If it stays in wait, is it worth spinning for a while and seeing if gets to ES16?

jordancarlin:
Also, presumably we don't want to accept DEAD. What about reading `seed` repeatedly 100 times. On each loop, if it reads `DEAD`, fail the test. If it reads `BIST` or `WAIT`, check that the other bits are 0 and continue to the next loop. If it reads `ES16`, check that the entropy field is different than what we read the previous time and continue the loop.

davidharrishmc:
Bandwidth on seed could be much lower.  Maybe read it up to 10k times?  Only give one signature at the end.
If it remains in WAIT the whole time, still pass?  Should BIST pass, or should the DUT boot sequence be required
to initialize the entropy source to a non-BIST state?

AAgarwal04:
I feel like these additions could be useful but should be a different test to check for liveness. According to the spec [Section 31.1.7.1](https://docs.riscv.org/reference/isa/v20260120/unpriv/scalar-crypto.html#31-1-7-1-checklists-for-design-and-self-certification)

> ISA Architectural Tests. Verify, to the extent possible, that RISC-V ISA requirements in this specification are correctly implemented. This includes the state transitions ([31.1.4. Entropy Source](https://docs.riscv.org/reference/isa/v20260120/unpriv/scalar-crypto.html#crypto_scalar_es) and [31.1.7.6. Suggested GetNoise Test Interface](https://docs.riscv.org/reference/isa/v20260120/unpriv/scalar-crypto.html#crypto_scalar_es_getnoise)), access control ([31.1.4.3. Access Control to seed](https://docs.riscv.org/reference/isa/v20260120/unpriv/scalar-crypto.html#crypto_scalar_es_access)), and that seed ES16 entropy words can only be read destructively. The scope of RISC-V ISA architectural tests are those behaviors that are independent of the physical entropy source details. A smoke test ES module may be helpful in design phase.

As such, I think that the current test should solely focus on correctness (i.e. if BIST/DEAD/WAIT states yield an entropy of 0x0; ES16 is also a pass). If the DUT stays in WAIT the whole time, that should still pass. Requiring ES16 after polling the DUT _n_ number of times seems like a test for the system's bandwidth, which is not an ISA requirement.

If the `OPST == DEAD`, although this means that something is wrong with the system, this is still valid behavior if the entropy bits are all 0. I would not require consecutive ES16 values to differ. The spec allows the same 16-bit pattern twice.

Also, if the same number is returned twice, this does not necessarily mean that the architecture is faulty. As per [Section 31.1.4.1](https://docs.riscv.org/reference/isa/v20260120/unpriv/scalar-crypto.html#crypto_scalar_seed_csr):

> Each returned seed[15:0] = entropy value represents unique randomness when OPST=ES16 (seed[31:30] = 10), even if its numerical value is the same as that of a previously polled entropy value.

jordancarlin:
Good point about consecutive reads being allowed to be the same. I guess we don't have a great of checking that the entropy bits are actually randomized on each read, so we may need to skip that.

Agreed that staying in WAIT the whole time should pass. I think all three of us are in agreement about that.

I'm inclined to say that certification tests should never see DEAD and if they do it isn't unreasonable for them to fail, but I could be convinced otherwise on that one.

jordancarlin:
@AAgarwal04 it would be great to get this updated based on the discussion in this thread so we can merge it.

AAgarwal04:
`generators/testgen/src/testgen/priv/extensions/ZkrCommon.py` is edited so that the test now fails when `OPST == DEAD`. In this scenario, a value of `0xDEAD` shows up as the signature word instead of the expected `0xB0BA`, prompting a failure.

I did not implement a WAIT/BIST polling loop. As discussed above, it is legal behavior to have `OPST == WAIT` or `OPST == BIST` as long as entropy is 0x0. In both scenarios, entropy is checked to be 0x0.

jordancarlin:
> I did not implement a WAIT/BIST polling loop. As discussed above, it is legal behavior to have `OPST == WAIT` or `OPST == BIST` as long as entropy is 0x0. In both scenarios, entropy is checked to be 0x0.

Agreed that it is legal for OPST to stay in WAIT or BIST, but I don't think that means we shouldn't add a bounded polling loop. The point of this coverpoint is to verify that when OPST is WAIT or BIST, the rest of the bits are zero. We are more likely to encounter WAIT and/or BIST if we read the `seed` CSR a bunch of times in rapid succession.

I'm still in favor of reading seed repeatedly 100 times. On each loop, if it reads DEAD, fail the test. If it reads BIST or WAIT, check that the other bits are 0 and continue to the next loop if they are; fail the test if they are not. If it reads ES16, continue to the next loop without checking the entropy bits. Do you see any potential issues with that approach?

AAgarwal04:
I agree with your approach. I have added a loop (arbitrarily at 100 iterations) which checks if entropy == 0x00 when OPST is in the WAIT or BIST stage. If OPST == ES16, entropy is not checked and the loop continues. If at any instance OPST == DEAD, the loop is exited and the test fails.

jordancarlin:
@AAgarwal04 thanks for continuing to work on this. Your commits will need to be cryptographically signed before this can be merged. See https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#ssh-commit-signature-verification for instructions.

jordancarlin:
@AAgarwal04 looks like you signed the commits but haven't uploaded your public key to your GitHub account, so it still isn't possible to merge.

AAgarwal04:
Sorry about that. This should now be fixed. Please let me know if it still isn't possible to merge.

davidharrishmc (approved):

PR #2220, opened 2026-08-31
Ssstrict Fixes

davidharrishmc:
* Zalasr should not be declared in Sail UDB because it is not implemented
* Add RVTEST_CODE_END check that proper number of traps are produced even when RVTEST_FAST_TRAP_HANDLER is used.  Previously, fast trap handler silently ignored mismatching trap count, hiding some errors in Ssstrict.
* Test vector crypto with LMUL=4 to avoid problems with short VLEN, per spec 33.1.7 software portability recommendations
* Fix Whisper CI exclusions


jordancarlin:
Something seems to have gone wrong. This has 13 commits and 0 changed files.

config/whisper/ci.yaml
```diff
@@ -4,13 +4,12 @@
 
 # SsstrictSm, SsstrictS, SsstrictU excluded because of sail-riscv#1882, sail-riscv#1880
 #   Should be fixed in Sail 0.14
+#   Also https://github.com/tenstorrent/whisper/issues/52
```
jordancarlin: This issue was closed already. Should it be removed from this list?

tests/env/rvtest_setup.h
```diff
@@ -149,11 +149,24 @@
...
+    // Add a signature record with the signature pointer's own offset to catch
+    // mismatched trap count.
+    check_fast_sig_offset:
+    #if defined(RVTEST_USE_FAST_TRAP_HANDLER) && !defined(RVTEST_NOSIG)
+      LA(     T1, rvtest_sig_begin)
+      sub     T1, DEFAULT_SIG_REG, T1   // T1 = byte offset of the final signature pointer
+      // x5/x4 = DEFAULT_LINK_REG/DEFAULT_TEMP_REG, written literally because
+      // RVTEST_SIGUPD token-pastes them into the failedtest_x5_x4 entry name.
+      RVTEST_SIGUPD(DEFAULT_SIG_REG, x5, x4, T1, check_fast_sig_offset, fast_sig_offset_mismatch)
+    #endif
```
jordancarlin: I'd rather back out this change and add a more global check for the number of signature checks. Looks like this is an issue for regular tests too, not just fast trap handlers. I will push a PR soon that adds a check for the number of signature updates separately.

jordancarlin (changes requested):

tests/env/rvtest_failure_code.h
```diff
@@ -2316,6 +2316,8 @@
         .string "RVCP: XSTATUS: "
     trap_sig_offset_mismatch:
         .string "\"Trap count mismatch.\"";
+    fast_sig_offset_mismatch:
+        .string "\"Final signature pointer differs from reference: trap counts diverged.\"";
```
jordancarlin: Looks like you forgot to revert this change.

tests/env/rvtest_setup.h
```diff
@@ -154,6 +154,7 @@
       RVTEST_WORD_PTR abort_test
       RVTEST_WORD_PTR abortstr
       .word   CSR_MEPC
+
```
jordancarlin: Remove extra whitespace.

jordancarlin (approved):

PR #2221, opened 2026-08-31
T-SBI Implementation of ExceptionsZicbo*

amanmurad-10xe:
Factor M-mode tests into the new ExceptionsZicboSm suite
Refactor ExceptionsZicboSm/S/U tests to share common test logic
Refactor ExceptionsZicboS/U tests to use T-SBI calls


coverpoints/priv/ExceptionsZicboS_coverage.svh
davidharrishmc: We've learned that menvcfg and senvcfg are defined in Sm1.12.0.  All the coverpoints and tests that use them should be gated by SM1P12P0_OR_LATER_SUPPORTED. We should still tests access faults and misaligned even in 1.11, and just not check *envcfg.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboS.py
```diff
@@ -8,352 +8,97 @@
...
     test_chunks: list[TestChunk] = []
     tc = test_data.begin_test_chunk()
 
-    tc.code.extend(_generate_cbie_tests(test_data))
-    tc.code.extend(_generate_cbcfe_tests(test_data))
-    tc.code.extend(_generate_cbze_tests(test_data))
-    tc.code.extend(_generate_cbo_access_fault_tests(test_data))
-    tc.code.extend(_generate_cbo_misaligned_tests(test_data))
+    for mode in _MODES:
+        tc.code.append(goto_mode(mode))
```
davidharrishmc: This could become RVTEST_TSBI_GOTO_{mode}MODE instead of a function call that isn't used elsewhere.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
```diff
@@ -0,0 +1,248 @@
...
+}
+
+
+def _csrw(csr: str, reg: int, mode: str) -> str:
+    """Return a csrw instruction: direct in M-mode, routed through T-SBI otherwise."""
+    instr = f"csrw  {csr}, x{reg}"
+    return instr if mode == "Sm" else tsbi_call(instr)
+
+
+def goto_mode(mode: str) -> str:
```
davidharrishmc: Could be replaced by one line in ZicboS.py.  See later comment.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
```diff
@@ -0,0 +1,248 @@
...
+        guard="ZICBOM_SUPPORTED",
+    ),
+    "cbze": _CboField(shift=7, bins=["0", "1"], instrs=["cbo.zero"], guard="ZICBOZ_SUPPORTED"),
+}
+
+
+def _csrw(csr: str, reg: int, mode: str) -> str:
+    """Return a csrw instruction: direct in M-mode, routed through T-SBI otherwise."""
+    instr = f"csrw  {csr}, x{reg}"
+    return instr if mode == "Sm" else tsbi_call(instr)
```
davidharrishmc: Access to senvcfg doesn't require a tsbi_call from S mode.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
```diff
@@ -0,0 +1,248 @@
...
+
+    lines = ["#ifdef RVMODEL_ACCESS_FAULT_ADDRESS", comment_banner(coverpoint, description), ""]
+
+    for cbo in CBO_INSTRS:
+        lines.append("#ifdef ZICBOZ_SUPPORTED" if cbo == "zero" else "#ifdef ZICBOM_SUPPORTED")
+        lines.extend(
+            [
+                f"LA(x{addr_reg}, RVMODEL_ACCESS_FAULT_ADDRESS)",
+                f"LI(x{cfg_reg}, 240)",  # setting all relevant bits in menvcfg to 1
+                _csrw("menvcfg", cfg_reg, mode),
```
davidharrishmc: Move menvcfg and senvcfg writes outside the for loop to reduce the number of T-SBI calls.

davidharrishmc: Applies to all helpers.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
```diff
@@ -0,0 +1,248 @@
...
+
+    lines = [comment_banner(coverpoint, description), ""]
+
+    for cbo in CBO_INSTRS:
+        lines.append("#ifdef ZICBOZ_SUPPORTED" if cbo == "zero" else "#ifdef ZICBOM_SUPPORTED")
+        lines.extend(
+            [
+                f"LA(x{addr_reg}, scratch)",
+                f"addi x{addr_reg}, x{addr_reg}, 1",
+                f"LI(x{cfg_reg}, 240)",  # setting all relevant bits in menvcfg to 1
```
davidharrishmc: This would be clearer to express 240 in binary.

Access to menvcfg should involve csrs or src instead of csrw so that the tests don't disturb other fields in the register.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboU.py
```diff
@@ -8,309 +8,91 @@
...
+_MODE = "U"
 
 
 @add_priv_test_generator(
     "ExceptionsZicboU",
     required_extensions=["U", ["Zicbom", "Zicboz", "Zicbop"]],
     march_extensions=["Zicbom", "Zicboz", "Zicbop"],
-    # TODO: Remove BOOT_TO_MMODE when converting this test to T-SBI.
-    extra_defines=["#define BOOT_TO_MMODE"],
+    extra_defines=["#define BOOT_TO_UMODE"],
```
davidharrishmc: Not necessary to have the extra_defines.   System will automatically boot to U mode if supported, and it's on the required_extensions list.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboU.py
```diff
@@ -8,309 +8,91 @@
...
-    tc.code.extend(["#ifdef S_SUPPORTED", "    LI(x11, -1)", "    csrw senvcfg, x11", "#endif"])
-    tc.code.extend(_generate_cbie_tests(test_data))
-    tc.code.extend(_generate_cbcfe_tests(test_data))
-    tc.code.extend(_generate_cbze_tests(test_data))
-    tc.code.extend(_generate_cbo_access_fault_tests(test_data))
-    tc.code.extend(_generate_cbo_misaligned_tests(test_data))
+    senvcfg_reg = test_data.int_regs.get_registers(1)[0]
+    tc.code.extend(
+        [
+            "#ifdef S_SUPPORTED",
```
davidharrishmc: Not necessary any more.  rvtest_setup initializes senvcfg if applicable when booting to U mode.

davidharrishmc (changes requested):
Looking good.  We've learned a bit since the original files were written, and this is a good time to update.

coverpoints/priv/ExceptionsZicboS_coverage.svh
amanmurad-10xe: Done. menvcfg access now gated by SM1P12P0_OR_LATER_SUPPORTED and senvcfg also under S1P12P0_OR_LATER_SUPPORTED. Access-fault and misaligned tests stay outside those guards so they still run in 1.11.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboS.py
amanmurad-10xe: Fixed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
amanmurad-10xe: Fixed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
amanmurad-10xe: Fixed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
amanmurad-10xe: Done, moved outside the loop in all helpers.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
amanmurad-10xe: Fixed. Value's in binary now, and switched to csrc+csrs with a field mask instead of csrw so other bits aren't touched.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboU.py
amanmurad-10xe: Removed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboU.py
amanmurad-10xe: Removed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboSm.py
```diff
@@ -0,0 +1,86 @@
...
+_MODE = "Sm"
+
+
+@add_priv_test_generator(
+    "ExceptionsZicboSm",
+    required_extensions=["Sm", ["Zicbom", "Zicboz", "Zicbop"]],
+    march_extensions=["Zicbom", "Zicboz", "Zicbop"],
+    extra_defines=["#define BOOT_TO_MMODE"],
+)
+def make_exceptionszicbosm(test_data: TestData) -> list[TestChunk]:
```
davidharrishmc, line 25: A single emit_suite(test_data, cg, mode, cross_senvcfg) function in common could replace all of this code that is nearly identical across configs.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboU.py
```diff
@@ -8,309 +8,78 @@
...
-    tc.code.extend(_generate_cbcfe_tests(test_data))
-    tc.code.extend(_generate_cbze_tests(test_data))
-    tc.code.extend(_generate_cbo_access_fault_tests(test_data))
-    tc.code.extend(_generate_cbo_misaligned_tests(test_data))
+    tc.code.extend(
+        cbo_config_helper(
+            test_data,
+            _CG,
+            "cbie",
+            description="Execute cbo.inval in user mode with menvcfg.cbie = {00/01/11}, via T-SBI",
```
davidharrishmc: via T-SBI isn't important to the user understanding what the coverpoint does.  Suggest dropping everywhere.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
```diff
@@ -0,0 +1,260 @@
...
+                lines.append(_csr_op("csrc", "senvcfg", mask_reg, mode))
+                if int(s_val, 2):
+                    lines.extend(
+                        [
+                            f"LI(x{cfg_reg}, 0b{int(s_val, 2) << shift:0{mask_bits}b})",
+                            _csr_op("csrs", "senvcfg", cfg_reg, mode),
+                        ]
+                    )
+                lines.append("#endif // S1P12P0_OR_LATER_SUPPORTED")
+            lines.append("nop")
```
davidharrishmc: No longer needed.

amanmurad-10xe: Removed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboU.py
amanmurad-10xe: Removed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
davidharrishmc: Looks like it is still here.

amanmurad-10xe: Pushed changes now.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboSm.py
amanmurad-10xe: Makes sense. Done.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
```diff
@@ -0,0 +1,312 @@
...
+        f"LI(x{mask_reg}, 0b{field_mask:0{mask_bits}b})  # {field} field mask",
+    ]
+
+    senvcfg_bins: list[str] = bins if cross_senvcfg else [""]
+    for m_val in bins:
+        lines.extend(
+            [
+                "",
+                f"# menvcfg.{field} = {m_val}",
+                _csr_op("csrc", "menvcfg", mask_reg, mode),
```
davidharrishmc, line 107: This looks like it unconditionally tries to access menvcfg even if U is not supported.  That would be an error if only Sm and SM1P12 is supported but not U.

coverpoints/norm/ExceptionsZicboS.yaml
davidharrishmc: cp_cbo_address_misaligned

coverpoints/norm/ExceptionsZicboU.yaml
davidharrishmc: cp_cbo_address_misaligned

davidharrishmc: cp_cbo_access_fault

davidharrishmc:
Looking pretty good to me.  Run it through your favorite AI for a sanity check before you call it done.  Mine pointed out a few inconsistencies.

coverpoints/norm/ExceptionsZicboS.yaml
amanmurad-10xe: Fixed.

coverpoints/norm/ExceptionsZicboU.yaml
amanmurad-10xe: Fixed.

amanmurad-10xe: Fixed.

generators/testgen/src/testgen/priv/extensions/ExceptionsZicboCommon.py
amanmurad-10xe: Fixed. The U-mode iteration in the S suite's S/U loop is now gated by #ifdef U_SUPPORTED.

amanmurad-10xe:
Added missing `cp_cbie/cp_cbcfe/cp_cbze` fallback bins for the Sm1.12-without-S1.12 case, matching `cp_cbo_address_misaligned`. AI sanity check clean otherwise, ready from my end. Let me know if anything else needs fixing.

coverpoints/priv/ExceptionsZicboS_coverage.svh
```diff
@@ -91,6 +91,13 @@ covergroup ExceptionsZicboS_cg with function sample(ins_t ins);
             `endif
             cp_cbo_address_misaligned:  cross cbo_instrs, adr_misaligned, priv_mode_s_u, menvcfg_all_enable, senvcfg_all_enable;
         `else
+            `ifdef ZICBOM_SUPPORTED
```
davidharrishmc: Unnecessary.  According to UDB, SM1P12P0 implies S1P12P0 if S is supported, so this branch will never occur.

davidharrishmc (changes requested):
The U_SUPPORTED guard protects the wrong thing.  It is applied to ExceptionZicboS, but S implies U, so U_SUPPORTED will always be true in that case.  

U_SUPPORTED should apply to Sm for the case of just machine mode.  In that case, menvcfg should not be touched.  The access fault and misaligned cases should still execute (they are gated off right now by the U logic), but they shouldn't touch menvcfg.  Please revert most of the changes in this file and just gate accesses to menvcfg by U_SUPPORTED.

coverpoints/priv/ExceptionsZicboS_coverage.svh
amanmurad-10xe: Fixed. I reverted the unnecessary U_SUPPORTED guards and now only guard the menvcfg accesses in the Sm case. The access-fault and misaligned instruction execution remains unconditional.

davidharrishmc (approved):

PR #2224, opened 2026-08-31
Remove stale tests when regenerating tests

### LINKED ISSUE #2201: [Bug] unexpected key 'CONFIG_DEPENDENT' found
### Affected area

### Commit

f7ba06c3d1a01428b172fe851ab56931a0624fee

### Affected test or extension

Sm-00

### Steps to reproduce

Simply `make` with these configs: https://github.com/nazar-pc/abundance/tree/126c5888e1f15ffd99457a61ab9e3a34bd7aa81f/crates/execution/ab-riscv-act4-runner/res/abundance

### Expected behavior

Tests compile successfully

### Actual behavior

```
      Built act @ file:///act4/framework                    
      Built covergroupgen @ file:///act4/generators/coverage
      Built testgen @ file:///act4/generators/testgen       
Uninstalled 3 packages in 1ms
Installed 3 packages in 1ms
✓ Generated covergroups for 87 extension(s)
✓ Generated 186 test suite(s)
╭───────────── Test YAML Header Error ─────────────╮
│ Malformed YAML header in Sm-00.S                 │
│                                                  │
│ Problem: unexpected key 'CONFIG_DEPENDENT' found │
│ File:    /act4/tests/priv/Sm/Sm-00.S             │
╰──────────────────────────────────────────────────╯
make: *** [Makefile:170: elfs] Error 1
```

I'm upgrading from 482611e0918a79eb7d5dcf253f4f6d37a27f240c that worked fine.

jordancarlin:
When test names changed, the old version was left and could lead to
broken compilation or extra tests running. This removes old tests before
generating the new ones.

Fixes #2201

### DIFF
--- a/generators/testgen/scripts/vector-testgen-priv.py
+++ b/generators/testgen/scripts/vector-testgen-priv.py
@@ -16,7 +16,6 @@ import math
 import os
 import pathlib
 import re
-import shutil
 from random import randint, seed
 
 import priv  # priv coverpoint generator scripts
@@ -557,8 +556,7 @@ if __name__ == '__main__':
 
     testplans = readTestplans(priv=True)
     extensions = list(testplans.keys())
-    for extension in extensions:
-      shutil.rmtree(pathlib.Path(ARCH_VERIF) / "tests" / "priv" / extension, ignore_errors=True)
+    generated_files: dict[str, set[pathlib.Path]] = {extension: set() for extension in extensions}
 
     for xlen in xlens:
       for extension in extensions:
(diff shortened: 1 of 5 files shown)


davidharrishmc:
Does this increase runtime for tests that were not stale, or is that unchanged?

jordancarlin:
It increases test generation time by about 2 seconds because of additional IO. Test compilation time is not changed because we use a hash to determine what needs to be rebuilt instead of file modification time.

davidharrishmc:
Sorry to fuss over 2 seconds, but that’s an appreciable portion of the whole.  Is there a more clever way of making a list of everything in tests, a list of everything being generated, and only deleting the diff?

davidharrishmc (approved):

jordancarlin:
I updated this to only delete tests that are no longer generated and skip tests that are still supposed to be there. From a performance perspective, this doesn't seem to change much compared to the previous approach, but it is more "correct".

I tried to do some more benchmarking, but I am getting 15 second run-to-run variance even with no changes, so comparing changes that have a 1 to 2 second impact is impossible.

PR #2225, opened 2026-08-31
Remove `make clean-tests`

jordancarlin:
The generated tests are checked in, so this caused more problems than it
solved.


jordancarlin:
Depends on #2224 being merged first.

davidharrishmc (approved):

PR #2257, opened 2026-09-06
Add invisible trap handling

### LINKED ISSUE #1924: SsstrictSm-03 ignores TIME_CSR_IMPLEMENTED=false and tests time CSR 0xC01

jordancarlin:
Adds an invisible trap-and-emulate trap handler for illegal instructions. The default implementation emulates `time` and `timeh` accesses when Zicntr is supported and `UDB_TIME_CSR_IMPLEMENTED` is false. There is also an `RVMODEL_INVISIBLE_TRAP_HANDLER` hook for custom emulation. When the invisible trap handler is enabled, illegal instruction traps are not delegated and are instead forwarded manually to S-mode after checking for emulation.

Fixes #1924


docs/ctp/src/abstraction.adoc
```diff
@@ -123,6 +121,7 @@ The RVMODEL DUT-Specific Macros are given in <<t-rvmodel>>. The "Can Be Empty?"
...
 |RVMODEL_MAX_CYCLES_PER_TIMER_TICK|The highest possible ratio of the clock frequency to the timer frequency in any mode of operation. Defaults to 1 if not defined | If Machine timer interrupts are not tested. | n/a
 |RVMODEL_DATA_SECTION|DUT-specific data that will be placed in a .data section. Used for memory-mapped regions like `tohost`/`fromhost`.  |Yes, if not needed by DUT. | n/a
+|RVMODEL_INVISIBLE_TRAP_HANDLER(_PC_REG, _INSTRUCTION_REG, _VALUE_REG, _ACTION_REG, _SCRATCH_REG)|Optionally emulates an illegal instruction in the M-mode invisible trap handler. `_PC_REG` and `_INSTRUCTION_REG` are read-only and contain the trapping PR and the trapping instruction respectivelt. Set `_ACTION_REG` to 0 to use the normal illegal-instruction path, 1 when the instruction was handled without writing a result, or 2 to write `_VALUE_REG` to the rd field of the trapping instruction. `_SCRATCH_REG` may be trashed.|Omit if the DUT has no invisible trap emulation. | Machine
```
davidharrishmc: PR -> PC

tests/env/rvtest_trap_handler.h
```diff
@@ -1222,11 +1222,25 @@ common_\__MODE__\()handler:                      // entered with T6 = vector add
...
+        csrr    T5, CSR_XCAUSE                  // T5 = xcause
+
+  // Route illegal-instructions to invisible trap handler in M-mode
+  .ifc \__MODE__ , M
+      LI(T4, CAUSE_ILLEGAL_INSTRUCTION)
+      bne T5, T4, invisible_Mcontinue
+      #ifdef RVTEST_INVISIBLE_TRAP_HANDLER
+          LA(T4, invisible_Mhandler)
+      #else
+          LA(T4, invisible_Mcontinue)
```
davidharrishmc, line 1394: Why can't the jump be inside the #ifdef and nothing in the #else?  
Why can't the LI and ben be in the #ifdef as well?

tests/env/rvtest_trap_handler.h
davidharrishmc: What does Mbootrap mean?  Should it be Mboottrap?

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
+// rvtest_invisible_trap_handler.h
+// RISC-V Architecture Test Framework — Invisible Trap Handler
+//
+// Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Invisible trap handler for emulated instructions.
+.macro RVTEST_INVISIBLE_TRAP_HANDLER_CODE
```
davidharrishmc, line 65: Should we add a DUT with S_SUPPORTED that can't handle reading "time" so we can keep this code in CI?

davidharrishmc: Is there a problem that an invisible trap from VS mode would mret to vs-mode at the HS stvec and not have permission?  Should we at least have an #error when in H until that logic is complete?

tests/env/rvtest_trap_handler.h
davidharrishmc, line 332: Should the contents of these 8 regs be documented, including offset 0 for invisible traps?

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+        li      T5, 0
+        beq     T4, T5, invisible_Mtime
+
+      // An invalid action is an integration error; take the normal trap path.
+      li      T5, CAUSE_ILLEGAL_INSTRUCTION
+      j       invisible_Mnormal_trap
+    #endif
+
+  invisible_Mtime:
+  #ifdef RVTEST_EMULATE_TIME_CSR
```
davidharrishmc: Do we need to remove UDB_TIME_CSR_IMPLEMENTED from Sm_coverage.svh or anywhere else in the acts?

docs/ctp/src/abstraction.adoc
davidharrishmc: respectivelt -> respectively

tests/env/rvtest_trap_handler.h
```diff
@@ -2626,6 +2638,16 @@ excpt_\__MODE__\()hndlr_tbl:
         sub  a1, a1, a2                 // a1 = mepc - code_begin (wraps if mepc is below it)
         LREG a2, code_seg_siz(a0)       // a2 = code segment size
         bgeu a1, a2, fast_Mbootrap      // outside the test code — use the standard handler
+#ifdef RVTEST_INVISIBLE_TRAP_HANDLER
+        // Let the invisible handler try emulation before recording this trap.
+        li   a1, 1
+        SREG a1, rvmodel_sv_off(a0)     // Mark a declined M-mode trap for the fast path.
+        j    fast_Mbootrap              // Restore a2, which the range check clobbered.
+#else
```
davidharrishmc, line 2851: Add comment why number of cops has to match the INVISIBLE_TRAP_HANDLER path.

docs/ctp/src/abstraction.adoc
davidharrishmc: I think the documentation needs more about the principles of operation of invisible traps and how to add support for DUT-specific ones.

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+    #ifdef RVMODEL_INVISIBLE_TRAP_HANDLER
+      li      T4, 0
+      // T1=mepc and T2=instruction are read-only. T3=value to write to rd, T4=action, and T5=scratch.
+      RVMODEL_INVISIBLE_TRAP_HANDLER(T1, T2, T3, T4, T5)
+      // A handled instruction needs no register write.
+      li      T5, 1
+      bne     T4, T5, 1f
+      j       invisible_Mtrap_return
+      1:
+        // Write the result to the instruction's destination register.
```
davidharrishmc: I presume that the invisible handler would write directly to FPRs/vector registers/CSRs.  Worth mentioning in the documentation and explain why this code 2 is necessary to avoid conflicts with trap handler registers.

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+        li      T5, 2
+        bne     T4, T5, 1f
+        j       invisible_Mwrite_gpr
+      1:
+        li      T5, 0
+        beq     T4, T5, invisible_Mtime
```
davidharrishmc: What if we put a label invisible_Mcustom_done just before the #endif in this block, and jump to that instead of invisible_Mtime?  Seems like invisible_Mtime should be inside the #ifdef RVTEST_EMULATE_TIME_CSR.

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+    invisible_Mtime_check_access:
+      // Check counter permissions before emulating the missing CSR.
+      csrr    T4, mstatus
+      LI(     T3, MSTATUS_MPP)
+      and     T4, T4, T3
+      beq     T4, T3, invisible_Mtime_read      // If in M-mode, proceed to read the time CSR
+      csrr    T3, mcounteren
+      andi    T3, T3, MCOUNTEREN_TIME           // mcounteren.TIME
+      beqz    T3, invisible_Mnot_handled
+      #ifdef S_SUPPORTED
```
davidharrishmc: Add TODO about hcounteren also needing support?

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+        beqz    T3, invisible_Mnot_handled
+      #endif
+
+    invisible_Mtime_read:
+      LI(     T3, RVMODEL_MTIME_ADDRESS)
+      #if UDB_MXLEN == 32
+        srli    T4, T2, 27                        // timeh differs from time in CSR address bit 7
+        andi    T4, T4, 1
+        slli    T4, T4, 2
+        add     T3, T3, T4
```
davidharrishmc: Explain what this is doing to get high word of MTIME when reading timeh.

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+    li      T5, CAUSE_ILLEGAL_INSTRUCTION     // the custom hook may use T5 as scratch
+    #ifdef S_SUPPORTED
+      // Illegal instructions were removed from medeleg. Forward an unhandled
+      // lower-mode trap to S-mode as if hardware delegation had occurred.
+      csrr    T1, mstatus
+      LI(     T4, MSTATUS_MPP)
+      and     T3, T1, T4
+      beq     T3, T4, invisible_Mnot_handled_in_M
+
+      csrr    T4, mepc
```
davidharrishmc, line 133: Important logic, worth some comments about what is involved in forwarding to S-mode.

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+      LI(     T4, MSTATUS_MPP)
+      csrc    mstatus, T4
+      LI(     T4, MPP_SMODE)
+      csrs    mstatus, T4
+      csrr    T1, stvec
+      andi    T1, T1, -4                        // exceptions use the stvec base in either mode
+      csrw    mepc, T1
+      // Return through the saved M-mode trap state after installing S-mode state.
+      SREG    zero, rvmodel_sv_off(sp)           // clear the fast-handler handoff marker
+      LA(     T4, resto_Mrtn)
```
davidharrishmc, line 162: Is it too far to do j resto_Mrtn?

tests/env/rvtest_invisible_trap_handler.h
```diff
@@ -0,0 +1,260 @@
...
+  #endif
+
+  invisible_Mnormal_trap:
+    // Continue through the regular trap handler.
+    SREG    zero, rvmodel_sv_off(sp)           // clear the fast-handler handoff marker
+    LA(     T4, invisible_Mcontinue)
+    jr      T4
+
+  // Write the emulation result in T3 to the trapped instruction's rd.
+  invisible_Mwrite_gpr:
```
davidharrishmc, line 195: Explain logic more.

davidharrishmc: How do compressed instructions reliably identify rd, whose location may vary?

davidharrishmc (changes requested):
This is subtle and interesting.  Looks solid, but hard to rule out weird corners. Most of my comments are related to adding developer documentation to make it easier to maintain, and user documentation to facilitate adding an invisible trap handler.  TIME is treated as a special case.  It would be helpful to have an example of an invisible trap handler that the user could adapt; TIME could do that role if it is clear how the invisible handler would relate (maybe pull TIME out as a RVTEST macro with the same registers as RVMODEL_INVISIBLE_TRAP_HANDLER?)

tests/env/rvtest_trap_handler.h
jordancarlin: The code needs to be a constant size with or without invisible traps.

tests/env/rvtest_trap_handler.h
jordancarlin: This was an existing label. Agreed that it should be renamed.

tests/env/rvtest_invisible_trap_handler.h
jordancarlin: I already removed it from the tests, but looks like I missed a spot in Sm_coverage.svh. Fixed.

tests/env/rvtest_invisible_trap_handler.h
jordancarlin: Depending on the DUT, it can be.

tests/env/rvtest_failure_code.h
```diff
@@ -2281,6 +2281,8 @@
         .ascii "\nRVCP-SUMMARY: TEST FAILED - Test File \""
         .ascii TEST_FILE
         .asciz "\"\n"
+    invisible_Minvalid_action_str:
+        .string "RVCP ERROR: Invalid invisible trap action code: "
```
davidharrishmc, line 2285: Hard for an outsider to know what an invalid invisible trap action code of 42 means.  Is the point that the error should be in the range of 0-2?

davidharrishmc:
Looks like Whisper rv64-max now fails on  ExceptionsSm_medeleg_s-00, ExceptionsSm_medeleg_u-00, Sm_xret-00, ZicntrS-00, ZicntrU-00, Sm_mcsr_cntr-00, related to time traps.  Is that intentional?

jordancarlin:
> Looks like Whisper rv64-max now fails on ExceptionsSm_medeleg_s-00, ExceptionsSm_medeleg_u-00, Sm_xret-00, ZicntrS-00, ZicntrU-00, Sm_mcsr_cntr-00, related to time traps. Is that intentional?

It is an issue with traps when we turn delegation off. Working on it.

tests/env/rvtest_failure_code.h
jordancarlin: Yes. The code must be in the range 0-2.

davidharrishmc: Should the message say that?

davidharrishmc:
LGTM once broken traps are recovered.

tests/env/rvtest_setup.h
```diff
@@ -1211,7 +1211,7 @@
     // Delegate exceptions to S-mode, except those that must be directed to M-mode
     // medeleg[0] = 1: delegate instruction address misaligned exception
     // medeleg[1] = 1: delegate instruction access fault exception
-    // medeleg[2] = 1: delegate illegal instruction exception unless M-mode must check for invisible traps
+    // medeleg[2] = 1: logically delegate illegal instruction exceptions to S-mode
```
davidharrishmc, line 1214: Should there be a reference to where invisible traps may override this later?

tests/env/rvtest_trap_handler.h
```diff
@@ -340,6 +343,55 @@
...
+  csrci   CSR_MEDELEG, (1 << CAUSE_ILLEGAL_INSTRUCTION)
+#else
+  nop
+  nop
+  nop
+  nop
+#endif
+.endm
+
+.macro RVTEST_RESTORE_MEDELEG_ILLEGAL SAVE_AREA_REG, TMP_REG
```
davidharrishmc, line 362: No explanation of why we would restore MEDELEG.  This is complicated logic and you are going to be the only one who can follow it right now without substantial documentation of the principles of operation.

davidharrishmc (approved):

tests/env/rvtest_trap_handler.h
jordancarlin: Added more comments explaining these macros and how the medeleg emulation works.

PR #2267, opened 2026-09-08
Update to sail-riscv 0.14.1

jordancarlin:
- Config files updated
- ExceptionsSvZalrsc passing on Whisper


jordancarlin:
The new sail release is out, so this should be good to go assuming it passes CI.

davidharrishmc (approved):

Issue #2146, opened 2026-08-20
Potential privileged issues found by Claude code review


davidharrishmc:
# Privileged review — work queue

Turns `priv-generators-review.md` (133 findings across 8 shards) into a work queue. **71 issues**, every
finding mapped to exactly one; the review stays as the reference document and each issue points back at
its `###` heading.

Regrouped 2026-09-09 into four buckets so the immediately actionable work is separable from the work
that is blocked or in someone else's branch. **Issue numbers are unchanged** — they are stable IDs, not
an order. The old epic is kept as a tag on each line: `vacuous` (produces a passing test a broken DUT
also passes), `wrong-state` (runs, but exercises something other than what it names), `false-failure`
(fails on conforming hardware), `contract` (framework-contract violation or hang), `metadata`,
`refactor`.

| Bucket | Count | What it is |
|---|---:|---|
| [A. Sdtrig](#a--sdtrig-6) | 6 | The whole Sdtrig story, including its missing coverage model |
| [B. Virtual memory](#b--virtual-memory-sv-and-exceptionssv-3) | 3 | Anything whose fix is page tables or `satp` |
| [C. Covered by an open PR](#c--covered-by-an-open-pr-25) | 25 | Being rewritten right now; do not open a competing branch |
| [D. Ready to fix now](#d--ready-to-fix-now-29) | 29 | Independent, unblocked, one branch each |
| [Landed](#landed-5) | 5 | Closed — fixed by a merged PR |
| [Not a bug](#not-a-bug-3) | 3 | Intentional; documented so it is not re-filed |

Line numbers are as of `sm2` @ `b4cb93405`. Several files have been split or renamed since
(`ZkrCommon.py`, `ExceptionsZicboCommon.py`, `PrivCommon.py`), so **anchor on the quoted line or
function name, not the number**.

## Batching rule

An issue is one branch, one PR, one reviewer. Findings were merged when they share a fix site *and* a
verification step, and split when they don't: merged where one root cause has many symptoms (the
`t0`/`x5` bypass closes 5 findings in 4 files, item 48; `a0` being allocatable closes 6 from two shards,
item 49; the WARL exact-compare class closes 6, item 42); split where a bug and its refactor need different
reviewers (item 26 vs items 67, 28 vs items 65, 64 vs item 63).

## Dependencies

- **Items 3 and 4 block item 2.** Re-enabling Sdtrig's seven disabled generators is pointless until the suite
[...]

PR #2278, opened 2026-09-09
SmV: clear vill with the config's smallest supported SEW

davidharrishmc:
Issue #2146 item 36.

_gen_vsetvl_rs2_vill emits one vsetvl per SEW ascending, with no read-back and no branch, so the
vtype entering the measured instruction is always the last attempt: SEW=64, LMUL=1. On an ELEN=32
part that combination is unsupported, so the final vsetvl sets vill instead of clearing it, and all
four cp_vsetvl_rs2_vill bins become unreachable.

Fix: emit a single vsetvl at UDB_SEW_MIN with LMUL=1, selected with #if, which is legal by
construction for any (SEW_MIN, ELEN) pair. Reversing the sweep to end on SEW=8 was the first thing I
tried, but it reintroduces the same bug on a SEW_MIN >= 16 part. The dropped preamble vsetvls lose
nothing; their rs2 values are covered exhaustively by _gen_sew_lmul_vsetvl.

This is latent today. Every vector-capable config in config/ declares ELEN 64 and SEW_MIN 8, so the
existing preamble happens to end on a legal SEW. There is no coverage delta to look for.

https://claude.ai/code/session_01QMw4mDyFq6c1c2dTtfR7yF

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/SmV.py
+++ b/generators/testgen/src/testgen/priv/extensions/SmV.py
@@ -288,11 +288,21 @@ def _gen_vsetvl_rs2_vill(test_data: TestData, temp_reg: int) -> list[str]:
     lines.append(f"LI(x{msb_reg}, 0x8000000000000000)")
     lines.append("#endif")
     for sew_name, sew_v in _SEW_VALUES:
-        # clear vill: try every supported sew with lmul=1 to find one that sticks
-        for try_sew in (0, 1, 2, 3):
-            vt = try_sew << 3
-            lines.append(f"LI(x{rs2_reg}, 0x{vt:02x})  # try clear vill with SEW={try_sew}")
-            lines.append(f"vsetvl x{temp_reg}, x{rs1_reg}, x{rs2_reg}")
+        # Clear vill with the config's smallest supported SEW and lmul=1, which is legal
+        # everywhere. A sweep of several SEWs would not work: each vsetvl overwrites the
+        # previous, so only the final vtype survives, and ending on an SEW the part does
+        # not implement leaves vill set and makes the coverpoint's vtype_prev_vill_clear
+        # term unreachable.
+        lines.append("#if UDB_SEW_MIN == 8")
+        lines.append(f"LI(x{rs2_reg}, 0x00)  # SEW=8, LMUL=1")
+        lines.append("#elif UDB_SEW_MIN == 16")
+        lines.append(f"LI(x{rs2_reg}, 0x08)  # SEW=16, LMUL=1")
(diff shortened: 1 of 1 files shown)


UmerShahidengr (approved):

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -288,11 +288,21 @@ def _gen_vsetvl_rs2_vill(test_data: TestData, temp_reg: int) -> list[str]:
...
-        # clear vill: try every supported sew with lmul=1 to find one that sticks
-        for try_sew in (0, 1, 2, 3):
-            vt = try_sew << 3
-            lines.append(f"LI(x{rs2_reg}, 0x{vt:02x})  # try clear vill with SEW={try_sew}")
-            lines.append(f"vsetvl x{temp_reg}, x{rs1_reg}, x{rs2_reg}")
+        # Clear vill with the config's smallest supported SEW and lmul=1, which is legal
+        # everywhere. A sweep of several SEWs would not work: each vsetvl overwrites the
+        # previous, so only the final vtype survives, and ending on an SEW the part does
+        # not implement leaves vill set and makes the coverpoint's vtype_prev_vill_clear
+        # term unreachable.
```
jordancarlin: Trim down comment. No need to explain why a different implementation doesn't work, just focus on the version that does work.

jordancarlin (changes requested):

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -288,11 +288,20 @@ def _gen_vsetvl_rs2_vill(test_data: TestData, temp_reg: int) -> list[str]:
...
     lines.append("#endif")
     for sew_name, sew_v in _SEW_VALUES:
-        # clear vill: try every supported sew with lmul=1 to find one that sticks
-        for try_sew in (0, 1, 2, 3):
-            vt = try_sew << 3
-            lines.append(f"LI(x{rs2_reg}, 0x{vt:02x})  # try clear vill with SEW={try_sew}")
-            lines.append(f"vsetvl x{temp_reg}, x{rs1_reg}, x{rs2_reg}")
+        # Clear vill with the config's smallest supported SEW and LMUL=1, a combination
+        # every part implements.
+        lines.append("#if UDB_SEW_MIN == 8")
```
jordancarlin:
```suggestion
        # Clear vill with the config's smallest supported SEW and LMUL=1
        lines.append("#if UDB_SEW_MIN == 8")
```

jordancarlin (approved):

PR #2281, opened 2026-09-09
ExceptionsF: make csrrs/csrrc frm actually set and clear a bit

davidharrishmc:
Issue #2146 item 14.

add_csr_instructions runs csrrs and csrrc on frm with frm_reg, which is only ever loaded with 0, so
the set clears nothing and the clear sets nothing. frm stays 0 for the entire suite and the
read-modify-write path is untested at every mstatus.FS value. The tell is the two LI(x{check_reg},
...) lines above them, which are dead because check_reg is only the destination.

Copying the fcsr and fflags siblings is not the right fix: check_reg is loaded with -1, which would
write frm = 0b111, a reserved rounding mode. That is almost certainly why the original author routed
rs1 to a zero register, and it walks into the separate item 41. Instead frm_reg is loaded with RMM
(0b100) and the dead LIs dropped, so the sequence is frm = 0 | 4 = 4, read back as 4, then
frm = 4 & ~4 = 0, read back as 0. frm is never left in a reserved state.

Independent of #2277 and #2279, which touch the load and store helpers in the same file.

https://claude.ai/code/session_01QMw4mDyFq6c1c2dTtfR7yF

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/ExceptionsF.py
+++ b/generators/testgen/src/testgen/priv/extensions/ExceptionsF.py
@@ -237,14 +237,14 @@ def add_csr_instructions(
         "",
         tsbi_call(f"csrc mstatus, x{clear_mask_reg}"),
         tsbi_call(f"csrs mstatus, x{set_mask_reg}"),
-        f"LI(x{check_reg}, -1)",
+        # frm is 0 here, from the csrw above, so this sets it to RMM
         test_data.add_testcase(f"csrrs_frm_{fs_val}_{coverpoint[2:]}", coverpoint, covergroup),
         f"csrrs x{check_reg}, frm, x{frm_reg}",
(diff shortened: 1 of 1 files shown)


UmerShahidengr (approved):

generators/testgen/src/testgen/priv/extensions/ExceptionsF.py
```diff
@@ -274,6 +274,10 @@ def add_csr_instructions(
     return t_lines
 
 
+# rs1 for the csrrs/csrrc frm read-modify-write cases: RMM (0b100), a legal rounding mode.
+_FRM_RMW_BIT = 0b100
```
jordancarlin: This is only used one place, so seems simpler to just inline it.

jordancarlin (changes requested):

jordancarlin (approved):

PR #2286, opened 2026-09-09
Tsbi ssstateen

UmerShahidengr:


coverpoints/priv/Smstateen_coverage.svh
```diff
@@ -294,8 +294,36 @@ covergroup Smstateen_cg with function sample(ins_t ins);
...
+        }
+    `else
+        se0_zero: coverpoint ins.current.csr[CSR_MSTATEEN0H][31] {
+                bins se0_disabled  = {1'b0};
+        }
+    `endif
+
     cp_csr_illegal_accesses: cross priv_mode_maybes_u, mstateen_csrs, csrops;
     cp_walking_ones: cross priv_mode_m, mstateen_walk_csrs, csrops, csr_walk;
+    cp_mstateen0_se0_zero_controls_sstateen0: cross csrops, priv_mode_s, se0_zero, sstateen0_csr;
```
davidharrishmc: Seems like these could be combined to two coverings if se0 is also combined to have the default 0 and 1 bins rather than just 0, just 1.

generators/testgen/src/testgen/priv/extensions/Smstateen.py
```diff
@@ -24,9 +24,9 @@
 
 
 # RVTEST mode-switch macros, emitted as plain assembly (see other priv generators).
-GOTO_UMODE = "RVTEST_GOTO_LOWER_MODE Umode  # enter U-mode"
-GOTO_SMODE = "RVTEST_GOTO_LOWER_MODE Smode  # enter S-mode"
-GOTO_MMODE = "RVTEST_GOTO_MMODE  # return to M-mode"
+GOTO_UMODE = "RVTEST_TSBI_GOTO_UMODE  # enter U-mode"
```
davidharrishmc: How about just using these command directly instead of giving them alternate names.

generators/testgen/src/testgen/priv/extensions/Smstateen.py
```diff
@@ -557,6 +557,75 @@ def _generate_fcsr_lower_fp_instrs(test_data: TestData) -> list[str]:
...
+            f"csrr x{save_sstateen}, sstateen0  # save sstateen0",
+            f"LI(x{ones_reg}, -1)",
+            f"csrr x{save_mstateen}, mstateen0  # save mstateen0",
+            "#if __riscv_xlen == 32",
+            f"csrr x{save_mstatenh}, mstateen0h  # save mstateen0h on RV32",
+            "#endif",
+        ]
+    )
+    lines.extend(_write_se0(temp_reg, enable=bool(se0)))
+    lines.append(GOTO_SMODE)
```
davidharrishmc: Does this need to be gated by #ifdef S_SUPPORTED?

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
```diff
@@ -20,14 +21,13 @@
 CSR_OPS = ["csrrw", "csrrs", "csrrc", "csrr"]
 
 
-# RVTEST mode-switch macros, emitted as plain assembly
-GOTO_UMODE = "RVTEST_GOTO_LOWER_MODE Umode  # enter U-mode"
-GOTO_SMODE = "RVTEST_GOTO_LOWER_MODE Smode  # enter S-mode"
-GOTO_MMODE = "RVTEST_GOTO_MMODE  # return to M-mode"
+# T-SBI mode-switch macros, emitted as plain assembly
+GOTO_UMODE = "RVTEST_TSBI_GOTO_UMODE  # enter U-mode"
```
davidharrishmc: Use inline rather than alias

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
davidharrishmc: S_SUPPORTED must be true if Ssstaten is supported.  Why is there a priv_mode_maybes_u?
Same question for many tests.

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
```diff
@@ -284,7 +236,7 @@ def _generate_jvt(test_data: TestData) -> list[str]:
                         insn,
                     ]
                 )
-            lines.append(GOTO_MMODE)
+            lines.append(GOTO_SMODE)
```
davidharrishmc: I'm having trouble finding the GOTO_UMODE that forces a matching GOTO_SMODE here.  
Same question for many tests.

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
```diff
@@ -448,17 +400,16 @@ def _generate_fcsr_lower_fp_instrs(test_data: TestData) -> list[str]:
     "Ssstateen",
     required_extensions=["Ssstateen"],
     march_extensions=["Ssstateen", "Smstateen", "Zcmt", "Zfinx"],
```
davidharrishmc: Why is Smstateen needed?

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
```diff
@@ -448,17 +400,16 @@ def _generate_fcsr_lower_fp_instrs(test_data: TestData) -> list[str]:
...
 def make_ssstateen(test_data: TestData) -> list[TestChunk]:
     """Generate tests for Ssstateen state-enable extension testsuite."""
     test_chunks: list[TestChunk] = []
     tc = test_data.begin_test_chunk()
 
-    # Unconditional coverpoints — required by all Ssstateen targets
-    tc.code.extend(_generate_se0_controls_sstateen0(test_data, se0=0))
-    tc.code.extend(_generate_se0_controls_sstateen0(test_data, se0=1))
+    # Unconditional coverpoints — required by all Ssstateen targets.
+    # mstateen0.SE0 gating of sstateen0 lives in the Smstateen suite: its subject is an
```
davidharrishmc: Drop this comment.  Is test plan updated to match coverpoints?

davidharrishmc (changes requested):

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
davidharrishmc: These mstateen0 accesses seem to be from S-mode and will trap rather.

davidharrishmc: Also executed in S mode, undesired trap.

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
```diff
@@ -210,7 +162,7 @@ def _generate_walking_ones(test_data: TestData) -> list[str]:
     )
     lines.extend(csr_walk_test(test_data, ("sstateen0", 0x7), covergroup, coverpoint))
 
-    lines.append(GOTO_MMODE)
+    lines.append(GOTO_SMODE)
```
davidharrishmc: Aren't we already in S mode from 155?

coverpoints/priv/Smstateen_coverage.svh
davidharrishmc: gate coverpoint with SSSTATEEN_SUPPORTED because otherwise ssstateen0 may not exist.

UmerShahidengr: Done. One `se0` coverpoint with `se0_disabled`/`se0_enabled`, one cross `cp_mstateen0_se0_controls_sstateen0`. Also renamed the testcase coverpoint to match

generators/testgen/src/testgen/priv/extensions/Smstateen.py
UmerShahidengr: Yes. The SE0 tests drop to S-mode. The whole chunk is now gated

generators/testgen/src/testgen/priv/extensions/Smstateen.py
UmerShahidengr: Deleted the GOTO_* constants; RVTEST_TSBI_GOTO_{U,S,M}MODE now appear inline

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
UmerShahidengr: Dropped all three `#ifdef S_SUPPORTED` guards, and removed the guard from `priv_mode_maybes_u` in the covergroup so both bins are unconditional now. Thanks for pointing that out

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
UmerShahidengr: Dropped

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
UmerShahidengr: Testplan has already been updated

coverpoints/priv/Ssstateen_coverage.svh
```diff
@@ -24,11 +24,10 @@ covergroup Ssstateen_cg with function sample(ins_t ins);
     sstateen_walk_csr: coverpoint ins.current.insn[31:20] {
             bins sstateen0 = {CSR_SSTATEEN0};
     }
+    // Ssstateen implies S-mode, so both bins are always reachable.
     priv_mode_maybes_u: coverpoint {ins.prev.mode_virt, ins.prev.mode} {
```
davidharrishmc, line 28: There is already a priv_mode_s_u in RISCV_coverage_standard_coverpoints.svh that does this, so I think you can drop this coverpoint definition and rely on it.

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
```diff
@@ -20,103 +21,61 @@
...
-# (label, line that switches into the mode, whether an S_SUPPORTED guard is needed).
-# GOTO_MMODE returns to M-mode afterward.
+# Mode dispatch for the coverpoints that sample in both S- and U-mode. Each entry is
+# (label, line entering the mode, line returning to S-mode). Ssstateen implies S-mode, and
+# the suite boots there, so the S-mode leg needs no switch at all.
 _LOWER_MODES = [
-    ("smode", GOTO_SMODE, True),
-    ("umode", GOTO_UMODE, False),
+    ("smode", "", ""),
+    ("umode", "RVTEST_TSBI_GOTO_UMODE", "RVTEST_TSBI_GOTO_SMODE"),
```
davidharrishmc, line 29: Confirm this is intentional to drop the TRUE/FALSE and have two commands in umode.

davidharrishmc (approved):

davidharrishmc:
- Other than minor comments above, looks good to me. Merge when resolved.

generators/testgen/src/testgen/priv/extensions/Ssstateen.py
jordancarlin: If it is intentional, then it might be easier to just inline this and do a check in the Python for if umode.

UmerShahidengr: Yes, intentional. The third field was `needs_guard` from when the suite booted to M-mode. It now boots to S-mode and `Ssstateen` implies S, so the S leg needs no switch or guard. Inlined as suggested.

PR #2288, opened 2026-09-09
T-SBI support for Sscounterenw

UmerShahidengr:


generators/testgen/src/testgen/priv/extensions/Sscounterenw.py
```diff
@@ -19,29 +19,20 @@ def _generate_scounteren_tests(test_data: TestData) -> list[str]:
         comment_banner(
             coverpoint,
             "Set and clear each bit individually in scounteren.\n"
-            "Must execute in S-mode so that priv_mode_s is sampled in the cross.",
+            "Runs in S-mode, which the suite boots into, so priv_mode_s is sampled in the cross.",
```
jordancarlin: Drop the second line of the comment now that it is not needed to justify the `RVTEST_GOTO_LOWER_MODE Smode` line.

davidharrishmc (approved):

jordancarlin (approved):

PR #2290, opened 2026-09-09
SmV, SmVF: commit what each testcase actually produced

davidharrishmc:
Issue #2146 item 6. Neither generator read back any state, so every testcase checked nothing despite docstrings promising checks. Each instruction now commits the state it changes: `mstatus`, the CSR after each walking-1s write, rd and `vtype` after every `vsetvl`/`vsetvli`/`vsetivli` (including the rs2-vill "expect vl=0" case), `vstart`, `vl`, and `fflags`. Each vset testcase gets a priming vset with the same SEW and LMUL so `cp_sew_lmul_vset_i_vli`, which bins from the previous instruction, stays at 28 bins. `misa` is checked only after `csrs` (bit 21), because whether V can be cleared is implementation-defined; the FS-nonaffecting case commits `fflags` rather than `mstatus.FS`; the trapping VS-off and FS-off cases get no readback.

### DIFF
--- a/generators/testgen/src/testgen/priv/extensions/SmV.py
+++ b/generators/testgen/src/testgen/priv/extensions/SmV.py
@@ -8,7 +8,8 @@
 
 """SmV privileged test generator: vector CSRs and vtype/vl/vstart behavior in M-mode."""
 
-from testgen.asm.helpers import comment_banner
+from testgen.asm.csr import gen_csr_read_sigupd, gen_csr_write_sigupd
+from testgen.asm.helpers import comment_banner, write_sigupd
 from testgen.data.state import TestData
 from testgen.data.test_chunk import TestChunk
 from testgen.priv.registry import add_priv_test_generator
@@ -16,6 +17,19 @@ from testgen.priv.registry import add_priv_test_generator
 _CG = "SmV_cg"
 
 _VS_MASK = 3 << 9  # mstatus.VS = bits [10:9]
+
+
+def _check_vset(rd_reg: int, check_reg: int, test_data: TestData) -> list[str]:
+    """Commit what a vset* produced: the vl written to rd, and the resulting vtype.
+
+    Given the config's VLEN, ELEN and SEW_MIN the spec fully determines both, so they are
+    exact checks rather than WARL noise: an illegal request yields vill=1 with the rest of
+    vtype zero, and vl follows from the AVL and VLMAX.
+    """
+    return [
+        write_sigupd(rd_reg, test_data),
+        gen_csr_read_sigupd(check_reg, ("vtype", None), test_data),
+    ]
 
 
 def _set_vs(vs: int, temp_reg: int) -> list[str]:
@@ -43,20 +57,23 @@ def _gen_vcsrrswc(test_data: TestData, temp_reg: int) -> list[str]:
         ),
     ]
     lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
-    save_reg = test_data.int_regs.get_register()
+    save_reg, check_reg = test_data.int_regs.get_registers(2)
     lines.append(f"LI(x{save_reg}, -1)  # all 1s mask for csr ops")
     for csr in _VECTOR_CSRS:
         for op_name, op in (("csrrs", "csrs"), ("csrrc", "csrc"), ("csrrw", "csrw")):
             lines.append(test_data.add_testcase(f"{csr}_{op_name}", coverpoint, _CG))
             lines.append(f"{op} {csr}, x{save_reg}  # {op_name} {csr}")
-    test_data.int_regs.return_registers([save_reg])
+            # Read the CSR under test back. vl, vtype and vlenb are read-only, so the
+            # write traps and this shows the CSR kept its value.
+            lines.append(gen_csr_read_sigupd(check_reg, (csr, None), test_data))
+    test_data.int_regs.return_registers([save_reg, check_reg])
     return lines
 
 
 def _gen_vcsrs_walking1s(test_data: TestData, temp_reg: int, test_chunks: list[TestChunk]) -> None:
     """cp_vcsrs_walking1s: csrrw with walking-1s rs1 against each writable vector CSR."""
     coverpoint = "cp_vcsrs_walking1s"
-    walk_reg, mask_reg = test_data.int_regs.get_registers(2)
+    walk_reg, mask_reg, check_reg = test_data.int_regs.get_registers(3)
     for idx, csr in enumerate(_VECTOR_CSRS_WR):
         tc = test_data.new_test_chunk(test_chunks, "vcsr")
         if idx == 0:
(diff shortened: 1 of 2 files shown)


generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -187,15 +214,21 @@ def _gen_sew_lmul_vset_i_vli(test_data: TestData, temp_reg: int) -> list[str]:
         comment_banner(coverpoint, "vsetvli/vsetivli over all (sew, lmul) immediate combos"),
     ]
     lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
-    rs1_reg = test_data.int_regs.get_register()
+    rs1_reg, check_reg = test_data.int_regs.get_registers(2)
     lines.append(f"LI(x{rs1_reg}, 1)  # vl = 1")
     for sew_name, _ in _SEW_VALUES:
         for lmul_name, _ in _LMUL_VALUES:
+            # The coverpoint reads SEW/LMUL from ins.prev.insn, so prime it with a matching
```
fourth-bit, line 221: This seems like a coverpoint error, not something that should be accommodated in testgen.

davidharrishmc: Agreed, and fixed at the coverpoint in c0f99d9ae rather than in testgen.

`cp_sew_lmul_vset_i_vli` crosses `vset_i_vli_instructions`, which samples `ins.current.insn`, with `vset_sew` and `vset_lmul`, which sampled `ins.prev.insn`. The SEW and LMUL being written to vtype are immediate fields of the vsetvli/vsetivli under test, so they belong to the current instruction. The sibling `cp_sew_lmul_vsetvl` already reads `ins.current.rs2_val`, so the two halves of the same idea disagreed.

```systemverilog
-    vset_lmul: coverpoint ins.prev.insn[22:20] {
+    vset_lmul: coverpoint ins.current.insn[22:20] {
-    vset_sew: coverpoint ins.prev.insn[25:23] {
+    vset_sew: coverpoint ins.current.insn[25:23] {
```

With that corrected the priming vsets in `_gen_sew_lmul_vset_i_vli` are unnecessary and are gone, along with the comment explaining why they were there:

| | before | after |
|---|---|---|
| `vsetvli`/`vsetivli` in SmV_vset-00.S | 352 | 296 |
| lines in SmV_vset-00.S | 2717 | 2661 |
| `SmV_cg`, sail-rv64-max and sail-rv32-max | 100% | 100% |

The other `ins.prev` coverpoints in the file are untouched: `vsetivli_prev_instruction` and `vsetivli_prev_sew_lmul` serve `cp_vl_walking1s_sew_lmul`, which measures a `csrrw` to `vl` and genuinely needs the preceding `vsetivli`.

`make lint` is clean.

fourth-bit (approved):
LGTM

davidharrishmc:
Closed the SmVF coverage hole too: **SmVF_cg goes from 90.38% to 100%** on sail-rv64-max and sail-rv32-max (3d5b75273).

The 7 missing bins were all in `cp_vectorfp_mstatus_fs_state_nonaffecting` (4/8 bins) and `cp_vectorfp_mstatus_fs_off` (1/4 bins), and they had a single cause. Both crosses include `vs1_zero` and `vs2_zero`:

```systemverilog
vs1_zero : coverpoint ins.current.vs1_val == 0;
vs2_zero : coverpoint ins.current.vs2_val == 0;
```

which compare the **whole** vector register. The generator already swept all four zero/nonzero combinations, but the fills ran under the `vl=4, tu` configuration the measured instruction wants, so `vmv.v.i v2, 0` only zeroed the first four elements and the tail kept whatever it held. `v2` never read as zero, so the `auto[1]` half of each coverpoint was unreachable.

The fills now run at VLMAX and then restore `vl=4`:

```asm
vsetvli xN, x0, e32, m1, ta, ma  # vl = VLMAX so the fills cover whole registers
vfmv.v.f v1, f1
vmv.v.i  v2, 0
...
vsetivli xN, 4, e32, m1, tu, mu  # vill=0, vl=4
```

Since `_load_v_zero_one` now leaves that configuration itself, the `_vector_setup` call that preceded it at each of its five call sites was redundant and is gone.

Combined state of this PR: `SmV_cg` and `SmVF_cg` are both at 100% on sail-rv64-max and sail-rv32-max. `make lint` is clean.

jordancarlin:
@fourth-bit can you confirm that this covers all of the parts of #1577 that we wanted to include?

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -43,20 +56,23 @@ def _gen_vcsrrswc(test_data: TestData, temp_reg: int) -> list[str]:
         ),
     ]
     lines.extend(_set_vs(vs=3, temp_reg=temp_reg))
-    save_reg = test_data.int_regs.get_register()
+    save_reg, check_reg = test_data.int_regs.get_registers(2)
     lines.append(f"LI(x{save_reg}, -1)  # all 1s mask for csr ops")
     for csr in _VECTOR_CSRS:
         for op_name, op in (("csrrs", "csrs"), ("csrrc", "csrc"), ("csrrw", "csrw")):
             lines.append(test_data.add_testcase(f"{csr}_{op_name}", coverpoint, _CG))
             lines.append(f"{op} {csr}, x{save_reg}  # {op_name} {csr}")
```
jordancarlin, line 64: Why are op and op_name different?

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -43,20 +56,23 @@ def _gen_vcsrrswc(test_data: TestData, temp_reg: int) -> list[str]:
...
-    test_data.int_regs.return_registers([save_reg])
+            # Read the CSR under test back. vl, vtype and vlenb are read-only, so the
+            # write traps and this shows the CSR kept its value.
+            lines.append(gen_csr_read_sigupd(check_reg, (csr, None), test_data))
```
jordancarlin: This may fail for `vstart` because `vstart` is WARL and can hold arbitrary values. There is no guarantee that it will actually read back all ones.

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -71,19 +87,22 @@ def _gen_vcsrs_walking1s(test_data: TestData, temp_reg: int, test_chunks: list[T
             tc.code.append(f"csrc {csr}, x{mask_reg}  # clear all bits")
             tc.code.append(test_data.add_testcase(f"{csr}_bit_{i}", coverpoint, _CG))
             tc.code.append(f"csrw {csr}, x{walk_reg}  # walking-1 bit {i}")
+            tc.code.append(gen_csr_read_sigupd(check_reg, (csr, None), test_data))
```
jordancarlin: Same issue with vstart in the walking ones test.

generators/testgen/src/testgen/priv/extensions/SmV.py
```diff
@@ -145,12 +165,18 @@ def _gen_misa_v(test_data: TestData, temp_reg: int) -> list[str]:
...
+    # No signature word here: whether misa.V can be cleared is implementation-defined
+    # (Sail keeps V set, QEMU clears it), so the value after the clear is not determined.
     lines.append(test_data.add_testcase("misa_v_csrrs", coverpoint, _CG))
     lines.append(f"csrs misa, x{temp_reg}")
-    lines.append("nop")
+    # After the set, V must read 1 on any machine that implements V, whether or not the
+    # clear above took effect. temp_reg already holds the bit-21 mask.
+    lines.append(gen_csr_read_sigupd(check_reg, ("misa", 0x200000), test_data, temp_reg))
+    test_data.int_regs.return_registers([check_reg])
     return lines
```
jordancarlin, line 180: Seems like this belongs in the regular Sm suite where we test `misa` more generally. I would drop it from SmV.

generators/testgen/src/testgen/priv/extensions/SmVF.py
```diff
@@ -115,16 +109,26 @@ def _gen_fs_state_nonaffecting(test_data: TestData, temp_reg: int) -> list[str]:
...
+        ("v2", "v1", "vs1_0_vs2_n"),  # vs1=0, vs2!=0
+        ("v1", "v4", "vs1_n_vs2_n"),  # both nonzero
+    ]
     for fs in (1, 2):
-        for vs2_reg, vs1_reg, name in _VFADD_PATTERNS:
+        for vs2_reg, vs1_reg, name in pattern_pairs:
             lines.extend(_set_fs_vs(fs=3, vs=3, temp_reg=temp_reg))
-            lines.extend(_vector_setup(temp_reg))
             lines.extend(_load_v_zero_one(temp_reg))
+            lines.append("csrwi fcsr, 0  # clear fcsr under FS=Dirty, as the sibling generator does")
```
jordancarlin, line 124: Sibling generator?

Issue #2258, opened 2026-09-07
[Bug] `vand.vi` alignment causes index collision in unordered indexed store tests (vsuxseg)


hugo-teng:
### Affected area

Tests — Test generation or individual test files

### Commit

bdd226e1

### Affected test or extension

vsuxseg

### Steps to reproduce

1.Build and run Vls16-vsuxseg7ei64.v on hardware that executes unordered vector stores
2.In `indexed_store_vector_type.py`, the index post-processing logic applies `vremu.vx` followed by `vand.vi` to ensure SEW-alignment. However, `vand.vi` halves the effective address space, causing index collisions that lead to memory address overlap in unordered indexed store instructions.
The relevant code (lines 169-198):

```python
if not params.ignore_vector_safety:
    # Ensure that the data in vs2 is within the range [0, 2*vlmax) and sew-aligned
    setup.extend([
        f"vsetvli x{params.temp_reg}, x0, e{params.sew}, m{get_lmul_flag(params.lmul)}, ta, ma",
        f"add x{params.temp_reg}, x{params.temp_reg}, x{params.temp_reg}",  # temp_reg = 2*vlmax
    ])
    # ... vsetvli for index EEW ...
    sew_alignment_factor = -params.sew // 8
    setup.extend([
        f"vremu.vx v{params.vs2}, v{params.vs2}, x{params.temp_reg}",
        f"vand.vi v{params.vs2}, v{params.vs2}, {sew_alignment_factor}",
    ])
```

### Expected behavior

All generated index values should produce **unique memory addresses** after SEW-alignment,
so that each element of an unordered indexed store writes to a distinct location. The test
should pass on any RISC-V Vector implementation regardless of store execution order.

Specifically, for `vsuxseg7ei64` with SEW=16, VLEN=256, LMUL=mf4:
- 4 indices should map to 4 **distinct** even-aligned addresses: e.g. `{0, 2, 4, 6}`

### Actual behavior

**Step-by-step for `vsuxseg7ei64` with SEW=16, LMUL=mf4, VLEN=256:**

1. `vsetvli temp_reg, zero, e16, mf4` → `temp_reg = vlmax = 256×(1/4)/16 = 4`
2. `add temp_reg, temp_reg, temp_reg` → `temp_reg = 8` (i.e., 2×vlmax)
3. `vremu.vx vs2, vs2, temp_reg` → values mapped to `{0,1,2,3,4,5,6,7}` (range `[0, 2×vlmax)`)
[...]

PR #2293, opened 2026-09-10
Fix: scale mismatch index by eew_bytes in failedtest_saveregs

hugo-teng:
## Summary

This PR fixes incorrect diagnostic output from `failedtest_saveregs` when a vector test fails. The actual value extraction was missing index scaling by element width, causing the reported **Bad Value** and **Expected Value** to be unreliable.

## Problem

In `tests/env/rvtest_failure_code.h`, the actual value extraction in
`failedtest_saveregs` (around line 766) used:

```asm
slli x8, x8, 0        # no-op
```

instead of scaling the mismatch index by `eew_bytes`. The comment `already scaled above` was incorrect.

As a result, the diagnostic reads the actual value from the wrong memory offset:

- Expected value: `sigptr + failing_index * eew_bytes` ✅
- Actual value: `vecreg_scratch + vd_offset + failing_index` ❌

For example, with `e16` (2 bytes per element) and `index = 1`:

- Expected reads from `sigptr + 1 * 2 = sigptr + 2` ✅
- Actual reads from `vecreg_scratch + 1` ❌, but should read from
  `vecreg_scratch + 2`

This causes the UART diagnostic to report the same value for both **Bad Value** and **Expected Value**, even though the **Mismatch Mask** shows multiple mismatched elements. The diagnostic becomes misleading and useless for debugging.

### Before fix

```text
RVCP: Element Index: 1
RVCP: Bad Value:      0x0003
RVCP: Expected Value: 0x0003
RVCP: Mismatch Mask: 0x000d000d000d000d000d000d000d000d000d000d000d000d000d000d000d000e
```

## Root Cause

The expected value extraction correctly scales the index:

```asm
mul  x8, x8, x17       # failing_index * eew_bytes
add  x6, x6, x8        # sigptr + failing_index * eew_bytes
```

But the actual value extraction does not:

```asm
slli x8, x8, 0         # no-op, does not scale
add  x6, x6, x8        # vecreg_scratch + vd_offset + failing_index
```

`x17` already contains `eew_bytes`, computed earlier from the SEW encoding.

## Fix

Replace the no-op shift with the correct scaling:

```diff
         # offset by mismatch index
-        slli x8, x8, 0            # already scaled above
+        mul  x8, x8, x17          # failing_index * eew_bytes
         add  x6, x6, x8
```

This makes the actual value extraction use the same element granularity as the expected value extraction.

## Expected behavior after fix

```text
RVCP: Element Index: 1
RVCP: Bad Value:      0x0017
RVCP: Expected Value: 0x0003
RVCP: Mismatch Mask: 0x000d000d000d000d000d000d000d000d000d000d000d000d000d000d000d000e
```

### DIFF
--- a/tests/env/rvtest_failure_code.h
+++ b/tests/env/rvtest_failure_code.h
@@ -763,7 +763,7 @@
         add  x6, x7, x6
 
         # offset by mismatch index
-        slli x8, x8, 0                    # already scaled above
+        mul  x8, x8, x17                  # failing_index * eew_bytes
         add  x6, x6, x8
 
         # store SEW-length expected value bytewise


jordancarlin:
@fourth-bit can you please review this one?

fourth-bit:
This doesn't seem quite right to me. `x8` already holds the scaled value, so this would scale it twice which is not the desired behavior, unless I'm mistaken. Can you attach a trace so we can look into this more?

hugo-teng:
Hi @fourth-bit , thanks for checking — you're right, `x8` already holds the scaled value, so the `slli` there is intentional. I'm withdrawing that part of the change (and reverting it on my side too).

The reason I went looking there is a symptom I hit. I am not sure what is behind it, so here is just what I captured on hardware (`Vls16-vsuxseg7ei64.v-00`, via GDB):

```
# UART output (framework failure report)
RVCP-SUMMARY: TEST FAILED - Test File "Vls16-vsuxseg7ei64.v-00.S"
RVCP: Test Info: "test: 92; cg: Vls16_vsuxseg7ei64.v_cg; cp: cp_masking_edges; bin: cp_masking_edges_maskval_PresetMask.ONES"
RVCP: Instruction: 0xc5027327
RVCP: Approximate address (failure may be slightly after this): 0x0000000080007e24
RVCP: Register: v6
RVCP: Element Index: 1
RVCP: VL: 0x0000000000000004
RVCP: VTYPE: 0x000000000000008e
RVCP: Bad Value:      0x0003
RVCP: Expected Value: 0x0003
RVCP: Mismatch Mask: 0x000d000d000d000d000d000d000d000d000d000d000d000d000d000d000d000e

# test code
80007e24: vsuxseg7ei64.v v6,(x4),v16,v0.t    # failing instruction
80007e28: vluxseg7ei64.v v24,(x4),v16,v0.t   # read the same locations back

# registers at that point (16-bit elements)
v6  = {1,3,7,0}     # store data source, v6[1] = 0x3
v24 = {1,1,1,1}     # value read back, v24[1] = 0x1
v31 = {0xe,0,0,0}   # mismatch mask: elements 1,2,3
v16 = {0,2,6,4}     # indices
```

The rest of the log is only the GDB load/detach output.

Both fields print `0x3`: that is the value being stored, while the value read back at the same element is `0x1`. Having the two fields identical is what makes a failing case look like a pass.

I have not worked out why the reference and the store source coincide here, so I am not proposing anything for it in this PR.  Should I open a separate issue to track this? Happy to run whatever checks you would like.

Thanks again for the review.

fourth-bit:
Let's open an issue to track this

jordancarlin:
Could this be caused by #2258 (which is fixed in #2348)?

fourth-bit:
The problem is in the error reporting, so I think its something in the sigupd or failure code

jordancarlin:
Got it. I saw that it was the same test but didn't look much further than that.

jordancarlin:
Closing now that #2369 is merged instead.

PR #2304, opened 2026-09-10
Zama16b: check the value each AMO returns, and make amocas actually swap

davidharrishmc:
Issue #2146 item 16. `_generate_amo_tests` used `dest_reg` as both the AMO's rd and the read-back scratch register, so the old-memory value in rd was never checked, `amocas` never swapped (its compare operand held stale data), and `amocas.q` left the odd half of its source pair uninitialized. rd is now signed immediately after the AMO, `amocas` rd is preloaded with the scratch-pattern value at that offset (sign-extended for `amocas.w`) so the swap happens, and the upper half of the 16-byte source pair is initialized. No extra register is needed, which matters because the privileged register set has none spare.


davidharrishmc:
@Ammarahwakeel could you review?

Ammarahwakeel:
Looking good to me! Both coverage and tests are updated now.

Issue #2147, opened 2026-08-20
Potential unprivileged issues found by Claude code review


davidharrishmc:
# Unprivileged review — issue decomposition

Turns `unpriv-review.md` (108 verdict lines, ~101 distinct defects) into a work queue.

**59 issues in 6 epics.** Addressed items have moved to the end so the remaining work is visible at a glance; their numbers are unchanged, since they are stable IDs. Every verdict line in the review maps to exactly one issue; the review
stays as the reference document and each issue links back to its `###` heading.

## Batching rule

An issue is **one branch, one PR, one reviewer**. Findings were merged when they share a fix
site *and* a verification step, and split when they don't:

- **Merged** where two shards found the same defect from opposite ends — `amocas`/`rdval`
  (A5 + B2), `cp_misalign` FS (B1 + B3), `MisalignD` FLEN (B1 + B3), the mask SIGUPD reading one
  register (B4 + B6), RMM ties (A4 + B3), `vstart` never written (A8 + B6).
- **Merged** where one root cause has many symptoms — the dead `INTEGER_EDGES` table closes four
  findings across three shards.
- **Split** where a finding needs two different fixes in two files — e.g. the `Zbkb` gaps split
  into "missing instructions" (CSV rows) and "missing coverpoint columns" (different review step).
- **Batched by class, not by file**, for the ~500 norm-YAML citation findings; per-file issues
  would be 30+ issues of near-identical one-line edits.

## Ordering and dependencies

Only two hard dependencies:

- **#4 (MisalignD FLEN) blocks #3 (`cp_misalign` FS float load)** — the float load needs a correct
  `flen` or the data section emits `.word` for a value read as 8 bytes.
- **#17 (wire up `INTEGER_EDGES`) should land before #38, #43, #59** — those testplan edits want
  the `_w` edge variant to exist.

Everything else is independent. Epic 1 items are the ones that change *test outcomes*, so they
want the closest review; Epic 6 is documentation-only.

---

## Epic 1 — Vacuous tests: the test runs but cannot fail (7 issues)

Highest priority. Each of these currently produces a passing test that a broken DUT also passes.

- [ ] 5. **~1,440 testcases write their result to x0** — `make_rs1`/`make_rs2`/`cp_uimm`/`cp_fs1`/
   `cp_fs2`/`cp_offset` omit the `exclude_regs=[0]` their siblings pass. Includes 19 compressed
[...]

PR #2312, opened 2026-09-11
Say why the reserved FENCE encodings write no signature

davidharrishmc:
Issue #2147 item 6.

`I-fence-00.S` declares `SIGUPD_COUNT 10` and writes no signature, which reads like an oversight. It
is not: the `rd` and `rs1` fields of FENCE are reserved for finer-grain fences in future extensions
and base implementations ignore them, so these encodings produce no architectural result to capture.
The only failure they can exhibit is a trap, which the framework already detects.

Comment only; generated output is unchanged.

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/special/cp_custom_fence.py
+++ b/generators/testgen/src/testgen/coverpoints/special/cp_custom_fence.py
@@ -7,9 +7,58 @@
 
 """cp_custom_fence coverpoint generator."""
 
+from testgen.asm.helpers import write_sigupd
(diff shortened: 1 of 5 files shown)


jordancarlin (changes requested):
How about actually doing a sigupd of rd for the fences with rd != x0 to ensure it does not write to rd?

davidharrishmc:
Added sigupds.  They don't check much.  I think this was a waste of time.  If you don't like the sigupds, let's just close this PR as unimportant.

jordancarlin (approved):

PR #2314, opened 2026-09-11
Derive FLEN from the same name components as MARCH

davidharrishmc:
Issue #2147 item 4. Blocks item 3.

`get_flen_for_extension` looked up the whole suite name, so `MisalignD` missed `FLEN_64_EXTENSIONS`
and computed FLEN=32 while its MARCH — built from the *parsed* name components — declared
`rv32ifd`. On RV32 that desynced two things:

- The data section was sized `max(xlen, flen)=32` and emitted `.word` entries, but the runtime
  advances the data pointer by `SIG_STRIDE`, keyed on `TEST_FLEN` from the march string, so it
  strides 8. Later loads walked off the table into the testcase string block.
- `write_sigupd` allocated 2 signature slots per value instead of 3, leaving `SIGUPD_COUNT` short by
  9 — it survived only on the 10-word margin. This was the only such undercount in the repo.

Deriving FLEN from the same `re.findall` components MARCH uses makes them agree by construction, and
covers any future `Misalign*` form rather than needing another name added to a set.

Across all 71 unpriv testplans this changes exactly one suite. Only the two rv32 `MisalignD` files
regenerate: data section becomes 4 `.dword` entries, `SIGUPD_COUNT` 28 → 37. MisalignD and MisalignF
pass 4/4 on spike-rv64-max and spike-rv32-max — they passed before as well, since the reference model
supplies the signature, so a test reading the wrong data still passes.


jordancarlin:
Superseded by #2366.

PR #2315, opened 2026-09-11
Load the FP store operand as a floating point register

davidharrishmc:
Issue #2147 item 3. Stacked on #2314 (item 4) — review that first.

The misaligned FP-store tests loaded the value into `x{r2}` and then stored `f{r2}`, so what reached
memory was whatever the boot code left in that FP register rather than the `0x0f1e2d3c...` pattern
chosen to differ from the bytes already in `scratch`. 36 testcases across MisalignF and MisalignD
stored an unintended value, and the intended load was dead code.

The `rs2` load moves into each per-type branch so the FS branch can load a float at the suite's FP
width. S, CS and CSS keep the integer load in the same position and their output is byte-identical,
which is the check that the move was position-preserving.

Only the four FS files regenerate. Misalign, MisalignD, MisalignF and MisalignZca pass 20/20 on
spike-rv64-max and 13/13 on spike-rv32-max.

Depends on #2314: the FP load width comes from the suite's FLEN, which is wrong for MisalignD until
that lands.


jordancarlin:
Superseded by #2366.

PR #2316, opened 2026-09-11
Use a fixed 6-bit immediate for CBP-type c.andi

davidharrishmc:
Issue #2147 item 23.

CBP is only `c.andi`, whose immediate is a fixed 6-bit signed field, `[-32,31]` on both RV32 and
RV64. The config declared `imm_bits="xlen_log2"`, which resolves to 5 on RV32, confining RV32 random
immediates to `[-16,15]`. It also set `imm_nonzero=True`, but `c.andi` reserves nothing — `imm=0` is
a legal encoding that clears the register (UDB: `X[xd] = X[xd] & $signed(imm)`, no constraint).

Both settings are correct for CBS, the shift type `c.srli`/`c.srai`, where the immediate really is an
`xlen_log2` shift amount and `shamt=0` is a HINT. They were copied to CBP, which shares neither
property. Of the five types setting `imm_nonzero`, CBP was the only one whose instruction has no
architectural nonzero requirement — `c.addi4spn`, `c.lui` and `c.addi16sp` reserve a zero immediate,
and the shift types make it a HINT.

Two files regenerate, RV32 only; RV64 already resolved `xlen_log2` to 6 and is byte-identical. Random
immediates move from `[-16,13]` to `[-32,27]`; the 156 testcases under `cp_imm_edges` and
`cr_rs1_imm_edges_6bit` pin their immediates and already spanned the full range. Zca passes 32/32 on
spike-rv64-max and 26/26 on spike-rv32-max.

Thanks @copilot-pull-request-reviewer for catching `imm_nonzero` — it is the same copy-paste as the
width. Worth noting for the record that it changes no generated output today: no random draw in the
current stream lands on 0, so the re-roll never fired, and `imm=0` is already exercised 16 times per
file by the edge cross. The value is preventing a future seed or edit from silently suppressing a
legal encoding in the random path.

### DIFF
--- a/generators/testgen/src/testgen/formatters/types/cbp_type.py
+++ b/generators/testgen/src/testgen/formatters/types/cbp_type.py
@@ -13,9 +13,8 @@ from testgen.formatters.registry import InstructionTypeConfig, add_instruction_f
 cbp_config = InstructionTypeConfig(
     required_params={"rs1", "rs1val", "immval"},
     reg_range=range(8, 16),
-    imm_bits="xlen_log2",
+    imm_bits=6,  # c.andi immediate is a fixed 6-bit signed field on RV32 and RV64
     imm_signed=True,
-    imm_nonzero=True,
 )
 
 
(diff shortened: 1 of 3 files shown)


generators/testgen/src/testgen/formatters/types/cbp_type.py
```diff
@@ -13,9 +13,8 @@
 cbp_config = InstructionTypeConfig(
     required_params={"rs1", "rs1val", "immval"},
     reg_range=range(8, 16),
-    imm_bits="xlen_log2",
+    imm_bits=6,  # c.andi immediate is a fixed 6-bit signed field on RV32 and RV64
```
jordancarlin:
```suggestion
    imm_bits=6,
```

jordancarlin (approved):

Issue #1221, opened 2026-04-07
Zicbom improvements


davidharrishmc:
cbo.inval tests could be improved.

Right now, cbo.inval does not check the result because it is nondeterministic.  
A better approach would be to write a known pattern, cbo.clean, write a new 
pattern that differs in one bit in each word, cbo.inval, read back the values,
and mask off the bit before writing to signature so that either the old or new
value is acceptable.

In some priv modes, controlled by m/senvcfg.CBIE, invalid becomes flush, and
this should be tested that only the new value is read back.

davidharrishmc:
Resolved with #2370.

PR #2320, opened 2026-09-11
Check that cbo.zero actually zeroes memory

davidharrishmc:
Issue #2147 item 2.

Load the first byte back after a cbo.zero and check that it is actually zero.

### DIFF
--- a/generators/testgen/src/testgen/formatters/types/cbo_type.py
+++ b/generators/testgen/src/testgen/formatters/types/cbo_type.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: Apache-2.0
 ##################################
 
+from testgen.asm.helpers import write_sigupd
 from testgen.data.params import InstructionParams
 from testgen.data.state import TestData
 from testgen.formatters.registry import InstructionTypeConfig, add_instruction_formatter
@@ -25,4 +26,12 @@ def format_cbo_type(
     test = [
         f"{instr_name} (x{params.rs1}) # perform operation",
     ]
-    return (setup, test, [""])
+    # cbo.zero must zero the block holding rs1; the other CBOs may legally be no-ops, so they have no check.
+    # rs1 is the load destination so no extra register is needed; setup reloads it for each testcase.
+    check = [""]
+    if instr_name == "cbo.zero":
+        check = [
+            f"lbu x{params.rs1}, 0(x{params.rs1}) # read back a byte inside the zeroed block; must be 0",
+            write_sigupd(params.rs1, test_data, "int"),
+        ]
+    return (setup, test, check)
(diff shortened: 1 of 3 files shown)


jordancarlin:
@davidharrishmc is this intentionally still a draft?

davidharrishmc:
Nope.

generators/testgen/src/testgen/formatters/types/cbo_type.py
```diff
@@ -25,4 +26,12 @@ def format_cbo_type(
...
     ]
-    return (setup, test, [""])
+    # cbo.zero must zero the block holding rs1; the other CBOs may legally be no-ops, so they have no check.
+    # rs1 is the load destination so no extra register is needed; setup reloads it for each testcase.
+    check = [""]
+    if instr_name == "cbo.zero":
+        check = [
+            f"lbu x{params.rs1}, 0(x{params.rs1}) # read back a byte inside the zeroed block; must be 0",
+            write_sigupd(params.rs1, test_data, "int"),
+        ]
```
jordancarlin, line 37: We know the cache block size from `UDB_CACHE_BLOCK_SIZE`. Is it worth checking the full cache block instead of just a single byte? And possibly even slightly more than the size of the cache block to ensure it didn't zero more than the expected amount of memory.

generators/testgen/src/testgen/formatters/types/cbo_type.py
```diff
@@ -25,4 +26,12 @@ def format_cbo_type(
     test = [
         f"{instr_name} (x{params.rs1}) # perform operation",
     ]
-    return (setup, test, [""])
+    # cbo.zero must zero the block holding rs1; the other CBOs may legally be no-ops, so they have no check.
+    # rs1 is the load destination so no extra register is needed; setup reloads it for each testcase.
+    check = [""]
```
jordancarlin, line 31: #1221 suggests a better way of checking the other cache block instructions instead of having no check. We can defer that to a different PR if you prefer, but since you are here anyway, it might make sense to do now.

jordancarlin (changes requested):

generators/testgen/src/testgen/formatters/types/cbo_type.py
davidharrishmc: Let's defer to another PR (in progress, targeting ExceptionsZicboS/U priv tests that already toggle xenvcfg.CBIE.

generators/testgen/src/testgen/formatters/types/cbo_type.py
davidharrishmc: cp_custom_cbo checks the full scratch area that the proper parts got zeroed.  I don't think we need to repeat that for all the rs1 cases.

However, I've added something to write a byte to rs1 before the cbo so we do confirm each cbo at least clears that byte for all 31 rs1 choices.

jordancarlin (approved):

PR #2321, opened 2026-09-11
Exclude x0 as the CSR scratch register in CSR and CSRI tests

davidharrishmc:
Issue #2147 item 9.

`csr_type.py` and `csri_type.py` use `params.rs2` as the scratch register for both the CSR seed (`csrrw x0, <csr>, x{rs2}`) and the read-back (`csrrs x{rs2}, <csr>, x0`). x0 was not excluded, so a testcase that drew rs2 = x0 seeded nothing and signed a constant 0.

**Fix:** `excluded_regs={"rs2": {0}}` on both configs. `instructions/params.py` applies operand exclusions only when the operand is unset, so explicitly requested x0 bins still work.

**Measured:** collapsed `csrrw x0, <csr>, x0` testcases 27 -> 2 (the issue's 81 counts emitted asm lines: `zicsr_access` emits one per preprocessor branch — fflags/vxsat/mepc). The 2 that remain are the genuine `cmp_rd_rs1` b0 bin (rd = rs1 = x0) in rv32i/rv64i `Zicsr-csrrw`, and their seed line now uses a real register.

**Blast radius:** 12 generated files change, all Zicsr (rv32i and rv64i x csrrw/csrrwi/csrrs/csrrsi/csrrc/csrrci); SIGUPD_COUNT is unchanged everywhere. Zicntr and Zihpm do list Type `CSR`, but their only coverpoint `cp_cntr` allocates its own registers and never goes through the CSR config's parameter generation, so they are untouched.

### DIFF
--- a/generators/testgen/src/testgen/formatters/types/csr_type.py
+++ b/generators/testgen/src/testgen/formatters/types/csr_type.py
@@ -10,7 +10,8 @@ from testgen.data.params import InstructionParams
 from testgen.data.state import TestData
 from testgen.formatters.registry import InstructionTypeConfig, add_instruction_formatter
 
-csr_config = InstructionTypeConfig(required_params={"rd", "rs1", "rs1val", "rs2", "rs2val"})
+# rs2 is the scratch register used to seed and read back the CSR, so it must not be x0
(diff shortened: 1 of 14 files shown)


jordancarlin:
Replaced by #2367. Closing.

PR #2322, opened 2026-09-11
Skip the destination register check when rd is x0 for c.j and c.jr

davidharrishmc:
Issue #2147 item 8.

`cp_imm_edges_jal.py` correctly passes `rd=0` for `c.j` (it is `c.jal` with rd hard-wired to x0), but the destination-register check called `write_sigupd(params.rd, ...)` unconditionally, signing 21 constant-zero words per test file (10 forward alignment bins + 11 backward). `cp_offset.py` has the same pattern for `c.jr`, which sets `rd=0` deliberately because it has no link register: 5 more constant-zero words per file (1 in the main negative-offset sequence, 4 in `cp_offset_lsbs`).

**Fix:** guard both checks with `if params.rd != 0`. For `c.jr` the `auipc`/`sub` that compute the position-independent return address are guarded with the signature write, since they are dead without it.

**Confirmed not to fire on the link-register cases:** `c.jal` uses `rd=1` and `c.jalr` uses `rd=1` (both in `cp_offset` and in `cp_offset_lsbs`), so their checks are untouched — neither `Zca-c.jal-00.S` nor `Zca-c.jalr-00.S` changes. `jalr` (Type JR) also draws a random rd that could in principle be x0; it happens not to be in any generated file, so `I-jalr` and `E-jalr` are unchanged too.

**Measured:** 8 files change (c.j and c.jr for rv32i/rv32e/rv64i/rv64e). SIGUPD_COUNT: c.j 51 -> 30 (-21) on all four, c.jr 51 -> 46 on rv32i/rv64i and 35 -> 30 on rv32e/rv64e (-5). 104 constant-zero signature words removed in total; no `RVTEST_SIGUPD` with check register x0 remains in any of the 8 files.

**Consequence worth knowing:** the c.j forward-align-1 and backward-align-(-1) bins already skip their jump-taken self-check because the offset is too small to fit the inline check, so with this guard they emit no signature word at all and become trap-only. Nothing real is lost — their only current assertion is that x0 reads zero.

### DIFF
--- a/generators/testgen/src/testgen/coverpoints/cp_imm_edges_jal.py
+++ b/generators/testgen/src/testgen/coverpoints/cp_imm_edges_jal.py
@@ -74,8 +74,12 @@ def make_cp_imm_edges_jal(instr_name: str, instr_type: str, coverpoint: str, tes
                 write_sigupd(params.temp_reg, test_data)
                 if not skip_check
                 else f"{INDENT}# offset too small, skipping self-check",
-                f"{INDENT}# Check destination register",
-                write_sigupd(params.rd, test_data),
+                # c.j hardwires rd to x0, so its link check would sign a constant 0
+                *(
+                    [f"{INDENT}# Check destination register", write_sigupd(params.rd, test_data)]
+                    if params.rd != 0
+                    else []
+                ),
                 "",
             ]
         )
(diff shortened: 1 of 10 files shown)


jordancarlin:
I don't see the issue that this solves. While checking that xd=0 isn't very meaningful, it also doesn't hurt and doesn't seem worth special casing just to avoid a few signature entries. Closing.

PR #2328, opened 2026-09-11
F/D/Zfh: load floating point edge values with flw, fld, and flh

davidharrishmc:
Issue #2147 item 45. FP loads never loaded a chosen bit pattern, so the `*_bits_maintained` rules for `flw`/`fld`/`flh` were untested. A new `cp_fd_edges` coverpoint (with `_D`/`_H` variants and matching `.sv` templates) passes each `FLOAT_EDGES` entry as `temp_fval`, enabled on the `flw`, `fld` and `flh` rows of `F.csv`, `D.csv` and `Zfh.csv`. `fld_bits_maintained` in `coverpoints/norm/D.yaml` now maps to `D_fld_cg/cp_fd_edges_D` instead of `TODO`.


davidharrishmc:
Copilot feedback unimportant - don't need to reference every coverpoint in norm rules

PR #2333, opened 2026-09-11
Add exact-tie FP operands so RMM is distinguishable from RNE

davidharrishmc:
Addresses #2147 item 19. RMM and RNE differ only on an exact tie, and `FLOAT_EDGES` has no exactly-halfway operand, so no current test distinguishes them. This adds per-format `FLOAT_EDGES.ties_*` (`+-0.5`, `+-2.5`, `+-2^p`) and `tie_partners_*` (`+-1.0`) tables, a linear `cp_fs1_edges_ties` on `fcvt.<int>.<fp>` and `fsqrt.*`, and `cr_fs1_fs2_edges_ties` on `fadd.*`, `fsub.*`, `fmul.*`, `fdiv.*` and the `fmadd` family (with `fs3` pinned to `1.0`), all sweeping only RNE and RMM. The ties are kept out of the main edge tables because `cr_fs1_fs2_edges` grows quadratically with them; this way F, D and Zfh gain 1,440 testcases in total. For `fmul`/`fdiv`/`fsqrt` the `+-2^p`/`+-1.0` pair is not an exact tie, so per-operation tie pairs could follow.


jordancarlin:
I think this will already be covered by the coverfloat tests. Is there any value in having this too?

davidharrishmc:
Rather err on the side of complete.  It’s not a ton of tests.

generators/testgen/src/testgen/data/edges.py
```diff
@@ -391,6 +391,53 @@ class FLOAT_EDGES:
...
+    # +-0.5 and +-2.5 are float-to-integer ties; +-2^p (p = significand bits) is an
+    # arithmetic tie when combined with +-1.0 from tie_partners_*.
+    ties_single = (
+        0x3F000000,  # 0.5
+        0xBF000000,  # -0.5
+        0x40200000,  # 2.5
+        0xC0200000,  # -2.5
+        0x4B800000,  # 2^24
+        0xCB800000,  # -2^24
+    )
```
jordancarlin, line 405: Rather than adding new coverpoints, I'd lean towards modifying the existing edge values. If we don't want to extend the list of existing values too much, we can consider modifying some of them instead. For example, it is probably fine to replace the existing 1.5 with 2.5.

PR #2334, opened 2026-09-11
Test that fflags accrues rather than being overwritten

davidharrishmc:
Issue #2147 item 46.

`fflags` accrues (sticky OR) rather than being overwritten by an FP operation,
and that was untested. All 732,247 emitted `fsflagsi` were `0b00000`, so every
FP operation started from a cleared state. The covergroup could not observe it
either: `templates/cp_csr_fflags_voun.sv` uses transition bins between
consecutive samples, so its `NV1` bin (`5'b1???? => 5'b1????`) only ever means
"two consecutive NV-raising tests", never "the flag was already set and
survived".

## Approach

A new `cp_custom_fflags_accrue` coverpoint on four rows — `feq.s`, `flt.s`
(F) and `feq.d`, `fle.d` (D). Each testcase:

1. pre-sets `fflags` via `fsflagsi` to one of `0b00001`, `0b00010`, `0b00100`,
   `0b01000`, `0b01111` — all NV-free,
2. runs the comparison against a signaling NaN, which raises only NV,
3. lets the existing `RVTEST_SIGUPD_FFLAGS` in the check sequence capture the
   result.

The pre-set flags and the raised flag are disjoint, so "the operation raised
it" and "it was already set" cannot be confused. The covergroup bins the
post-operation `fflags` on the five accrued values (`5'b10001` … `5'b11111`);
none of them is reachable unless the pre-set bits survive.

The comparison rows were chosen so this change does not collide with the
rounding-tie work in issue #2147 item 19, which touches the arithmetic and
conversion rows.

## No re-goldening

The shared `"fsflagsi 0b00000 # clear all fflags"` line in the 15 scalar FP
formatters is untouched — its occurrence count across `tests/` is identical
before and after. The only changes are 40 appended testcases (10 per suite per
XLEN) and the two regenerated covergroup files.

`RVTEST_SIGUPD_F`/`RVTEST_SIGUPD_FFLAGS` reads `fflags` with `csrr` and does
not clear it, so the new coverpoint composes with the existing fflags
signature entry rather than duplicating it.


jordancarlin:
This seems like an attempt at fixing the issue raised in #2347, but it does it by leaving the problematic coverpoints in place and only testing an arbitrary subset of instructions (based on what wouldn't conflict with other PRs which is not a strong reason). I'm going to close this and I think we need a more holistic solution to this. I will open another PR with a fix soon.

PR #2341, opened 2026-09-11
ebreak straddling 64B boundary

davidharrishmc:
CVW generates the wrong xtval on this test, so it is worth having in the suite.


davidharrishmc:
Needs updated CVW #1852 before this will pass CI.  This has merged into cvw.  Passing and ready to review.

jordancarlin (approved):

PR #2344, opened 2026-09-11
Select test CSR for Zicsr tests with priv mode fallback

### LINKED ISSUE #1572: Zicsr tests fail to compile for U-mode cores without F, V, or Zicntr

jordancarlin:
Fixes #1572
Replaces #1935

### DIFF
--- a/generators/testgen/src/testgen/formatters/types/csri_type.py
+++ b/generators/testgen/src/testgen/formatters/types/csri_type.py
@@ -18,7 +18,7 @@ def zicsr_accessi(instr_name: str, rd: int, rs2: int, immval: int) -> str:
     """Helper function to determine which CSR to use for testing based on supported extensions."""
     # instret requires special treatment because it is not writable, and the value is not initialized
     if instr_name in ["csrrw", "csrrwi"] or rs2 == 0 or rd == 0:
-        read_only_access = f"li x{rd}, 0 # avoid write to read-only CSR, or inconsistent result with rs2 or rd = 0"
+        read_only_access = f"li x{rd}, 0x42 # avoid write to read-only CSR, or inconsistent result with rs2 or rd = 0"
     else:
         read_only_access = (
             f"{instr_name} x{rs2}, RVTEST_TEST_CSR, 0\n"
--- a/generators/testgen/src/testgen/formatters/types/csri_type.py
+++ b/generators/testgen/src/testgen/formatters/types/csri_type.py
@@ -16,31 +16,21 @@ csri_config = InstructionTypeConfig(required_params={"rd", "rs2", "rs2val", "imm
 
 def zicsr_accessi(instr_name: str, rd: int, rs2: int, immval: int) -> str:
     """Helper function to determine which CSR to use for testing based on supported extensions."""
-    # Use writable unprivileged extension CSRs if any exist,
-    # else use mepc if U is not supported
-    # else use instret (which is not writable, but at least can be accessed)
-
     # instret requires special treatment because it is not writable, and the value is not initialized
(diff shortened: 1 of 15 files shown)


generators/testgen/src/testgen/formatters/types/csri_type.py
```diff
@@ -16,31 +16,21 @@
...
 def zicsr_accessi(instr_name: str, rd: int, rs2: int, immval: int) -> str:
     """Helper function to determine which CSR to use for testing based on supported extensions."""
-    # Use writable unprivileged extension CSRs if any exist,
-    # else use mepc if U is not supported
-    # else use instret (which is not writable, but at least can be accessed)
-
     # instret requires special treatment because it is not writable, and the value is not initialized
     if instr_name in ["csrrw", "csrrwi"] or rs2 == 0 or rd == 0:
-        instret_access = f"li x{rd}, 0 # avoid write to read-only instret, or inconsistent result with rs2 or rd = 0"
+        read_only_access = f"li x{rd}, 0 # avoid write to read-only CSR, or inconsistent result with rs2 or rd = 0"
```
davidharrishmc: Should this be some magic number other than 0 just in case a CSR is ROZ?

generators/testgen/src/testgen/formatters/types/csri_type.py
```diff
@@ -16,31 +16,21 @@
 
 def zicsr_accessi(instr_name: str, rd: int, rs2: int, immval: int) -> str:
     """Helper function to determine which CSR to use for testing based on supported extensions."""
-    # Use writable unprivileged extension CSRs if any exist,
-    # else use mepc if U is not supported
-    # else use instret (which is not writable, but at least can be accessed)
-
     # instret requires special treatment because it is not writable, and the value is not initialized
```
davidharrishmc: still says instret

generators/testgen/src/testgen/formatters/types/csri_type.py
jordancarlin: I had just left it as it was before, but I agree that a non-zero value would be better. Updated.

tests/env/utils.h
```diff
@@ -225,6 +225,23 @@
...
+// Determine the appropriate CSR to test based on the
+// supported extensions and set boot mode if necessary.
+#if defined(F_SUPPORTED)
+  #define RVTEST_TEST_CSR fflags
+#elif defined(ZVE32X_SUPPORTED)
+  #define RVTEST_TEST_CSR vxsat
+#elif defined(ZICNTR_SUPPORTED) && defined(U_SUPPORTED)
+  #define RVTEST_TEST_CSR instret
+  #define RVTEST_READ_ONLY_TEST_CSR
+#elif defined(S_SUPPORTED)
```
davidharrishmc, line 234: Why isn't this a higher priority than a read-only CSR?

davidharrishmc (approved):

davidharrishmc (approved):

tests/env/utils.h
jordancarlin: Currently, we prioritize booting to the standard mode for unpriv tests (the lowest supported mode). Using `sepc` requires stopping the boot at S-mode instead. Both approaches has pros and cons. Happy to switch the order if you think having a writable CSR is more valuable.

davidharrishmc: My personal feeling is that if S exists, it's better to test a writable CSR.  I've approved, so you can make the call and merge when ready.

jordancarlin: Updated to prioritize writable CSRs.

PR #2349, opened 2026-09-13
Vector Floating Point in Main Testgen

fourth-bit:
Passes all sims and gets 100% coverage. Includes support for Vf16, Vf32, Vf64, Zvfbfmin, Zvfbfwma, and Zvfhmin.

Small fixes to bf16 testplans (overflow case) and a small fix to coverage for `cp_custom_vfp_NaN_input` (ensures widening operations are respected and `cp_custom_vfncvt_rup_overflow` (coverage failed to build before).


jordancarlin:
@fourth-bit please update this now that #2387 is merged and then I will take a look. Excited to get this one in. Looks like we are close to dropping the old vector testgen!

davidharrishmc:
Some minor AI findings, could be left as an issue for a followup PR after this is merged:

cp_custom_vfncvt_rup_overflow is not tested for Zvfbfmin

vfrec7 sNaN marked as non-flag-setting, but they should raise NV on sNaN

cp_custom_vfncvt_rup_overflow's register_vector_data(label, 64, …) should use 2 * config.sew rather than 64 (affects bf16 path, still functions ok because of little-endian)

fp_mask_prod_type.py is a ~100-line copy of mask_prod_type.py and the two have already diverged (force_full_length_check, no_dot_t, the "vd" in mask_registers assert). Also: the 5-line csr_frm_val assert appears 9 times, fvvfm_config is defined twice in fvvf_type.py, and fvmvfs_type.py asserts maskval is None then has three dead if params.maskval: branches.

 fsrmi 0x0 escapes the #if guard. handle_vector_fp appends to check after handle_parameter_exclusions added #endif, so the restore sits outside the block while fsrmi <frm>/fsflagsi are inside. One-line fix: swap the call order.

unordered_sum_rng returns element_count + 2 elements (both halves add a trailing piece). Harmless but it makes the exactness bound harder to audit; to_fp_words would also raise on a negative shift if a magnitude ever reached 2**mantissa_length

Nits: Zvfhmin missing from get_vector_base_extension's map (its tests come out as march=rv64im_zicsr_zifencei_zvfhmin, no Zve32f/F, unlike its siblings); cp_custom_fmv_fs_vs2_all_lmul lacks the sew > flen guard its sibling has; cp_fp_reg_edges_v.py's docstring/error message are copy-paste from elsewhere; cp_custom_vfp_flags returns [] while 86 testplan rows list it; generate_vector_special_cases adds an extra random test for every vector instruction with an fflags coverpoint (with bin_name = ""), not just the four FMA variants that need the UF case.

fourth-bit:
1. `cp_custom_vfncvt_rup_overflow` is in the testplan for `Zvfbfmin` in the `cp_custom` column. Now it has been modified to `cp_custom_vfncvt_rup_overflow_bf16` so that it can collect coverage, independent of the `VFCUSTOM32` macro
2. Fixed.
3. Fixed.
4. The mask_prod_type code duplication is unfortunate, but the mask producing generator is already complex in how it handles building instructions, as it assembles instructions with register and immediate operands. I didn't want to add more complexity with the floating point cases, so a separate generator was chosen. It could be worthwhile to refactor the duplicated code in a future PR. Fixed the other two issues
5. Fixed the element count issue. The negative shift is impossible to reach in this code (and is also a sign that the desired integer is not in the exactly representable range)
6. `cp_custom_vfp_flags` returns `[]` because it doesn't need any special cases to hit the fflags values it desired. Its unclear as to how this coverpoint is distinct from `cp_csr_fflags`, but deciding the fate of a coverpoint family is not in the scope of this refactor.  Fixed the other nits

davidharrishmc (approved):

PR #2357, opened 2026-09-13
Convert Vector Failure Code to T-SBI

fourth-bit:
The VS mstatus bits need to be enabled for the failure code to work as some exceptions tests include disabling VS.


tests/env/rvtest_failure_code.h
```diff
@@ -348,8 +348,11 @@
 
     #ifdef RVTEST_VECTOR
         # We need to ensure that VS is set in mstatus here, as VS off is an exceptions test
-        LI (x6, 0x600)
-        csrs mstatus, x6
+        # T-SBI call to execute instruction: csrs mstatus, a1
+        LI (a1, 0x600)
+        add a2, x0, x0
+        LI(a0, 0x3005a073) # csrs mstatus, a1
+        ecall # T-SBI call to execute instruction at suitable privilege level
```
jordancarlin: Can this be simplified to use the RVTEST_TSBI_CSR_SET macro?

jordancarlin (changes requested):

jordancarlin (approved):

Issue #13, opened 2018-11-02
compressed tests are using non-compressed instructions


hajekra:
Following macros in test_macros.h are used for testing compressed instructions but do not use it directly:
TEST_CLWSP, TEST_CSWSP, TEST_CADDI16SP, TEST_CADDI4SPN.

eroom1966:
Hi hajekra

As I recall, I could not get the assembler we were using to accept some of the compressed instructions by name, for example c.addi16sp

I have found that constructing an assembler instruction correctly, it will map to the compressed instruction
are you saying that you have been able to assemble code using the mnemonic 
c.addi16sp ?

If so, can you tell me what version of the assembler you used ?
Thx

hajekra:
Hi,

I am using riscv64-unknown-elf-gcc (GCC) 7.1.1 20170509. In asm file I used "c.addi16sp sp, 0x20".

Radek H.

Imperas:
@hajekra 
Thanks for clarification
fixed in commit 57049ee

hajekra:
@Imperas 
Thanks, closing this issue.

PR #2363, opened 2026-09-14
Enable and fix the hypervisor trap handler and T-SBI paths

UmerShahidengr:
`H_SUPPORTED` stays off by default. Fixed while bringing up the new H suite:

- `init_Xsatp` used #elseif, which is not a preprocessor directive. On RV64 the `LI` was dropped entirely and on RV32 it was emitted plus three clobbering copies, so `xSATP.MODE` has always ended up Bare on both. It now says so explicitly. 

- `INSTANTIATE_MODE_MACRO` emitted M, S, H, V while every cross-area offset in the file is written against M, H, S, V. The prolog and epilog call sites in rvtest_setup.h now match, which also settles which of `HS` and `S` owns `stvec` and `sscratch`: they are the same registers, so the last prolog wins and it must be `S`.

- The S-mode dispatch only matched cause 8, so a guest's ecall (cause 10) was recorded as an ordinary trap. A guest could never leave VS-mode, and the `RVTEST_TSBI_GOTO_MMODE` in cleanup_epilogs never reached M-mode.

- `GOTO_SMODE/UMODE` did not clear `hstatus.SPV`, so an `sret` out of a guest went straight back into it. `GOTO_VSMODE/VUMODE` are now serviced in HS-mode rather than forwarded, which also drops the `sepc` limitation the forwarding path documented.

- The S/HS handler reserved six-word exception entries but never wrote words 4 and 5. It now records `htval` and `htinst`, which is the only place a test can observe a faulting guest physical address: every later trap into HS-mode overwrites `htval`.

- The S-mode `xEPC` relocation ignored `hstatus.SPV`, `vsatp` and `hgatp`, so a guest's `xEPC` was relocated against the host's segment bases.

- The instruction-width fetch used a plain lhu at `xEPC`. HS-mode runs with V=0, so at a guest virtual address that resolves through the host's satp: the wrong memory once the guest has a VS-stage map, and a nested fault when it is unmapped in the host. It now uses `hlvx.hu`.

- The hypervisor and VS-mode CSRs were missing from `tsbi_instr_table`, and `RVTEST_BOOT_TO_SMODE` did not initialize `hstatus`, `hideleg`, `hvip`, `hcounteren` or `henvcfg`, none of which are architecturally defined at reset.

The new H suite generates two chunks. `H_hsmode` covers hypervisor CSR access and T-SBI from HS-mode, an hlv/hsv round trip, and entry into VS-mode with a virtual instruction exception and a guest T-SBI call (RV32 and RV64). H_twostage runs a guest under Sv39x4 over Sv39 and covers a load guest-page fault and a non-identity guest view, where one trap uses an instruction whose upper halfword is the illegal compressed encoding so that a mis-sized skip shows up as an extra trap signature entry rather than passing silently (RV64).


generators/testgen/src/testgen/io/templates.py
```diff
@@ -67,6 +67,10 @@ def insert_header_template(
         all_extensions = flat_ext_components
         march = generate_march_string(all_extensions, xlen)
     all_defines = [*(extra_defines or []), *generate_defines_from_extensions(all_extensions)]
+    # Opt in to the hypervisor only for suites that require H. A suite that lists H
+    # in march_extensions just needs the assembler to accept H CSR names.
+    if "H" in flat_ext_components:
+        all_defines.append("#define RVTEST_HYPERVISOR")
```
davidharrishmc, line 73: I'm not following why this is needed?  Is it to save instantiating the hypervisor trap handler when it is not supported?  Can we drop RVTEST_HYPERVISOR and just gate things with H_SUPPORTED where necessary?

generators/testgen/src/testgen/priv/extensions/H.py
```diff
@@ -0,0 +1,386 @@
...
+_HOLE_GIB = 0x40000000
+_ALIAS_GIB = 0xC0000000
+
+# G-stage PTEs are always checked as user accesses, so PTE_U must be set.
+_G_PERMS = "(PTE_V | PTE_R | PTE_W | PTE_X | PTE_U | PTE_A | PTE_D)"
+# VS-stage pages belong to the guest supervisor, so PTE_U is clear.
+_VS_PERMS = "(PTE_V | PTE_R | PTE_W | PTE_X | PTE_A | PTE_D)"
+
+
+def _case(test_data: TestData, bin_name: str, coverpoint: str) -> str:
```
davidharrishmc, line 34: unnecessary wrapper.

tests/env/rvtest_failure_code.h
```diff
@@ -2354,6 +2354,10 @@
         .string "\"Mismatch in mtval2 value! Trap was being handled in M-Mode.\"";
     sv_Mtinst_str:
         .string "\"Mismatch in mtinst value! Trap was being handled in M-Mode.\"";
+    sv_Htval2_str:
```
davidharrishmc, line 2357: htval2->htval?

tests/env/rvtest_trap_handler.h
```diff
@@ -2378,6 +2533,22 @@ skp_\__MODE__\()tval:
         csrr    T3, CSR_MTINST
         TRAP_SIGUPD(T4, T3, 5, sv_\__MODE__\()Mtinst, sv_Mtinst_str) // write word 5: mtinst
       #endif
+  .else
+    .ifnc \__MODE__ , V
+      #ifdef H_SUPPORTED
+        // HS-mode is where a guest's traps land, so htval (the faulting guest
+        // physical address, shifted right by 2) and htinst are the fields that
```
davidharrishmc, line 2540: Do we need a check that accepts all legal values on htval/mtval2, or does Sail have enough parameters for the reference to match the DUT?

tests/env/rvtest_trap_handler.h
```diff
@@ -1966,6 +2045,39 @@ tsbi_\__MODE__\()handle_forwarded:
...
+        TSBI_CSR_INSTR_TABLE(0x604) // hie
+        TSBI_CSR_INSTR_TABLE(0x606) // hcounteren
+        TSBI_CSR_INSTR_TABLE(0x607) // hgeie
+        TSBI_CSR_INSTR_TABLE(0x60A) // henvcfg
+        TSBI_CSR_INSTR_TABLE(0x643) // htval
+        TSBI_CSR_INSTR_TABLE(0x644) // hip
+        TSBI_CSR_INSTR_TABLE(0x645) // hvip
+        TSBI_CSR_INSTR_TABLE(0x64A) // htinst
+        TSBI_CSR_INSTR_TABLE(0xE12) // hgeip
+        TSBI_CSR_INSTR_TABLE(0x680) // hgatp
```
davidharrishmc, line 2063: Is htimedelta needed?

generators/testgen/src/testgen/priv/extensions/H.py
```diff
@@ -0,0 +1,386 @@
...
+        "#else",
+        f"{INDENT}hsv.d x{check}, (x{temp})",
+        f"{INDENT}li x{check}, 0",
+        f"{INDENT}hlv.d x{check}, (x{temp})",
+        "#endif",
+        write_sigupd(check, test_data),
+    ]
+
+
+def _gen_vsmode_tests(test_data: TestData, check: int, temp: int) -> list[str]:
```
davidharrishmc, line 140: Should there be some smoke tests from VU-mode to test TSBI_GOTO_VUMODE, TSBI accesses from VU mode to HS and M mode registers, and 2-stage address translation from VU?

generators/testgen/src/testgen/priv/extensions/H.py
```diff
@@ -0,0 +1,386 @@
...
+
+# Invisible trap emulation cannot yet handle traps from VS or VU mode, so
+# check_defines.h rejects it on a hypervisor build. It is enabled whenever the
+# time CSR is emulated, which is what TIME_CSR_IMPLEMENTED=false means.
+_PARAMS = ["TIME_CSR_IMPLEMENTED: true"]
+
+
+@add_priv_test_generator(
+    "H",
+    required_extensions=["S", "H"],
```
davidharrishmc, line 345: H implies S so S shouldn't be needed.

generators/testgen/src/testgen/priv/extensions/H.py
```diff
@@ -0,0 +1,386 @@
...
+    tc.raw_data.extend([".p2align 3", "H_guest_scratch:", "  .dword 0"])
+    test_chunks.append(test_data.end_test_chunk())
+
+    test_data.int_regs.return_registers([check, temp])
+    return test_chunks
+
+
+@add_priv_test_generator(
+    "H",
+    required_extensions=["S", "H"],
```
davidharrishmc, line 369: no S

tests/env/rvtest_failure_code.h
davidharrishmc: Is there any more state that the failure code should be printing now that H is supported?

tests/env/rvtest_macros.h
```diff
@@ -18,9 +18,13 @@
 #define LEVEL3 0x03
 #define LEVEL4 0x04
 
+// Selector tags for the S-stage/VS-stage translation modes. They are only ever
```
davidharrishmc, line 21: Not sure I understand why these are necessary.  There are already parameters for modes supported.

tests/env/rvtest_setup.h
davidharrishmc, line 365: All of these will need VS and VU equivalents, and the RVTEST macros in utils.  We will also need set/clr vsei/vsti/vssi interrupts and sgeip.  This could wait for a follow-up PR.

davidharrishmc, line 954: Should we initialize mtinst, mtval2 if H_SUPPORTED?

davidharrishmc (commented):
Initial comments. Still reviewing rvtest_setup and trap handler.

tests/env/rvtest_setup.h
```diff
@@ -1296,6 +1306,45 @@
...
+             HSTATUS_VTVM | HSTATUS_VTW | HSTATUS_VTSR | HSTATUS_VSBE)
+      csrc hstatus, t0
+
+      // Delegate nothing to VS-mode by default: a guest trap goes to HS-mode,
+      // where the framework's trap handler is, unless a test opts in by writing
+      // hedeleg/hideleg itself. hedeleg is also cleared by RVTEST_TRAP_PROLOG H,
+      // which saves the incoming value; this keeps hideleg consistent with it.
+      csrw hideleg, zero
+      csrw hvip, zero     // no guest-visible interrupts pending at boot
+
```
davidharrishmc, line 1383: #ifdef SSSTATEEN_SUPPORTED
     #if __riscv_xlen == 64
        li t0, HSTATEEN_SSTATEEN | HSTATEEN0_SENVCFG  # alternate names for SE0 and ENVCFG in encoding.h
        csrs hstateen0, t0  // Set these fields
      #else    // RV32
        li t0, HSTATEENH_SSTATEEN | HSTATEEN0H_SENVCFG   # alternate names for SE0 and ENVCFG in encoding.h
        csrs hstateen0h, t0 // Set these fields
      #endif

      li t0, SSTATEEN0_JVT | SSTATEEN0_FCSR
      csrs hstateen0, t0 // enable access from lower privilege mode
    #endif

tests/env/rvtest_setup.h
```diff
@@ -1296,6 +1306,45 @@
...
+      //   HU    = 0: no hypervisor loads/stores from U-mode
+      //   VGEIN = 0: no guest external interrupt selected
+      //   VTVM  = 0: guest may use sfence.vma and satp
+      //   VTW   = 0: guest wfi does not trap
+      //   VTSR  = 0: guest sret does not trap
+      //   VSBE  = 0: VS-mode is little endian
+      // Tests that need any of these set them themselves.
+      LI(t0, HSTATUS_SPV | HSTATUS_SPVP | HSTATUS_HU | HSTATUS_VGEIN | \
+             HSTATUS_VTVM | HSTATUS_VTW | HSTATUS_VTSR | HSTATUS_VSBE)
+      csrc hstatus, t0
```
davidharrishmc, line 1340: Does vssstatus need any initialization? 
Are hie and vise and hgeie cleared in the trap prologue?
initialize htimedelta to 0
initialize hgatp to 0

tests/env/rvtest_setup.h
```diff
@@ -1296,6 +1306,45 @@
...
+
+      // Delegate nothing to VS-mode by default: a guest trap goes to HS-mode,
+      // where the framework's trap handler is, unless a test opts in by writing
+      // hedeleg/hideleg itself. hedeleg is also cleared by RVTEST_TRAP_PROLOG H,
+      // which saves the incoming value; this keeps hideleg consistent with it.
+      csrw hideleg, zero
+      csrw hvip, zero     // no guest-visible interrupts pending at boot
```
davidharrishmc, line 1376: Do hip/vsip need initialization

davidharrishmc:
We should wait until Sail 0.14.1 is deployed to fully test this in CI before merging.

tests/env/rvtest_trap_handler.h
```diff
@@ -408,26 +408,72 @@
...
+//   sved_hgatp_off  = +1 area (H)   sved_mpp_off = +2 areas (S)
+//   RVTEST_TRAP_EPILOG  reaches its own area at +1 (H), +2 (S), +3 (V)
+//   RVTEST_GOTO_LOWER_MODE reads code_bgn_ptr at +1 (H/U), +2 (S), +3 (VS/VU)
+//   \__MODE__\()trap_sig_sv reaches M's shared trap_sigptr at -1 (H), -2 (S),
+//                            -3 (V) once the -1 area sp bias is included
+//
+// H sits directly above M because the hypervisor CSRs it owns (hgatp, hedeleg)
+// are the M-mode-adjacent half of the S CSR space; S then sits above it so the
+// S prolog runs last and wins the CSRs the two modes genuinely share.
+// Emitting S before H silently points every one of the offsets above at the
```
davidharrishmc: Trim comment

tests/env/rvtest_trap_handler.h
```diff
@@ -1100,17 +1146,15 @@ init_\__MODE__\()edeleg:
...
-        LI(T3, (SATP64_MODE) & (SATP_MODE_SV57 << 60))  // RV64 SV57
-      #endif
+      // MODE is left at Bare. The root page tables are still all-invalid here
+      // (RVTEST_DATA_END only reserves them) and the identity map in
+      // rvtest_identity_map is not built until after the boot chain has already
+      // dropped into S-mode, so enabling translation at this point would fault
+      // on the very next fetch. Tests that want translation enable it
+      // themselves once they have populated a page table.
+        li      T3, 0                             // MODE = Bare
         or      T4, T4, T3                        // combine MODE bits with PPN
```
davidharrishmc: Spec says "Attempting to select MODE=Bare with a nonzero pattern in the remaining fields has an UNSPECIFIED effect..."  I think we should just do csrrw T4, CSR_XSATP, T4 at this point and set up the PPN later.

tests/env/rvtest_trap_handler.h
```diff
@@ -1687,7 +1733,15 @@ tsbi_\__MODE__\()goto_vu:
         LI(T4,(1<<(UDB_MXLEN-1))+((1<<12)-1))        // T4 = int_bit + cause[11:0] mask
         and     T4, T4, T5                        // T4 = masked xcause
         addi    T3, T4, -CAUSE_USER_ECALL          // T3 = masked_cause - 8 (U-mode ecall = cause 8)
-        bnez    T3, \__MODE__\()trapsig_ptr_upd   // not a U-mode ecall -> normal trap sig recording
+        beqz    T3, \__MODE__\()goto_sishere      // U-mode ecall -> T-SBI candidate
```
davidharrishmc: what does sishere mean?

tests/env/rvtest_trap_handler.h
```diff
@@ -1787,9 +1866,9 @@ tsbi_\__MODE__\()forward_to_m:
         ecall                                      // trap to M-mode with a0/a1 intact
         // For CSR_ACCESS: M-mode executes the CSR op and mrets back here in S-mode with the
         //   result in a0; sret returns to the caller (sepc was bumped past its ecall already).
-        // For GOTO_VS/VU (H only): M-mode sets MPP/MPV and mrets back here in the target virtual
-        //   mode; the sret then uses the virtual sepc, which is NOT the caller's address -- known
-        //   limitation, use tsbi_*forward_goto_m-style handling when H support is completed.
+        // GOTO_VS/VU no longer come through here: they are serviced locally above, which also
```
davidharrishmc: don't need to keep history in comments.

davidharrishmc (changes requested):

davidharrishmc:
Finished my review pass. Please get @jordancarlin 's input after updating.

tests/env/rvtest_setup.h
davidharrishmc: We initialize mepc/mtval/mcause.  Do we need to the same with s and vs if they exist, in their boot parts?  If not, then do we really need to init m?

tests/env/rvtest_setup.h
davidharrishmc: Good chance most interrupt regs do not need initialization, but worth thinking through and documenting why not if not.

generators/testgen/src/testgen/io/templates.py
UmerShahidengr: It keeps the HS/VS trap handlers, their prologs and the H boot init out of tests that don't need H. I'd like to keep it until H_SUPPORTED can actually be honoured on every config that declares H, which isn't the case today. I will remove `RVTEST_HYPERVISOR` once we move to Sail 0.14.1 and enables H in the configs; at that point it will be a one-line change.

generators/testgen/src/testgen/priv/extensions/H.py
UmerShahidengr: Added.
H_hsmode (RV32 and RV64): RVTEST_TSBI_GOTO_VUMODE; sscratch access from VU raises a virtual instruction exception (so it really is VU, not U); ECALL_TEST from VU; a T-SBI read of hstatus serviced by the HS handler (reads back SPV=1, SPVP=0); a T-SBI read of mstatus forwarded to M-mode (MPP reads back as S); then GOTO_SMODE and a check that hstatus.SPV was cleared.

H_twostage (RV64): after the VS-mode tests, the VS-stage mappings get PTE_U and the guest drops to VU for a load/store through both stages, an ECALL_TEST (the HS handler has to fetch the instruction width through the guest's translation at VU privilege), and a load guest-page fault.

While doing this I also found that each ecall test's signature word was attributed to the previous testcase, so a mismatch would have named the wrong check. Fixed by creating the testcase label where the check is emitted.

generators/testgen/src/testgen/priv/extensions/H.py
UmerShahidengr: Done, required_extensions is now ["H"] for both generators.

tests/env/rvtest_failure_code.h
UmerShahidengr: The name follows the sv_<mode><field>_str scheme. 
sv_Htval_str already exists and is HS-mode's stval word, so sv_Htval2_str is HS-mode's second trap value, i.e. htval, matching sv_Mtval2_str for mtval2. Renaming it to sv_Htval_str collides with the existing symbol. The message text itself says "htval".

UmerShahidengr: htval and htinst are already covered: they are words 4 and 5 of the trap signature entry. 

For a guest trap, the useful state that isn't printed is hstatus (SPV, SPVP, GVA), which says whether the trap came from VS or VU. Should we add it in signatures too? I cant think of any other state to add as signature

tests/env/rvtest_trap_handler.h
UmerShahidengr: Good question, and I dont know the exact answer.
Currently `htval` is covered by parameters: UDB's `REPORT_GPA_IN_HTVAL_ON_GUEST_PAGE_FAULT` corresponds to Sail's `extensions.H.guest_page_fault_writes_htval`, so the reference matches as long as the config generator maps it.

`htinst` (and the existing `mtinst` recording) is not. UDB has per-cause `TINST_VALUE_ON_*` parameters that allow non-zero values (e.g. a transformed instruction), but Sail 0.14 has no corresponding parameter and always writes 0. Spike and Sail both write 0 on the load guest-page fault in these tests, but a DUT that legally writes a transformed instruction would mismatch.
We need some Sail parameters here

tests/env/rvtest_trap_handler.h
UmerShahidengr: Yes, added `htimedelta` and `htimedeltah`. Thank you. I also added `hstateen0/hstateen0h` and the AIA `hvien, hvictl, hviprio1, hviprio2 and hvienh`, whose RV32 high halves were already listed, and fixed 0x655, which was labelled hvienh but is hviph.

tests/env/rvtest_macros.h
UmerShahidengr: These aren't capability parameters. They are selector values passed as the MODE argument to the page table macros, e.g. SATP_SETUP_RV64(sv39) and VS_PTE_SETUP(sv39, ...), and compared with .if inside them. sv39/sv48/sv57 already existed for the Sv suite. The VS-stage macros in `rvtest_macros_hypervisor.h` also compare against sv32, which was never defined. GNU as silently treats the undefined symbol as 0, but clang's assembler rejects it ("expected absolute expression"), which is what broke the clang CI configs. This change only adds the missing sv32.

tests/env/rvtest_setup.h
UmerShahidengr: Yes to both. Their reset values are UNSPECIFIED, and a test can read one before its first trap into that mode. Boot takes no trap into S or VS, and an M-mode test can read `mtval` before any trap. 
All of them are now zeroed together in `RVTEST_BOOT_TO_MMODE`: mepc/mtval/mcause, sepc/stval/scause, and with H also mtval2, mtinst, htval, htinst, vsepc, vstval and vscause.

UmerShahidengr: Agreed, We will need VS/VU interrupt macros and set/clear for vsei/vsti/vssi and sgeip in a follow-up PR.

tests/env/rvtest_setup.h
UmerShahidengr: Added under SSSTATEEN_SUPPORTED, using the HSTATEEN0_JVT | HSTATEEN0_FCSR names.

tests/env/rvtest_setup.h
UmerShahidengr: No write needed, and it's now documented in `rvtest_setup.h`. hip is a view of mip: VSSIP aliases hvip.VSSIP, and VSTIP/VSEIP are hvip ORed with platform sources. Boot clears mip and hvip. vsip is a view of hip through hideleg, which is zero.

tests/env/rvtest_setup.h
UmerShahidengr: - vsstatus: now initialized the same way as mstatus: all fields zero, UXL = 64 on RV64, and FS/VS dirty when F/V are supported, since guest FP/vector use is gated by both `vsstatus` and `sstatus`. hstatus.VSXL is also set to 64 on RV64.
- hie/vsie/hgeie: the prolog doesn't touch them, but they don't need it. hie is a view of mie, which boot clears, and vsie is a view of hie through hideleg, which is zero. hgeie is a separate register, so it is now cleared explicitly.
- htimedelta (and htimedeltah on RV32): now initialized to 0.
- hgatp: zeroed by RVTEST_TRAP_PROLOG H, which now writes the whole xSATP as 0.

tests/env/rvtest_trap_handler.h
UmerShahidengr: Agreed, done. The prolog now writes xSATP = 0 (csrrw T4, CSR_XSATP, x0). Nothing depended on the PPN it used to install: every test that enables translation writes the whole CSR.

tests/env/rvtest_trap_handler.h
UmerShahidengr: Nothing useful. Renamed to tsbi_candidate.

UmerShahidengr:
@davidharrishmc thanks for the detailed review. I have addressed your comments but kept a few comments open for discussion or to park them as issues for later (e.g, requiring VS/VU interrupt macros in a follow-up PR).

CI is expected to stay red until Sail 0.14.1. But one thing to note here: spike-rv32/64-max and whisper-rv32-max are not enabling H on their command lines, so they will stop at the first H CSR in the trap prolog. Moving to Sail 0.14.1 alone won't be enough; those command lines and each H config's sail.json also need H enabled.

tests/env/rvtest_failure_code.h
davidharrishmc: Those sound useful.

tests/env/rvtest_trap_handler.h
davidharrishmc: skip testing htinst/mtinst

jordancarlin:
@UmerShahidengr now that #2267 is merged you, should be able to update this PR and get CI passing.

generators/testgen/src/testgen/io/templates.py
jordancarlin: Since Sail 0.14.1 is merged, I think this is no longer necessary?

UmerShahidengr: I tried this now that 0.14.1 is in. Removing this gate and honoring
H_SUPPORTED everywhere hits two things, so I'd like to do it as a follow-up.

1. Assembler march. With H_SUPPORTED on for every test, the shared trap handler emits
   hlvx.hu into suites whose MARCH has no h:

     tests/env/rvtest_trap_handler.h:2446: Error: unrecognized opcode `hlvx.hu x7,(x8)',
     extension `h' required   (ZicntrS, and every other non-H suite on an H config)

   Easy fix: add h to the assembler march for all priv tests, the way ZawrsS already does
   with march_extensions=["H", ...]. Harmless on non-H configs because H_SUPPORTED still
   gates what gets emitted. 

2. whisper. whisper-rv32-max and whisper-rv64-max declare H in UDB, but whisper.json never
   enables h, so with the gate gone every test there runs the H boot code on a model without
   H. whisper-rv64-max also trips the invisible-trap #error in check_defines.h; that only
   shows up in the self-check build, since sail_macros.h undefines RVTEST_EMULATE_TIME_CSR
   in the reference build.

   Making whisper's config honest instead cascades: dropping the H extension makes UDB
   validation reject ~60 H parameters that are still in the file (VSXLEN, MUTABLE_MISA_H,
   the TINST_VALUE_ON_* set, REPORT_*_VSTVAL_*, SV*X4_TRANSLATION, ...), which is really a
   whisper-side change, not one for this PR.

   Separately, whisper's H looks incomplete: with h enabled, the mie/mip VS bits (2, 6, 10,
   12) read as zero, which fails Sm_mcsr_access-00, Sm_mcsr_walk-01/02 and Sm_shadow-00.
   That is why H is excluded for whisper in ci.yaml on this branch.

So the gate is not really the blocker, the whisper config is. Once whisper.json enables h and
whisper implements those mie/mip bits, this will become the three-line removal plus the march
change, and I'm happy to do it later.

davidharrishmc:
# PR #2363 AI review

Line numbers refer to 89dc897c2 (the branch after act4 was merged in).

## Fix before merge

1. **No functional coverage for H.** H.py:16 names covergroup `H_cg` and coverpoints (cp_hgatp_mode, cp_tsbi_ecall,
   cp_twostage_access, cp_guest_page_fault, cp_htval, ...) that don't exist. `EXTENSIONS=H make coverage` fails with
   "No coverage entries found in report", so `make regression` fails.
   Fix: add `coverpoints/priv/H_coverage.svh`, `H_coverage_init.svh` and `coverpoints/norm/H.yaml`, or point the
   testcases at the existing ExceptionsH_cg / ExceptionsHV_cg / SvH_cg / EndianH_cg. [DH: ok to skip this - these are temporary test cases]

2. **A legacy `RVTEST_GOTO_SMODE` from VS/VU returns to VS.** `Srtn2smode` (rvtest_trap_handler.h:2864-2870) sets SPP
   but doesn't clear hstatus.SPV. A probe confirmed that callers from both VS and VU land in VS.
   Fix: add `RVTEST_SET_SPV 0, T3` in `Srtn2smode`, or take the a0==0 path only for cause 8 with SPV=0.

3. **M-mode EPC relocation uses hgatp for MPV=0 traps.** rvtest_trap_handler.h:2294-2300 replaces satp with hgatp
   whenever misa.H=1, so a trap from translated HS/U into M is relocated as a PA and aborts if outside the segments.
   Fix: test MPV first. MPV=0 → satp; MPV=1 → vsatp, then hgatp (same shape as the S handler at 2327-2346).

4. **hstatus is only initialized in `RVTEST_BOOT_TO_SMODE`** (rvtest_setup.h:1324-1343). The S handler records
   hstatus[8:6] on every trap (rvtest_trap_handler.h:2246). SPVP has no reset value and is unchanged on V=0 traps
   (norm:hstatus_spvp_op), so a `BOOT_TO_MMODE` H test records reset state.
   Fix: move the hstatus/hideleg/hvip/hgeie/htimedelta/vsstatus init into `RVTEST_BOOT_TO_MMODE` next to the
   mtval2/htval zeroing (rvtest_setup.h:977).

5. **The S/HS handler assumes misa.H stays set.** It always reserves 5 words (2122) and reads hstatus
   (2246, 2337, 2451), vsatp/hgatp (2340-2345) and htval (2508), and executes hlvx.hu (2454). If a test clears
   misa.H, the handler traps recursively.
   Fix: gate these on misa.H at run time as the M handler does (2111-2115, 2489-2498), or #error if H suites
   clear misa.H.

6. **The hstatus overlay in word 0 overwrites SIE/MIE and conflicts with #2429.** rvtest_trap_handler.h:2185 ORs
   hstatus[8:6] into word0[16:14]. Word 0 holds xstatus[17:0] in bits 30:13, so bits 16:14 are MIE/WPRI/SIE.
   #2429 moves the GVA/MPV/SPVP overlay to bits 29:27. The two PRs conflict in rvtest_trap_handler.h and
   trap_report.py.
   Fix: use #2429's layout (bits 29:27) for the S/HS overlay, and update `_decode_xstatus` in trap_report.py to
   match.

7. **The imperas-rv64-max sail.json has the old medeleg mask.** config/imperas/imperas-rv64-max/sail.json:98 is
   still `0x0000_0000_000c_b3af`, so Sail takes the virtual-instruction trap in M while Imperas delegates it, and
   both H tests fail.
   Fix: change it to `0x0000_0000_00fc_b7af` (both tests then pass).

8. **`spp_not_in_hs` always passes.** H.py:198-204 reads sstatus.SPP right after the T-SBI `sret`. xRET always sets
   SPP=0 (norm:mstatus_xret_op).
   Fix: read `vsstatus` from HS and expect SPP=1, or compare sstatus and vsstatus directly.

9. **The hedeleg/hideleg tests skip the required-writable bits.** The comment at H.py:81-84 says the writable bits
   are implementation choice. The spec requires hedeleg 1-8, 12, 13, 15, 18, 19 and hideleg 2, 6, 10 to be
   writable, so a DUT with these registers hard-wired to 0 passes.
   Fix: after writing all ones, also record `readback & required_writable_mask`, excluding bit 0, which depends
   on IALIGN.

10. **H_twostage assumes the image is in the gigapage at 0x80000000.** H.py:24-26 and 279-283 hard-code the
    0x80000000 identity map and the 0xC0000000 alias.
    Fix: derive the gigapage from `rvtest_code_begin` (align down to 1 GiB) and place the alias and hole relative
    to it.

## Smaller fixes

11. **AGENTS.md rules in H.py.**
    - Bare `li` at lines 149, 153, 186, 232, 322, 329, 341, 352, 368, 410 and 419: use `LI()`.
    - Hard-coded `t0` at line 279: take it from a TestData allocator.
    - The `.word 0x000024F3` at line 354 hard-codes rd=x9: add a comment that it always traps.
12. **Wrong comment.** H.py:344 says "reached only if the handler skipped four bytes". It always runs after a
    correct skip.
13. **Coverage gaps in the tests.**
    - hlv/hsv run only with both stages Bare (H.py:131-157), never with SPVP=1 or under two-stage translation.
    - There is no store or instruction guest-page fault test.
    - hstatus.GVA is never checked.
14. **Guest EPC relocation checks the host satp before SPV** (rvtest_trap_handler.h:2327-2331). Branch on SPV
    first, and for a guest with Bare/Bare, relocate against the VS save area (sp+sv_area_sz) as the M handler does.
15. **An htval mismatch reports "unknown field".** rvtest_failure_code.h adds `sv_Htval2_str` (2348) but has no
    compare for it. Add the Htval2 compare (subtype 6, mode 1). trap_report.py labels S/HS word 4 as "MTVAL2";
    label it HTVAL.
16. **trap_report.py:293-294 is wrong for M-mode.** The docstring says bits 1 and 3 read zero on trap entry, which
    is false: an M-mode trap leaves mstatus.SIE set, so "GVA" in M entries is SIE|GVA.
    - On non-H builds, the S handler's word 0 ORs in a stale T4 (the signature pointer), so address bits appear as
      SPVP/SPV/GVA. Use `li T4, 0` in the non-H path.
    - On RV64 non-H, the M handler's `csrr T4, CSR_MSTATUS` (2228) is under `#ifdef H_SUPPORTED`. Make it
      unconditional.
17. **Guest GOTO_S/U and `forward_goto_m` (1864-1865) jump to guest VAs in host context.** They only work with
    identity guest maps. Document this in the T-SBI section, or relocate through the VS save area as rtn2mmode does.
18. **T-SBI CSR access from a guest to an S CSR name (e.g. sstatus 0x100) runs in HS and changes host state**
    (1868-1878). Document that guests must use the vs* names.
19. **Dead code and stale comments.**
    - Remove the dead `.ifc \__MODE__ , H` xTVEC block (rvtest_trap_handler.h:1139-1145).
    - Update the "M, S, H, or V" and "3, 4, or 6 words" (2085) comments.
    - `sved_mpp_off` (318) now points at an unwritten VS-area slot; remove the MPRV read at 2281-2285 or comment it.
20. **CI comments.**
    - qemu ci.yaml: cite QEMU commit 18645f1957 ("do not allow mstatus MPV/GVA writes", v11.1.0) or file a QEMU
      issue.
    - whisper ci.yaml: the H exclusion only has an effect on whisper-rv32-max, because whisper-rv64-max never
      selects H (`TIME_CSR_IMPLEMENTED: false`). Say so, or remove H from the whisper UDB yamls.
21. **PR description.** Update it: entries are five words with htval only (no htinst/mtinst), and the mode order is
    M, S(=HS), V.
22. **Follow-up issues.** Open issues for the items deferred in review:
    - VS/VU interrupt macros, and set/clear for vsei/vsti/vssi/sgeip.
    - Removing `RVTEST_HYPERVISOR` once whisper enables h.
    - Adding `h` to the priv-test MARCH.

davidharrishmc (approved):
Ready for Jordan review

davidharrishmc:
One more round of Claude feedback

davidharrishmc:
- #13 partly done. He added the store guest-page fault and a GVA check, but:
  - The new SPVP=1 round-trip test proves nothing: both translation stages are Bare there, so SPVP has no effect.
  - His reason for skipping the instruction guest-page fault is wrong. He says the handler would loop on the faulting fetch, but the handler returns to ra for fetch faults (rvtest_trap_handler.h:2427-2433), so a jalr ra into the hole works.
- #22: the follow-up issues haven't been filed.

New bug from the fix for #4 (medium): hstateen0 now ends up 0. Moving the hstatus init into BOOT_TO_MMODE also moved the hstateen0 writes (rvtest_setup.h:~1063-1076) to before mstateen0 is set. hstateen0 bits read as zero while the matching mstateen0 bit is 0, so the writes are dropped; Spike logs hstateen0 = 0x0. A guest touching senvcfg, sstateen0, jvt or fcsr would now take a virtual-instruction exception. No current H test does that. The fix is to redo those writes after the mstateen0 writes.

Minor:
- A dead li T6, 0 // no H is left inside #ifdef H_SUPPORTED (~2316).
- One new comment assumes a save slot is zero; the agent didn't check it.
- Interaction with #2430: on H builds, M-mode traps with MPRV=1 (the SvSm tests) now record an unrelocated xEPC, as non-H builds already did. Every Sv and SvSm suite still passes.

config/qemu/ci.yaml
```diff
@@ -26,7 +26,13 @@ ci_enabled: true
 #     on RV32 the IMSIC bit is writable although Sail does not model IMSIC
 # ZawrsU and ZawrsS fail on qemu due to stval mismatching, filed QEMU issue https://gitlab.com/qemu-project/qemu/-/work_items/4080
 # Sdtrig*: Sail does not implement Sdtrig yet (https://github.com/riscv/sail-riscv/issues/222)
-exclude_extensions: "Sm,S,SsstrictSm,SsstrictS,SsstrictU,Zacas,ZacasZabha,ExceptionsZalrsc,ExceptionsZicboU,ExceptionsZicboS,ExceptionsZicboSm,SvZicbo,ExceptionsZaamo,ExceptionsSvZalrsc,Zama16b,SsstrictV,Smstateen,ZawrsS,ZawrsU,SdtrigSm,SdtrigS,SdtrigU"
+# H: mstatus.MPV and mstatus.GVA read back as zero after being written from M-mode
+#    even with misa.H set, so mret never enters VS/VU: a guest's ecall is reported
+#    as cause 9/8 instead of 10, and a hypervisor CSR access from VS raises an
+#    illegal instruction instead of a virtual instruction exception. QEMU made both
+#    fields read-only in commit 18645f1957 ("target/riscv/csr.c: do not allow mstatus
+#    MPV/GVA writes", v11.1.0), which is what a test needs to write to enter a guest.
```
jordancarlin: Are there issues opened for these? If so, link the issue. No need to mention what commit broke QEMU here, just the issue link to track progress towards a fix.

config/whisper/ci.yaml
```diff
@@ -5,8 +5,13 @@
 # SsstrictSm, unknown
 # Smstateen Sail does not yet support 1P13 bit
 # Sdtrig* not yet implemented in Sail (https://github.com/riscv/sail-riscv/issues/222)
+# H: whisper.json does not enable h, and enabling it exposes incomplete H support:
+#    the mie/mip VS interrupt bits (2, 6, 10, 12) read as zero, which fails the Sm
+#    mcsr access/walk/shadow tests. Re-enable once whisper implements them. Only
```
jordancarlin: Whisper H support works. I think the issue is just that the `whisper.json` config file needs updated `mask` values for `mie` and `mip`. Please update the config file and then test again.

generators/testgen/src/testgen/io/templates.py
jordancarlin: These both seem like issues in this PR and I think they should be solved here rather than avoiding the problem by leaving the code that this PR introduces turned off.
1. The simplest solution is probably to use `.option arch +h` around these instructions in the trap handler, similar to what is done for F instructions in the failure code.
2. Part of adding support for H is updating the invisible trap handler to support it. Ideally that would be fixed in this PR.

Issues with one config should not prevent us from enabling the code for everything else. If that one config still has issues, let's disable that config rather than disabling all of this new code.

tests/env/rvtest_setup.h
```diff
@@ -949,23 +949,133 @@
...
-      csrw mcause, zero
-
       // Set up trap handlers for all modes
       // S and H-mode setup could be deferred to RVTEST_BOOT_TO_SMODE, but that is upsetting the linker
       // and there is no harm setting up all the trap handlers here
       RVTEST_TRAP_PROLOG M
       #ifdef S_SUPPORTED
+        // Order matches INSTANTIATE_MODE_MACRO: M, S, V. There is no HS-mode
+        // prolog: HS uses the S-mode trap CSRs, so the S prolog covers both and
+        // also saves hedeleg and hgatp when H is supported.
```
jordancarlin: No need to mention the lack of HS prolog here.

tests/env/rvtest_setup.h
```diff
@@ -949,23 +949,133 @@
...
+      //   VSXL  = 64: VS-mode XLEN is 64 (RV64 only)
+      // Tests that need any of these set them themselves.
+      LI(t0, HSTATUS_SPV | HSTATUS_SPVP | HSTATUS_HU | HSTATUS_VGEIN | \
+             HSTATUS_VTVM | HSTATUS_VTW | HSTATUS_VTSR | HSTATUS_VSBE)
+      csrc hstatus, t0
+      #if __riscv_xlen == 64
+        LI(t0, HSTATUS_VSXL)
+        csrc hstatus, t0
+        LI(t0, 0x0000000200000000)  // VSXL = 2
+        csrs hstatus, t0
```
jordancarlin: Clearing hstatus.VSXL seems like a bad idea. On a DUT with writable VSXL, that would switch to RV32 mode and potentially throw off other things.

tests/env/rvtest_setup.h
```diff
@@ -949,23 +949,133 @@
...
+      // on RV64, and FS/VS dirty when supported (as INIT_FLOAT_VECTOR_STATE does
+      // for mstatus), since a guest's FP and vector use is gated by both.
+      li t0, 0
+      #if __riscv_xlen == 64
+        LI(t0, 0x0000000200000000)  // UXL = 2
+      #endif
+      #if defined(F_SUPPORTED) || defined(ZFINX_SUPPORTED)
+        LI(t1, SSTATUS_FS)
+        or t0, t0, t1
+      #endif
```
jordancarlin: This isn't right. From the spec:

> `mstatus` field FS is hardwired to 0 if the Zfinx extension is implemented

tests/env/rvtest_setup.h
```diff
@@ -949,23 +949,133 @@
...
+      csrw htimedelta, zero
+      #if __riscv_xlen == 32
+        csrw htimedeltah, zero
+      #endif
+
+      // Delegate nothing to VS-mode by default: a guest trap goes to HS-mode,
+      // where the framework's trap handler is, unless a test opts in by writing
+      // hedeleg/hideleg itself. hedeleg is also cleared by RVTEST_TRAP_PROLOG S,
+      // which saves the incoming value; this keeps hideleg consistent with it.
+      csrw hideleg, zero
```
jordancarlin: Should we handle hideleg in the same place as hedeleg then?

tests/env/rvtest_setup.h
```diff
@@ -949,23 +949,133 @@
...
+          li t0, HSTATEEN_SSTATEEN | HSTATEEN0_SENVCFG
+          csrs hstateen0, t0
+        #else
+          li t0, HSTATEENH_SSTATEEN | HSTATEEN0H_SENVCFG
+          csrs hstateen0h, t0
+        #endif
+        li t0, HSTATEEN0_JVT | HSTATEEN0_FCSR
+        csrs hstateen0, t0
+      #endif
+    #endif // H_SUPPORTED
```
jordancarlin: Has this been tested? It looks like we are setting `hstateen0` before `mstateen0`. If `mstateen0.SE0` is still zero, then these writes will all trap becuase `hstateen0` is not accessible. Even if it is accessible, the fields in `hstateen0` are read-only zero if the corresponding field in `mstateen0` is set to 0.

jordancarlin: A similar issue applies to the `henvcfg` accesses slightly higher up. `henvcfg` is not accessible if `mstateen0.envcfg` is not set to 1 first.

jordancarlin: Let's move the whole block with all of the H-mode CSR initialization after all of the M-mode initialization to avoid these issues and any others that might pop up in the future with CSR interactions.

generators/testgen/src/testgen/priv/extensions/H.py
```diff
@@ -0,0 +1,552 @@
...
+        "",
+        "# The VS-stage macros take their VA as an immediate, so the three windows are",
+        "# fixed constants, which only works while the image really does live in the",
+        f"# 0x{_TEST_GIB:08X} gigapage. Derive that gigapage from rvtest_code_begin and stop",
+        "# the test if a link script ever moves the image somewhere else.",
+        f"{INDENT}LA(x{base}, rvtest_code_begin)",
+        f"{INDENT}srli x{base}, x{base}, 30",
+        f"{INDENT}slli x{base}, x{base}, 30   # align down to the 1 GiB the image sits in",
+        f"{INDENT}LI(x{check}, 0x{_TEST_GIB:08X})",
+        f"{INDENT}bne x{base}, x{check}, abort_test",
```
jordancarlin: AI says:
generators/testgen/src/testgen/priv/extensions/H.py:320-328: H_twostage still requires the image to be in
   the 0x80000000 GiB and aborts on any other valid link address. This makes the H suite fail for legal
   external DUT configurations. Add a VA-in-register page-table helper or derive all mapping windows from
   rvtest_code_begin.

generators/testgen/src/testgen/priv/extensions/H.py
jordancarlin: AI comments:

 - H.py:174-190: the SPVP test uses Bare/Bare translation and identical accessible memory. A DUT that ignores
   SPVP still passes. Use mappings or permissions that distinguish VS from VU access.
 - H.py:414-429: load and store guest-page faults are tested, but instruction guest-page fault is not. The
   handler already returns to ra for cause 20, so a jalr ra into the hole can test it.

tests/env/rvtest_trap_handler.h
```diff
@@ -118,6 +118,16 @@
 //    S-mode handler:  handles ECALL_TEST, GOTO_S/U, S-mode CSR_ACCESS locally;
 //                     forwards GOTO_M/VS/VU, M-mode CSR_ACCESS, and loads/stores to M-mode via ecall
 //
+//  Two limits apply to calls made from a guest (V=1):
+//    - GOTO_S/U and the GOTO_M forwarding resume the caller by jumping to xEPC,
+//      which is a guest virtual address, while the hart is already in host
+//      context. That only lands on the caller when the guest's VS-stage and
+//      G-stage maps are identity, which is what the H tests set up.
```
jordancarlin: This seems like a potentially significant issue. Any reason they can't do translation similar to what we already do when jumping back to S?

tests/env/rvtest_trap_handler.h
```diff
@@ -489,6 +499,33 @@
...
+.macro RVTEST_SET_SPV VAL, TMP_REG
+#ifdef H_SUPPORTED
+        LI(     \TMP_REG, HSTATUS_SPV)
+  .if (\VAL != 0)
+        csrs    CSR_HSTATUS, \TMP_REG
+  .else
+        csrc    CSR_HSTATUS, \TMP_REG
+  .endif
+#endif
+.endm
```
jordancarlin: How often is this macro used and is it really worthwhile to add the indirection? We access all other CSRs directly without a macro. Why does this one need a macro?

tests/env/rvtest_trap_handler.h
```diff
@@ -2118,36 +2199,40 @@ tsbi_\__MODE__\()instr_found:
         li      T2, 3*REGWIDTH                    // cause < timer -> SW or timer int (3 words)
         j       \__MODE__\()trap_sig_sv            // go to pointer update
 
-\__MODE__\()xcpt_sig_sv:                          // exception: check for hypervisor (6-word entry)
+\__MODE__\()xcpt_sig_sv:                          // exception: check for hypervisor (5-word entry)
+// An exception entry carries one extra word on a hypervisor build: the second
+// trap value (mtval2 / htval, the faulting guest physical address). M decides at
+// run time from misa.H, which a test can clear. xtinst is deliberately not
+// recorded: its value is implementation-defined (UDB TINST_VALUE_ON_*), so a
+// legal DUT may disagree with the reference model.
```
jordancarlin: `xtinst` has very clearly defined requirements in the H chapter. For all traps it is allowed to be either zero or a specific defined value. Seems like something we can and should check. The trap handler should check that it either matches the reference model value or is zero.

tests/env/rvtest_trap_handler.h
```diff
@@ -2118,36 +2199,40 @@ tsbi_\__MODE__\()instr_found:
         li      T2, 3*REGWIDTH                    // cause < timer -> SW or timer int (3 words)
         j       \__MODE__\()trap_sig_sv            // go to pointer update
 
-\__MODE__\()xcpt_sig_sv:                          // exception: check for hypervisor (6-word entry)
+\__MODE__\()xcpt_sig_sv:                          // exception: check for hypervisor (5-word entry)
+// An exception entry carries one extra word on a hypervisor build: the second
+// trap value (mtval2 / htval, the faulting guest physical address). M decides at
+// run time from misa.H, which a test can clear. xtinst is deliberately not
```
jordancarlin: Seems like maybe we should just drop the `misa` check and condition this off H_SUPPORTED. Keep in mind `misa` is allowed to be read-only zero even on a DUT that supports H, so this is not a reliable check.

jordancarlin (changes requested):

PR #2364, opened 2026-09-14
fix: skip mcountinhibit test when counter inhibition is unavailable

### LINKED ISSUE #2248: [Feature] support for cores without mcountinhibit
### Affected area

### Problem or motivation

As norm:mcountinhibit_not_impl states:

> If the mcountinhibit register is not implemented, the implementation behaves as though the register were set to zero.

ACT4 has tests for count inhibition:
- `cp_inhibit_minstret` - not a problem if sail supported lack of `mcountinhibit.IR`
- `cp_inhibit_mcycle` - without `mcountinhibit` is non-predictable

Additionally `mcountinhibit` causes invalid instruction exceptions

### Proposed solution

1. sail supports subset of WARL bits in `mcountinhibit` as well as fully non-accessible `mcountinhibit` https://github.com/riscv/sail-riscv/issues/1933
1. `cp_inhibit_mcycle` is skipped when UDB parameter `MCOUNTINHIBIT_IMPLEMENTED == false` or `MCOUNTINHIBIT[CY] == false`

### Additional context

_No response_

cryptokvc:
Fixes #2248

When `mcountinhibit` or its CY inhibit bit is unavailable, the generated `cp_inhibit_mcycle` sequence could access an unsupported CSR. The generated sequence and its `Sm_mcsr_cg` coverpoint/cross are now enabled only when both `UDB_MCOUNTINHIBIT_IMPLEMENTED` and `UDB_COUNTINHIBIT_EN_0` are defined. Implementations with both capabilities retain the test and coverage.

Validation: generated the `Sm` suite with `testgen` on Python 3.10; Ruff and Pyright passed for `Sm.py`. Preprocessed the generated mcycle sequence with neither macro, either macro alone, and both macros: the sequence appears only with both. The PR diff has no whitespace errors. No ISA simulation was run. Full CI is awaiting maintainer approval (`ci_pass` is pending).


davidharrishmc:
These names are incorrect and the PR has not been tested.

An improvement may be
"#if defined(UDB_MCOUNTINHIBIT_IMPLEMENTED) && defined(UDB_COUNTINHIBIT_EN_0)",


Please test PRs before indicating they are ready for review.

jordancarlin:
This also needs to update the coverpoints.

cryptokvc:
Thank you for the review. I changed the generated mcycle guard to `#if defined(UDB_MCOUNTINHIBIT_IMPLEMENTED) && defined(UDB_COUNTINHIBIT_EN_0)` and gated both `old_mcountinhibit_cy` and the `cp_inhibit_mcycle` cross with the same capabilities.

I generated the `Sm` suite with `testgen` on Python 3.10, ran Ruff and Pyright on `Sm.py`, and preprocessed the generated mcycle sequence under all four macro combinations. The sequence is emitted only when both macros are defined. The PR diff also passes the whitespace check. I have not run an ISA simulator. The required `ci_pass` check is still waiting for the maintainer-approved workflows; could a maintainer approve them to run?

jordancarlin (changes requested):
There are several other places in Sm and rvtest_setup.h where `mcountinhibit` is accessed. Can you update this PR to gate all of them? As is, a DUT that does not implement `mcountinhibit` would still fail with these changes.

davidharrishmc:
Please take the time to go through the full verification loop before presenting a PR as ready.  If you follow the instructions in the README for this repo, you can install Verilator and various simulators and test your design on all of them.  If your goal is to support no mcountinhibit, you should also temporarily disable it in one of the configs, and check that the existing tests fail and that your new tests pass.  Verification is much more work than code changes, and the PR developer needs to be the one to carry that load.

jordancarlin:
Superseded by #2449.

PR #2369, opened 2026-09-15
Update Vector Failure Code to Derive vd from SIGUPD

fourth-bit:
In segmented instructions and stores, the relevant `vd` in SIGUPD is not the `vd` derived from the instruction. This means in the failure code the wrong register was being read back for the display of a bad element vs the expected element, leading to the problems that #2293 tried to address. 

This PR adds a dummy instruction after the `RVTEST_WORD_PTR` entries that encodes the `vd` register so that it can be retrieved in the failure code.

In addition, this PR ensures that segmented unordered indexed loads have no potential overlap.


fourth-bit:
@hugo-teng Can you verify that these changes resolve the problems you were seeing?

hugo-teng:
> hugo-teng Can you verify that these changes resolve the problems you were seeing?

Hi @fourth-bit 

  yes —I rebuilt and re-ran on hardware:
  1. 2dbab5a: all 79 previously failing vsuxseg tests now pass (79/79).
  2. d957978: failure report now shows the correct register/value.

Thanks!

jordancarlin (approved):

PR #2372, opened 2026-09-15
Zicfilp tests only

EmanNasar001:
Adds privileged test generators for Zicfilp (landing pads) covering all four
privilege configurations, plus a no-S DUT config for the U-without-S variant.


config/sail/sail-rv64-zicfilp-nos/link.ld
```diff
@@ -0,0 +1,52 @@
+/* Set these to match the DUT memory map and hart count. */
```
davidharrishmc, line 1: I think we should drop this sail config in favor of #2291 before merging.  Fine to leave this in while experimenting.  Please pull #2291 into your branch (don't commit it to this PR) and test all the MU and M-only configs from that as well.

coverpoints/priv/Zicfilp_coverpoints.svh
```diff
@@ -73,7 +73,10 @@
         bins label_zero    = {20'h0};
         bins label_nonzero = {[20'h1:20'hFFFFF]};
     }
-    lpl_match: coverpoint (ins.current.imm[31:12] == ins.prev.x_wdata[7][31:12]) { }
+    lpl_match: coverpoint (ins.current.imm[31:12] == ins.prev.x_wdata[7][31:12]) {
```
davidharrishmc: It was better the original way, which implied both 0 and 1 bins without having to state them.  Drop this change.

generators/testgen/src/testgen/priv/extensions/ZicfilpSU.py
```diff
@@ -0,0 +1,79 @@
+##################################
```
davidharrishmc: The convention we've been following is to not have a SU suite.  We should move this either into S (run in U-mode) or U (#ifdef S_SUPPORTED).

generators/testgen/src/testgen/priv/extensions/ZicfilpU.py
```diff
@@ -0,0 +1,56 @@
...
+    emit_mode,
+)
+from testgen.priv.registry import add_priv_test_generator
+
+
+@add_priv_test_generator(
+    "ZicfilpU",
+    required_extensions=["Zicfilp", "Zicsr", "U"],
+    march_extensions=["I", "A", "F", "D", "C", "V", "Zabha", "Zacas", "Zicbom", "Zicbop", "Zicboz", "Zca"],
+    extra_defines=["#define BOOT_TO_MMODE"],
```
davidharrishmc: It's time to move everything to T-SBI.  A U suite should boot directly to U-mode.  Drop the BOOT_TO_MMODE and it will go straight to U.

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,77 @@
...
+    emit_mode,
+)
+from testgen.priv.registry import add_priv_test_generator
+
+
+@add_priv_test_generator(
+    "ZicfilpS",
+    required_extensions=["Zicfilp", "Zicsr", "S"],
+    march_extensions=["I", "A", "F", "D", "C", "V", "Zabha", "Zacas", "Zicbom", "Zicbop", "Zicboz", "Zca"],
+    extra_defines=["#define BOOT_TO_MMODE", "#define TRAP_SIGUPD_COUNT 40000"],
```
davidharrishmc: Should boot to S_MODE.  Why are so many trap counts used.  Is that expected to need so many?

generators/testgen/src/testgen/priv/extensions/ZicfilpSU.py
```diff
@@ -0,0 +1,79 @@
...
+    def build_su(xlen: int, mode: str = satp_mode) -> list[str]:
+        # trampoline_section=".text.rvtest": this mode's LPAD mismatch
+        # exception is genuinely reachable (senvcfg.LPE really gates it),
+        # so a real fault can land on these trampolines. .text.rvtest is
+        # required so the trap handler's EPC segment-recognition logic
+        # records the fault instead of treating it as unrecognized.
+        #
+        # skip_trampoline_fallthrough=True: without it, mode-entry code
+        # falls straight through into _tgt_lpad_zero's `c.jr x7` with x7
+        # uncontrolled, which lands back on GOTO_UMODE's own ecall and
```
davidharrishmc: GOTO_UMODE is pre-TSBI.  Switch everything TSBI

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+NON_LINK_COMP_REGS = [r for r in COMP_REGS if r not in LINK_REGS]
+REP_NON_LINK = 28
+REP_LINK = 1
+
+# SATP Modes
+MODES = ["bare", "sv39", "sv48", "sv57"]
+MODE_GUARDS = {m: None if m == "bare" else f"{m.upper()}_SUPPORTED" for m in MODES}
+LEVELS_BELOW_ROOT = {"sv39": 2, "sv48": 3, "sv57": 4}
+
+GOTO_MMODE = "RVTEST_GOTO_MMODE"
```
davidharrishmc: Drop these macros and just use the RVTEST_TSBI_GOTO_* directly.  Don't use RVTEST_GOTO_MMODE anymore.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+    if mode == "bare":
+        return ["csrwi satp, 0", "sfence.vma"]
+    lines = _grant_umode_code_access(xlen) if grant_umode_access else []
+    lines += [f"SATP_SETUP_RV{'64' if xlen == 64 else '32'}({mode})", "sfence.vma"]
+    return lines
+
+
+def teardown_vm(has_satp: bool = True) -> list[str]:
+    if not has_satp:
+        return [GOTO_MMODE, ""]
```
davidharrishmc: TSBI

davidharrishmc:
Be sure to set up your signing and sign this so DCO passes. Put Zicfilp in QEMU CI exclusion, and anywhere else necessary to get through CI.

coverpoints/priv/ZicfilpSU_coverage.svh
```diff
@@ -23,22 +23,22 @@ covergroup Zicfilpsu_cg with function sample(ins_t ins);
...
     lpe_disabled: coverpoint get_csr_val(ins.hart, ins.issue,
-                                `SAMPLE_CURRENT, "menvcfg", "lpe") {
+                                `SAMPLE_CURRENT, "senvcfg", "lpe") {
         bins disabled = {0};
     }
     xtval_lpad: coverpoint ins.current.csr[12'h343] {
         bins code_2 = {2};
     }
-    cp_zicfilp_indirect_elp_state_update: cross priv_mode_u, menvcfg_lpe, indirect_ct_prev, rs1_all_prev, lpad_dest, spelp;
+    cp_zicfilp_indirect_elp_state_update: cross priv_mode_u, senvcfg_lpe, indirect_ct_prev, rs1_all_prev, lpad_dest, spelp;
```
davidharrishmc, line 38: From U mode, if S is supported, it seems we should test a cross of mseccfg, menvcfg, and senvcfg.  Other suites should test a subset of this, such as mseccfg x menvcfg from S mode, or from U mode with no S.  The test plan is getting out of date with the tests and is unspecific about LPE.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+# Author : Eman Nasar  email:fatehulnasareman@gmail.com (UET, May 2026)
+# SPDX-License-Identifier: Apache-2.0
+##################################
+
+"""Shared Zicfilp extension test infrastructure.
+Common code for ZicfilpSm (M-mode), ZicfilpS (S-mode and U+S), ZicfilpU (U-S)
+test generators.
+"""
+
+from __future__ import annotations
```
davidharrishmc: No longer needed or desired.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+            pelp_csr=pelp_csr,
+            pelp_bit=_MSTATUS_MPELP_BIT if xlen == 64 else _MSTATUSH_MPELP_BIT,
+            guard="#ifndef S_SUPPORTED",
+            pelp_needs_elevation=True,  # mstatus/mstatush isn't readable from U-mode
+        )
+    raise ValueError(f"Unknown mode: {mode}")
+
+
+# ═══════════════════════════════════════════════════════════════════════════
+# ASSEMBLY HELPERS
```
davidharrishmc: Put all these short helpers inline rather than making them functions.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+            rd = 10  # temp
+            return _fixed(f"jalr x{rd}, 0(x{rs1})")
+    elif name == "c.jr":
+        return [_emit_c_jr(rs1)]
+    elif name == "c.jalr":
+        return [_emit_c_jalr(rs1)]
+    raise ValueError(f"Unknown branch: {name}")
+
+
+# CSR addresses for the T-SBI CSR-access encodings below. mseccfg isn't
```
davidharrishmc: This shouldn't be necessary.  Use tsbi_call from asm.py instead of raw machine code.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+    lines = []
+    if guard:
+        lines.append(guard)
+    if xlpe.csr == "mseccfg":
+        act = "csrs" if en else "csrc"
+        msk = 1 << xlpe.bit
+        lines += [
+            f"# {xlpe.csr}.{xlpe.name} = {int(en)}",
+            f"LI(x{reg}, {hex(msk)})",
+            f"{act} {xlpe.csr}, x{reg}",
```
davidharrishmc: This whole function can be greatly simplified.  See ZawrsCommon.py m_csr for an example.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+
+    Gated on SV39_SUPPORTED/SV32_SUPPORTED: only meaningful for callers whose
+    build() actually cycles through paged satp modes (sv39/sv48/sv57). For a
+    bare-only body, use both_xlens_bare instead -- gating bare content behind
+    a translation-support define excludes it on any DUT that doesn't also
+    happen to support paging (e.g. a true no-S config), even though the body
+    itself never touches satp.
+    """
+    return [
+        "#if __riscv_xlen == 64",
```
davidharrishmc: Shouldn't be needed because SV39_SUPPORTED implies xlen64.

coverpoints/priv/ZicfilpSm_coverage.svh
davidharrishmc: I think this needs to be UDB_MXLEN_64 now.  There is no longer XLEN64 so you won't test these coverpoints.  Many places, 32 as well

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,1193 @@
...
+    td.int_regs.return_registers([chk])
+    return out
+
+
+def _build_faults(pfx: str, td: TestData, cg: str, xlpe: XLPEConfig, xlen: int) -> list[str]:
+    """CP_LPAD_MISSING, CP_LPAD_MISMATCH, CP_EXC_DELIVERY, CP_LPAD_SCENARIO."""
+    if xlpe.csr == "mseccfg":
+        # Every case here deliberately traps with LPE=1. In M-mode that
+        # recurses into the shared trap dispatcher's own unguarded indirect
+        # jump (see the note in _build_elp_update); skip until that's fixed.
```
davidharrishmc: Fix this so ZicfilpSm tests LPAD faults

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,77 @@
...
+    tc.code.extend(both_xlens(build))
+    if guard:
+        tc.code.append(f"#endif // {guard}")
+    return td.end_test_chunk()
+
+
+@add_priv_test_generator(
+    "ZicfilpS",
+    required_extensions=["Zicfilp", "Zicsr", "S", "U"],
+    march_extensions=["I", "A", "F", "D", "C", "V", "Zabha", "Zacas", "Zicbom", "Zicbop", "Zicboz", "Zca"],
```
davidharrishmc: Are these all used?

generators/testgen/src/testgen/priv/extensions/ZicfilpSm.py
```diff
@@ -0,0 +1,67 @@
+##################################
+# priv/extensions/ZicfilpM.py
```
davidharrishmc: Sm

davidharrishmc (commented):
A helper function such as ZawrsCommon.py m_csr/s_csr will save you a bunch of code related to T-SBI.

ZicfilpCommon has so many functions and their names are not obviously related to the coverpoints.  Would it be possible to simplify to functions whose names match the coverpoints?

Reduce the indirection and just print the stuff in the coverpoints where possible.

Focus comments on what the code does, not its history of what changed.  Keep the comments succinct, usually 1 or 2 lines not a paragraph.

davidharrishmc (changes requested):

config/sail/sail-rv64-zicfilp-nos/link.ld
EmanNasar001: Tested the Zicfilp suites on the #2291 MU and M-only configs (pulled locally) rv32-mu and all -m configs pass. rv64-mu fails in Sail
"architecture(0b00) is invalid";#2291 has no sail -mu config, and the spike -mu configs hang under coverage , so dropping sail-rv64-zicfilp-nos would leave no way to
collect ZicfilpU coverage. Drop it anyway, or keep it for now? 
Screenshot 2026-09-16 212827

coverpoints/priv/Zicfilp_coverpoints.svh
EmanNasar001: Reverted

generators/testgen/src/testgen/priv/extensions/ZicfilpSU.py
EmanNasar001: DONE

generators/testgen/src/testgen/priv/extensions/ZicfilpU.py
EmanNasar001: Dropped BOOT_TO_MMODE, ZicfilpU now boots straight to U-mode and reaches
menvcfg/mstatus through tsbi_call. Passes on the rv32-mu configs from #2291.

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
EmanNasar001: the 40000 was left over from debugging a fault  where
ELP stayed armed after a fault. The trap report shows 127 traps per file, all
cause 18 at the landing-pad targets, so I dropped TRAP_SIGUPD_COUNT;                                                   Now boots to S-mode. It still steps up briefly with RVTEST_TSBI_GOTO_MMODE to
clear medeleg, since the landing-pad coverpoints sample mcause/mtval, so the
software-check has to be taken in M-mode, and medeleg isn't in
tsbi_instr_table. If it rather  be delegated to S, the coverpoints would
need to read scause/stval instead.

generators/testgen/src/testgen/priv/extensions/ZicfilpSU.py
EmanNasar001: Switched everything to T-SBI

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
EmanNasar001: only Zca was needed

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
EmanNasar001: Looked into ZicfilpSm LPAD faults. When the fault is taken in M-mode,
mseccfg.MLPE is still 1 in the handler, so its own indirect jump re-arms ELP and faults on the next instruction, recursing
forever. So this needs a trap handler change, not a test change. Simplest option
I see is to clear MLPE on handler entry and restore it before mret, guarded for
Zicfilp. The alternative is lpad 0 at every indirect jump target in the
handler, but that shifts the dispatch table offsets. Is clearing MLPE in the
handler okay?

coverpoints/priv/ZicfilpSm_coverage.svh
EmanNasar001: Switched to UDB_MXLEN_64 everywhere

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
EmanNasar001: Removed

UmerShahidengr:
#2382 is superseding this PR, so I am closing this one

PR #2380, opened 2026-09-16
Install prek hooks even with a global core.hooksPath

davidharrishmc:
One-line change to `.mise.toml`. Writing this up in full because it is a build-environment change and the trade-off is worth a look rather than a rubber stamp.

## Symptom

With a global `core.hooksPath` configured, **no make target runs at all** in a git worktree:

```
[deps.prek] error: Refusing to install hooks because `core.hooksPath` is configured outside this repository.
mise ERROR deps providers failed:
  prek: sh exited with non-zero status: exit code 2
make: *** [Makefile:195: work/stamps/covergroupgen.stamp] Error 1
```

## Why it happens, and why it is usually invisible

Two things combine:

1. `prek install` **refuses** (exit 2) whenever `core.hooksPath` points outside the repository, because git would then run hooks from that directory instead of the repo's own.
2. The dep declares `outputs = [".git/hooks/pre-commit"]`. In a normal clone that file exists after the first install, so the dep is satisfied and never runs again. **In a worktree `.git` is a file, not a directory**, so that path can never exist — the dep is considered stale on every invocation and fails every time.

So a maintainer working in a normal clone will never see this, while anyone using worktrees is hard-blocked. A global hooks directory is a common setup (org-wide DCO sign-off, commit-message policy, secret scanning), and the usual pattern delegates to each repo's own hook, so per-repo hooks still work — prek just has no way to know that.

## The change

```toml
run = "prek install --prepare-hooks || prek install --force --prepare-hooks"
```

Retry with `--force` **only if** the plain install refuses.

## Why not just `--force`

That was my first attempt and it is worse. Verified in a scratch repo:

- Plain `prek install` with a pre-existing foreign hook: `moved it to .git/hooks/pre-commit.legacy` … `Migration mode: prek will also run legacy hook`. The old hook is preserved and still runs.
- `prek install --force`: `Overwriting existing hook` and `pre-commit.legacy` is **deleted**.

So unconditional `--force` would silently discard a pre-existing hook for every contributor, including those who never had the problem. The `||` fallback keeps the normal path untouched and only escalates in the case that is otherwise a hard failure.

## What a reviewer may want to weigh

- In the fallback case prek's refusal message is still printed before the retry succeeds, so the output looks alarming even though the build proceeds. Suppressing it would hide a genuine warning, so I left it.
- `--force` in the fallback path does still drop a legacy hook — but only for users whose hooks git was not going to run from that directory anyway, which is the situation prek is warning about.
- The alternative is to fix `outputs` so the dep is satisfied in worktrees. I tried and it is a trap: `git rev-parse --git-path hooks` resolves *through* `core.hooksPath` and returns the global directory, which always exists, so mise would consider the hook installed when it is not. `--git-common-dir` gives the right path, but mise gates re-runs on the **sources hash**, not on outputs (`seen_outputs` is informational), so fixing the path does not fix the failure. Hence the change is on `run`, not `outputs`.
- Doing nothing is defensible if the project would rather tell worktree users to run `prek install --force` once by hand. That works, but it has to be redone whenever the mise dep state is cleared.

## Testing

From a fully clean state (mise dep state and make stamps deleted, hook removed) inside a worktree with a global `core.hooksPath`: `make tests` exits 0, the dep reports `✓ done`, and the prek shim is installed into the repo's common hooks directory. Re-running is a no-op. Behaviour in a normal clone is unchanged, including legacy hook migration.


davidharrishmc:
Superseded by 2384

PR #2381, opened 2026-09-16
Enable Smstateen and Ssstateen on the cvw-rv32/64gc configs

davidharrishmc:
CORE-V Wally now implements the state enable CSRs (openhwgroup/cvw#1881), so declare Smstateen and Ssstateen on the `cvw-rv32gc` and `cvw-rv64gc` configs.

Two halves are needed, and the second is easy to miss:

- **`cvw-rv*gc.yaml`** gains the two extensions and `MSTATEEN_ENVCFG_TYPE: rw`. Wally implements `mstateen0.ENVCFG` as read/write; every other `mstateen0` bit is read-only zero because the state it controls is not implemented.
- **`sail.json`** gains the matching `Stateen.Smstateen.supported` and `Stateen.Ssstateen.supported`. Without this the reference model does not implement `sstateen0` and traps on the very first machine-mode `csrr x15, sstateen0` in `Ssstateen-00`. That trap becomes the first entry in the expected trap signature, so every later entry is shifted by one and a conforming DUT fails with a confusing mode mismatch:

```
RVCP: Mismatching field: Vector+Mode+Status word (trap signature word 0)
RVCP: Expected value: 0x000000000f000663   (mstatus.MPP = 11, trap from M-mode)
RVCP: Actual value:   0x000000000d000663   (mstatus.MPP = 01, trap from S-mode)
```

With both halves in place, `Smstateen-00` and `Ssstateen-00` pass on `cvw-rv32gc` and `cvw-rv64gc`:

```
make EXTENSIONS=Smstateen,Ssstateen cvw-rv64gc
make EXTENSIONS=Smstateen,Ssstateen cvw-rv32gc
```


davidharrishmc:
Closing; will add all new features together later.

PR #2382, opened 2026-09-16
Zicfilp test generator

UmerShahidengr:


coverpoints/priv/ZicfilpS_coverage.svh
davidharrishmc: should this have lpe_enabled?

davidharrishmc: lpe_enabled?  Applies to all coverage files.

davidharrishmc: lpe_enabled. Also, does the test plan at entry C8 need to be updated to list xLPE=1?

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
         ignore_bins ig_no_lp = binsof(elp_before.no_lp_expected);
     }
 
-    cp_zicfilp_lpad_label_exception_delivery: cross priv_mode_s, elp_before, sw_check_exc, xtval_lpad {
-        ignore_bins ig_no_lp = binsof(elp_before.no_lp_expected);
+    cp_zicfilp_lpad_label_exception_delivery: cross priv_mode_s, elp_before, sw_check_exc, xtval_lpad, spelp {
```
davidharrishmc: lpe_enabled?  Also, does the test plan at entry C9 need to be updated to list xLPE=1?

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
+        ignore_bins ig_no_lp   = binsof(elp_before.no_lp_expected);
+        ignore_bins ig_no_pelp = binsof(spelp.no_lp_expected);
     }
 
-    cp_disabled_zicfilp: cross priv_mode_s, lpe_disabled, lpad_lpl_nonzero;
+    cp_disabled_zicfilp: cross priv_mode_s, lpe_disabled, indirect_ct_prev, lpad_lpl_nonzero, lpl_match {
+        ignore_bins ig_match = binsof(lpl_match.match);
+    }
 
     cp_lpad_no_sw_exception_elp_clear_zicfilp: cross priv_mode_s, elp_before, lpad_lpl_zero {
```
davidharrishmc: lpe_enabled? Also, does the test plan at entry C11 need to be updated to list xLPE=1?

coverpoints/priv/ZicfilpS_coverage.svh
davidharrishmc: Do a review of coverpoints against test plan for S/Sm/U.

davidharrishmc: We don't usually list expected outputs (sw_check_exc, xtval) in the cross because they make coverpoints more fussy to hit and are checked by the signature, but I suppose they are harmless since you are getting the coverage.

coverpoints/priv/ZicfilpS_coverage.svh
davidharrishmc: testplan here mentions x7, mismatched LPL.  This coverpoint checks the outcomes but not the input conditions.

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
-    cp_zicfilp_lpad_label_exception_delivery: cross priv_mode_s, elp_before, sw_check_exc, xtval_lpad {
-        ignore_bins ig_no_lp = binsof(elp_before.no_lp_expected);
+    cp_zicfilp_lpad_label_exception_delivery: cross priv_mode_s, elp_before, sw_check_exc, xtval_lpad, spelp {
+        ignore_bins ig_no_lp   = binsof(elp_before.no_lp_expected);
+        ignore_bins ig_no_pelp = binsof(spelp.no_lp_expected);
     }
 
-    cp_disabled_zicfilp: cross priv_mode_s, lpe_disabled, lpad_lpl_nonzero;
+    cp_disabled_zicfilp: cross priv_mode_s, lpe_disabled, indirect_ct_prev, lpad_lpl_nonzero, lpl_match {
+        ignore_bins ig_match = binsof(lpl_match.match);
```
davidharrishmc: why is the match bin necessary to ignore?

coverpoints/priv/ZicfilpS_coverage.svh
davidharrishmc: where is lpad_lpl_zero coming from?  I don't see it in the test plan.  The address alignment and match seem to be missing.

davidharrishmc: The test plan specifies target address 4-bytes aligned.  I don't see anything in the test plan about target address misaligned (it's been moved to gray).  Should misaligned be added back to the testplan somewhere?

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
 
     cp_lpad_no_sw_exception_elp_clear_zicfilp: cross priv_mode_s, elp_before, lpad_lpl_zero {
-        ignore_bins ig_no_lp    = binsof(elp_before.no_lp_expected);
+        ignore_bins ig_no_lp = binsof(elp_before.no_lp_expected);
     }
 
     `ifdef RVMODEL_ACCESS_FAULT_ADDRESS
-        cp_exception_priority_zicfilp: cross priv_mode_s, elp_before, pc_fault_addr {
-            ignore_bins ig_no_lp = binsof(elp_before.no_lp_expected);
+        instr_access_fault: coverpoint ins.current.csr[12'h142]
```
davidharrishmc: Use CSR_SCAUSE and CSR_SEPC instead of 142 & 141

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
+                            iff (ins.current.csr_wb[12'h141] && (ins.current.csr[12'h141] == `RVMODEL_ACCESS_FAULT_ADDRESS)) {
+            bins cause_1 = {1};
+        }
+        cp_exception_priority_zicfilp: cross priv_mode_s, lpe_enabled, lp_jalr_to_fault_addr, instr_access_fault, spelp {
```
davidharrishmc: I don't see the illegal instruction part of the coverpoint.  May not be easy to measure, but there should be a comment if you are intentionally omitting it.

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
+
+    // ELP before the current instruction in U-mode, which uses senvcfg.LPE
+    elp_before_u: coverpoint (`ZICFILP_LP_BRANCH(ins.prev.insn) &&
+                              get_csr_val(ins.hart, ins.issue, `SAMPLE_PREV, "senvcfg", "lpe")) {
+        bins no_lp_expected = {1'b0};
+        bins lp_expected    = {1'b1};
+    }
+
+    // Trap from S-mode into S-mode saves ELP in SPELP
+    cp_pelp_trap_entry_s_zicfilp: cross priv_mode_s, s_trap_here, elp_before, spelp {
```
davidharrishmc: lpe_enabled?

davidharrishmc: EBREAK is mentioned in test plan but doesn't appear here.  Should it be broken into two coverpoints?

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
+    }
+    `ifdef ZCA_SUPPORTED
+        cp_pelp_trap_entry_s_guarded_zicfilp_c: cross priv_mode_s, lpe_enabled, indirect_ct_prev_c, rs1_link_prev_c, ebreak, s_trap_here, spelp {
+            ignore_bins ig_pelp = binsof(spelp.lp_expected);
+        }
+    `endif
+
+    // Trap from U-mode into S-mode saves ELP in SPELP
+    `ifdef U_SUPPORTED
+        cp_elp_state_preservation_s_zicfilp: cross priv_mode_u, s_trap_here, elp_before_u, spelp {
```
davidharrishmc: Doesn't seem to be checking inputs of non-LPAD instruction

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
+            ignore_bins ig_nolp_lp = binsof(elp_before_u.no_lp_expected) && binsof(spelp.lp_expected);
+            ignore_bins ig_lp_nolp = binsof(elp_before_u.lp_expected)    && binsof(spelp.no_lp_expected);
         }
     `endif
 
+    // SRET sets ELP from SPELP only if xLPE of the mode it returns to is 1, and clears SPELP
+    sret_mode: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_PREV, "sstatus", "spp") {
+        bins S_mode = {1'b1};
+        `ifdef U_SUPPORTED
```
davidharrishmc: Non need to gate; S implies U

coverpoints/priv/ZicfilpS_coverage.svh
```diff
@@ -61,24 +82,100 @@ covergroup Zicfilp_s_cg with function sample(ins_t ins);
...
+        ) {
+        bins disabled = {0};
+        bins enabled  = {1};
+    }
+    // SPELP written through the sstatus view, which is how the test sets it up before SRET
+    spelp_before: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_PREV, "sstatus", "spelp") {
+        bins no_lp_expected = {0};
+        bins lp_expected    = {1};
+    }
+    cp_pelp_trap_return_s_zicfilp: cross priv_mode_s, sret, sret_mode, sret_lpe, spelp_before, spelp {
```
davidharrishmc: set sepc to non-LPAD instruction?  cross menvcfg.LPE={0/1} and senvcfg.LPE={0/1}

docs/ctp/src/privmisc23.adoc
```diff
@@ -117,14 +117,18 @@ See Ssnpm for normative rules.
...
 
-In Privileged section because it has both unpriv and privileged behaviors.
+In Privileged section because it has both unpriv and privileged behaviors.  The landing pad tests run once per
+privilege mode: ZicfilpSm in M-mode, ZicfilpS in S-mode, and ZicfilpU in U-mode.  ZicfilpSm also covers
+MPELP on trap entry into M-mode and ELP on MRET, which require M-mode control of medeleg, mstatus.MPP, and mstatus.MPELP.
 
 [[t-Zicfilp-coverpoints]]
 .Zicfilp Coverpoints
 //,===
 //include::{testplansdir}/Zicfilp.adoc[]
```
davidharrishmc: Add hyperlink to four google sheet test plans:
https://docs.google.com/spreadsheets/d/1ED9ontJZP-Dg2MAF-4mkhqKFgpjf6Fp6yeXKuW61FOY/edit?gid=1319034416#gid=1319034416
https://docs.google.com/spreadsheets/d/1ED9ontJZP-Dg2MAF-4mkhqKFgpjf6Fp6yeXKuW61FOY/edit?gid=2113591295#gid=2113591295
https://docs.google.com/spreadsheets/d/1ED9ontJZP-Dg2MAF-4mkhqKFgpjf6Fp6yeXKuW61FOY/edit?gid=779370838#gid=779370838
https://docs.google.com/spreadsheets/d/1ED9ontJZP-Dg2MAF-4mkhqKFgpjf6Fp6yeXKuW61FOY/edit?gid=84584846#gid=84584846

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,193 @@
...
+@add_priv_test_generator(
+    "ZicfilpS",
+    required_extensions=["S", "Zicfilp"],
+    extra_defines=["#define BOOT_TO_SMODE"],
+)
+def make_zicfilps(test_data: TestData) -> list[TestChunk]:
+    """Generate Zicfilp landing pad tests in S-mode."""
+    return [
+        *make_zicfilp_tests(test_data, "S", covergroup),
+        _trap_entry_s_tests(test_data),
```
davidharrishmc: Do any of these have enough in common with m tests to share in ZicfilpCommon with a priv input?

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,329 @@
...
+            "With xLPE=1 and rs1 other than x1/x5/x7, a non-LPAD target raises a software-check exception.",
+        )
+    ]
+    if kind != "jalr":
+        lines.append("#ifdef ZCA_SUPPORTED")
+    for lpe in (0, 1):
+        lines.extend(set_lpe(priv, priv, temp_reg, bool(lpe)))
+        for rs1 in range(1, 32):
+            # jalr is an indirect jump/return (rd=x0) for odd rs1, including x1/x5/x7, and an indirect call otherwise
+            rd = 0 if rs1 % 2 else 1
```
davidharrishmc, line 161: This is kind of cute to alternate rd 0 and nonzero.  But do you think it's worthwhile crossing 0 and nonzero on every rs1 and adding it to the coverpoint and test plan?

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,193 @@
...
+    tc = test_data.begin_test_chunk()
+    tc.code.extend(
+        [
+            comment_banner(
+                coverpoint, "ebreak in S-mode with menvcfg.LPE=1 and ELP=NO_LP_EXPECTED saves sstatus.SPELP=0"
+            ),
+            *set_lpe("S", "S", temp_reg, True),
+            test_data.add_testcase("ebreak_no_lp_expected", coverpoint, covergroup),
+            "ebreak",
+            "lpad 0 # resume point if the ebreak were wrongly a software-check exception",
```
davidharrishmc: I think lpad 0 should be replace with *RESUME to trap in case ELP was wrongly restored.

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,193 @@
...
+                [
+                    "",
+                    *([f"mv x{save_reg}, x{link_reg} # save x{link_reg}"] if rs1 == link_reg else []),
+                    f"LA(x{rs1}, {label}_target)",
+                    f"{label}:",
+                    jump(kind, rs1),
+                    ".p2align 2",
+                    f"{label}_target:",
+                    "ebreak",
+                    "lpad 0 # resume point if the ebreak were wrongly a software-check exception",
```
davidharrishmc: I think lpad 0 should be replace with *RESUME to trap in case ELP was wrongly restored.

davidharrishmc: Same in Sm (1 place after ecall at 99)

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,193 @@
...
+                        f"{'csrs' if pelp else 'csrc'} sstatus, x{temp_reg}",
+                        f"LA(x{temp_reg}, {label}_target)",
+                        f"csrw sepc, x{temp_reg}",
+                        f"{label}:",
+                        "sret",
+                        ".p2align 2",
+                        f"{label}_target:",
+                        "addi x0, x0, 0 # software-check exception only if SPELP=1 and xLPE=1",
+                        "lpad 0 # resume point after the software-check exception",
+                        *(["RVTEST_TSBI_GOTO_SMODE"] if mode != "S" else []),
```
davidharrishmc: Didn't check that SPELP is restored to 0 by the sret.  Consider sigupd'ing it explicitly:

f"csrr x{temp_reg}, sstatus",
f"LI(x{other}, SSTATUS_SPELP)",
f"and x{temp_reg}, x{temp_reg}, x{other}",
write_sigupd(temp_reg, test_data),

davidharrishmc (changes requested):
Still reviewing Python.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,329 @@
...
+        ("lp_expected_lpl_zero_x7_zero", "cp_zicfilp_lpad_zero_label_bypass", 0, "lpad 0"),
+        ("lp_expected_lpl_zero_x7_nonzero", "cp_zicfilp_lpad_zero_label_bypass", LABEL << 12, "lpad 0"),
+        ("lp_expected_lpl_match", "cp_zicfilp_lpad_valid_execution", LABEL << 12, f"lpad 0x{LABEL:x}"),
+        (
+            "lp_expected_lpl_match_all_ones",
+            "cp_zicfilp_lpad_valid_execution",
+            ALL_ONES_LABEL << 12,
+            f"lpad 0x{ALL_ONES_LABEL:x}",
+        ),
+        ("lp_expected_lpl_mismatch", "cp_zicfilp_lpad_label_mismatch", LABEL << 12, f"lpad 0x{OTHER_LABEL:x}"),
```
davidharrishmc: Reverse order of missing_instruction and mismatch coverpoints to match test plan.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,329 @@
...
+        ),
+        (
+            "lp_expected_not_lpad",
+            "cp_zicfilp_lpad_missing_instruction_exception",
+            LABEL << 12,
+            "addi x0, x0, 0 # non-LPAD target",
+        ),
+        (
+            "lp_expected_illegal_instruction",
+            "cp_zicfilp_lpad_missing_instruction_exception",
```
davidharrishmc: I like this test but am having trouble matching it up with the test plan.  Do we need to add a bin to the testplan?

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,193 @@
...
+
+SPP = {"S": 1, "U": 0}
+
+
+def _trap_entry_s_tests(test_data: TestData) -> TestChunk:
+    """A trap from S-mode into S-mode with ELP=NO_LP_EXPECTED saves SPELP=0.
+
+    ELP=LP_EXPECTED is saved on the software-check exceptions of the shared S-mode tests.
+    """
+    coverpoint = "cp_pelp_trap_entry_s_zicfilp"
```
davidharrishmc: The test plan part (1) mentioned an Indirect_CT that doesn't appear here.  Is that a duplicate of the common tests that should be deleted from the test plan?

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
```diff
@@ -0,0 +1,193 @@
...
+def _trap_return_s_tests(test_data: TestData) -> TestChunk:
+    """SRET sets ELP=SPELP only if xLPE of the new mode is 1, and clears SPELP."""
+    coverpoint = "cp_pelp_trap_return_s_zicfilp"
+    temp_reg = test_data.int_regs.get_register()
+    tc = test_data.begin_test_chunk()
+    lines = [
+        comment_banner(
+            coverpoint,
+            "SRET to S- and U-mode x new mode's xLPE = {0, 1} x SPELP = {0, 1}, returning to a non-LPAD\n"
+            "instruction. Only SPELP=1 with xLPE=1 restores ELP=LP_EXPECTED and raises a software-check exception.",
```
davidharrishmc: As mentioned on coverpoint, I think this might as well cross menvcfg.LPE and senvcfg.LPE 4 bins.

generators/testgen/src/testgen/priv/extensions/ZicfilpSm.py
```diff
@@ -0,0 +1,215 @@
...
+            )
+        if kind != "jalr":
+            lines.append("#endif")
+    lines.extend(set_lpe("M", "M", temp_reg, False))
+    tc.code.extend(lines)
+    test_data.int_regs.return_registers([save_reg, temp_reg])
+    return test_data.end_test_chunk()
+
+
+def _elp_state_preservation_tests(test_data: TestData) -> TestChunk:
```
davidharrishmc: Move this up to top to match order of coverpoints in test plan and coverage file.

davidharrishmc (changes requested):

davidharrishmc:
The misaligned landing pad rule seems to never be tested.

The Zicflip normative rule misspellings have been fixed upstream and the rules in this PR can be fixed to match.

Add rows deferring mnstatus_mnpelp_op, mnstatus_mnpelp_update, and mnret_zicfilp 

claude says
 norm:Zicfilp_pelp_trap_return has two clauses; the second, xPELP is set to NO_LP_EXPECTED, is required to hit cp_pelp_trap_return_* (the surviving mpelp bin is no_lp_expected) but is not asserted by any signature — a DUT that leaves MPELP set loses a coverage bin rather than failing.

davidharrishmc (commented):
Re-reviewed at ffb74259b. I regenerated the three suites and they pass on spike-rv64-max, spike-rv32-max and sail-rv64-max, and Questa functional coverage closes at 100% for all three covergroups on both sail-rv64-max and sail-rv32-max (so the new mstatush.MPELP / sstatus.SPELP decoding works in both XLENs). `make lint` is clean and every CI check is green. The trap reports show exactly the intended traps (56 software checks per jump kind = 28 non-guarded rs1 x 2, so x1/x5/x7 correctly do not trap) and no trap lands inside the trap handler, so the new landing pad at common_<MODE>entry is sufficient for these tests. Three comments below.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
```diff
@@ -0,0 +1,650 @@
...
+        lines.extend([f"#if {condition}", *case, "#endif"] if condition else case)
+    lines.extend(set_lpe(priv, priv, temp_reg, False))
+    tc.code.extend(lines)
+    test_data.int_regs.return_registers([jump_reg, save_reg, temp_reg])
+    return test_data.end_test_chunk()
+
+
+def disabled_tests(test_data: TestData, priv: str, covergroup: str) -> TestChunk:
+    """With xLPE=0, an LPAD reached by an indirect jump never checks its label."""
+    coverpoint = "cp_disabled_zicfilp"
```
davidharrishmc, line 263: The "lpad is a no-op when ELP=NO_LP_EXPECTED" arm is never tested. Per the spec listing (src/unpriv/cfi.adoc, lpad operation), with xLPE=1 but ELP=NO_LP_EXPECTED an lpad must be a no-op *even if* its LPL mismatches x7[31:12] or its PC is not 4-byte aligned:

```
if (xLPE == 1 && ELP == LP_EXPECTED)
    if pc[1:0] != 0                      -> software-check exception
    else if (inst.LPL != x7[31:12] && inst.LPL != 0) -> software-check exception
    else ELP = NO_LP_EXPECTED
else
    no-op
```

In the generated tests every mismatching-label lpad executes with ELP=LP_EXPECTED, and every lpad that executes with ELP clear has LPL=0 (the RESUME sequence). `cp_disabled_zicfilp` covers only the xLPE=0 arm. So a DUT that checks the label without consulting ELP passes all three suites.

Suggest one case per suite: xLPE=1, no preceding indirect jump, `LI(x7, LABEL << 12)` then `lpad OTHER_LABEL` (and, with Zca, a misaligned `lpad 0`), followed by `check_trap_count` — plus a bin in the coverpoint and a row in the test plan.

coverpoints/norm/Zicfilp.yaml
```diff
@@ -1,92 +1,283 @@
...
+  # The Zicfilp extension adds the SPELP field that holds the previous ELP.
+  - name: vsstatus_spelp_op
+    coverpoint: ["Hypervisor: belongs to ZicfilpH"]
+
+  # On a trap caused by a software-check exception, the stval register holds the cause for the exception.
   - name: stval_exception_info
     coverpoint:
-      ["TODO: Zicfilp and Zicfiss should both cause exceptions and read xtval"]
+      [
+        "{ZicfilpS_cg,ZicfilpU_cg}/{cp_zicfilp_lpad_missing_instruction_exception,cp_zicfilp_lpad_label_mismatch,cp_zicfilp_lpad_label_exception_delivery}",
```
davidharrishmc, line 275: `mtval_swchk` (normative_rule_defs/machine.yaml — "On a trap caused by a software-check exception, the mtval register holds the cause for the exception", with the 0/2/3 encodings) is the M-mode counterpart of this rule and is not mapped anywhere in the repo, even though ZicfilpSm already covers mtval=2 via `xtval_lpad` in `cp_zicfilp_lpad_label_exception_delivery`. Worth adding:

```yaml
  - name: mtval_swchk
    coverpoint:
      [
        "ZicfilpSm_cg/{cp_zicfilp_lpad_missing_instruction_exception,cp_zicfilp_lpad_label_mismatch,cp_zicfilp_lpad_label_exception_delivery}",
      ]
```

Every other one of the 39 rule names in this file checks out against the canonical norm-rules.json (fetched 2026-09-21, 3307 rules). The only other unmapped Zicfilp-named rule is `Zicfilp_ssp_below_m_mode`, which despite its name is a Zicfiss rule.

coverpoints/priv/ZicfilpS_coverage.svh
UmerShahidengr: Done. Every cross in the S, Sm and U files now includes the mode's xLPE (`lpe_enabled`, `lpe_disabled`, or `lpe` where both values are swept). The crosses now list the input conditions the row describes (Indirect_CT kind via `lp_branch_prev`, LPL, x7 label, target alignment) instead of `elp_before`, and I took the expected outputs out of them. `sw_check_exc`/`xtval_lpad`/`spelp` stay only in `cp_zicfilp_lpad_label_exception_delivery`, whose row is about where the exception is delivered and what xtval and xPELP hold.

UmerShahidengr: Done in all three files. This one is now `cross priv_mode_s, lpe_enabled, lp_branch_prev, lpad_lpl_nonzero, lpl_match, x7_label`, ignoring the match and x7=0 bins.

UmerShahidengr: Added `lpe_enabled`. Yes, C8 should say xLPE=1. I have updated it along with C9 and C11.

coverpoints/priv/ZicfilpS_coverage.svh
UmerShahidengr: Fixed. `lpad_lpl_zero` was wrong for this row. The cross is now `lpe_enabled, lp_branch_prev, lpad_lpl_nonzero, lpl_match, pc_aligned` with the mismatch bin ignored, so it covers a matching nonzero LPL at a 4-byte-aligned target, which is what C11 describes.

On misaligned: yes, I think it should come back. The gray row says the instruction-address-misaligned exception takes priority, but that's only true without Zca. With Zca, a jump to pc[1:0]=2 is a legal fetch, and the software-check exception is the only thing that catches the misaligned LPAD. I added `cp_lpad_alignment_exception_zicfilp` (ZCA_SUPPORTED only): jalr, c.jr and c.jalr to an `lpad 0` at pc[1:0]=2. This also answers your general comment that the misaligned landing pad rule was never tested. I have un-grayed `cp_lpad_alignment_exception_zicflip` in the Common sheet, rename it to `_zicfilp`, and replace the "Redundant" note with the Zca condition. The duplicate `cp_lpad_misaligned_address_exception_zicfilp` row can stay gray.

coverpoints/priv/ZicfilpS_coverage.svh
UmerShahidengr: It wasn't necessary. The test only uses a mismatching LPL, because with xLPE=0 that's the case that would trap if the DUT ignored LPE. I dropped `lpl_match` from the cross: `cross priv_mode_s, lpe_disabled, lp_branch_prev, lpad_lpl_nonzero`. The test still jumps to `lpad 0x54321` with x7[31:12]=0x12345.

coverpoints/priv/ZicfilpS_coverage.svh
UmerShahidengr: Thanks for reminding it. Added `lpe_enabled` and `ebreak`: `cross priv_mode_s, lpe_enabled, elp_before, ebreak`, where `elp_before` has only the no_lp_expected bin. 

I didn't split it. The ELP=LP_EXPECTED half of trap entry is the software-check exception at a non-LPAD target, and `cp_zicfilp_lpad_label_exception_delivery` already covers that with SPELP=1. Sm is the same, using ecall.

coverpoints/priv/ZicfilpS_coverage.svh
UmerShahidengr: The cross is now `priv_mode_u, u_lpe_enabled, trap_case`. `trap_case` has two bins: `lp_expected_not_lpad` (the previous instruction is an ELP-setting Indirect_CT and the current one is not LPAD) and `no_lp_expected_ecall` (an ecall with no Indirect_CT before it). Sm does the same for S and U origins.

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
UmerShahidengr: Yes, all four. `trap_entry_tests`, `guarded_trap_entry_tests`, `elp_state_preservation_tests` and `trap_return_tests` are now in ZicfilpCommon and take `priv`, so ZicfilpS.py and ZicfilpSm.py are down to their registration functions. Two differences remain. The trap instruction is ecall in M and the delegated ebreak in S, because an S-mode ecall has to reach M for the T-SBI. And the xPP/xPELP writes differ (mstatush.MPELP on RV32).

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
UmerShahidengr: I'd rather not cross it. The ELP rule only looks at rs1, and the row already says rd=random for JALR, so the alternation is just a deterministic version of that. It exercises both the call form and the jump form without doubling the 124 jalr cases. I left it out of the coverpoint and testplan.

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
UmerShahidengr: Done for returns to the same mode. `_check_pelp_cleared` records SPELP after the SRET, and MPELP after the MRET (mstatush on RV32), as you suggested. For a return to a lower mode, the T-SBI ecall back up is itself a trap and rewrites xPELP from ELP before the test can read it. So the cleared value can't be observed there, and the same-mode cases cover the clause.

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
UmerShahidengr: I moved it into `cp_exception_priority_zicfilp` as the `illegal_instruction` bin of `priority_case`, since it's really a priority check (software-check exception over illegal instruction). The exception priority row already lists it as case (2), next to the instruction access fault.

generators/testgen/src/testgen/priv/extensions/ZicfilpS.py
UmerShahidengr: Yes, it's a duplicate. Part (1) is the software-check exception at a non-LPAD target, and `cp_zicfilp_lpad_label_exception_delivery` already covers it with SPELP=LP_EXPECTED. I'll remove part (1) from the S and M trap-entry rows and change their bins to 1 (ELP=NO_LP_EXPECTED).

generators/testgen/src/testgen/priv/extensions/ZicfilpCommon.py
UmerShahidengr: Good catch. Added in 929adac37. `lpad_nop_tests` now runs in all three suites with xLPE=1 and no Indirect_CT before the LPAD: with x7[31:12]=0x12345 it executes a sequential `lpad 0x54321` and, with Zca, an `lpad 0` at pc[1:0]=2, each followed by `check_trap_count`. The new cross `cp_lpad_nop_no_lp_expected_zicfilp` (`lpe_enabled, no_lp_branch_prev, lpad_nop_case`) is at 100% on sail-rv64-max and sail-rv32-max, and the suites pass on Sail, Spike, Whisper and QEMU (apart from ZicfilpSm on qemu-rv32, which is already excluded for the MPELP bug). It's mapped to `Zicflip_lpad_exception` and `Zicflip_elp_lpad_expected`, and it gets its own row in the Common testplan sheet.

coverpoints/norm/Zicfilp.yaml
UmerShahidengr: Added in 929adac37 as you suggested, mapped to the three ZicfilpSm crosses.

davidharrishmc (approved):
I have no further comments.  Ready for @jordancarlin review.

PR #2385, opened 2026-09-16
Migrate all openhwgroup links to openhwfoundation

jordancarlin:
The openhwgroup GitHub org was renamed to openhwfoundation. Update all
of the issue links, github repo clone locations, etc. accordingly.


.markdown-link-check.config
```diff
@@ -3,10 +3,6 @@
     {
       "pattern": "^https://lists.riscv.org/g/sig-arch-test",
       "reason": "Random CI False Negative"
-    },
-    {
-      "pattern": "^https://(?:[a-z0-9-]+\\.)?openhwgroup\\.org",
```
davidharrishmc, line 8: confirm this is no longer needed.

davidharrishmc (approved):

.markdown-link-check.config
jordancarlin: Not sure. I figured I would remove it and if the link checker starts failing again then we can add it back in. Hopefully the new URL is more stable and it's no longer necessary.

PR #2387, opened 2026-09-16
Fix FFLAGS coverage sampling

### LINKED ISSUE #2347: [Bug] cp_csr_fflags coverage transition bins do not detect transitions in fflags
### Affected area

Coverpoints — Coverage model or coverage generation

### Commit

9d85ed6414859a0533f89b5abc6fa837c62d61d9

### Affected test or extension

F, D, Zfh, Vf16, Vf32, Vf64

### Steps to reproduce

`make EXTENSIONS=F coverage`

### Expected behavior

cp_csr_fflags_vdoun (for example) consists of the following coverage:
```systemverilog
    cp_csr_fflags_vdoun : coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_AFTER, "fcsr", "fflags") iff (ins.trap == 0 )  {
        // Value of FCSR.fflags
        wildcard bins NV   = (5'b0???? => 5'b1????);
        wildcard bins NV1  = (5'b1???? => 5'b1????);
        wildcard bins DZ   = (5'b?0??? => 5'b?1???);
        wildcard bins DZ1  = (5'b?1??? => 5'b?1???);
        wildcard bins OF   = (5'b??0?? => 5'b??1??);
        wildcard bins OF1  = (5'b??1?? => 5'b??1??);
        wildcard bins UF   = (5'b???0? => 5'b???1?);
        wildcard bins UF1  = (5'b???1? => 5'b???1?);
        wildcard bins NX   = (5'b????0 => 5'b????1);
        wildcard bins NX1  = (5'b????1 => 5'b????1);
    }
```

Unless I'm mistaken, it looks like the coverage is meant to cover a transition from 0 --> 1 and a transition from 1 --> 1 in fflags from before an instruction runs and after an instruction retires, however this is not the behavior of systemverilog transition bins (from the systemverilog spec):

> A _trans_list_ specifies one or more sets of ordered value transitions of the coverage point. A single value transition is thus specified as follows:
> > value1 => value2
>
> It represents the value of coverage point at two successive sample points, that is, `value1` followed by `value2` at the next sample point

ACT4 functional coverage only samples instructions when they retire. This means that the value that the transition bins is sampling is the difference between when the last instruction retired and when the current instruction retired.

As written all floating point tests currently execute
```asm
fsflagsi 0b00000 # clear all fflags
```
before every instruction (verified for `fadd.s`). This means that each instruction does not have a case where a bit in fflags can transition between 1 and 1 as the instruction executes, so the floating point tests should fail coverage.

### Actual behavior

Right now, everything passes coverage with an `fflags` reset before every instruction execution.

jordancarlin:
Fixes #2347.

Sample FFLAGS before and after each instruction. Add generated cases that keep relevant FFLAGS bits set, and regenerate affected coverpoints and tests.

Generated with AI assistance.


jordancarlin:
@fourth-bit I fixed the coverpoints for vector floating point but not the tests. Can you handle the tests in #2349 once this is merged?

davidharrishmc (approved):

PR #2388, opened 2026-09-16
Python Sv generator

jordancarlin:
This can definitely still be improved further, but I think it is a good enough start and an improvement over what we had. Let's get this in and then continue cleaning it up from here.


generators/testgen/src/testgen/asm/helpers.py
```diff
@@ -76,7 +76,11 @@ def load_float_reg(
 
 
 def write_sigupd(
-    check_reg: int | None, test_data: TestData, sig_type: Literal["int", "fflags", "float"] = "int"
+    check_reg: int | None,
+    test_data: TestData,
+    sig_type: Literal["int", "fflags", "float"] = "int",
+    *,
```
davidharrishmc, line 82: Unfamiliar with this syntax, and don't see it used below.  Is it necessary?

generators/testgen/src/testgen/priv/extensions/sv/Svpbmt.py
```diff
@@ -0,0 +1,107 @@
...
+    return tests
+
+
+for sv in RV64_SV_MODES:
+    add_priv_test_generator(
+        "Svpbmt",
+        name=f"make_svpbmt_{sv.name}",
+        required_extensions=["I", sv.extension, "Svpbmt"],
+        march_extensions=_MARCH,
+        extra_defines=["#define BOOT_TO_MMODE"],
```
davidharrishmc: Should we transition to T-SBI now?

davidharrishmc (approved):
I'll approve now so you can merge when you want.  If not switching to T-SBI now, then soon?

generators/testgen/src/testgen/priv/extensions/sv/Svpbmt.py
jordancarlin: I'd like to do the T-SBI transition as a separate PR after the Python generator is merged.

generators/testgen/src/testgen/asm/helpers.py
jordancarlin: Which syntax? The `*`? If so, we use that lots of places. It means that the subsequent arguments have to be passed by keyword instead of position.

davidharrishmc:
LGTM

UmerShahidengr (approved):

PR #2392, opened 2026-09-17
Fix trap save-area instantiation order

WeiShen1011:
The trap save areas were instantiated in M-S-H-V order, while the mode-switching, trap-signature, CSR-save, and cleanup code calculates sv_area_sz offsets for an M-H-S-V layout.

This mismatch made HS-mode handlers read the S-mode save area. In particular, an HS trap could update Strap_sig instead of the shared Mtrap_sig pointer and place a correct HS signature in an M-mode slot.

Instantiate H before S so the physical layout matches the existing offset model used throughout the trap framework. Changing the single ordering point is safer than rewriting several related offsets and keeps M, H, S, and V save-area consumers consistent.


davidharrishmc:
Is this a duplicate of #2363?

WeiShen1011:
Yes, #2363  already has the same commit, I will close this request, thanks for reminder.

Issue #1871, opened 2026-07-14
Remove all Occurences of RVTEST_VECTOR


fourth-bit:
RVTEST_VECTOR is an old macro that should be dropped from testgen. It is currently present in the failure code.

JyoshnaKarri:
I'd like to work on this. I've traced the current `RVTEST_VECTOR` usage across `rvtest_failure_code.h` and the test generator's extension files, and confirmed `ZVL32B_SUPPORTED` is already checked via `#ifdef` in `rvtest_setup.h`/`rvtest_failure_code.h`. However, I can't find where `ZVL32B_SUPPORTED` actually gets defined (likely a compiler flag from the build system rather than source) — could you point me to that, so I make sure the replacement is wired up correctly?"

jordancarlin:
`ZVL32B_SUPPORTED` is in the UDB generated header.

JyoshnaKarri:
Thanks, that clears it up! I'll go ahead and replace the `RVTEST_VECTOR` references with `ZVL32B_SUPPORTED` in the test env files and remove the manual `#define RVTEST_VECTOR` lines from the generators, since UDB already provides `ZVL32B_SUPPORTED` automatically.

PR #2397, opened 2026-09-17
Drop CI exclusions that now pass

davidharrishmc:
Remove CI exclusions for suites that now pass.

- CVW: `Sv` passes on `cvw-rv32gc` and `cvw-rv64gc`.
- Whisper: `SsstrictS` and `SsstrictU` pass.
- Keep the exclusions for `SsstrictSm`, `Smstateen`, and `Sdtrig*`.


jordancarlin:
I'm concerned that something has gone wrong with the Sv tests. The commit of CVW we are using in CI is still pinned to a version of CVW from before the fixes mentioned in this PR, so they should not be passing.

davidharrishmc:
Are you sure?  I thought I saw an automatic update.

jordancarlin:
CVW is currently pinned to commit `79562b5677599c9a731e21b3da2aba33baa27c27`, which is the last commit from Sept 13. 1871 was merged Sept 15.

davidharrishmc:
The test started passing on July 20 with #1822 that changed rvtest_setup.h, breaking the conditions for the failure to manifest.  

Then cvw #1871 fixed the root cause, then cvw #1875 oversimplified the fix and reintroduced a failure on old RTL.  See  https://github.com/openhwfoundation/cvw/issues/1885.

jordancarlin (approved):

PR #2399, opened 2026-09-17
Align misaligned LR/SC fault ordering across simulator configs

davidharrishmc:
## Problem

`ExceptionsSvZalrsc` test case 6 issues a misaligned `lr.w` to a page whose PTE has been made invalid. Two faults compete, and which one is reported depends on whether the model resolves misalignment before or after address translation.

The expected signature is produced by Sail from each config's `sail.json`, which has two separate controls, as its own comments explain:

- `memory.misaligned.exceptions` is evaluated **before** translation, so a fault there outranks a page fault.
- `regions[].attributes.misaligned_exceptions` is evaluated **after** translation.

## What this changes

**imperas-rv64-max.** Its `imperas.ic` sets `cpu*/unaligned_low_pri=T`, so the ISS translates first, but its `sail.json` asked Sail to fault on the alignment first. The global `lrsc` entry moves to `{ "None": null }` so the reference matches the ISS.

**imperas-rv32gck and imperas-rv32gcv are deliberately left alone.** Neither sets that override, and both already declare alignment-first in `sail.json`, so they are internally consistent and pass as they are. Leaving them keeps one Imperas config on each ordering, matching the existing practice in this directory of splitting `unalignedV` across configs "to test both cases".

**QEMU.** A misaligned LR/SC raises cause 4, not cause 5. Confirmed with `qemu-system-riscv{32,64} -d int`: for the same invalid PTE, the aligned access reports `load_page_fault` and the misaligned one reports `misaligned_load`. `LRSC_MISALIGNED_BEHAVIOR` is corrected to `"always raise misaligned exception"`, matching the `AlignmentException` already in its `sail.json`.

**Imperas CI list.** The note claimed no Imperas parameter changes this priority. `unaligned_low_pri` does, so the stale exclusion and note are removed. Imperas CI does not run publicly, so this list only tracks incomplete features. The QEMU exclusion stays, since its note already records the real cause.

## Verified correct, unchanged

`spike-rv32-max`, `spike-rv64-max`, `whisper-rv32-max`, `whisper-rv64-max`, `cvw-rv32gc`, `cvw-rv64gc`. Each was checked by flipping its global `lrsc` entry and confirming the run went from pass to fail, so the pass is a real signal and not a skipped case.

The two CVW configs legitimately disagree. `trap.sv` orders the misaligned cause on `ZICCLSM_SUPPORTED`, which is 0 on rv32gc and 1 on rv64gc, so rv32gc is alignment-first and rv64gc is translation-first.

## Results

Verified on a clean checkout of this branch against current `act4`.

| config | suite | before | after |
| --- | --- | --- | --- |
| imperas-rv64-max | ExceptionsSvZalrsc | 3 fail | 3 pass |
| imperas-rv32gck | ExceptionsSvZalrsc | 3 pass | 3 pass |
| spike-rv64-max | ExceptionsSvZalrsc | 3 pass | 3 pass |
| whisper-rv64-max | ExceptionsSvZalrsc | 3 pass | 3 pass |
| cvw-rv64gc | ExceptionsSvZalrsc | 3 pass | 3 pass |
| qemu-rv64-max | ExceptionsSvZalrsc | 3 fail | 3 fail |

Wider checks on `imperas-rv64-max` with this branch: all eleven Sv privileged suites pass, 175 tests, and a full unfiltered run passes 1005 of 1005. `ExceptionsSvZaamo` passes on every config listed above, including QEMU.

QEMU's remaining `ExceptionsSvZalrsc` failure is test case 3, an `sc.w` to `RVMODEL_ACCESS_FAULT_ADDRESS` returning 1 where 0 is expected. It hits M-mode as well as S-mode and U-mode, involves an aligned access to a valid PTE, and is already recorded in the QEMU CI exclusion. It is unrelated to this change.

## Note for reviewers

`MISALIGNED_LDST_EXCEPTION_PRIORITY` cannot express any of this. It is defined over load, store and AMO exceptions, and UDB validation forbids `high` whenever `MISALIGNED_MAX_ATOMICITY_GRANULE_SIZE` exceeds 1, which is true of every simulator config here. A misaligned LR/SC always faults and is never split, so the atomicity argument behind that restriction does not apply to it, and its ordering is currently expressible only in `sail.json`. A dedicated parameter would let the yaml and the reference be checked against each other rather than drifting. Happy to open a UDB issue if reviewers agree.

Two unrelated things found while investigating, not addressed here:

- `whisper-rv32-max-spike-ref` and `whisper-rv64-max-spike-ref` fail UDB validation outright on undefined parameters `STVEC_MODE_DIRECT`, `STVEC_MODE_VECTORED`, `VSTVEC_MODE_DIRECT` and `VSTVEC_MODE_VECTORED`, so they cannot build. Note that they use Spike as the reference model, and Spike and Whisper genuinely disagree on this ordering, so they will likely still fail this suite once they build.
- `cvw-rv32imc` declares a global `lrsc` of `AccessFault` while its RTL would raise cause 4. Dead today, since that config has neither Zalrsc nor Zaamo.


config/imperas/imperas-rv32gck/imperas.ic
```diff
@@ -61,6 +61,7 @@
 
 # Misaligned access (Zicclsm)
 --override cpu*/unaligned=T
+--override cpu*/unaligned_low_pri=T
```
jordancarlin: Can we leave the Imperas config alone and get the sail config to match it?

jordancarlin (approved):

PR #2401, opened 2026-09-17
Fix RV32 trap diagnostics; make the Imperas configs match the model

davidharrishmc:
This PR fixes mismatches exposed by `Ssstrict` and corrects RV32 trap diagnostics.

- Store the fast-trap xEPC, xCAUSE, xTVAL, and xSTATUS values at the fixed 8-byte offsets used by the diagnostic data layout. RV32 failures now report the correct fields.
- Align the Imperas configs with the model by removing unsupported H, Sh*, Ssqosid, and Smcntrpmf extensions, and disabling Ssdbltrp on `imperas-rv32gck`.
- Mark unsupported vector extensions as unavailable in the `imperas-rv32gcv` Sail config.
- Keep the remaining `Ssstrict` CI exclusions for known Imperas and Sail differences.


jordancarlin:
> The residual `Ssqosid` / `Smcntrpmf` mismatch is real — the configs declare both and `imperas.ic` enables neither. Fixing it means either enabling them in `imperas.ic` or dropping them from the configs, which is a separate decision from this PR.

Seems like something that should be fixed in this PR too.

davidharrishmc:
Agreed.  Ready.

jordancarlin (approved):

PR #2408, opened 2026-09-18
Add `CERTIFICATE` option

jordancarlin:
Add an option to specify a certificate. When a certificate is selected,
only tests that are part of that certificate are build. M-mode tests and
tests for extensions the DUTs supports that are not part of the
certificate are skipped.


jordancarlin:
The actual list of tests will need to be checked carefully as one of the last steps before launch since the names of some suites keep changing and new suites are added, but it should be close to accurate and good enough for now.

davidharrishmc (approved):

davidharrishmc:
Looks good.  Doesn't have to be this PR, but eventually there should also be a list of required UDB parameter values.  For example, MXLEN=64.  Probably some of the XTVAL parameters, etc.

jordancarlin:
> Looks good. Doesn't have to be this PR, but eventually there should also be a list of required UDB parameter values. For example, MXLEN=64. Probably some of the XTVAL parameters, etc.

That should be handled by UDB. There is as `profile` option in the UDB config that will validate the config against the specified profile. I believe there are a few bugs in that feature that still need to be cleaned up.

PR #2413, opened 2026-09-18
Sdtrig: A-extension trigger tests and icount coverpoints

pclarkhmc:
Implements the Sdtrig testplan's A-extension trigger coverpoints and enables the icount ones, for all three modes (`Sm`/`S`/`U`).
## Icount
- `cp_sdtrig_icount_trap` is commented out. It needs the trap taken in a mode where the trigger is *not* also enabled; for `Sm` (M→M) and `S` (S→S) the breakpoint would fire inside a trap handler that is not reentrant. I am waiting on 10x to help out with their reeentrancy for for itrigger/etrigger.
- `cp_sdtrig_icount_hardwired` stays gated on `UDB_ICOUNT_HARDWIRED_1`, which no config defines, so it contributes no executable testcases currently.

## Verification
`EXTENSIONS=SdtrigSm,SdtrigS,SdtrigU EXCLUDE_EXTENSIONS= DEBUG=True   make whisper-rv64-max-spike-ref whisper-rv32-max-spike-ref`

## Note for reviewers
`config/whisper/whisper-rv{32,64}-max-spike-ref/*.yaml` still carry `STVEC_MODE_DIRECT`/`STVEC_MODE_VECTORED` and `VSTVEC_MODE_*`, which UDB has renamed to `STVEC_MODES: [0, 1]` (+ `STVEC_BASE_ALIGNMENT_VECTORED`) and `VSTVEC_MODES: [0, 1]`. Fixed in PR 2414


pclarkhmc:
Both blockers in the "Note for reviewers" section are now fixed in #2414 — the UDB stvec parameter rename and the Sail platform defines. With that PR applied, this one builds and runs on `whisper-rv{32,64}-max-spike-ref` directly.

davidharrishmc:
Claude review.  Some of this is not applicable.

# Review: riscv/riscv-arch-test#2413 — Sdtrig A-extension and icount tests

- PR head 379237060 (pclarkhmc, branch sdtrig-icount-aext), 44 commits behind act4; not yet reviewed by anyone else.
- Compared against the Sdtrig testplan (Google Sheet gid 1209377334): rows cp_lrsc_addr, cp_lrsc_data, cp_amo, cp_icount_hardwired/instr/trap/eq0.
- Reviewed 2026-09-24. Not posted.

## What I ran

- Sail 0.14.1 has no Sdtrig. On spike-rv64-max (Sail as the reference model), every `csrw tdata1` traps as an illegal instruction:
  - SdtrigSm_Icount's "signature" is 40 illegal-instruction traps.
  - SdtrigS_Icount aborts inside T-SBI.
  - So only the Spike-reference configs can validate this PR.
- On whisper-rv{32,64}-max-spike-ref, with #2414 cherry-picked (as the PR asks): all 18 tests pass on each XLEN.
  - The trap reports have exactly the breakpoint counts I derived by hand:
    - AExt: 88 on RV64, 44 on RV32.
    - Icount: 4 per mode.
  - So Spike and Whisper both fire where the generator expects.
- The checked-in tests/ match the generator.

## Per coverpoint

| Plan row | PR | Verdict |
|---|---|---|
| cp_lrsc_addr | tdata2 {scratch,0} × XSL {100,010,001} × LR/SC × {.w, .d on RV64}, per trigger | Matches the plan's stimulus. Weak evidence (see 2). |
| cp_lrsc_data | select=1, tdata2 {dataval, 0}, .w uses the low 32 bits | Matches (a random value instead of 0x1234…, which is fine). Same weakness. |
| cp_amo | 9 Zaamo ops × {.w,.d} × 2 tdata2 × 3 XSL, rd≠x0 | Superset of the plan (amoswap only). The extra ops fit no planned bin (see 5). |
| cp_icount_hardwired | gated on `UDB_ICOUNT_HARDWIRED_1`, which is never defined; sets only the current mode's bit | **Not implemented in practice**; also deviates from the plan (see 3). |
| cp_icount_instr | count=9, priv {0, current mode}, reads tdata1 | Stimulus matches. **Does not check the count** (see 1). tval is checked implicitly by the trap handler. |
| cp_icount_trap | entirely commented out | **Not implemented**; the stated reason is wrong (see 4). |
| cp_icount_eq0 | pending {0,1}, count=0, marker addi, reads tdata1 | Good: the marker proves whether and where it fired. |

## Findings, most important first

1. **cp_icount_instr does not check that the trigger fires after 9 instructions.**
   - All Sdtrig tests define `SDTRIG_IMPRECISE_XEPC`, so the trap handler does not record xEPC for breakpoints.
   - The only evidence left is the tdata1 readback (count=0, pending=0) and the fact that one breakpoint occurred.
   - A DUT that fires after 8 or 10 instructions, or that decrements on the wrong events, passes.
   - This defeats Sdtrig_icount_match1 and match_gt1, which are the point of the row.
   - Fix: use the marker-sum technique that eq0 and the commented-out trap test already use. Put distinct-weight `addi` markers around the 9th and 10th positions and sigupd the sum. The handler skips the instruction the trigger fires on, so the sum names the fire position.

2. **LR/SC tests cannot tell which instruction fired.**
   - Each testcase runs LR then SC under one trigger configuration, with no sigupd in between.
   - The only signature content is the trap record (cause, tval=scratch).
   - A DUT that treats LR as a store and SC as a load gives a bit-identical signature. So does one that fires a load trigger on SC instead of LR.
   - Sdtrig_Aext_lr_op and Aext_sc_op are therefore not actually distinguished.
   - Fix: preload rd with a sentinel and sigupd after each of LR and SC. For SC, record only "skipped or not" (rd == sentinel), because SC success after a trapped LR is implementation-defined.
   - AMO is fine: each testcase has one instruction under one XSL setting.

3. **cp_icount_hardwired never runs and differs from the plan.**
   - `//#define UDB_ICOUNT_HARDWIRED_1` is hard-coded and commented out, and nothing lets a DUT set it. The PR says so, but the title still says it "enables the icount coverpoints".
   - The converse problem: on a DUT with count hard-wired to 1 (which the spec explicitly permits), cp_icount_instr and cp_icount_eq0 will fail, with no opt-out.
   - The plan says set m/s/u/vs/vu = 11111 and check that all are cleared.
     - The PR sets only the current mode's bit, so the "all cleared" part of Sdtrig_icount_match_eq1 is not tested.
     - In Sm, 0b11111 is safe: unsupported bits are WARL 0.
     - In S/U, m must stay 0, because the tdata1 write runs in M via T-SBI. But setting s|u (and vs|vu) would still test that the lower-mode bits clear.

4. **cp_icount_trap is dropped for a reason the spec already handles.**
   - The PR says Sm (M→M) and S (S→S) are infeasible because "the breakpoint would fire inside a trap handler that is not reentrant".
   - But Sdtrig's native-trigger rules (option 1, which this suite relies on through `_global_ie` setting MIE/SIE) forbid action=0 triggers from matching or firing in M while MIE=0.
   - Walkthrough:
     - The illegal-instruction trap from M with MIE=1 matches, and count 1→0 sets pending.
     - The handler runs with MIE=0, so nothing fires.
     - After mret, pending fires before the next instruction.
   - This is exactly the case the spec's `pending` note describes. It is what the plan asks for ("illegal instruction trap in mmode").
   - The 10x itrigger/etrigger reentrancy issue the PR cites is a different mechanism: those triggers fire at handler entry.
   - The U-mode variant (U→S with the trigger disabled in S) was written and works, but is commented out too.
   - Note for Sm: the commented code's `csrr 0x300` is not illegal in M. A real illegal encoding is needed.
   - Result: Sdtrig_icount_match2, multiple_match, match_non_reexecuted and all_traps_match remain uncovered.

5. **cp_amo scope.**
   - The plan specifies amoswap only (2 × 3 bins).
   - The PR adds 8 more ops × 2 widths × 2 triggers: 72 testcases on RV64 where the plan has 6 bins.
   - This is defensible, since the rule says "each AMO instruction". But either the plan and covergroup should add an op dimension, or the test should trim to amoswap. Otherwise most of these testcases map to no coverage bin.
   - Plan erratum: the cp_amo expectation says "mem access **data** matches tdata2", but the stimulus is select=0 (address). The PR correctly uses address.

7. **Hard-coded capability defines** (existing pattern, extended here).
   - `UDB_ICOUNT_TRIG{n}_AVAILABLE` is emitted into every test for every trigger. Any DUT with Sdtrig but without icount (many implement only mcontrol6) fails the Icount tests, with no switch to turn them off.
   - Same for `UDB_SDTRIG_MCONTROL6_SELECT_AVAILABLE`: cp_lrsc_data uses select=1 with no guard at all.
   - UDB has no Sdtrig parameters yet, so the proper fix is upstream. Meanwhile the defaults could at least be overridable.


## Answer

- **Partially.**
  - The LR/SC/AMO stimulus matches the plan (AMO over-delivers).
  - cp_icount_eq0 is done properly.
- **But:**
  - cp_icount_instr's core check (fires after N) is not observable.
  - LR versus SC is not distinguished.
  - cp_icount_hardwired and cp_icount_trap are not implemented, and the trap case is dropped for a reason that doesn't hold in M-mode.

pclarkhmc:
Thanks for the feedback, I'll update the test plan and tests.

pclarkhmc:
Looks like Claude is referencing an older version of the test plan for 5. I
already updated the test plan with the new AMO scope. You might want to
update its context. Otherwise, I implemented the changes you suggested.

PR #2418, opened 2026-09-18
Sdtrig mcontrol6 Covergroup

AngelaZzzzz:
Add the SdtrigSm_mcontrol6_cg covergroup, and moved some common coverpoints for later Sdtrig covergroups into the shared sdtrig include.


AngelaZzzzz:
cp_sdtrig_mcontrol6_chain_adr currently only checks Trigger i+1 since sw would only expose the state of Trigger i+1 while the test plan tests the Trigger i and Trigger i+1 pair for chaining. The pair is only implicitly checked via triggernum_chain, which checks that the tselect values sweep 1 through NUM_TRIGGERS-1. Would this be sufficient for this coverpoint, or should the chain be more explicitly tested? I'm thinking about creating another point with ins.prev to explicity test that the previous instruction's t_select is i, but aren't sure how helpful this would be.

coverpoints/general/RISCV_coverage_sdtrig_coverpoints.svh
```diff
@@ -12,11 +12,61 @@
...
+    bins size[] = {[0:6]};
+}
+
+// tdata2 of the selected trigger against the sampled access
+tdata2_adr: coverpoint (ins.current.csr[CSR_TDATA2] == ins.current.mem_addr) {
+    type_option.weight = 0;
+    bins scratch = {1'b1};
+    bins zero    = {1'b0} iff (ins.current.csr[CSR_TDATA2] == '0);
+}
+tdata2_data: coverpoint (ins.current.csr[CSR_TDATA2][31:0] == (ins.current.has_rd ? ins.current.rd_val[31:0] : ins.current.rs2_val[31:0])) {
```
davidharrishmc: Why does this one only match on 32 bits?

davidharrishmc (changes requested):
starting review

coverpoints/priv/SdtrigSm_coverage.svh
```diff
@@ -74,6 +65,195 @@ covergroup SdtrigSm_trig_module_reg_cg with function sample(ins_t ins);
...
+        bins mask_high     = {4'd5};
+        bins not_mask_low  = {4'd12};
+        bins not_mask_high = {4'd13};
+    }
+    tdata1_chain_disabled: coverpoint ins.current.csr[CSR_TDATA1][11] {
+        bins disabled = {1'b0};
+    }
+    tdata2_pc: coverpoint (ins.current.csr[CSR_TDATA2] == ins.current.pc_rdata) {
+        bins pc   = {1'b1};
+        bins zero = {1'b0} iff (ins.current.csr[CSR_TDATA2] == '0);
```
davidharrishmc: Why is the iff needed?

davidharrishmc:
AI review# ACT PR #2418 — Sdtrig mcontrol6 covergroup (AngelaZzzzz)

Reviewed 2026-09-24. Head `8254848b5`, base act4 `54cfe21bb`. 3 files: shared sdtrig coverpoints, SdtrigSm_coverage.svh (+SdtrigSm_mcontrol6_cg, 11 crosses), init.
Test plan: Google sheet 1jvJCSf9..., gid 1209377334 (Sdtrig tab), rows 43-61 (mcontrol6 + chain). Exported as CSV (Drive connector could not see it).

Coverage cannot be measured: Sdtrig not in Sail (excluded in CI). Review is static: coverpoints vs test plan vs the existing generator `SdtrigCommon._generate_mcontrol6_tests` (already on act4).

Checked: compiles (Questa vlog, +define+SDTRIGSM_COVERAGE, rv64 and rv32 UDB headers: 0 errors). Norm map coverpoint names == PR crosses (52 refs, none missing). `ifndef` with no macro name in shared file fixed by PR (real bug).

## Crosses vs stimulus actually generated

| cross | PR requires | generator emits | result |
|---|---|---|---|
| priv_mode | sw, xsl 010, adr, m 0/1 | same | OK, but fire bin needs mem_addr of a trapped sw |
| execute_adr | `nop` insn | `add x0,x1,x2` at label | pc bin never; zero bins hit only by nop spacers of other sections |
| execute_data | tdata2 {NOP,0}, insn NOP | tdata2 {ADDI_HINT 0x8013, TDATA2_NOFIRE 0x2}, insn addi x0,x1,0 | 0% |
| load_store_data | tdata2 zero bin iff tdata2==0 | nofire tdata2 = match^1 | zero bins never; LW fire uses rd_val of trapped load |
| execute_size | xsl 111, tdata2/insn NOP | xsl 100, ADDI_HINT | 0% (c.nop half also fails on xsl) |
| load_store_size | size 0..6, Q bins | sizes 0,1,2,3,5; no Q | size 4,6 and Q holes |
| match | xsl 010 | xsl 011 | 0% |
| match_napot | xsl 010; tdata2 walking zero | xsl 011; tdata2 all ones, data walks (TODO) | 0% |
| match_mask | xsl 010 | xsl 011 | 0% |
| chain_adr | tdata1/2 of trig i+1, imm 0/8, data | same | OK; trig i never observed |

Plan says nop / XSL 010 / XSL 111 / tdata2 walk — PR follows plan, generator deviates (ADDI_HINT: U-mode T-SBI nops trip triggers).

## Other findings
- No per-trigger / per-feature gating: triggernum bins all N; generator gates on UDB_SDTRIG_MCONTROL6_SUPPORTED{n}, *_SIZE/MATCH/CHAIN/EXECUTE/LOAD/STORE_AVAILABLE.
- Fire cases of address/data load-store crosses use mem_addr / rd_val of an instruction that took a breakpoint before executing; use rs1_val+imm.
- Crosses don't pin m=1, match=0, size=0, chain=0; with persistent tdata1 other sections' stimulus fills bins.
- NAPOT: tdata2 WARL (maskmax6) makes high zero-bit bins unreachable on smaller maskmax6.

coverpoints/general/RISCV_coverage_sdtrig_coverpoints.svh
AngelaZzzzz: Ah, I think it's cuz claude read the XLEN parameter and took 32 from there. Will change to the entire CSR

PR #2420, opened 2026-09-19
fcov: let the covergroups compile for a DUT without the vector extension

davidharrishmc:
## Problem

Two things stop the functional coverage sources compiling for a DUT that has no vector extension.

**1. `UDB_VLEN` expanded unguarded.** `coverage/RISCV_instruction_base.svh` expands `` `UDB_VLEN `` in `get_vr_val`, `get_vr_val_lmul4` and `get_vr_val_lmul8` (via `` `SIGNED_VLEN_BITS `` and directly). A configuration only defines `UDB_VLEN` when the DUT implements the vector extension, so `vlog` reports:

```
(vlog-2163) Macro `UDB_VLEN is undefined.
```

Guard the three helpers with `` `ifdef ZVL32B_SUPPORTED ``, the extension every vector configuration carries and the macro the generated coverpoints already gate their vector code on. The one existing `` `ifdef UDB_VLEN `` around the vector include in `RISCV_coverage_base.svh` switches to the same macro, so vector code is gated on a single name throughout.

**2. A stale generated covergroup.** The committed `coverpoints/priv/ExceptionsZicboSm_coverage.svh` contains two stray backticks, one on a line of its own and one immediately before an `` `endif ``. The preprocessor reads them as macro references:

```
(vlog-13483) Illegal space between ` and compiler directive or macro name.
(vlog-2163) Macro `cbo_instrs is undefined.
```

This is **not** a generator bug: running `make covergroupgen` regenerates the file without the backticks, and the result is byte-identical to the fix here. The checked-in copy had simply fallen out of step with the generator. No other generated file changes.

## Effect

With both fixed, the covergroups compile cleanly for a scalar DUT. Verified by building all 97 coverage groups applicable to CORE-V Wally's `rv64gc` configuration into its RTL testbench and sampling them off the RVVI trace: 0 compile errors, 540 covergroup types, and 99.45% covergroup coverage across the riscv-arch-test suite.


jordancarlin:
Why does ExceptionsZicboSm coverage work in this repo with those extra backticks? Presumably that should be throwing an error for us too.

davidharrishmc:
It was introduced by https://github.com/riscv/riscv-arch-test/commit/ba53eb88ad021dd69419cacfcfc9a3f8c104f988#diff-416f4ede136cc10c4028d13c4997e27f08fc44e70487d632587d48dfa385b142

coverage is not in CI, so we missed it.

UmerShahidengr (approved):

PR #2425, opened 2026-09-19
fcov: emit one write per register in the RVVI trace and tokenize it in one pass

davidharrishmc:
## Summary

Two changes that together make the SEW=8 vector coverage groups roughly 4x faster and cut tens of GB from a coverage build.

**1. `sail_to_rvvi.py` emitted every element write.** Sail logs each element a vector instruction writes as a separate whole-register update, and the converter copied each one into the trace. A `vle8.v` at LMUL=8 with VLEN=1024 emitted 1,024 `V` entries of 256 hex digits each, plus an `mstatus` and a `vstart` write per element — a **345 KB line for one instruction**. `testbench.sv` keeps only the final value of each register (`v_wdata[regNum] = vRegVal`), so all but eight of those entries were dead bytes.

Keep the last write per register, in first-write order. Output is byte-identical to the current traces with duplicate register writes removed.

| Trace | Before | After | |
|---|---|---|---|
| `Vx8-vssra.vv-00` | 154 MB | 7.2 MB | 21.5x |
| `Vls8-vsoxseg8ei8.v-00` | 99 MB | 3.5 MB | 28.4x |
| `Vx32-vadd.vv-00` | 19.4 MB | 3.1 MB | 6.2x |

Whole groups in `sail-rv64-max`: `Vx8` 26 GB, `Vls8` 19 GB.

**2. `splitLine` was quadratic in the line length.** For each word it `$sscanf`-scanned the remaining string and then copied it with `substr` — roughly 1.4 GB of copying for a 345 KB line. Scan the characters once instead.

## Measurements

One `vsim` over the 154 MB `Vx8-vssra.vv-00` trace, compile excluded:

| Reader | Trace | Time | Bins hit |
|---|---|---|---|
| current | as shipped | 13 s | 1,925 / 89,955 |
| linear tokenizer | as shipped | 7 s | 1,925 |
| current | deduplicated | 3 s | 1,925 |
| linear tokenizer | deduplicated | 4-5 s | 1,925 |


jordancarlin (approved):
This is all going to change dramatically when we switch to RVVI-Text in the next few weeks, but seems like a reasonable speedup for now.

PR #2430, opened 2026-09-19
Port Sv suites to T-SBI and split M-mode coverpoints into Sm suites

UmerShahidengr:
Converts the virtual-memory suites to T-SBI. Tests boot to S-mode and use the standard
`RVTEST_TSBI_GOTO_*` macros. Coverpoints that need M-mode move to new Sm suites.

- **Suites converted:** Sv, Svade, Svadu, Svbare, Svinval, Svnapot, Svpbmt, SvZicbo.
  menvcfg is written through T-SBI CSR calls, MXR/SUM through sstatus.
- **New Sm suites:**
  - SvSm: MPRV, TVM, SBE and M-mode satp access
  - SvbareSm: MPRV with satp.MODE=Bare
  - SvinvalSm: Svinval with mstatus.TVM=1
- **Unchanged:** SvPMP, SvPMPZicbo and SvaduPMP stay M-boot because they program PMP.
- **Trap handler:** the T-SBI GOTO paths now relocate the resume address when satp.MODE != Bare
  and the test has registered a code alias with `SAVE_AREA_SETUP`. GOTO_UMODE moves into the
  U-accessible alias, and GOTO_S/MMODE from the alias move back to the identity-mapped code.
  Suites that never register an alias are unaffected.
- **Coverage:** Sv-family fault coverpoints sample scause/stval, since faults are now delegated
  to S. The satp access cross is split between Sv (S/U) and SvSm (M).
- **Normative rules:** Sv*.yaml files are mapped to coverpoints (165 rules), stale rule names
  and coverpoint references are fixed, and YAMLs are added for Svade, SvPMP, SvaduPMP and the
  new Sm suites. The CTP chapter `sv.adoc` includes all of them.


davidharrishmc:
CI troubles

davidharrishmc:
Claude review:

# Review: riscv-arch-test PR #2430 — Port Sv suites to T-SBI and split M-mode coverpoints into Sm suites

- **PR:** https://github.com/riscv/riscv-arch-test/pull/2430
- **Author:** Umer Shahid · **Head:** `c16503c25` (`sv-tsbi`) · **Base:** `act4` @ `fcf1067c1` (5 commits behind tip)
- **Size:** 256 files, +8499 / −9486 (206 of them generated tests)
- **Reviewed:** 2026-09-19

---

## Summary

The port itself is solid: generated tests are in sync with the generators, everything lints, and the
whole privileged regression passes on two reference models. The normative-rule mapping work bundled
with it is thorough and mechanically correct.

One real defect: `TSBI_RELOCATE_EPC` reads `satp` from the S-mode trap handler, which is an illegal
instruction when `mstatus.TVM=1`. It is not hit today only by luck. One coverage hole:
`SvinvalSm_cg/cr_svinival` has cross bins that no test can reach.

---

## Verification performed

Worktree: `/tmp/claude-1001/-home-harris-cvw/33e4d8ec-ac13-4c3c-a9f5-b9acefe57d53/scratchpad/act-pr2430`
Baseline worktree (for the pre-existing-failure check): `.../scratchpad/act-base` (`origin/act4`).

| Check | Result |
|---|---|
| `uv run testgen testplans -o …` regen vs. committed `tests/priv/**`, all 14 `Sv*` suites | **identical** — generators and checked-in tests agree |
| `uv run ruff check` / `ruff format --check` / `pyright` | clean (0 errors, 0 warnings) |
| Sail 0.14.1 `sail-rv64-max`, `Sv*` suites | **199/199 pass** |
| Sail 0.14.1 `sail-rv32-max`, `Sv*` suites | **60/60 pass** |
| Sail `sail-rv64-max`, **all 63 privileged suites** (shared trap-handler blast radius) | **650/650 pass** |
| Sail `sail-rv32-max`, all privileged suites | **506/506 pass** |
| Spike `spike-rv64-max`, all privileged suites | 370/371 — only `Smstateen-00` fails, and it fails identically on the `act4` baseline → **pre-existing, unrelated** |
| Dangling covergroup/coverpoint references in changed `coverpoints/norm/*.yaml` | **none** (the 9 my checker flags are pre-existing `{s, u}` spacing typos in the untouched `Svnapot.yaml`) |
| Normative-rule names vs. canonical `norm-rules.json` | **all resolve**, including `satp-asidlen` → `ASIDLEN` and `satp-mode_roz_sfence_illegal` → `SATP_MODE_ROZ_SFENCE_ILLEGAL` (correctly rehomed to `Svbare.yaml`) |
| Canonical `Svinval`/`Svnapot`/`Svpbmt`/`Svadu`/`Svade` rules left unmapped | **none** |
| Empty `coverpoint: [""]` entries in touched YAMLs | **113 → 0** |

Repro for the model runs:

```bash
EXTS=$(uv run python -c "from testgen.priv import get_priv_test_suites; print(','.join(sorted(get_priv_test_suites())))")
make sail-rv64-max  EXTENSIONS=$EXTS JOBS=16
make sail-rv32-max  EXTENSIONS=$EXTS JOBS=16
make spike-rv64-max EXTENSIONS=$EXTS JOBS=16
```

---

## 1. (High) `TSBI_RELOCATE_EPC` reads `satp` from the S-mode handler — fatal under `mstatus.TVM=1`

`tests/env/rvtest_trap_handler.h:432` does `csrr T4, CSR_SATP`, and the macro is instantiated into the
**S-mode** `tsbi_goto_mode` at line 1798. Reading `satp` in S-mode with `mstatus.TVM=1` raises an
illegal instruction, so any `RVTEST_TSBI_GOTO_*` issued from S- or U-mode by a test that has both TVM
set *and* a code alias registered re-enters the S trap handler and never comes back.

### Reproduced

Took `tests/priv/Sv/Sv_sv39_spage_access_Umode-00.S`, set `mstatus.TVM` through a T-SBI CSR call, and
appended a `GOTO_UMODE` / `GOTO_SMODE` pair just before `RVTEST_CODE_END`, with `TRAP_SIGUPD_COUNT`
raised to 300. The byte-identical control (same excursion, TVM left clear) **passes**; with TVM set
the test runs away and dies on trap-signature overflow even at 300 entries.

Sail trace pins it exactly:

```
[3331] [HS]: 0x800023A8  beq   t2, s0, ...      tsbi_Sgoto_mode+60   ; alias != code_begin → no skip
[3332] [HS]: 0x800023AC  csrrs x9, satp, x0     tsbi_Sgoto_mode+64   ; <-- TSBI_RELOCATE_EPC
[3333] [HS]: 0x80002040  ...                    Strampoline+0        ; <-- illegal instruction
```

### Why nothing hits it today

Only by luck. The two tests that set TVM — `SvinvalSm.make_svinvalsm_tvm` and
`SvSm.make_svsm_mstatus_tvm` — do not call `begin_sv_test`, so the S save area still holds its static
initializer (`.dword rvtest_code_begin`, line 3185). `code_bgn(S) == rvtest_code_begin`, the `beq` at
line 431 short-circuits, and the `csrr` is never reached. The next test someone writes — "Sv with
TVM", "Svinval with translation on" — trips it immediately.

### Suggested fix

The `satp` read is only load-bearing for the `GOTO_UMODE` direction (deciding whether an EPC inside
the code region should move into the alias). The `GOTO_M/SMODE` direction is already safe on the range
check alone: an EPC in the identity-mapped code region cannot fall inside the alias range. Either:

- gate the whole macro on an explicit opt-in word the test writes alongside the alias, or
- drop the `csrr` and treat "an alias is registered" as implying translation is on — true for every
  `begin_sv_test` test, which enables `satp` immediately after `SAVE_AREA_SETUP` and never returns to
  Bare.

---

## 2. (Medium) `SvinvalSm_cg/cr_svinival` has unreachable cross bins

`generators/testgen/src/testgen/priv/extensions/sv/SvinvalSm.py:29-31` sets `mstatus.TVM` before *any*
Svinval instruction executes, so `ins.prev.csr[MSTATUS][20]` is 1 for every sample this suite produces.
But the cross only ignores lower-privilege × TVM=0:

```systemverilog
cr_svinival : cross cp_instr, cp_priv, cp_tvm {
    // each instruction executed in M mode with each TVM, and in S and U mode with TVM set
    ignore_bins lower_tvm_zero = (binsof(cp_priv.S_mode) || binsof(cp_priv.U_mode)) && binsof(cp_tvm.zero);
}
```

`M_mode × cp_tvm.zero` survives for all four instructions. `sfence.vma` gets picked up incidentally
from M-boot suites, but `sinval.vma`, `sfence.w.inval` and `sfence.inval.ir` in M-mode with TVM=0
appear nowhere in the repo: this PR dropped the `M_mode` bin from `Svinval_cg`, and `SvinvalSm` is the
only other producer. Three bins can never close, and the comment above the cross claims the opposite.

**Fix:** emit one `add_operations()` in M-mode *before* `csrs mstatus, MSTATUS_TVM`.

---

## 3. (Medium, design) Two overlapping EPC-relocation mechanisms in the shared trap handler

`Mrtn2mmode` (`rvtest_trap_handler.h:2724`) already relocates the resume EPC through the caller-mode
save area's `code_bgn_off` — that is how the legacy `RVTEST_GOTO_MMODE` worked for these same Sv tests.
`TSBI_RELOCATE_EPC` re-implements the idea for the T-SBI paths with different rules: direction-aware,
gated on `satp != Bare`, hardcoded `+1*sv_area_sz` slot. Both are now reachable, reading the same
save-area field, with different semantics. Worth unifying, or at minimum documenting why they differ.

Related stale comment at line 1578:

> *"For GOTO_MMODE, we reuse the existing rtn2mmode path which handles EPC relocation across address
> spaces. The simpler GOTO_S/U/VS/VU paths don't need relocation…"*

It was already wrong about the code path (`tsbi_goto_m` jumps to `resto_rtn`, not `rtn2mmode`); this PR
makes the second half wrong too.

Also: the trigger is implicit. The macro fires for any suite whose test happens to have called
`SAVE_AREA_SETUP(..., code, ...)` with translation on — not because the test asked for relocation.
That is a behavior change to infrastructure every suite shares, and it is the same implicitness that
makes finding #1 a landmine.

---

## 4. Smaller items

- **`keep_image_mapped()`** (`sv/generate.py:83`) hardcodes `sd` and `0x1FF`, so it is RV64 / 9-bit-VPN
  only. `range(sv.levels - 2, 1, -1)` happens to be empty for Sv32/Sv39, so it is safe today, but in
  that case it returns a bare comment line with no body. It also silently assumes the boot map already
  links root → `rvtest_slvl{levels-2}_pg_tbl`. Use `SREG` and an xlen-derived index mask, and
  early-return for `levels <= 3` to match the call-site guard.
- **`begin_sv_test` `sig_init` reordering** (`sv/generate.py:39` and `:77`): moving `sig_init` after
  `setup_asm` only when `driver_mode is not None` looks arbitrary. The real reason is that `setup_asm`
  may contain a `tsbi_call`, which marshals through a0–a2 and clobbers the signature pointer (Svadu's
  `menvcfg` write). One comment saves the next reader the hunt.
- **`SvSm_satp_cg.tvm_mstatus`** has only `bins zero = {0}`, yet `make_svsm_mstatus_tvm` deliberately
  executes `csrw satp, zero` in M-mode with TVM=1 — exactly the "TVM does not affect M-mode" case the
  test appears to target. That sample is dropped. Consider adding `bins set = {1}`.
- **M-mode satp access narrowed for sv48/sv57**: `SvSm._t_satp_access` early-returns unless
  `sv.name in ("sv32","sv39")`, so `cp_access_m` now only comes from the Sv32/Sv39 generators, where the
  old `Sv_sv48/57_satp_access_test` also existed. Fine in practice (Sv48 configs implement Sv39), but
  worth a comment.
- **The code alias is a top-level superpage**, so `va_rvtest_code_begin` exposes the whole 1 GiB (Sv39)
  containing the image — data and signature included — as U-readable and U-executable. Harmless for the
  current tests, but broader than "a U-accessible alias of the code region" suggests; `sv/modes.py`'s
  docstring should say so.
- **`make_svsm_mstatus_tvm`** declares `required_extensions=["I", "Sm", "S"]` with no Sv requirement,
  though it lives in the SvSm suite and writes `satp`.
- The CTP text points at "the SvSm / SvbareSm / SvinvalSm sheet of the Sv testplan" — those sheets need
  to exist in the external Google Sheet.

generators/testgen/src/testgen/priv/extensions/sv/Svadu.py
```diff
@@ -151,7 +158,7 @@ def _make_svadu(test_data: TestData, sv: SvMode) -> list[TestChunk]:
     "Svadu",
     required_extensions=["I", "Sv32", "Svadu"],
     march_extensions=_MARCH,
-    extra_defines=["#define BOOT_TO_MMODE"],
+    extra_defines=[BOOT_SMODE],
```
jordancarlin: `BOOT_SMODE` adds extra indirection for minimal gain. Remove this and just use `#define BOOT_TO_SMODE` directly. Global comment.

generators/testgen/src/testgen/priv/extensions/sv/Svadu.py
```diff
@@ -57,35 +59,39 @@ def _add_adu_access(test_data: TestData, sv: SvMode, mode: str, level: int, numb
...
-    stride = sv.xlen // 8
+    index_bits = 10 if sv.xlen == 32 else 9
+    offsets = [
+        ((int(va, 16) >> sv.page_offset_bits(level)) & ((1 << index_bits) - 1)) * (sv.xlen // 8)
+        for va in _VAS[sv.name][level]
+    ]
     return [
-        *virtual_address(sv, f"va_data_l{level}_w", level, destination="a0", scratch="t0", merge_sv32_base_page=True),
-        *virtual_address(sv, f"va_data_l{level}_r", level, destination="a1", scratch="t0", merge_sv32_base_page=True),
+        *virtual_address(sv, f"va_data_l{level}_w", level, destination="s0", scratch="t0", merge_sv32_base_page=True),
```
jordancarlin, line 69: I wonder if we are better off switching to using the testgen register allocator so that we don't run into issues like this again in the future. Maybe something to do in a future PR.

generators/testgen/src/testgen/priv/extensions/sv/modes.py
jordancarlin, line 1: Are these function really needed? Seems like we might be better off dropping this file.

generators/testgen/src/testgen/priv/extensions/sv/access.py
```diff
@@ -39,6 +40,16 @@ def virtual_address(
     ]
 
 
+def lower_mode_entry(mode: str, driver_mode: str | None) -> list[str]:
+    """Enter the mode under test; driver_mode None selects the legacy M-mode macros."""
```
jordancarlin: Why are you adding code that emits the legacy macros? We should be fully moving away from the legacy macros and only using the TSBI calls going forward. I think you can also get rid of these functions if you only use the new TSBI macros.

generators/testgen/src/testgen/priv/extensions/sv/generate.py
```diff
@@ -30,12 +30,13 @@ def begin_sv_test(
     va_code_override: str | None = None,
     pre_va_asm: tuple[str, ...] = (),
     setup_asm: tuple[str, ...] = (),
+    driver_mode: str | None = None,
```
jordancarlin: Why would driver_mode ever be None? These tests are always coming from some known mode.

generators/testgen/src/testgen/priv/extensions/sv/SvbareSm.py
```diff
@@ -0,0 +1,43 @@
+##################################
+# priv/extensions/sv/SvbareSm.py
+#
+# SvbareSm suite: M-mode MPRV accesses with satp.MODE=Bare.
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
jordancarlin: Use your own Copyright for files you create, not Qualcomm's.

generators/testgen/src/testgen/priv/extensions/sv/SvbareSm.py
```diff
@@ -0,0 +1,43 @@
...
+from testgen.priv.extensions.sv.Svbare import bare_rwx, begin_bare_test
+from testgen.priv.registry import add_priv_test_generator
+
+_MARCH = ["I", "Zicsr", "Zifencei"]
+
+
+@add_priv_test_generator(
+    "SvbareSm",
+    required_extensions=["I", "Sm", "S"],
+    march_extensions=_MARCH,
```
jordancarlin: No need to specify march if there are no special extensions needed. All tests already include Zicsr and Zifencei.

generators/testgen/src/testgen/priv/extensions/sv/SvbareSm.py
```diff
@@ -0,0 +1,43 @@
...
+from testgen.priv.extensions.sv.modes import BOOT_MMODE
+from testgen.priv.extensions.sv.Svbare import bare_rwx, begin_bare_test
+from testgen.priv.registry import add_priv_test_generator
+
+_MARCH = ["I", "Zicsr", "Zifencei"]
+
+
+@add_priv_test_generator(
+    "SvbareSm",
+    required_extensions=["I", "Sm", "S"],
```
jordancarlin: Why does this require I? Shouldn't it work with E too?

generators/testgen/src/testgen/priv/extensions/sv/SvinvalSm.py
```diff
@@ -0,0 +1,46 @@
...
+            "LI(a0, MSTATUS_TVM)",
+            "csrs mstatus, a0",
+            f"{tvm_label}:",
+            f"RVTEST_SIGUPD_CSR_READ(mstatus, a4, {tvm_label}, {tvm_label}_str)",
+            *add_operations(test_data, 1),
+            *add_operations(test_data, 2),
+            "RVTEST_TSBI_GOTO_SMODE",
+            *add_operations(test_data, 3),
+            "RVTEST_TSBI_GOTO_MMODE",
+            "RVTEST_TSBI_GOTO_UMODE",
```
jordancarlin: You should be able to go straight to U mode.

tests/priv/SvinvalSm/SvinvalSm_Svinval_mstatus_tvm-00.S
```diff
@@ -0,0 +1,136 @@
...
+  sfence.inval.ir
+  nop
+
+SvinvalSm_cg_cp_svinval_test1_sfence_vma:
+  sfence.vma x0, x0
+  nop
+
+SvinvalSm_cg_cp_svinval_test2_sfence_w_inval:
+  sfence.w.inval
+  nop
```
jordancarlin: Why are the nops needed after each instruction? Is this a remnant from when the trap handler skipped the next instruction? If so, remove them.

jordancarlin (changes requested):
I still haven't reviewed most of this, but here are some initial comments.

coverpoints/priv/SvinvalSm_coverage.svh
```diff
@@ -0,0 +1,39 @@
...
+        bins S_mode     = {2'b01};
+        bins U_mode     = {2'b00};
+    }
+    cp_tvm : coverpoint ins.prev.csr[CSR_MSTATUS][20] {
+        bins zero = {0};
+        bins set  = {1};
+    }
+    cr_svinival : cross cp_instr, cp_priv, cp_tvm {
+        // each instruction executed in M mode with each TVM, and in S and U mode with TVM set
+        ignore_bins lower_tvm_zero = (binsof(cp_priv.S_mode) || binsof(cp_priv.U_mode)) && binsof(cp_tvm.zero);
```
davidharrishmc: Just drop these from cp_priv and cp_tvm instead of ignoring.

davidharrishmc: Any reason not to test all of these with tvm = {0/1}, dropping the ignore cp_tvm?

coverpoints/priv/SvSm_coverage.svh
```diff
@@ -0,0 +1,200 @@
...
+////////////////////////////////////////////////////////////////////////////////////////////////
+
+`define COVER_SVSM
+
+covergroup SvSm_satp_cg with function sample(ins_t ins);
+    option.per_instance = 0;
+    `include  "general/RISCV_coverage_standard_coverpoints.svh"
+
+    tvm_mstatus: coverpoint ins.current.csr[CSR_MSTATUS][20]{
+        bins zero = {0};
```
davidharrishmc, line 19: Why only zero?

docs/ctp/src/sv.adoc
```diff
@@ -28,6 +50,16 @@ https://docs.google.com/spreadsheets/d/1LGwD2e01j4SN9tg85nt6gPwpp7nXyXMF7hqbw3jv
 
 include::norm/Svbare_norm_rules.adoc[]
 
+==== SvbareSm Virtual Memory Off with MPRV
+
+[[t-SvbareSm-coverpoints]]
+.SvbareSm Coverpoints
+,===
+See the SvbareSm sheet of the Sv testplan.
```
davidharrishmc, line 58: I've imported these to testplans/priv as SvbareSm.yaml etc.  See an open PR from today Sept 19.  Check other suites for the syntax.

generators/testgen/src/testgen/priv/extensions/sv/Sv.py
```diff
@@ -1020,17 +957,11 @@ def _make_sv(test_data: TestData, sv: SvMode) -> list[TestChunk]:
...
-    _t_mstatus_sbe(test_data, test_chunks, sv)
-    return test_chunks
-
-
 @add_priv_test_generator(
     "Sv",
     required_extensions=["I", "Sv32"],
     march_extensions=_MARCH,
-    extra_defines=["#define BOOT_TO_MMODE"],
+    extra_defines=[BOOT_SMODE],
```
davidharrishmc: How about just saying #define BOOT_TO_SMODE?  Global change.

generators/testgen/src/testgen/priv/extensions/sv/Sv.py
```diff
@@ -1020,17 +957,11 @@ def _make_sv(test_data: TestData, sv: SvMode) -> list[TestChunk]:
...
 
-def _make_sv_sbe(test_data: TestData, sv: SvMode) -> list[TestChunk]:
-    test_chunks: list[TestChunk] = []
-    _t_mstatus_sbe(test_data, test_chunks, sv)
-    return test_chunks
-
-
 @add_priv_test_generator(
     "Sv",
     required_extensions=["I", "Sv32"],
```
davidharrishmc: Drop I because these should work with E.  Global change.

generators/testgen/src/testgen/priv/extensions/sv/Svadu.py
```diff
@@ -28,7 +30,7 @@
 _VAS = {
     "sv32": {1: ("0x00400000", "0x00800000", "0x00C00000"), 0: ("0x01001000", "0x01002000", "0x01003000")},
     "sv39": {
-        2: ("0x040000000", "0x080000000", "0x0C0000000"),
+        2: ("0x0C0000000", "0x100000000", "0x140000000"),
```
davidharrishmc, line 34: Why did these need to change?

generators/testgen/src/testgen/priv/extensions/sv/Svadu.py
```diff
@@ -57,35 +59,39 @@ def _add_adu_access(test_data: TestData, sv: SvMode, mode: str, level: int, numb
...
         "nop",
-        "RVTEST_GOTO_MMODE",
+        *lower_mode_exit(mode, "Smode"),
         write_sigupd(12, test_data, label=labels["store"]),
         write_sigupd(13, test_data, label=labels["load"]),
         write_sigupd(14, test_data, label=labels["exec"]),
         f"LA(a0, {table})",
         f"{labels['read_store_pte']}:",
-        f"{load} a4, {stride}(a0)",
+        f"{load} a4, {offsets[0]}(a0)",
```
davidharrishmc, line 89: Why did this need to change?

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
```diff
@@ -68,52 +63,27 @@ def _begin_bare_test(test_data: TestData, split_name: str) -> TestChunk:
...
 def make_svbare_smode(test_data: TestData) -> list[TestChunk]:
-    return _make_bare_mode(test_data, "S")
+    chunk = begin_bare_test(test_data, "Svbare_Smode")
+    chunk.code.extend(bare_rwx(test_data, "test1"))
+    return [test_data.end_test_chunk()]
 
 
 @add_priv_test_generator(
     "Svbare",
     required_extensions=["I", "S"],
```
davidharrishmc: No I global change.

generators/testgen/src/testgen/priv/extensions/sv/Svinval.py
```diff
@@ -18,15 +19,15 @@
     ("sfence_inval_ir", "sfence.inval.ir"),
     ("sfence_vma", "sfence.vma x0, x0"),
 )
-_MARCH = ["I", "Zicsr", "Zifencei", "Svinval"]
+MARCH = ["I", "Zicsr", "Zifencei", "Svinval"]
```
davidharrishmc: No I, Zicsr.  Compiler automatically supports those.  Not sure if Zifencei is needed in march.

generators/testgen/src/testgen/priv/extensions/sv/SvSm.py
```diff
@@ -0,0 +1,289 @@
...
+
+def _make_svsm(test_data: TestData, sv: SvMode) -> list[TestChunk]:
+    test_chunks: list[TestChunk] = []
+    _t_mstatus_mprv(test_data, test_chunks, sv)
+    _t_upage_mprv(test_data, test_chunks, sv)
+    _t_satp_access(test_data, test_chunks, sv)
+    return test_chunks
+
+
+def _make_svsm_sbe(test_data: TestData, sv: SvMode) -> list[TestChunk]:
```
davidharrishmc: Is this wrapper necessary?

generators/testgen/src/testgen/priv/extensions/sv/SvSm.py
```diff
@@ -0,0 +1,289 @@
...
+    test_chunks: list[TestChunk] = []
+    _t_mstatus_sbe(test_data, test_chunks, sv)
+    return test_chunks
+
+
+@add_priv_test_generator(
```
davidharrishmc, line 180: I've never seen this method with multiple decorated make functions.  I presume it works, but unusual.

tests/env/rvtest_trap_handler.h
```diff
@@ -408,6 +410,53 @@
 1:
 .endm
 
+// Relocate the T-SBI GOTO_M/S/UMODE resume address (xEPC, already past the ecall)
```
davidharrishmc: This is pretty subtle.  Can you give an example?

tests/env/rvtest_trap_handler.h
```diff
@@ -408,6 +410,53 @@
...
+// between the code region's two mappings when a test uses address translation.
+//
+// A test registers a virtual alias of its code region by writing the alias of
+// rvtest_code_begin into code_bgn of the save area after M's, which is what
+// SAVE_AREA_SETUP(VA, rvtest_code_begin, code, LEVEL) does with a0 = Mtramptbl_sv.
+// When that alias differs from rvtest_code_begin and satp.MODE is not Bare:
+//   - GOTO_UMODE from the code region resumes at the same offset in the alias,
```
davidharrishmc: There aren't GOTO_*MODE anymore.  Everything should be going through TSBI_GOTO.

davidharrishmc (changes requested):
Generally minor changes / queries.  Looking good to me.

generators/testgen/src/testgen/priv/extensions/sv/Svinval.py
jordancarlin: Zifencei is automatically always included in the march.

generators/testgen/src/testgen/priv/extensions/sv/SvSm.py
jordancarlin: I think I introduced it when creating either the PMP or Sv generators. It is no different than if they were all in separate Python files but is just easier to group them together for things like virtual memory tests.

generators/testgen/src/testgen/priv/extensions/sv/Svadu.py
UmerShahidengr: Because the suite now boots to S-mode and runs translated out of the boot identity map. On sv39 the
image sits in root slot 2 (`0x80000000` → VPN[2] = 2), and the old VAs `0x040000000 / 0x080000000 /
0x0C0000000` are slots 1/2/3 — slot 2 would unmap the code that is running. The new VAs
`0x0C0000000 / 0x100000000 / 0x140000000` are slots 3/4/5. Comment added at `_VAS`.

tests/env/rvtest_trap_handler.h
UmerShahidengr: Rewrote the block with a worked Sv39 example (0x80000000 mapped at 0xC0000000; GOTO_UMODE at
0x80000210 resumes at 0xC0000210, the GOTO_SMODE that ends the excursion at 0xC0000260 resumes at
0x80000260), plus why each direction has to move and what the range checks are for.

generators/testgen/src/testgen/priv/extensions/sv/Svadu.py
UmerShahidengr: a0–a2 are the T-SBI argument registers and don't survive `RVTEST_TSBI_GOTO_*`; s0/s1 do. The readback
offsets were hardcoded `stride * 1..3`, which only worked because the old VAs happened to land in root
entries 1/2/3; they're now derived from the VAs, which reproduces the old offsets for levels 0 and 1
and gives the correct ones for the new level-2 VAs.

coverpoints/priv/SvinvalSm_coverage.svh
UmerShahidengr: `SvinvalSm` now runs all four instructions in M, S and U mode with `mstatus.TVM`
clear *and* set, so the full 4 × 3 × 2 cross closes and the `ignore_bins` is gone. `SvinvalSm_cg` is
100 % on both rv64 and rv32. This also closes the hole the Claude review found: `M_mode × tvm.zero` for
`sinval.vma`, `sfence.w.inval` and `sfence.inval.ir` had no producer anywhere in the repo.

generators/testgen/src/testgen/priv/extensions/sv/Svade.py
```diff
@@ -8,20 +8,22 @@
...
+from testgen.asm.tsbi import tsbi_call
 from testgen.data.state import TestData
 from testgen.data.test_chunk import TestChunk, trap_sigupd_count
-from testgen.priv.extensions.sv.access import add_rwx_test
+from testgen.priv.extensions.sv.access import add_rwx_test, enter_mode, leave_mode
 from testgen.priv.extensions.sv.generate import begin_sv_test, sv_data
 from testgen.priv.extensions.sv.page_tables import SV32, SV39, SV48, SV57, PteFlags, SvMode, create_page_mapping
 from testgen.priv.registry import add_priv_test_generator
 
+DRIVER = "Smode"
```
jordancarlin, line 19: Get rid of `DRIVER` and just use `Smode` directly. The extra indirection here doesn't help.

generators/testgen/src/testgen/priv/extensions/sv/access.py
```diff
@@ -39,6 +47,31 @@ def virtual_address(
...
+    Unlike the T-SBI macros, RVTEST_GOTO_LOWER_MODE also moves the caller into the code
+    alias, which those suites rely on because their test VAs overwrite the root slot that
+    maps the test image.
+    """
+    return [f"RVTEST_GOTO_LOWER_MODE {mode}"]
+
+
+def legacy_leave_mode() -> list[str]:
+    """Return to M-mode after legacy_enter_mode."""
+    return ["RVTEST_GOTO_MMODE"]
```
jordancarlin, line 72: We should be getting rid of the legacy macros entirely, even in Sm suites. They should still be able to use the T-SBI mode switching macros.

generators/testgen/src/testgen/priv/extensions/sv/access.py
```diff
@@ -39,6 +47,31 @@ def virtual_address(
     ]
 
 
+def enter_mode(mode: str, driver_mode: str) -> list[str]:
+    """Switch from the mode the test booted to into the mode under test."""
+    return [] if mode == driver_mode else [_GOTO[mode]]
```
jordancarlin, line 52: Get rid of the mapping and just do something like `else [f"RVTEST_TSBI_GOTO_{mode.upper()}"]`. The mapping doesn't really simplify anything.

generators/testgen/src/testgen/priv/extensions/sv/access.py
```diff
@@ -39,6 +47,31 @@ def virtual_address(
...
+
+
+def leave_mode(mode: str, driver_mode: str) -> list[str]:
+    """Return to the boot mode after enter_mode."""
+    return [] if mode == driver_mode else [_GOTO[driver_mode]]
```
jordancarlin, line 57: I'm not convinced these functions are worth having. They add extra indirection and don't really reduce duplicate code. It might be better to just use something like the following inline:
`f"RVTEST_TSBI_GOTO_{MODE}" if mode != driver else ""`

docs/tsbi-changes.md
jordancarlin, line 1: The T-SBI port is done after this, so no need to update this file. Another PR is already removing it. This was just for the transition. The official TSBI docs are in the CTP.

docs/ctp/src/sv.adoc
```diff
@@ -5,6 +5,14 @@
 
 The UDB configuration file determines whether to run tests for Sv32, or for Sv39 and possibly Sv48 and Sv57.
 
+The Sv suites use T-SBI. Each test boots to S-mode and runs from the boot identity map of the test image.
+Each test registers a U-accessible virtual alias of the code region, and `RVTEST_TSBI_GOTO_UMODE` runs
+U-mode accesses from that alias; `RVTEST_TSBI_GOTO_SMODE` returns to the identity-mapped code.
+M-mode CSRs such as `menvcfg` are written through T-SBI CSR calls.
+Coverpoints that need M-mode state (`mstatus.MPRV`, `mstatus.TVM`, `mstatus.SBE`, or an M-mode CSR access)
+are in the `Sm` suites: SvSm, SvbareSm and SvinvalSm.
+The PMP suites (SvPMP, SvPMPZicbo, SvaduPMP) boot to M-mode because they program PMP entries.
```
jordancarlin, line 14: A lot of this can be removed from the CTP. All tests use T-SBI, so there is no need to call it out for the Sv tests in particular.

coverpoints/priv/Svinval_coverage.svh
```diff
@@ -18,14 +18,14 @@ covergroup Svinval_cg with function sample(ins_t ins);
         wildcard bins sfence_vma      = {SFENCE_VMA}; // not essential, but might as well cross it
     }
     cp_priv : coverpoint {ins.prev.mode} {
-        bins M_mode     = {2'b11};
         bins S_mode     = {2'b01};
         bins U_mode     = {2'b00};
     }
```
jordancarlin, line 23: You should be able to use the standard `priv_mode_s_u` coverpoint and drop this.

coverpoints/priv/SvinvalSm_coverage.svh
```diff
@@ -0,0 +1,39 @@
...
+        wildcard bins sfence_inval_ir = {SFENCE_INVAL_IR};
+        wildcard bins sfence_w_inval  = {SFENCE_W_INVAL};
+        wildcard bins sinval_vma      = {SINVAL_VMA};
+        wildcard bins sfence_vma      = {SFENCE_VMA}; // not essential, but might as well cross it
+    }
+    cp_priv : coverpoint {ins.prev.mode} {
+        bins M_mode     = {2'b11};
+        bins S_mode     = {2'b01};
+        bins U_mode     = {2'b00};
+    }
```
jordancarlin, line 25: Use the standard `priv_mode_m_s_u` coverpoint instead. Global comment.

coverpoints/priv/SvbareSm_coverage.svh
```diff
@@ -0,0 +1,55 @@
...
+
+    `ifdef UDB_MXLEN_64
+        satp_bare: coverpoint ins.current.csr[CSR_SATP][63:60] {
+            bins bare = {4'b0000};
+        }
+    `else
+        satp_bare: coverpoint ins.current.csr[CSR_SATP][31] {
+            bins bare = {1'b0};
+        }
+    `endif
```
jordancarlin, line 43:
```suggestion
    satp_bare: coverpoint get_csr_val(ins.hart, ins.issue, `SAMPLE_BEFORE, "satp", "mode") {
        bins bare = {'0};
    }
```

generators/testgen/src/testgen/io/templates.py
```diff
@@ -268,8 +268,8 @@ def generate_march_string(ext_components: list[str], xlen: int) -> str:
     single_letter: list[str] = []
     multi_letter: list[str] = []
     for ext in ext_components:
-        if ext in ["Sm", "S", "U"]:
-            continue  # Skip privilege modes in march string
+        if ext in ["Sm", "S", "U", "Sv32", "Sv39", "Sv48", "Sv57", "NORUN"]:
+            continue  # Skip privilege modes, address-translation modes, and NORUN in march string
```
jordancarlin, line 274: Revert. This is already covered by S.

generators/testgen/src/testgen/priv/extensions/sv/SvZicbo.py
```diff
@@ -27,11 +28,13 @@
 )
 from testgen.priv.registry import add_priv_test_generator
 
-_MARCH = ["I", "Zicsr", "Zifencei"]
+DRIVER = "Smode"
```
jordancarlin, line 31: Inline.

generators/testgen/src/testgen/priv/extensions/sv/Svnapot.py
```diff
@@ -24,8 +24,9 @@
 )
 from testgen.priv.registry import add_priv_test_generator
 
+DRIVER = "Smode"
```
jordancarlin, line 27: Inline.

generators/testgen/src/testgen/priv/extensions/sv/Svpbmt.py
```diff
@@ -8,15 +8,17 @@
...
+from testgen.asm.tsbi import tsbi_call
 from testgen.data.state import TestData
 from testgen.data.test_chunk import TestChunk
-from testgen.priv.extensions.sv.access import add_rwx_test
+from testgen.priv.extensions.sv.access import add_rwx_test, enter_mode, leave_mode
 from testgen.priv.extensions.sv.generate import begin_sv_test, sv_data
 from testgen.priv.extensions.sv.page_tables import SV39, SV48, SV57, PteFlags, SvMode, create_page_mapping
 from testgen.priv.registry import add_priv_test_generator
 
+DRIVER = "Smode"
```
jordancarlin, line 19: Inline.

generators/testgen/src/testgen/priv/extensions/sv/SvPMP.py
```diff
@@ -28,7 +28,6 @@
     (pmp.cfg_byte("0011", "napot", pmp.cfg_shift(1)), "pmpcfg0_rw", 1),
     (pmp.cfg_byte("0100", "napot", pmp.cfg_shift(1)), "pmpcfg0_x", 2),
 )
-_MARCH = ["I", "Zicsr", "Zifencei"]
 _PARAMS = ["NUM_PMP_ENTRIES: '>0'"]
 _DEFINES = ["#define BOOT_TO_MMODE"]
```
jordancarlin, line 32: Inline to match all of the other tests.

generators/testgen/src/testgen/priv/extensions/sv/SvPMP.py
```diff
@@ -11,7 +11,7 @@
 from testgen.data.state import TestData
 from testgen.data.test_chunk import TestChunk, trap_sigupd_count
 from testgen.priv.extensions.pmp import helpers as pmp
-from testgen.priv.extensions.sv.access import add_rwx_test
+from testgen.priv.extensions.sv.access import add_rwx_test, legacy_enter_mode, legacy_leave_mode
```
jordancarlin, line 14: Why can't these use the TSBI mode switching macros? We should ideally be getting rid of the legacy mode switching macros after this is merged.

generators/testgen/src/testgen/priv/extensions/sv/generate.py
```diff
@@ -63,8 +61,8 @@ def begin_sv_test(
             ),
             "sfence.vma",
             "",
-            "// Set up the code-region save area.",
-            "csrr a0, mscratch",
+            "// Register the alias so the T-SBI GOTO macros can move an excursion into it.",
```
jordancarlin, line 64: Not clear what this comment means. What is an excursion referring to in this context?

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
```diff
@@ -68,52 +60,25 @@ def _begin_bare_test(test_data: TestData, split_name: str) -> TestChunk:
...
-    chunk.code.extend(_add_rwx(test_data, "test1", mode=f"{mode}mode"))
-    return [test_data.end_test_chunk()]
-
-
 @add_priv_test_generator(
     "Svbare",
-    required_extensions=["I", "S"],
-    march_extensions=_MARCH,
-    extra_defines=["#define BOOT_TO_MMODE"],
+    required_extensions=["S"],
```
jordancarlin:
```suggestion
    required_extensions=["Svbare"],
```

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
```diff
@@ -68,52 +60,25 @@ def _begin_bare_test(test_data: TestData, split_name: str) -> TestChunk:
...
+    chunk.code.extend(bare_rwx(test_data, "test1"))
+    return [test_data.end_test_chunk()]
 
 
 @add_priv_test_generator(
     "Svbare",
-    required_extensions=["I", "S"],
-    march_extensions=_MARCH,
-    extra_defines=["#define BOOT_TO_MMODE"],
+    required_extensions=["S"],
```
jordancarlin:
```suggestion
    required_extensions=["Svbare"],
```

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
```diff
@@ -15,45 +15,37 @@
...
-        ]
-    )
-    return lines
+    return [
+        *enter,
+        "LA(a5, rvtest_data_1)",
+        "addi a2, a2, 16",
+        f"{labels['store']}:",
+        "sw a2, 20(a5)",
+        "nop",
```
jordancarlin, line 32: Why are the `nop`s needed?

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
```diff
@@ -15,45 +15,37 @@
...
+        "lw a3, 20(a5)",
+        "nop",
+        f"{labels['exec']}:",
+        "jalr ra, a5, 0",
+        "nop",
+        *leave,
+        write_sigupd(12, test_data, label=labels["store"]),
+        write_sigupd(13, test_data, label=labels["load"]),
+        write_sigupd(14, test_data, label=labels["exec"]),
+    ]
```
jordancarlin, line 43: Could this all be simplified to do each `sigupd` right after the relevant store/load/execute? Then you could put the calls to `add_testcase` inline and avoid the extra dictionary and the slightly hacky string processing to remove the colons.

coverpoints/priv/SvSm_coverage.svh
```diff
@@ -0,0 +1,202 @@
...
+
+    Mcause: coverpoint ins.current.csr[CSR_MCAUSE][31:0] {
+        bins illegal_ins  = {32'd2};
+        bins no_exception = {32'd0};
+    }
+
+    cp_ins: coverpoint ins.current.insn {
+        wildcard bins csrrs = {32'b000110000000_?????_010_?????_1110011};
+        wildcard bins csrrw = {32'b000110000000_?????_001_?????_1110011};
+        wildcard bins csrrc = {32'b000110000000_?????_011_?????_1110011};
```
jordancarlin, line 31: Use the named constants from the decode package instead of raw encodings.

coverpoints/priv/SvSm_coverage.svh
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    cp_access_m: cross priv_mode_m, cp_ins, Mcause, tvm_mstatus { //sat.1
+        ignore_bins ig1 = binsof(Mcause.illegal_ins);
```
jordancarlin, line 35: What is the point of defining `Mcause` with `illegal_ins` if it is ignored the only place it is used?

generators/testgen/src/testgen/priv/extensions/sv/SvinvalSm.py
```diff
@@ -0,0 +1,49 @@
...
+
+"""Generate Svinval tests in every mode with mstatus.TVM clear and set."""
+
+from testgen.data.state import TestData
+from testgen.data.test_chunk import TestChunk
+from testgen.priv.extensions.sv.access import enter_mode, leave_mode
+from testgen.priv.extensions.sv.Svinval import add_operations
+from testgen.priv.registry import add_priv_test_generator
+
+DRIVER = "Mmode"
```
jordancarlin, line 17: Inline.

generators/testgen/src/testgen/priv/extensions/sv/SvinvalSm.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@add_priv_test_generator(
+    "SvinvalSm",
+    required_extensions=["Sm", "S", "Svinval"],
+    extra_defines=[f"#define BOOT_TO_{DRIVER[0]}MODE"],
```
jordancarlin, line 23: Much easier to read if this is just

```suggestion
    extra_defines=["#define BOOT_TO_MMODE"],
```

generators/testgen/src/testgen/priv/extensions/sv/Sv.py
```diff
@@ -25,14 +26,14 @@
     PteExpression,
     PteFlags,
     SvMode,
-    create_leaf_pte,
     create_page_mapping,
-    create_page_walk,
 )
 from testgen.priv.registry import add_priv_test_generator
 
+DRIVER = "Smode"
```
jordancarlin, line 33: Inline.

generators/testgen/src/testgen/priv/extensions/sv/Sv.py
```diff
@@ -152,7 +164,7 @@ def _emit_access(test_data: TestData, sv: SvMode, level: int, style: str, name:
             True,
         )
     elif style == "sum":
-        setup = ("LI(t0, MSTATUS_SUM)", "csrs mstatus, t0")
+        setup = ("LI(t0, MSTATUS_SUM)", "csrs sstatus, t0" if driver_mode == "Smode" else "csrs mstatus, t0")
```
jordancarlin, line 167: Is this necessary or can we just always use `sstatus`? `sstatus` is accessible from M-mode.

generators/testgen/src/testgen/priv/extensions/sv/SvSm.py
```diff
@@ -0,0 +1,275 @@
...
+from testgen.priv.extensions.sv.Sv import (
+    change_pte_to_be,
+    emit_access,
+    level_header,
+    satp_access_ops,
+    satp_csr_read,
+)
+from testgen.priv.registry import add_priv_test_generator
+
+DRIVER = "Mmode"
```
jordancarlin, line 35: Inline.

generators/testgen/src/testgen/priv/extensions/sv/SvSm.py
```diff
@@ -0,0 +1,275 @@
...
+    required_extensions=["Sm", "Sv32", "NORUN"],
+    extra_defines=["#define BOOT_TO_MMODE"],
+)
+def make_svsm_sv32_sbe(test_data: TestData) -> list[TestChunk]:
+    return _t_mstatus_sbe(test_data, SV32)
+
+
+@add_priv_test_generator(
+    "SvSm",
+    required_extensions=["Sm", "Sv39", "NORUN"],
```
jordancarlin, line 227: Add a TODO comment to all tests that have a `NORUN` documenting why and what the condition for removing it is.

jordancarlin (changes requested):

jordancarlin:
@UmerShahidengr it looks like you requested my review on this but haven’t addressed my most recent round of comments from yesterday.

generators/testgen/src/testgen/priv/extensions/sv/access.py
UmerShahidengr: `legacy_enter_mode` and `legacy_leave_mode` are gone, and most of the PMP tests now use the T-SBI macros: all of SvPMP's `pmp_on_pa`, all of SvPMPZicbo's `pmp_on_pa`, and everything else in the suites.

Three tests still need `RVTEST_GOTO_LOWER_MODE`, and I left the macro inline with a comment: `SvPMP pmp_on_pte`, `SvPMPZicbo pmp_on_pte` and `SvaduPMP no_pmp_perm`. They all map `va_data` at VA 0, which on Sv48/Sv57 is the root slot the boot tables identity map the test image through, so once the mapping is installed only the code alias is executable. `RVTEST_GOTO_LOWER_MODE` mrets straight into the alias. The T-SBI relocation moves *into* the alias only for `GOTO_UMODE` on the M side, and back *out* of it for `GOTO_SMODE`/`GOTO_MMODE` on the S side, so a `GOTO_SMODE` issued from M-mode resumes at the identity PC that the test just unmapped.

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
UmerShahidengr: Two reasons, now written in a comment next to them. 
The one after `jalr ra, a5, 0` is where the call returns: `ra` points at it and the data region ends in `jr ra`. The others are the slot the trap handler resumes at if the access faults, which is the same convention `add_rwx_test` uses. In Svbare nothing should fault, so they only matter there if the DUT is wrong, which is exactly when you want the resume to land somewhere harmless.

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
UmerShahidengr: Done. `bare_rwx` is a loop over `(operation, register, instruction)` that emits the testcase label, the instruction and its `sigupd` together, so the dictionary is gone. The colon stripping is gone too: `test_data.current_testcase_label` gives the bare label. The signature updates now happen in the mode under test rather than after the switch back; translation is off for the whole suite, so U-mode reaches the signature region.

coverpoints/priv/SvSm_coverage.svh
UmerShahidengr: No point at all, so it is gone. Thanks for raising that. In `SvSm_satp_cg` the accesses are from M-mode, where `satp` is always accessible and no exception can be raised, so `Mcause` now has only `no_exception` and the `ignore_bins` is dropped. The other covergroup in that file keeps its own `Mcause` with both bins, where the illegal-instruction case really is reachable.

generators/testgen/src/testgen/priv/extensions/sv/SvSm.py
UmerShahidengr: Added above all four `_sbe` registrations: drop `NORUN` once a DUT and the reference model implement big-endian implicit page-table accesses (`mstatus.SBE=1`); until then no configuration can run them.

UmerShahidengr:
@jordancarlin, all comments are now addressed, please give it another look

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
```diff
@@ -68,52 +47,27 @@ def _begin_bare_test(test_data: TestData, split_name: str) -> TestChunk:
...
-    return [test_data.end_test_chunk()]
-
-
 @add_priv_test_generator(
     "Svbare",
-    required_extensions=["I", "S", "Svbare"],
-    march_extensions=_MARCH,
-    extra_defines=["#define BOOT_TO_MMODE"],
+    required_extensions=["Svbare"],
+    march_extensions=["Svbare"],
```
jordancarlin: You shouldn't need to specify `march_extensions` unless it uses extensions that are not included in `required_extensions`. By default, `march_extensions` is set to `required_extensions`. Applies to lots of tests.

generators/testgen/src/testgen/priv/extensions/sv/Svbare.py
jordancarlin: I'm still not clear on why the `nop` is needed though. Why not just have the trap handler resume after it at the sigupd?

generators/testgen/src/testgen/priv/extensions/sv/access.py
jordancarlin: Does this imply we need to further enhance the T-SBI mode switching macros? I really think we should avoid having multiple different ways of switching modes if at all possible. Is there a good reason the T-SBI versions can't do the relocation?

jordancarlin (changes requested):
Almost there. A few final comments.

jordancarlin (approved):

davidharrishmc (approved):

PR #2431, opened 2026-09-19
Close the remaining non-Sv coverage holes in the Sail coverage run

davidharrishmc:
This closes four non-Sv covergroup holes left by `make coverage` on sail-rv64-max. **Smstateen:** `mstateen0.P1P13` only controls hedelegh, so the P1P13 coverpoints and cross are now guarded on `H_SUPPORTED` and `UDB_MXLEN_32`. **Sm:** the mstatus and menvcfg walks skipped SDT, MDT, DTE and CDE; `csr_walk_test` gains `gated_mask_bits` so those bits are walked when Ssdbltrp, Smdbltrp or Smcdeleg is present, and the covergroup ignores them otherwise. **Svinval:** the test now also runs the invalidation instructions in M-mode, so the M-mode/TVM-clear column is sampled. **Vls:** `vs2_ls_edges_check` walked the index register by vtype.vsew instead of the index EEW, and the generator left tail index elements unclamped; the check now walks by the index width, the clamp runs with vl = VLMAX, and the Vls covergroups are regenerated. Sm_mcsr stays at 96.6% because of a separate `cp_satp_from_*`/`mstatus_tvm` hole. The SmVF `vs1_zero`/`vs2_zero` hole is fixed in #2290 and not repeated here.


generators/testgen/scripts/vector_testgen_common.py
```diff
@@ -2307,7 +2310,10 @@ def loadVecReg(instruction, register_argument_name: str, vector_register_data, s
       writeLine(f"csrr x{avlReg}, vl",                                          "# save vl register for after load")
       writeLine(f"vsetvl x{vlmaxReg}, x0, x{vtypeReg}",                         "# set vl to vlmax")
       writeLine(f"add x{vlmaxReg}, x{vlmaxReg}, x{vlmaxReg}",                   "# save vlmax * 2")
-      writeLine(f"vsetvli x0, x{avlReg}, e{eew}, m{getLmulFlag(vs2_emul)}, tu, mu", "# setting sew to vs2 eew")
+      # vl = VLMAX at the index EEW so every index slot of the register group is clamped, not just
```
fourth-bit: The change should be in the main vector testgen, not here where `make vector-tests` applies.

framework/src/act/fcov/coverage/RISCV_coverage_vector.svh
```diff
@@ -291,25 +291,27 @@ typedef enum {
     None
 } edge_vs2_ls_values_t;
 
-function edge_vs2_ls_values_t vs2_ls_edges_check (int hart, int issue, `VLEN_BITS val);
+// val is the index register of an indexed load or store.  Its elements are as wide as the
+// instruction's width field says (the index EEW), not vtype.vsew: an ei8 index vector under an
+// e64 vtype packs eight indices into every 64-bit chunk, so the walk below goes by the index EEW.
+function edge_vs2_ls_values_t vs2_ls_edges_check (int hart, int issue, `VLEN_BITS val, logic [2:0] width);
```
fourth-bit: The check needs to be expanded to support non-overlaping index values for different numbers of segments and eew alignment. Instead of being constrained by 2vlmax, the check here should probably be `2vlmax*segments*EEW/8`

PR #2432, opened 2026-09-20
Vector Crypto in Main Testgen

fourth-bit:
Adds support for the Vector Crypto extensions to the main testgen. This includes Zvbb, Zvkb, Zvbc, and all of the Zvk* crypto algorithms with full support for EGS.

Passes all sims, and gets 100% coverage.


davidharrishmc:
Can you enable the vector extension in the DUTs so that the tests run?

davidharrishmc:
Claude thinks the NIST vectors are in reversed order.  It has some smaller suggestions too.


1. GCM vectors are loaded byte-reversed and block-misaligned. I implemented vghsh from the spec pseudocode and confirmed it reproduces every X1..X6 and GHASH value in the bundled NIST CSV when each block is loaded in memory order (first byte in the low byte of element 0). The module instead does int(hex, 16) and splits from the low end, which puts the last NIST byte first. In the generated Zvkg test the H constant is emitted as .word 0xca342b2e, 0x884cfa59, 0xef8a2c3b, 0x66e94bd4; memory order would start with 0xd44be966. On top of that, get_128_bits is applied to the unpadded C string and takes blocks from the low end, so for TC4 "C1" is a misaligned slice and "A1" is actually the second A block. Under the PR's convention no chain step produces the next NIST X value, so feeding X_i back in as the accumulator is not the worked-example chain the file comment promises. The tests still run correctly against the reference model, but they are effectively random constants. Fix: int.from_bytes(bytes.fromhex(s), "little"), pad before slicing, and take blocks from the start of the string.
2. SHA state words are registered in reversed element order. The spec reads {a @ b @ e @ f} = get_velem(vs2) and {c @ d @ g @ h} = get_velem(vd), so element 0 is f and h. The module registers [a, b, e, f] and [c, d, g, h] as elements 0..3, and the generated Zvknha test shows custom_nist_sha_1_vs2 starting with a. Using the NIST intermediate values in the file, vsha2cl from state t with W[t..t+3]+K reaches state t+2 only with the spec order. The message-schedule case (vsha2ms) is correct.
3. The cross-edge overlap change applies to every vector instruction. In cr_vs_reg_edges.py the additional_no_overlap argument went from the crossed pair to all three pairs for all types, not just EGS. That silently changes the Vx and Vls suites' register choices, and since those suites are not checked in, CI cannot show it. Scope it to the EGS branch.

Smaller points
- sections.py adds a 256-bit repack path that nothing registers; the 128-bit .octa path is used. Drop the dead branch or say what will use it.
- The ACT3 crypto tables are now duplicated between cp_custom_crypto.py and the standalone scripts/custom/cp_custom_act3_crypto.py. Delete the standalone copy now that the main testgen owns Zvkned.
- handle_label keeps a module-level nist_count; a counter on test_data would match the rest of the generator.
- vror.vi takes a 6-bit unsigned immediate, but the VVIU type keeps imm_bits=5, so values 32 to 63 are never generated. Pre-existing, but this PR is the one touching that type.

jordancarlin:
> Can you enable the vector extension in the DUTs so that the tests run?

They are already enabled in all of the -max configs.

> The ACT3 crypto tables are now duplicated between cp_custom_crypto.py and the standalone scripts/custom/cp_custom_act3_crypto.py. Delete the standalone copy now that the main testgen owns Zvkned.

Let's leave the old vector generator alone until the port is complete and then delete it all at once.

generators/testgen/src/testgen/asm/sections.py
```diff
@@ -69,7 +69,12 @@ def generate_vector_data_section(vector_data_labels: list[tuple[str, list[int],
             continue
         seen_labels.add(label)
 
-        directives = {8: ".byte", 16: ".short", 32: ".word", 64: ".dword"}
+        # Repack into .dword chunks
+        if eew == 256:
+            data = [(val >> (64 * i)) & (2**64 - 1) for val in data for i in range(4)]
+            eew = 64
```
jordancarlin, line 75: Should we generalize this so that it works for any eew larger than 64?

generators/testgen/src/testgen/io/testplans.py
```diff
@@ -28,7 +28,7 @@ def get_extensions(testplan_dir: Path) -> list[str]:
 def expand_vector_extension(extension: str) -> list[str]:
     """Expands a vector extension by adding SEW suffixes."""
 
-    if not extension.startswith(("Vx", "Vls")):
+    if not extension.startswith(("Vx", "Vls", "Zvbb", "Zvbc", "Zvk")):
         # Only Vx and Vls are supported for now
```
jordancarlin: Comment outdated.

generators/testgen/src/testgen/generate/unpriv.py
```diff
@@ -63,6 +63,9 @@ def generate_unpriv_extension_tests(
         assert match is not None, f"Unable to Parse Vector Extension {testsuite} into a Testplan and SEW Pair"
 
         testplan = match.group(1)
+        if testplan == "Zvbc":  # Special case where the testplan gets the SEW suffix
+            testplan = testsuite
```
jordancarlin: Is there any reason for this special case? Should we just rename `Zvbc64.csv` to `Zvbc.csv`? It is the only vector crypto testplan with a suffix.

generators/testgen/src/testgen/asm/sections.py
fourth-bit: Maybe in the future. I don't think that there is a good reason that any generator should emit elements higher than 256 bits. I think having it throw an error instead of handling large elements is a good idea (256 is also the maximum EEW a VLEN=32 configuration can support, so it doesn't seem reasonable that the spec would ever allow for something bigger, and if it does this should require consideration from someone writing a test).

fourth-bit:
Fixed the first two Claude comments. The third is a non-issue and should have been part of the randomization call in the first place. For the smaller points, the EGS8 edges should be registering 256-bit values. I agree with Jordan on the second one. Fixed the third one. The last smaller point sounds like a hallucination. The immediate is 5 bits long.

PR #2435, opened 2026-09-20
Converted Google Sheets testplans to YAML, improved landscape tables

davidharrishmc:
Checked that text is unchanged from Google sheets except for spelling corrections.


testplans/priv/SdtrigSm.yaml
```diff
@@ -0,0 +1,937 @@
...
+    bins: 2 priv bits
+    rules:
+      - Sdtrig_mcontrol6_m
+    notes: "comments on test results: Exceptions when m.1"
+  - name: cp_mcontrol6_execute_adr
+    description: |-
+      tdata2 = {PC address of nop/0}
+      tdata1.select = 0 (adr)
+      tdata1.XSL = 8 combinations
+      perform nop
```
AngelaZzzzz: addi instead since nop triggers other nops, we need a unique one

jordancarlin:
@davidharrishmc heads up that I updated several of the Exceptions testplans while working on #2508, so this will need to be updated based on the Google Sheet again.

PR #2437, opened 2026-09-20
Cleanup T-SBI dispatch

jordancarlin:
The S-mode T-SBI dispatch still used self-modifying code instead of the instruction table. Update it to use the same approach to accessing CSRs as M-mode.


davidharrishmc (approved):

davidharrishmc:
LGTM.  @UmerShahidengr would you confirm?

UmerShahidengr (approved):

PR #2438, opened 2026-09-20
Remove legacy mode-switching macros

jordancarlin:
All tests are now ported to T-SBI, so these macros are no longer needed.
Depends on #2430.


davidharrishmc (approved):
Whoo!

PR #2441, opened 2026-09-21
Add ACT configuration for the CHIPS Alliance VeeR EH1 core

davidharrishmc:
Adds `config/cores/veer/veer-eh1/`, an ACT configuration for the CHIPS Alliance [VeeR EH1](https://github.com/chipsalliance/Cores-VeeR-EH1) core (RV32IMC, machine mode only, its maximum feature set), plus CI wiring and a per-test runner. Draft: the exclusions below are worth discussing before merge. `ci.yaml` excludes the failing suites with reasons inline. Porting also surfaced ACT issues: tests cannot be linked at address 0 (so `run-veer.sh` prepends a boot stub, as the testbench reset vector is 0), stale ELFs are not pruned when a config's extension set shrinks, and `rvtest_boot_to_mmode_csr_init` writes `mcountinhibit` and `mhpmevent3..31` without consulting `MCOUNTINHIBIT_IMPLEMENTED`.

- EH1 does not implement `mstatush` (0x310); ACT's RV32 trap handler reads it unconditionally, so every trap recurses (`Sm`, `ExceptionsSm`, `ExceptionsZc`, `I-fence-00`).
- `mip` is pin-driven and read-only to software, and a `mcycle` write does not stick (`Sm`).
- `Zifencei` fails because the testbench loads separate instruction and data arrays and stores reach only the data array; not attributable to the core.


davidharrishmc:
Asked offline why the fence tests fail, and whether `fence` traps because it is unimplemented. Traced both on the DUT; the answer is different for each, and neither is what the original description implied. Pushed a commit sharpening the `ci.yaml` reasons.

## `I-fence-00`: plain `fence` is fine, `fence.tso` is not

`fence` itself executes and retires — both `0x0ff0000f` (`fence iorw,iorw`) and `0x0330000f`
(`fence rw,rw`). The instruction that traps is `0x8330000f`, which is **`fence.tso`** (FENCE with
`fm=1000`, pred=RW, succ=RW).

EH1's decode table hardcodes the `fm` field to zero, so `fence.tso` matches no pattern and decodes
as an illegal instruction:

```
fence =   [0000........00000000000000001111]
```
https://github.com/chipsalliance/Cores-VeeR-EH1/blob/d04b1c7a/design/dec/decode#L56

The unprivileged spec does not allow that:

> The `fence.tso` instruction was incorrectly described as optional in 2019. However, `fence.tso` is
> encoded within the standard `fence` encoding such that implementations must treat it as a simple
> global fence if they do not support TSO optimizations.

with the companion note that "it is correct to ignore the _fm_ field and implement `fence.tso` as
`fence[RW,RW]`". So the minimum conforming behaviour is to ignore `fm`; EH1 traps instead. This is
not listed as a deviation in the VeeR EH1 PRM.

Trap order from the trace:

| # | PC | Instruction | Traps |
|---|---|---|---|
| 1 | 0x39294 | `csrw 0x310` (mstatush), boot CSR init | yes |
| 2 | 0x39298 | `csrw mcountinhibit`, boot CSR init | yes |
| 3 | **0x1048** | **`8330000f` = `fence.tso`** | **yes** |
| 4 | 0x20dc | `csrrs s1,0x310` inside the trap handler | 3750x |

The two boot traps are survivable by themselves — `I-add-00` takes them too (3 trap entries, 0
`mstatush` reads) and passes. What kills `I-fence-00` is that the `fence.tso` trap gets *recorded*,
and the trap-recording path builds a status word by reading `mstatush`, which recurses. So this one
test exposes two independent bugs.

## `Zifencei-fence.i-00`: `fence.i` does not trap either

`fence.i` retires normally and the next fetch simply returns the stale instruction:

```
#390  0x1088  sw   a6,0(ra)  [0x1090]   <- stores 0x005c0c13 over the code
#391  0x108c  fence.i                   <- executes, retires, no trap
#392  0x1090  addi s8,s8,1              <- fetched the OLD 0x001c0c13
```

so x24 ends at 4 instead of 8. Two checks show this is the testbench, not the core:

- rebuilding with `-set icache_enable=0` fails identically, so it is not a stale I-cache;
- the testbench's instruction-memory instance is tied off with `.HWDATA(64'h0)` and sits only on the
  fetch bus, while both memories are loaded once from `program.hex` at time 0 — so no store can ever
  reach fetched memory.

EH1 does separately document a real `fence.i` I/D-ordering deviation, but this test is not currently
reaching it. Happy to leave `Zifencei` excluded on that basis.

PR #2443, opened 2026-09-21
Add ACT configuration for the CHIPS Alliance VeeR EL2 core

davidharrishmc:
Adds `config/cores/veer-el2/`, an ACT configuration for the CHIPS Alliance [VeeR EL2](https://github.com/chipsalliance/Cores-VeeR-EL2) core: RV32IMC with U-mode and 64 PMP entries, declared `Sm 1.11.0`. VeeR's draft-0.94 Zb* is left off, as is Smepmp. Independent of #2441 (VeeR EH1); the two share no files and can merge in either order. The core must be built with `fast_interrupt_redirect` off so interrupts reach the ACT trap handler through `mtvec`. `ci.yaml` excludes the failing suites with reasons inline.

- `PMPZca_misaligned_{napot,na4,tor}-00`: EL2 appears not to fault on a compressed fetch crossing out of an executable PMP region (not yet confirmed against `design/el2_pmp.sv`).
- `PMPSm_cfg_A_tor_zero-00`: a TOR entry with `pmpaddr=0` does not raise the expected store/AMO access fault.
- `mtval` is inconsistent: EL2 writes the instruction encoding for most illegal instructions but zero for an illegal CSR access (`PMPU_csr_access-00`); the config models the majority behaviour.
- Other `Sm`/`U` trap-count, `mepc` and `mtval` mismatches, plus the same `fence.tso` and `fence.i` items as #2441.


davidharrishmc:
Two follow-up commits, from tracking down why the two Zihpm tests fail here.

That failure is a configuration bug, not a core bug. The yaml claims Zihpm — the M+U build decodes
`hpmcounter3..6` and their high halves — while `sail.json` still said `"Zihpm": { "supported": false }`,
so the reference model raised an illegal instruction on *every* `hpmcounter` read the tests make and
the trap handler resumed past each one, leaving the six `addi` instructions between the two reads as
the recorded difference. That is where "`hpmcounter3h` reads 0, the reference expects 6" came from;
the low-counter entries hid the same artifact because the testcase collapses any non-negative
difference to 0. With Zihpm supported in `sail.json`, the reference signature for both tests is all
zeros, which is what the core reports.

The second commit models `mcounteren`, which this U-mode build does implement: `csrdecode_mu` decodes
0x306 and `el2_dec_tlu_ctl.sv` makes CY, IR and HPM3..HPM6 writable, with TM read-only zero because
EL2 has no `time` CSR. `sail.json` carried 0 for that mask and now carries 0x7D. `MCOUNTENABLE_EN`
marks bits 3..6, the most UDB allows, since it requires `MCOUNTENABLE_EN[0:2]` to be false unless
Zicntr is implemented — and EL2 cannot claim Zicntr for the same missing-`time` reason.

That second one matters alongside #2462: every test on this configuration currently boots to M-mode,
because ACT keyed the boot privilege mode off which CSR it could use for the Zicsr tests and this hart
has no F, V, S or Zicntr. With #2462 the unprivileged suites run in U-mode, where `mcounteren` decides
whether the `hpmcounter` shadows are readable at all.

The pass/fail counts in the description predate all three changes (this pair, #2462 and #2446) and a
DUT re-run.

PR #2445, opened 2026-09-21
AGENTS.md: DUT bring-up lessons, process traps, and the prek step

davidharrishmc:
Three sections added to `AGENTS.md`, all learned the hard way while porting the CHIPS Alliance VeeR
cores (#2441, #2443, #2444). Documentation only — no code changes, and independent of those PRs.

### Bringing Up A New DUT

Audit every field inherited from another core's config. I copied CV32E20's `sail.json` `mtvec`
block (`direct.supported: false`, 256-byte vectored alignment) onto a core that supports both modes
with 4-byte alignment. Sail then legalized `MODE` to 1 on write; ACT reads `mtvec` back to decide
whether it is writable, saw the mismatch, and fell into its trampoline-relocation fallback — which
copies to the address the *original* `mtvec` pointed at. That was the reset value `0x1`, and
`0x1 & ~3 == 0`, so every test died loading from address 0 with "possible trap loop". Nothing in
the symptom pointed at `mtvec`.

Also recorded: `TEST_BASE = 0` does not work, and a trap diagnostic showing `XCAUSE=0, XTVAL=0` with
a tiny `XEPC` usually means no trap happened at all and the handler is printing reset values.

### Before Committing

Run `prek` and fix what it reports. The hooks are not cosmetic-only — `shellcheck` found a runner
script that parsed a `--stub` option it never used while unconditionally prepending another core's
boot stub. `shfmt` and `prettier` rewrite files in place, so re-stage and re-run the build
afterwards: `prettier` reflows YAML, and a config that reformats still has to validate.

### Process Traps

- The build **stops at the first failure**, so `✗ Build failed: 130 succeeded, 1 failed` does not
  mean the other 130 passed — they were never attempted. This misled me for several iterations.
- Editing a UDB or Sail config regenerates `rvtest_config.h` but does **not** rebuild existing
  `.sig.elf` artifacts, so tests keep failing for the previous configuration until
  `work/<config>` is deleted.
- Stale ELFs are not pruned when a config's extension set shrinks, so `run_tests.py` still runs and
  counts suites the config no longer claims.

Happy to trim any of this if it is more detail than `AGENTS.md` wants.


jordancarlin:
> * Editing a UDB or Sail config regenerates `rvtest_config.h` but does **not** rebuild existing
>   `.sig.elf` artifacts, so tests keep failing for the previous configuration until
>   `work/<config>` is deleted.

This isn't true. The config file and linker script are both tracked as dependencies. Changing them locally triggers a rebuild for me. I'll update AGENTS.md, but you might need to look into something strange happening in your local environment.

jordancarlin:
Replaced by #2477.

PR #2446, opened 2026-09-21
Zihpm: only test implemented hpmcounters

davidharrishmc:
`cp_cntr.py` accessed `hpmcounter3` through `hpmcounter31` unconditionally, but Zihpm allows any subset to be implemented, and [`norm:hpm_unimplemented_counter_access`](https://riscv.github.io/riscv-isa-manual/snapshot/unprivileged/index.html#norm%3Ahpm_unimplemented_counter_access) lets an access to a missing counter either trap or return a constant. The sweep therefore required the constant-value option: a hart that traps stopped at its first missing counter. Each counter's test is now wrapped in `#ifdef UDB_HPM_COUNTER_EN_<n>`, generated from `HPM_COUNTER_EN`, so a configuration tests exactly the counters it declares. The `cp_cntr_hpm` bins are gated the same way, so coverage no longer demands counters the configuration does not implement. Harts with all 29 counters are unaffected; configurations with a subset (CVA6 `cv64a6_imafdcv_sv39`, the whisper configs) no longer depend on the rest reading as a constant. `coverpoints/norm/Zihpm.yaml` records that `norm:hpm_unimplemented_counter_access` is not coverable by a signature-comparing test, since neither permitted response can be required.


davidharrishmc:
This PR is backed by this ISA manual.

"The implemented number and width of these additional counters, and the set of events they count, are
platform-specific. Accessing an unimplemented counter may cause an illegal-instruction exception or may
return a constant value."  

I think we should not test unimplemented counters.  If we did, we would have to accept either illegal instruction or returning a value, which is pretty much the only thing that could happen. We could read multiple times and check for constant values, but that also depends on mhpmevent*, and seems hard to put the system into a state that would incorrectly count when the UDB says the counter isn't implemented.

This was not found previously because unimplemented counters in DUTs tested previously just read 0.

I believe we need this gate to be spec compliant.

jordancarlin:
The reason this one is sort of strange is that Zihpm says that the number of implemented counters is implementation defined, but Sm says that you need to implement all of them. So I guess mhpmcounter3-31 are all required but hpmcounter3-31 are not and you can implement some subset of them even if you have Zihpm. I agree that we should move forward with this PR despite the subtlety there though.

jordancarlin (approved):

jordancarlin (changes requested):
The coverpoints need to be updated too.

jordancarlin (approved):

PR #2449, opened 2026-09-21
Gate accesses to unimplemented mcountinhibit

### LINKED ISSUE #2248: [Feature] support for cores without mcountinhibit

davidharrishmc:
`mcountinhibit` is optional and is reserved when not implemented. Guard the following accesses to `mcountinhibit` on `UDB_MCOUNTINHIBIT_IMPLEMENTED`:

- Boot code: `csrw mcountinhibit, zero` gated on `UDB_MCOUNTINHIBIT_IMPLEMENTED`. A hart without the register behaves as though it were zero, which is what the write was asking for anyway.
- `Sm`: `mcountinhibit` moves out of the unconditional `csrm` list into its own guarded access and walk chunks, and the two `cp_inhibit_*` testcases plus the counter re-enable are guarded as well.
- The trap handler's `misa` read in the `mtval2`/`mtinst` path only decided whether to run a block that is itself inside `#ifdef H_SUPPORTED`, so it was a dead CSR read on every trap on every hart without the hypervisor extension. Moving the guard above it removes the read.


davidharrishmc:
LGTM.

generators/testgen/src/testgen/priv/extensions/Sm.py
```diff
@@ -651,6 +649,9 @@ def _generate_mcsr_tests(test_data: TestData, test_chunks: list) -> None:
         ("mhpmevent30", 0),  # mask all bits because they are WARL and can all be ROZ
         ("mhpmevent31", 0),  # mask all bits because they are WARL and can all be ROZ
     ]
+    # mcountinhibit is optional, so it is accessed under UDB_MCOUNTINHIBIT_IMPLEMENTED
+    # TODO: remove mcountinhibit mask when Sail gets parameters for writable bits
```
jordancarlin, line 653: This config option has been merged into sail-riscv and will be available in the next release.

jordancarlin (approved):

jordancarlin (changes requested):
The coverpoints need to be updated too.

jordancarlin (approved):

PR #2450, opened 2026-09-21
Add ACT configuration for the lowRISC Ibex core

davidharrishmc:
Adds `config/cores/ibex/ibex-opentitan/`, an ACT configuration for the Ibex `opentitan` configuration (`BaseIsa=RV32I`), run on RTL under Verilator 5.036 as Ibex Simple System and pinned to `lowRISC/ibex@e9f55342edbd27e9e17a0e41b1c95a81abb5eac8`. It declares I, M, C, Zcb, Zba/Zbb/Zbc/Zbs, Zicntr, Zihpm, U and `Sm 1.12.0` with 16 PMP entries; `Zcmp` is implemented but not declared because sail-riscv 0.14.1 cannot model it. `RVMODEL_ACCESS_FAULT_ADDRESS` is deliberately kept, which exposes the first finding below at the cost of 2 tests. `ci.yaml` excludes the four failing tests.

- Instruction fetches never raise an access fault (3 of 4 failures): Simple System ties `instr_err` to `'0` and aliases every fetch into RAM. This is an integration limitation, not a core defect.
- `mhpmcounter3h`-`12h` are read-only zero (`MHPMCounterWidth=32`). Spec-legal, but UDB and `sail.json` cannot express HPM counter width.
- `mhpmevent3`-`31` are hardwired read-only and do not match the event counted (no ACT test covers it).
- Draft-0.93 bitmanip encodings (Zbf/Zbp/Zbr/Zbt) execute instead of trapping.
- `mip.MTIP` is pending out of reset; worked around in `RVMODEL_BOOT`.


davidharrishmc:
Defer instruction fetch failure.

#2457 records need for mhpmcounter  width

mhpmevent is intentionally not tested because it could be read-only zero.

Bit manip not tested.

mip.MTIP DUT variation on reset is handled in RVMODEL_BOOT

Looks ok to review.

PR #2454, opened 2026-09-21
Add ACT configuration for CVA6 + Ara (vector)

davidharrishmc:
Adds `config/cores/cva6/cv64a6_imafdcv_sv39/`, ACT's first vector-capable RTL config and the first runnable `cva6` config, run on RTL under Verilator and pinned to Ara `34bd3bc1` / CVA6 `99eac9a6`. It uses `nr_lanes=4` and `vlen=512`, which is much faster than the default 4096, and needs Ara's hierarchical Verilator build (>= 5.046). `setup-ara.sh` enlarges the testbench's 1 MB ELF loader window. The config defines `RVMODEL_ACCESS_FAULT_ADDRESS` because the vector, `Exceptions*`, `Sstvala` and `SvZicbo` tests reference it unguarded, so a DUT without one cannot build. DUT nonconformances:

- `vfirst.m` and `vcpop.m` ignore `vl`. ACT's vector self-check uses `vfirst.m`, so nearly every vector test fails on the check itself.
- `vsetivli` does not clamp AVL to VLMAX.
- AMOs do nothing and raise no fault although `misa.A` = 1; `sc.d` reports failure but commits the store.
- Misaligned scalar accesses neither trap nor work.
- `vrgather*`, `vmadc.*m`, `vmsbc.*m` and `vcompress.vm` hang, and Ara raises exceptions on masked-off elements.
- 32 scalar FP failures, mostly `fflags`, in `fdiv`/`fsqrt`/FMA/`fcvt`.


davidharrishmc:
The DUT findings from this configuration are now filed with the CVA6 project as openhwfoundation/cva6#3585, with a reproducer package (failing ELFs, console logs, directed probes, this configuration and the runner): https://github.com/davidharrishmc/riscv-arch-test/releases/download/cva6-ara-act-findings-2026-09-21/cva6-ara-act-findings.tar.gz

The run behind the issue is on this branch at dc78dca53 (no A, no misaligned support declared), 2337 tests with `EXCLUDE_EXTENSIONS=ExceptionsSm` (#2459): 347 pass, 1951 fail, 37 hang, 2 stop on an RTL assertion. Outside the vector suites 343 of 422 pass; 1875 of the 1915 vector tests fail on ACT's `vmsne.vv` + `vfirst.m` self-check because Ara's `vfirst.m` ignores `vl`.

Three failures are not attributed to the core in that issue and may be a question for this side: `Sm_mcsr_walk-01`, `Sm_mcsr_walk-02` and `S_scsr-01` expect `mie`, `mip` and `sip` bit 1 (SSIE/SSIP) to read back 0 after being set, and the core keeps it. The Sail reference produced those expectations, so either the config should declare that S-mode software interrupts are not implemented, or that expectation needs a look.

Issue #2457, opened 2026-09-21
Ten gaps in expressing real DUTs: four block whole suites, six have no UDB/Sail representation


davidharrishmc:
While bringing up ACT configurations for seven new cores (#2450-#2456: Ibex, Sargantana, Hazard3, SERV, CVA6+Ara, OpenC910, PicoRV32) we hit ten distinct places where a **conforming** DUT cannot be described to ACT, UDB or sail-riscv. Filing them together because they were found the same way and several are one-line fixes.

They split into two groups. **Group A blocks suites outright** — no config value avoids it — and is fixable inside this repository. **Group B** is missing expressiveness in UDB or sail-riscv, where the model assumes behaviour the spec leaves optional, so a legal DUT is scored as failing.

---

### A1. `RVMODEL_ACCESS_FAULT_ADDRESS` is referenced unguarded — the `Vls*` suites are the only real case — PR #2458

**Corrected after re-measuring.** My original text listed eight affected suites; it conflated the files that *reference* the macro with the files that fail to *guard* it. The accurate picture:

- 247 files under `tests/` reference `RVMODEL_ACCESS_FAULT_ADDRESS`
- 196 do not guard it, of which 2 are `tests/env/` headers (`check_defines.h`, which only carries the explanatory comment, and `sail_macros.h`, which defines it)
- **all 194 remaining test files are in `Vls8`, `Vls16`, `Vls32`, `Vls64`** — nothing else

Every privileged suite I originally listed — `ExceptionsSv`, `ExceptionsSvZaamo`, `ExceptionsSvZalrsc`, `ExceptionsSm`, `ExceptionsZicboU`, `SvZicbo`, `Sstvala` — **already guards it correctly**.

The macro is optional by design: `README.md:248` says it *"can be omitted if DUT does not generate some/all access faults"*, and `tests/env/check_defines.h:53` says *"If `RVMODEL_ACCESS_FAULT_ADDRESS` is not defined, no access faults are tested"*.

**Root cause is one coverpoint generator.** `cp_custom_ffLS` at `generators/testgen/src/testgen/coverpoints/vector/cp_custom_vls.py:116` emits `LI (x{rs1}, RVMODEL_ACCESS_FAULT_ADDRESS)` with no guard, so a DUT without the macro fails *assembly* (`Error: non-constant expression in ".if" statement`) and loses the whole configuration build.

[...]

PR #2458, opened 2026-09-21
Skip the fault-only-first vl test on a DUT with no faulting address

davidharrishmc:
Fixes item A1 of #2457. `cp_custom_ffLS` emitted an unguarded load from `RVMODEL_ACCESS_FAULT_ADDRESS`, although the macro is optional (`README.md`, `tests/env/check_defines.h`). On a DUT that omits it, `Vls8`/`Vls16`/`Vls32`/`Vls64` fail to assemble and break the whole configuration build; CVA6 + Ara (#2454) and OpenC910 (#2455) had to define the macro to an address that does not fault just to build. The covergroup side was already guarded with `` `ifdef RVMODEL_ACCESS_FAULT_ADDRESS ``. This PR generalizes the existing `ZVL{n}B_SUPPORTED` wrapping in `make_cp_custom_ffLS` to a list of nested guards that includes the macro, and corrects the function's docstring, which said the load runs at rs1=0x0. No regenerated tests are included because the vector suites are gitignored.


davidharrishmc:
@fourth-bit can you review?

PR #2460, opened 2026-09-21
ExceptionsSm/ExceptionsS: raise TRAP_SIGUPD_COUNT to 5000

davidharrishmc:
Fixes A2 of #2457.

`ExceptionsSm` and `ExceptionsS` overflow the trap region when the core does not support misaligned accesses. Increase the trap count from 3000 to 5000.


jordancarlin:
Seems like we could 5000 here instead of 15000. 10,000 extra entries is a significant amount of wasted space.

jordancarlin (approved):

PR #2461, opened 2026-09-21
Let a sail config say the platform has no CLINT or interrupt generator

davidharrishmc:
Fixes A4 of #2457.

## Problem

`_sail_platform_base` in `framework/src/act/build_plan.py` refuses any sail config that sets `platform.clint.supported` or `platform.simple_interrupt_generator.supported` to `false`. It raises during DUT prep, before a single ELF is compiled — `make EXTENSIONS=I` on such a config dies, though no test in the I suite raises an interrupt.

The requirement comes from `tests/env/sail_macros.h`, which `#error`s without both base addresses and then redirects every `RVMODEL_*_INT` macro and the `mtime`/`mtimecmp`/`msip` addresses at those two devices when building the reference ELF.

SERV (#2453) has neither device and had to invent addresses for both, aimed at servant's GPIO window so a stray write only toggles a pin — deliberately not the timer window, where it would reprogram `mtimecmp` and corrupt later tests.

## Changes

- `_sail_platform_base` returns `None` for an unsupported device instead of raising; `_sail_platform_defines` then passes only the defines for the devices that exist.
- `sail_macros.h` gates each group of overrides on its device: the timer and software-interrupt macros on the CLINT, the external and supervisor-interrupt macros on the interrupt generator. With a device absent, the DUT's own definitions stand — on a DUT with no interrupt source those are stubs that do nothing, so the reference and the DUT agree instead of diverging.
- A DUT that defines `RVMODEL_MTIME_ADDRESS` or `RVMODEL_MSIP_ADDRESS` does need Sail to model a CLINT. That combination is now a specific error naming the parameter to set:

```
sail_macros.h:27:6: error: #error "The sail config has no CLINT but the DUT defines
RVMODEL_MTIME_ADDRESS. Set platform.clint.supported to true so the reference model can provide mtime."
```

## Scope

sail-riscv separately requires a CLINT whenever Zicntr is enabled — *"Zicntr is enabled but there is no source of time: a source of time (currently only `platform.clint.supported`) must be enabled"* — so this helps a DUT that has no counters either. That remains a sail-riscv limitation; ACT no longer adds one of its own on top of it.

## Verification

- SERV with **both** devices `supported: false`: all **575** tests build, and the reference ELFs contain no writes to the removed devices.
- cv32e20 with `clint.supported: false` and its `RVMODEL_MTIME_ADDRESS` intact: the new targeted `#error` fires.
- `sail-rv64-max` unchanged: **32/32** on `Sm` + `ZicntrSm` + `ExceptionsSm`, the suites that actually drive `mtime`, `msip` and the interrupt generator.
- `make lint` and `prek` clean.


jordancarlin:
I don't think we want this. We should have Sail always use a CLINT and SIG and the user should pick an address that either has no memory on their DUT or overlaps with other IO on their DUT. I thought that was already described somewhere in the README or CTP, but apparently not. I would prefer a documentation fix instead of this.

PR #2462, opened 2026-09-21
Boot to the lowest privilege mode even when the test CSR is privileged

davidharrishmc:
`tests/env/utils.h` picks `RVTEST_TEST_CSR` from the extensions the DUT implements, and in two of
its branches it also defined `BOOT_TO_SMODE` or `BOOT_TO_MMODE`. `rvtest_setup.h` honours those
defines for **every** test compiled for the configuration, not just the tests that use the CSR, so
the choice of a CSR for the Zicsr tests decided the privilege mode of the whole suite:

- a hart with U-mode but no F, V, S or Zicntr ran every test in M-mode;
- a hart with S-mode but no F or V ran every test in S-mode.

Either way all of that hart's lower-privilege coverage disappeared, silently. Only 12 of the 1997
tests reference `RVTEST_TEST_CSR`.

## The change

`utils.h` now records which mode the CSR it picked needs (`RVTEST_TEST_CSR_NEEDS_SMODE` /
`..._NEEDS_MMODE`) and turns that into `BOOT_TO_SMODE` / `BOOT_TO_MMODE` only when the test asks,
by defining `RVTEST_USES_TEST_CSR`. testgen emits that define for any test whose body references
the macro, so the opt-in cannot drift out of sync with the tests. Every other test boots to the
hart's lowest privilege mode.

Regenerating all 207 test suites changes exactly the 12 Zicsr files, each by that one define.

## Effect per configuration shape

| test CSR | Zicsr tests | every other test |
|---|---|---|
| `fflags` (F) | unchanged | unchanged |
| `vxsat` (Zve32x) | unchanged | unchanged |
| `instret` (Zicntr + U) | unchanged | unchanged |
| `sepc` (S, no F/V) | S-mode | **U-mode** (was S-mode) |
| `mepc` (no F/V/S/Zicntr) | M-mode | **lowest mode** (was M-mode) |

No configuration in tree lands in the last two rows except the VeeR EL2 port in #2443, which has
U-mode but no F, V or S and cannot claim Zicntr (it has the `cycle` and `instret` shadows but no
`time` CSR). It was running all 755 of its tests in M-mode; with this change its unprivileged
suites run in U-mode, which is where the `hpmcounter` shadows it claims Zihpm for actually matter.
The M-mode-only cores (cv32e20, cv32e40s, VeeR EH1/EH2) are unaffected, since M is already their
lowest mode.

## Verification

- Preprocessed one test of each kind against each configuration shape above and checked which
  `RVTEST_BOOT_TO_*` macro survives.
- Regenerated all 207 suites: only the 12 Zicsr files change.
- `make sail-rv32-max EXTENSIONS=Zicsr` — all 6 tests pass.
- Full build of the VeeR EL2 configuration (755 tests) with the fix: every test builds, the ELFs
  now contain `rvtest_boot_to_umode`, and the Zihpm reference signature reads the implemented
  `hpmcounter3..6` and their high halves correctly from U-mode.


jordancarlin:
Simpler solution in #2476.

PR #2464, opened 2026-09-22
Skip misa accesses on a DUT that does not implement the CSR

davidharrishmc:
Fixes item B5 of #2457. UDB has `MISA_CSR_IMPLEMENTED` and ACT already emits `UDB_MISA_CSR_IMPLEMENTED`, but the generators emitted every `misa` access unconditionally; on a hart without the CSR the access traps or, where the address aliases another register (SERV maps 0x301 onto `mtvec`, so `Sm_misa-00` was rewriting the trap vector), silently corrupts it. This guards the four `misa` regions in `Sm.py`, the `misa.V` sequences and `cp_misa_v_clear_set` in `SmV.py`, and the stray `csrr misa` in `Smstateen.py` with `#ifdef UDB_MISA_CSR_IMPLEMENTED`, per region rather than per chunk so a guard cannot be split across generated files. Verified: `Sm_misa-00` now passes on SERV, and Sm/SmV/Smstateen/Ssstateen are unchanged on sail-rv32-max and sail-rv64-max (36/36 each); prek clean.


davidharrishmc:
**Register-definedness review of each guard**, since gating a CSR read out could leave its destination undefined for later code. Checked against the generated assembly rather than the generator source. All four sites are safe, each for a slightly different reason.

**`Smstateen.py`** — the destination is overwritten before any use:

```asm
  LI(x15, -1)
  # Ensure misa.F is set (read misa, verify F bit, then proceed)
#ifdef UDB_MISA_CSR_IMPLEMENTED
  csrr x6, misa
#endif // UDB_MISA_CSR_IMPLEMENTED
  # bit 5 = F; test proceeds assuming F is present per MARCH
  # Clear mstateen0.fcsr so fcsr reads zero
  LI(x6, 2)                 <-- unconditional write to x6
  csrc mstateen0, x6
```

Nothing reads `x6` between the gated `csrr` and the `LI` that overwrites it, so no dependent code needs to move inside the guard. The rest of the test does not depend on the read either — as its own comment says, it "proceeds assuming F is present per MARCH".

**Update: that read was dead code even on a hart that does implement `misa`** — it loaded `x6` and discarded it, so the comment's claim to "verify F bit" was not true. Rather than delete it, d50ac53e5 makes it a real check, because looking the behaviour up in the spec shows the read matters more than the comment suggested:

> [`norm:mstateen0_fcsr_roz`] Whenever `misa.F` = 1, FCSR bit of `mstateen0` is read-only zero (and hence read-only zero in `hstateen0` and `sstateen0` too).

So **the coverpoint's premise is inverted**. "misa.F set, `mstateen0.fcsr` clear — fcsr reads as zero" cannot happen: with `misa.F` = 1 the bit is read-only zero, the `csrc mstateen0` is a no-op, and `fcsr` is a normal FP CSR governed by `mstatus.FS`. The effect the coverpoint is after exists only when `misa.F` = 0, and it is not "reads as zero" either — per `norm:stateen0_fcsr0_misa_f0_illegal_fpu_instr`, with `misa.F` = 0 and the FCSR bit clear, `fcsr` *and every floating-point instruction* raise illegal-instruction. The generator now masks bit 5, branches past the block when F is set, and says that in the banner.

Two things that follow, which I have not changed here:

1. `cp_fcsr_ro_zero` keeps only `binsof(misa_F.F_set) && binsof(mstateen0_fcsr_bit.fcsr_zero)` as `bins fcsr_reads_zero`. That is the unreachable combination, and the whole covergroup half is ``ifdef ZFINX_SUPPORTED` — Zfinx harts have no `f` registers, so `F_set` cannot occur there either. The retained bin should be the `F_clear` one and the name should reflect illegal-instruction rather than reads-zero.
2. **No config in the repo implements Zfinx**, so none of this has ever been compiled. Forcing `-DZFINX_SUPPORTED` on the generated `Smstateen-00.S` gives **24 assembler errors**, all `unrecognized opcode `csrc(...)` / `csrs(...)`` between lines 3951 and 4195 — i.e. `_generate_fcsr_lower` and `_generate_fcsr_lower_fp_instrs` emit `csrc(mstateen0, x9)` with parentheses. `cp_fcsr_ro_zero` itself assembles clean. Worth a separate fix.

**`SmV.py`, the two best-effort `misa.V` writes** — the guarded block is self-contained (`LI` + `csrs` both inside), and the next thing emitted is `_set_vs`, which opens with `LI(x{temp_reg}, …)`:

```asm
#endif // UDB_MISA_CSR_IMPLEMENTED
  LI(x8, 1536)              <-- unconditional write to x8
  csrc mstatus, x8  # clear VS
```

**`SmV.py`, `cp_misa_v_clear_set`** — the whole coverpoint is inside the guard, including its testcase labels, and `nop` follows. The register is not used afterwards.

**`Sm.py`** — each region's `csrr …, misa # save misa` and its matching `csrw misa, …  # restore` fall inside the *same* guard, so there is no save-inside/restore-outside split. Confirmed by the global check that no ungated `misa` access remains anywhere under `tests/priv` after regeneration.

jordancarlin:
> on a hart without the CSR the access traps or, where the address aliases another register (SERV maps 0x301 onto `mtvec`, so `Sm_misa-00` was rewriting the trap vector), silently corrupts it

This sounds like illegal behavior in the DUT that the test is correctly catching. `misa` must be implemented (it is allowed to be read-only zero, but cannot trap).

> The misa CSR is a WARL read-write register reporting the ISA supported by the hart. This register must be readable in any implementation, but a value of zero can be returned to indicate this feature has not been implemented, requiring that CPU capabilities be determined through a separate non-standard mechanism.

jordancarlin:
@davidharrishmc do you think there is anything from this PR worth keeping or should we close it?

davidharrishmc:
Agreed that this is a DUT bug, not a test problem.  Closing.

PR #2467, opened 2026-09-22
Trap prolog: save xsatp and force MODE=Bare instead of dead #elseif satp code

davidharrishmc:
`init_Xsatp` in `tests/env/rvtest_trap_handler.h` used `#elseif`, which is not a preprocessor directive, so on RV64 the satp.MODE selection was skipped and `or T4,T4,T3` used a stale register (objdump shows no `lui/addi` for MODE), while RV32 emitted all three `LI` bodies; S-mode has been running in Bare mode only because the reset `sscratch` high bits happen to be zero. Since the prolog runs before `rvtest_identity_map` fills the root page table, enabling translation there can never be right (a plain `#elif` leaves Sv57/Sv32 live on a DUT without PMP), so the block now saves satp and writes an explicit Bare (`csrrw T4, CSR_XSATP, zero`), the dead `_VA_SZ_` fallback is removed, and the prolog summary and VS-alignment comments are corrected; Sv* suites still program satp themselves. Verified: full priv regression unchanged on sail-rv64-max (649/649), sail-rv32-max (505/505), spike-rv64-max and spike-rv32-max (only the pre-existing `Smstateen-00` failure), plus spike-rv32g; prek clean. Note: #2363 makes the same satp change, so this reduces to the comment and `_VA_SZ_` cleanup once that merges.


davidharrishmc:
Closing: #2363 makes the same satp change (explicit Bare write in the prolog). The remaining comment and dead-`_VA_SZ_` cleanup moves to a small PR stacked on #2363.

PR #2468, opened 2026-09-22
check_defines.h: fix UDB_NUM_PMP_ENTIRES typo so PMP limitation guards can fire

davidharrishmc:
`tests/env/check_defines.h:146` tests `UDB_NUM_PMP_ENTIRES` (typo for `ENTRIES`), so the `#error`s for "PMP without NAPOT support" and "fewer than 8 usable PMP entries" could never fire; a config with `PMP_NAPOT_SUPPORTED: false` built silently and then failed 61 of 102 PMP tests because every region helper hard-codes NAPOT. Fix the spelling; no in-tree config trips the guards (every yaml with PMP has NAPOT and at least 16 usable entries). Verified: negative configs now hit the `#error`s, and the spike-rv64-max, sail-rv32-max, cvw-rv32imc, spike-rv32g and cv32e40s builds are unchanged; prek clean.


jordancarlin (approved):

jordancarlin:
@davidharrishmc this commit isn't signed so can't be merged.

PR #2469, opened 2026-09-22
covergroupgen: drop unsubstituted EFFEW header lines from priv vector covergroups

davidharrishmc:
`covergroupgen/generate.py` always chose the vector header template for vector suites but only substituted the `EFFEW` placeholder when `effew` was set, so the five committed priv vector covergroups (`ExceptionsVls`, `ExceptionsVx`, `ExceptionsVfmin`, `MisalignV`, `SsstrictV`) contain literal `` `ifdef UDB_ELEN_EFFEW `` lines that can never be defined. Pick the vector header only when the vector sample header is used.


jordancarlin:
@fourth-bit can you please review?
@davidharrishmc this one is missing a signature too.

fourth-bit (approved):

jordancarlin (approved):

PR #2470, opened 2026-09-22
Trap handler: read mstatus before extracting MPV/GVA into RV64 signature word 0

davidharrishmc:
## Bug

In the M-mode trap handler's signature word 0 code (`sv_\__MODE__\()vect` in `tests/env/rvtest_trap_handler.h`), the RV32 branch reads `mstatush` to get GVA/MPV for bits 15:14, but the RV64 branch only shifts T4:

```asm
.ifc \__MODE__ , M
  #if (UDB_MXLEN==64)
        srli    T4, T4, UDB_MXLEN-32                 // align to mstatush
  #else
        #ifdef SM1P12P0_OR_LATER_SUPPORTED
          csrr    T4, CSR_MSTATUSH
        ...
        andi    T4, T4, 0x1C0                        // extract bits 8:6 (SPVP?, MPV, GVA)
        slli    T4, T4, 14-6                         // position at bits 16:14
        or      T3, T3, T4
```

T4 never holds `mstatus` at that point. `mstatus` is read into **T2** (`csrr T2, CSR_XSTATUS`) and immediately shifted down to bits 16:0, discarding bits 39:38. T4 holds either the compressed vector offset (`slli T4, T6, 1`, exception path) or the extracted xIP bit (`slli T4, T4, 12`, interrupt path); both are far below bit 32, so `srli T4, T4, 32` always yields 0 and RV64 signature bits 15:14 (MPV, GVA) are always 0.

Objdump of `S_scsr_addr-00.elf` (sail-rv64-max, upstream act4 2dea4b02c), `sv_Mvect`:

```
800058fc: csrrs  x7,mstatus,x0      # mstatus -> T2 (x7), then shifted to bits 16:0
   ...
80005934: srli   x9,x9,0x20         # T4 (x9): last written by slli x9,x14,1 / slli x9,x9,12 -- never mstatus
80005938: andi   x9,x9,448
8000593c: slli   x9,x9,0x8
```

Today this is invisible because `tests/env/riscv_arch_test.h` does `#undef H_SUPPORTED`, so on every supported DUT `mstatus.MPV/GVA` read as 0 and the (accidentally) zero result happens to be right. It would silently mask a wrong MPV/GVA the moment the hypervisor environment is turned back on.

## Fix

Re-read `mstatus` into T4 before shifting on RV64:

```asm
  #if (UDB_MXLEN==64)
        csrr    T4, CSR_XSTATUS                      // re-read mstatus (T2 copy was shifted to [16:0])
        srli    T4, T4, UDB_MXLEN-32                 // mstatus[39:38] (MPV, GVA) -> bits 7:6, as in mstatush
```

The RV32 branch is unchanged; the `SM1P12P0_OR_LATER_SUPPORTED` alternatives inside it stay one instruction each, so `.text.rvtest` remains the same length across DUTs of a given XLEN (RV32 and RV64 builds are already different lengths, e.g. `LI` expansions).

After the fix (`sv_Mvect`, sail-rv64-max): `csrrs x9,mstatus,x0 ; srli x9,x9,0x20 ; andi x9,x9,448 ; slli x9,x9,0x8`.

## Regressions (priv suites, `EXTENSIONS=` from `testgen.priv.get_priv_test_suites()`, `JOBS=16`), identical to upstream act4 2dea4b02c

| config | result |
| --- | --- |
| sail-rv64-max | 649/649 |
| spike-rv64-max | 369/370 (`Smstateen-00` fails identically on act4; unrelated) |
| sail-rv32-max | 505/505 (RV32 branch untouched; sanity) |

No signature changes are expected on any current DUT because `H_SUPPORTED` is undefined everywhere, so mstatus[39:38] read as 0; the fix only makes the code do what its comment says.


davidharrishmc:
Closing: #2429 (trap-status-word) already fixes this — it keeps mstatus in T2 and does `srli T4, T2, UDB_MXLEN-32` for the RV64 GVA/MPV overlay, along with the documented bit positions.

PR #2471, opened 2026-09-22
Trap prolog: keep the selected xTVEC MODE when installing the fast trap handler

davidharrishmc:
The fast trap handler always used direct mode. Update it to set the mode based on the UDB parameters.


jordancarlin (approved):

jordancarlin:
@davidharrishmc this one isn't signed either.

PR #2472, opened 2026-09-22
Trap prolog: select the xTVEC MODE from this mode's UDB_*TVEC_MODES, not mtvec's

davidharrishmc:
`RVTEST_TRAP_PROLOG` now selects the `xTVEC` MODE from the CSR used by each trap-handler mode:

- M-mode: `MTVEC_MODES`
- S- and HS-mode: `STVEC_MODES`
- VS-mode: `VSTVEC_MODES`

Previously, every mode used `MTVEC_MODES`. This caused S-mode prologs to abort when `stvec` supported vectored mode but `mtvec` did not have the same mode set.


jordancarlin (commented):
~~Accidental review submission; no review comments.~~

I have no idea what created this or what added the comment above saying it was an accidental submission.

jordancarlin (approved):

PR #2473, opened 2026-09-22
Gate the boot mcountinhibit write and the Sm mcountinhibit tests on UDB_MCOUNTINHIBIT_IMPLEMENTED

davidharrishmc:
## Bug

Boot writes `mcountinhibit` unconditionally (`tests/env/rvtest_setup.h:1068`, `csrw mcountinhibit, zero`) before `mtvec` is programmed. `mcountinhibit` is optional (UDB `MCOUNTINHIBIT_IMPLEMENTED`), and on a DUT without it the access is reserved and typically traps, so the hart traps into an unprogrammed `mtvec` and never reaches the test. The TODO comment at `:1065-1067` asks for exactly the parameter that `udb-gen` already emits into `rvtest_config.h` as `UDB_MCOUNTINHIBIT_IMPLEMENTED`.

Evidence: UDB-parameter sweep on `whisper-rv64-max` with `MCOUNTINHIBIT_IMPLEMENTED: false` (`COUNTINHIBIT_EN` removed as UDB requires, whisper `csr.mcountinhibit.exists=false`): all 52 tests of `Zicntr,ZicntrS,ZicntrSm,ZicntrU,Zihpm,Sscounterenw,Sm,S,U,Ssu64xl` die at boot with `Failed stop: Hart 0: 16 consecutive illegal instructions`; the whisper trace shows the first exception at `csrrw x0, mcountinhibit, x0`.

## Fix

- `rvtest_setup.h`: wrap the write in `#ifdef UDB_MCOUNTINHIBIT_IMPLEMENTED`. The `mhpmevent3..31` writes that follow are left unconditional: the spec requires those CSRs to be implemented in M-mode (read-only zero is a legal implementation when `HPM_COUNTER_EN[n]` is false), so writing zero to them is always legal; the comment now says so.
- `Sm.py`: the `Sm` suite also reads and writes `mcountinhibit` in `cp_mcsr_access`/`cp_mcsrwalk` (the `csrm` list, `:623`) and `cp_inhibit_mcycle`/`cp_inhibit_minstret` (`Sm_mcsr_cntr`). Those are now emitted inside `#ifdef UDB_MCOUNTINHIBIT_IMPLEMENTED` (a small `_gate_csr` helper wraps the emitted lines for CSRs whose presence is parameterised, the same way the `SM1P12P0_OR_LATER_SUPPORTED`/`MSECCFG_SUPPORTED` blocks are gated), including the `csrw mcountinhibit, x0` that re-enables the counters before the wraparound tests. Generated output is unchanged for every DUT that implements the CSR apart from the added `#ifdef`/`#endif` lines (the `Sm` suite's generated files are not committed).

## Verification

Same experiment with this patch: 52/52 pass (`~/reviews/act/udb-param-experiments/MCOUNTINHIBIT_IMPLEMENTED=false-fixed/` locally, baseline `baseline-counters` 52/52). Sail has no knob to remove `mcountinhibit`, so the parameter value itself remains Sail-untestable, but the boot no longer dies and the Sm CSR tests no longer assert its presence.

Full regression, before and after (unchanged; `Smstateen-00` on spike is the known pre-existing trap-signature-word-0 failure):

```
EXTS=$(mise exec -- uv run python -c "from testgen.priv import get_priv_test_suites; print(','.join(sorted(get_priv_test_suites())))")
make -k sail-rv64-max sail-rv32-max spike-rv64-max spike-rv32-max EXTENSIONS=$EXTS JOBS=16
```

| target | result |
|---|---|
| sail-rv64-max | 649/649 |
| sail-rv32-max | 505/505 |
| spike-rv64-max | 369/370 (Smstateen-00, pre-existing) |
| spike-rv32-max | 225/226 (Smstateen-00, pre-existing) |


davidharrishmc:
Closing: #2449 (counter-init-gating) already gates the boot mcountinhibit write and the Sm mcountinhibit access/walk/inhibit tests on UDB_MCOUNTINHIBIT_IMPLEMENTED, with the same treatment of the mhpmevent writes, and additionally removes the dead misa read in the trap handler. Nothing here is missing from #2449.

PR #2474, opened 2026-09-22
Svbare/Sstvecd: require the suite's own extension in the test generators

davidharrishmc:
Add Svbare and Sstvecd as required extensions in their respective tests.


jordancarlin (approved):

jordancarlin:
@davidharrishmc this commit is not signature verified.

Issue #2010, opened 2026-07-28
Sm_mcause-00 assumes exact readback for unsupported WLRL mcause cause 19


farazahmad204:
Sm_mcause-00 fails on a StarFive VisionFive 2 / SiFive U74 at Sm_mcause_cg_cp_mcause_write_exception_b_19.

The generated test writes exception cause 19 to mcause, reads the CSR back, and requires the result to equal the Sail reference value:

li    x9, 19
csrrw x0, mcause, x9
csrrs x9, mcause, x0

Sail 0.13 retains 0x13, while the U74 reads back 0x03. ACT consequently reports:

RVCP: Test Info: "test: 18; cg: Sm_mcause_cg; cp: cp_mcause_write_exception; bin: b_19"
RVCP: Instruction: 0x34249073
RVCP: Approximate address (failure may be slightly after this): 0x0000000080000370
RVCP: Register: x9
RVCP: Bad Value:      0x0000000000000003
RVCP: Expected Value: 0x0000000000000013

This appears to be an ACT generation/configuration issue. The mcause Exception Code field is WLRL. Exact preservation is required only for supported exception codes. Cause 19 is not supported by this target:

The target's sail.json selects Privileged_ISA_1_11. Privileged ISA 1.11 reserves exception codes 16 through 23.

The U74 manual marks exception codes greater than or equal to 16 as reserved and describes mcause. EXCCODE as WLRL.

Therefore, requiring an implementation-independent exact readback of 0x13 for a write of unsupported value 19 is not valid. Reading back 0x03 does not, by itself, indicate a hardware violation.

there are other interrupt causes such as H/VS causes 2, 6, 10, 12 and counter-overflow cause 13 which are testing unconditionally. 
The U74 does not implement the Hypervisor extension, so these interrupt causes simply do not exist on the processor.

[act_sm_mcause_wlrl_jenkins_weekly_14.zip](https://github.com/user-attachments/files/30457998/act_sm_mcause_wlrl_jenkins_weekly_14.zip)

allenjbaum:
Actually, the spec says that "other bit encodings are reserved"  and that
they return a deterministic value dependent on the " illegal written value
and the value of the field prior to the write."
Basically, tests that write unsupported values should be skipped, since
they are reserved.
We need to define a parameter (if there isn't one already) about which
mcause values are supported; effectively, these would be identical to the
parameter schemas used for WARL fields

The defining normative rule is the sentence in 3.1.15 "The Exception Code
[...]

PR #2478, opened 2026-09-23
Do not merge until Sail supports Ssdbltrp: Gate the mcause double-trap bin on Ssdbltrp, not Smdbltrp

davidharrishmc:
`Sm_mcause` requires exception code 16 to read back from `mcause` whenever Smdbltrp is implemented, and the coverage bin uses the same gate. Only Ssdbltrp writes code 16 to `mcause`; Smdbltrp delivers an M-mode double trap as an RNMI with the cause in `mncause`, and `mcause` is WLRL, so a hart with Smdbltrp but not Ssdbltrp need not keep 16. This PR gates both the test and the bin on Ssdbltrp, following the pattern of #2010/#2134 for cause 19.


davidharrishmc:
Do not merge this until Sail supports Ssdbltrp.

jordancarlin (approved):
LGTM. Ssdbltrp will be in the upcoming Sail 0.15 release (approved today). We can merge this once it is released.

PR #2480, opened 2026-09-23
Record sstatus with mstatus in RVVI traces

davidharrishmc:
`sstatus` is a view of `mstatus`, but Sail logs a write only under the CSR it names, and it names `mstatus`. The coverage testbench never saw `sstatus`, so coverpoints that read `csr[CSR_SSTATUS]` (Sv, SvZicbo, Sscofpmf, Endian) sampled an unset entry: X in Questa, which hits no bin. The converter now records the matching `sstatus` value with every `mstatus` write, and the updated `mstatus` with any `sstatus` write.

In Questa, `sum_sstatus` and `mxr_sstatus` are now covered. Sv_mstatus_mprv_cg rises from 35.71% to 38.94% (rv64), Sv_vm_permissions_cg from 7.57% to 9.57%, and SvZicbo_cg from 9.78% to 11.64%, with similar gains on rv32.


jordancarlin:
The sail model has been moving in the direction of reporting all changed CSRs. This has already been fixed for fflags/fcsr, but apparently not yet for `sstatus`. Since we need to wait on the next Sail release for Sv and Sscofpmf coverage anyway, I think we should wait on that and then we can revisit this if it is still a problem. The script you are modifying here is going away with that next release anyway when we move to native RVVI-Text, so this doesn't seem like the right approach even if we do need to keep a workaround.

PR #2481, opened 2026-09-23
Split Sm csr_ro bins below Verilator's bin-array limit

davidharrishmc:
`bins readonly[] = {[12'hC00:12'hFFF]}` creates 1024 bins. Verilator drops a bin array larger than 1000, which removed `csr_ro` and `cp_csr_ro` from Sm_mcsr_cg. This splits it into four 256-bin arrays covering the same values.

Sm_mcsr_cg now matches between Questa and Verilator (96.50% rv64, 96.69% rv32), with all 1024 `csr_ro` bins covered in both.


jordancarlin:
Does Verilator silently drop the coverage or does it emit a warning/error of some kind? If it is the former, I would rather fix it in Verilator because this is inevitably going to come up again and may lead to silently missed coverage if we have fully switched over to Verilator.

davidharrishmc:
Verilator gives a SELRANGE warning, though Claude's current coverage script does not display warnings.

jordancarlin:
Does adding `-Wall` turn that warning into an error? I want to make sure we have a way of preventing this from sneaking back in in the future.

I have a local Verilator coverage script from when I was testing it out a few weeks ago. Once we are ready to actually enable everything, I can compare it to whatever Claude generated script you are using.

wsxarcher:
1000 was an artificial limit introduced at the beginning of the Coverage support, I think 1024 will work the same and it makes more sense as a power of 2, feel free to do a PR in Verilator  and ensure everything works @davidharrishmc

davidharrishmc:
-Wall does not turn the warning into an error.  We have no known way to prevent this from sneaking back in, other than to review warnings.  Warnings seems like it could be adequate.

I have many PRs open on Verilator right now and don't want to open more at this point.

davidharrishmc:
Correction to my earlier reply: the warning for the dropped `csr_ro` array is `COVERIGN` ("array 'bins' covering more than 1000 values ... bin ignored"), not `SELRANGE`. `-Wall` does not change it: `COVERIGN` is on and fatal by default, so the build already stops. It went unnoticed only because #2483 builds with `-Wno-fatal`, which it needs while `type_option.weight` and `with` still raise `COVERIGN`. When the Verilator `with` and weight work is finished, #2483's flow should add `-Werror-COVERIGN` so a dropped bin array fails the build again (tracked in #2484).

UmerShahidengr (approved):
LGTM

PR #2483, opened 2026-09-23
Add Verilator as a coverage simulator

davidharrishmc:
Adds Verilator as a third backend for `make coverage`, alongside Questa and VCS.

**Depends on Verilator PRs [#8459](https://github.com/verilator/verilator/pull/8459), [#8460](https://github.com/verilator/verilator/pull/8460), [#8463](https://github.com/verilator/verilator/pull/8463), and [#8464](https://github.com/verilator/verilator/pull/8464).** They fix `$sscanf` return values, `'0`/`'1` bin values, embedded covergroups calling enclosing-class methods, and enum automatic bins. Released Verilator versions compile many suites incorrectly or not at all, so use a Verilator built from `master` once these are merged.

To run:

```bash
make coverage COVERAGE_SIMULATOR=verilator
make coverage COVERAGE_SIMULATOR=verilator EXTENSIONS=ExceptionsSm
VERILATOR=~/verilator/bin/verilator make coverage COVERAGE_SIMULATOR=verilator
```

`docs/DeveloperGuide.md` has a new "Collecting Coverage with Verilator" section.

- `riscv-arch-test-verilator.sh` builds the testbench with `verilator --binary --coverage-user`, runs the traces, and writes `<suite>.dat`.
- The report reads covergroup bins from `coverage.dat`. Verilator ignores `type_option.weight`, so the report reads each item's weight from its declaration and computes covergroup coverage as the weighted average of its items, as Questa does.
- Build warnings that can affect coverage are listed at the top of each suite's report and in `<suite>_warnings.txt`, and the coverage summary names the suites that have them.
- The coverage testbench reads the trace list into its own variable, which removes a Verilator MULTIDRIVEN warning.

With the Verilator fixes above, a full `make coverage` on sail-rv64-max and sail-rv32-max builds and runs every suite. 4946 of 4955 covergroups match Questa exactly. #2480, #2481, and #2482 fix the other 9. Verilator's C++ compilation makes a first run several times slower than Questa.


jordancarlin:
> * Verilator ignores `type_option.weight`, so the report reads each item's weight from its declaration and computes covergroup coverage as the weighted average of its items, as Questa does.

I would rather see Verilator natively support `type_option.weight` instead of using this workaround.

davidharrishmc:
Runtime measurements for Verilator coverage, for reference. All runs cover both default configurations (sail-rv64-max and sail-rv32-max, 249 coverage suites) on a 16-core (32-thread) AMD Ryzen 9 9950X, using Verilator master plus verilator#8459, #8460, #8463, #8464, and #8467.

### Full `make coverage`, 16 jobs, cold ccache

| | Questa | Verilator |
|---|---|---|
| ELF and trace generation (shared) | 28 min | 28 min |
| Coverage wall time | 6.7 min | 28.5 min |
| Compile (summed over suites) | ~5 min | 330 CPU-min (C++ 325, Verilator 5) |
| Simulation (summed over suites) | ~77 min | 72 CPU-min |
| Build directories on disk | small | 49 GB (44 GB of precompiled headers) |

Verilator simulation is 1.0-1.7x Questa's per suite; the difference is the C++ build. Running 32 jobs instead of 16 gives the same wall time (28:44): the run is CPU-bound on the 16 physical cores. A warm ccache does not help either: the default 5 GB cache evicts during a full run, and precompiled headers make many compiles uncacheable.

Improvements measured so far: verilator#8467 cuts the U suite's C++ build from 497 s to 35 s and Vx16's from 196 s to 180 s; #2489 cuts Vx16's Verilator simulation from 252 s to 188 s.

### Prototype: one Verilator model per configuration

Every suite's coverage compiled into one model per configuration, with a plusarg selecting the suites to construct and sample (`+coverSuites=Sm`):

| | Per-suite builds (this PR) | One model per configuration |
|---|---|---|
| Build | 330 CPU-min | rv64 191 s + rv32 187 s (~30 CPU-min) |
| Runs (249 suites, 16 at a time) | included above | 366 s (69 CPU-min) |
| Coverage wall time | 28.5 min | ~12.5 min |
| Build directories on disk | 49 GB | 3.5 GB |

230 of 249 suites give identical coverage. The other 19 differ because some code depends on which suites are compiled: `COVER_E` sets `NUM_REGS` in the shared testbench, `disassemble.svh` checks `I_`, `ZIMOP_`, `ZCMOP_`, `ZICBOP_`, and `ZIHINT*_COVERAGE`, the Vls/Vx/Vf families check their siblings' `COVER_` defines, and Zicfilp checks `COVER_ZCA`. Those would need to depend on the configuration instead before a combined model is usable.

### Alternatives compared

- **Suite selection matters.** With every suite's covergroups enabled, each instruction costs ~1 ms instead of 30-50 us: Zba 1.5 s -> 43.5 s, I 7 s -> 223 s, Vx16 276 s -> 4875 s.
- **One run over all traces with all covergroups enabled** costs ~27 CPU-hours (97 M instructions at ~1 ms), ~14 h of latency in one process per configuration, or ~1.7 h split 16 ways. It also credits each suite with bins hit by other suites' tests.
- **Questa** compiles each suite in 1-7 s, so a combined model does not pay off: the combined rv64 design compiles in 57 s, but loading it adds ~9 s to every run (Zba 4 s -> 10.9 s). A single run over all traces costs ~0.35 ms per instruction more, ~9-10 CPU-hours in total.

For Verilator, one model per configuration with suite selection is the better long-term flow; for Questa, per-suite builds remain best. ELF and trace generation (28 min) is then the largest remaining cost for both.

